PR #42, opened 2015-01-06
Fix NPE when unboxing in JsonStream

mattprecious:


mattprecious:
Uhh... I think this one needs to be merged and released ASAP. I believe no crashes are being reported pre-5.0 without the `GET_TASKS` permission since `isInForeground()` returns `null`.

ConradIrwin:
Thanks @mattprecious, @JakeWharton!

loopj:
Released as v3.2.1, propagating through maven central now. Thanks again!

mattprecious:
Thanks for the quick release!

PR #43, opened 2015-01-07
Remove the util folder to fix debugging

mattprecious:
All the classes in the `util` folder use the package `com.bugsnag.android` so IntelliJ can't find the sources when debugging.


loopj:
Thanks @mattprecious, released in v3.2.1, propagating through maven central now.

PR #45, opened 2015-01-08
Changes for my unity experiments over xmas

loopj:


ConradIrwin:
@loopj any chance you could merge this?

PR #47, opened 2015-01-09
Prefer passed apiKeys to manifest apiKeys

loopj:


src/main/java/com/bugsnag/android/Client.java
```diff
@@ -61,7 +62,10 @@ public Client(Context androidContext, String apiKey, boolean enableExceptionHand
         // Attempt to load API key from AndroidManifest.xml
         try {
             ApplicationInfo ai = appContext.getPackageManager().getApplicationInfo(appContext.getPackageName(), PackageManager.GET_META_DATA);
-            apiKey = ai.metaData.getString("com.bugsnag.android.API_KEY");
+            String manifestApiKey = ai.metaData.getString("com.bugsnag.android.API_KEY");
+            if(!TextUtils.isEmpty(manifestApiKey)) {
```
JakeWharton: Oh, this will work. I was thinking
```java
if (TextUtils.isEmpty(apiKey)) {
  try { /* load from manifest */ } catch {}
}
```

Maybe it should crash if you specify it in both places? Just to prevent people from doing something stupid (like only changing one value in the future and getting unexpected behavior).

loopj: Interestingly @JakeWharton I cannot reproduce your issue, in my tests I always fall into the `catch` block when trying to load the api key from the manifest.

Also, I think your suggested approach is better.

JakeWharton: I think Play Services' library manifest got merged which means I have metadata, just not this key.

loopj: Bingo, that's it. Can reproduce now with non-bugsnag metadata present.

PR #60, opened 2015-02-19
Keep Async class (fixes #39)

### LINKED ISSUE #39: Dex doesn't like Async. Proguard config problem?
```
warning: Ignoring InnerClasses attribute for an anonymous inner class
(com.bugsnag.android.Async$1) that doesn't come with an
associated EnclosingMethod attribute. This class was probably produced by a
compiler that did not target the modern .class file format. The recommended
solution is to recompile the class from source, using an up-to-date compiler
and without specifying any "-target" type options. The consequence of ignoring
this warning is that reflective operations on this class will incorrectly
indicate that it is *not* an inner class.
```

ConradIrwin:


loopj:
nice

ConradIrwin:
self-:penguin:

PR #64, opened 2015-05-22
Add blocking API

### LINKED ISSUE #63: Bugsnag drops a lot of crashes
Here: https://github.com/bugsnag/bugsnag-android/blob/master/src/main/java/com/bugsnag/android/Client.java#L517

The crashes are written to disk asynchronously, and then the exception is handled to the default exception handler. 

If the app is in the foreground, the default handler will show the crash dialog and block there, then when the user clicks on it the default handler kills the process. See here: https://github.com/android/platform_frameworks_base/blob/master/core/java/com/android/internal/os/RuntimeInit.java#L88

If the app is in the background, there is no dialog, so the process dies immediately.

So this basically means that the process can die before the async work is done, so before the crash is saved to the file system.

We can't even work around that and use our own exception  handler, because we do not have any hook into the async work to wait for its completion.

Btw, `Async` uses `AsyncTask`. That's terrible, its internal behavior  has changed over Android versions. Don't use that.

pyricau:
When uploading a crash, it makes sense to do it from the crashing thread before delegating to the default handler, to guarantee that the crash is either uploaded or saved before the process is killed.

* This change introduces an API that allows one to write a blocking `UncaughtExceptionHandler`.
* I did not change the default exception handler, because this change might slightly delay the appearance of the crash dialog.
* Also fixed `Async` to use a Java Executor instead of the `AsyncTask` garbage. `AsyncTask` should be burnt to the ground.
* The async API is still needed when you want to log a non crashing throwable.

Fixes #63

### DIFF
--- a/src/main/java/com/bugsnag/android/Async.java
+++ b/src/main/java/com/bugsnag/android/Async.java
@@ -1,15 +1,13 @@
 package com.bugsnag.android;
 
-import android.os.AsyncTask;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 
 class Async {
-    static void run(final Runnable task) {
-        new AsyncTask <Void, Void, Void>() {
-            @Override
-            protected Void doInBackground(Void... voi) {
-                task.run();
-                return null;
-            }
-        }.execute();
+
+    private static final Executor executor = Executors.newCachedThreadPool();
+
+    static void run(Runnable task) {
+        executor.execute(task);
     }
 }
--- a/src/main/java/com/bugsnag/android/Client.java
+++ b/src/main/java/com/bugsnag/android/Client.java
@@ -17,6 +17,9 @@ import android.text.TextUtils;
  * @see Bugsnag
  */
 public class Client {
+
+    private static final boolean BLOCKING = true;
+
     private Configuration config;
     private Context appContext;
     private AppData appData;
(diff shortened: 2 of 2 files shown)


snmaynard:
Hey @pyricau, thanks for all your other PRs - really appreciate it!

I'd love to get your thoughts on when it would be appropriate to use the blocking notify call in the crash handler. Would you anticipate it being when the app is running as a service or in the background or something? I'm trying to figure out if there is anything appropriate that we can include in the notifier to have a better default.

Possibly we could always block until we have written the crash report to the disk and then asynchronously try to send the crash reports (or wait until the next app boot if that fails). This is very similar to what we do on iOS. That way the crash dialog appears reasonably quickly and it doesnt depend on network time.

pyricau:
SGTM. So the behavior would be:

1) Write to disk synchronously
2) Start sending to network asynchronously, and once done remove the crash written to disk
3) Delegate to default handler
4) On startup, send the crash if available

If you do that though, you need to make sure the crash uploads are idempotent. There is a chance that the process will die after the network has sent the crash but before that crash has been removed from the file system.

I would still keep the same general API (blocking vs non blocking), but I would change the blocking API to block on the FS only, and then go async for the network.

ConradIrwin:
Thanks @pyricau. I've merged your changes so far, as I think they're an improvement. We can come back and change the default crash handler as suggested in another PR.

PR #70, opened 2015-07-02
make setting and clearing of max breadcrumbs static methods

wuman:
The setting and clearing of max breadcrumbs are currently not static methods and hence can never be called.  This pull request corrects the error.


snmaynard:
Thanks @wuman - good catch!

PR #77, opened 2015-11-29
Jailbreak detection improvement

d4rken:
While just checking for the existence of certain files does not guarantee a device is rooted, it's something we can check with virtually no impact on the user.

I've extended the current type of check with a few more locations.
The previous check only worked if the device was using a specific superuser app.

Additionally, on newer Android versions the app folder hierarchy changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ proguard/
 
 # Log Files
 *.log
+
+# Android Studio #
+.idea
+*.iml
--- a/src/main/java/com/bugsnag/android/DeviceData.java
+++ b/src/main/java/com/bugsnag/android/DeviceData.java
@@ -1,13 +1,13 @@
 package com.bugsnag.android;
 
+import android.content.ContentResolver;
+import android.content.Context;
+import android.provider.Settings;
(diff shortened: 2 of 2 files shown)


d4rken:
The tests complete fine on my computer if a device is attached.
Why does the travis build expect a connected device?
```
com.android.builder.testing.ConnectedDevice > runTests[test(AVD) - 2.3.3] FAILED 
com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException
at com.android.builder.testing.ConnectedDevice.installPackage(ConnectedDevice.java:86)
```

kattrali:
Thank you for the contribution, @d4rken! The changes look good and the detection is a bit more robust now.

> Why does the travis build expect a connected device?

Per the [documentation](http://developer.android.com/tools/help/adb.html), `adb wait-for-device` doesn't wait for the device to fully boot:

> Note that this command does not cause adb to wait until the entire system is fully booted.

We may need to increase the timeout or otherwise wait for boot completion in some other way.

Merging the PR, though we can continue the discussion to fix the timeout issues.

Issue #78, opened 2015-12-08
Exception in Bugsnag.init() when getResouces() is null


duncanhewett:
In some circumstances an exception can be thrown during initialization when Context.getResources() is null.

It appears that this only happens when the App's onCreate is triggered by a widget refresh.

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.util.DisplayMetrics android.content.res.Resources.getDisplayMetrics()' on a null object reference
at com.bugsnag.android.DeviceData.<init>(DeviceData.java:33)
at com.bugsnag.android.Client.<init>(Client.java:87)
at com.bugsnag.android.Client.<init>(Client.java:45)
at com.bugsnag.android.Bugsnag.init(Bugsnag.java:35)
```

PR #84, opened 2015-12-09
Check for context resources before loading data

kattrali:
Potential fix for issue #78, adds a check for resources being non-null before caching device data.


snmaynard:
LGTM!

PR #85, opened 2015-12-11
Update UPGRADING.md

ben-j69:
Context missing in args


kattrali:
Nice catch! Thanks, @ben-j69.

PR #87, opened 2015-12-25
Remove the need for synthetic methods.

JakeWharton:
`javac` generates package-scoped methods when you access private fields and methods from inner to outer classes (and vise-versa).

```
$ dex-method-count before.dex
443

$ dex-method-count after.dex
436
```

Every bit counts for libraries!


kattrali:
Looks good, thanks @JakeWharton.

> Every bit counts for libraries!

Too true :+1:

Issue #50, opened 2015-01-20
Should log when no INTERNET permission


ConradIrwin:
At the moment we log to recommend people enable the other permissions, but if you don't have internet it just fails and writes the request to disk without telling you why.

kattrali:
Closing in favor of #90.

larsgrefer:
@kattrali This issue would have been closed automatically when #91 gets merged

PR #91, opened 2016-01-04
Distribute the library only as aar (not as jar)

### LINKED ISSUE #90: Distribute the library only as aar (not as jar)
Using an aar instead of a simple jar has two advantages:
- An aar can declare `<uses-permission />`-elements in its own manifest which gets merged to the apps manifest during the build.
- An aar can contain proguard rules which are used during the build of the app.
  - http://google.github.io/android-gradle-dsl/1.5/com.android.build.gradle.internal.dsl.ProductFlavor.html#com.android.build.gradle.internal.dsl.ProductFlavor:consumerProguardFiles

larsgrefer:
see and close #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,13 +43,12 @@ Configuring Your AndroidManifest
     </application>
     ```
 
--   Enable the `INTERNET`, `ACCESS_NETWORK_STATE` and `GET_TASKS` permissions:
+-   Recommended: Enable the `ACCESS_NETWORK_STATE` and `GET_TASKS` permissions:
 
     ```xml
-    <!-- Required: Used to deliver Bugsnag crash reports -->
-    <uses-permission android:name="android.permission.INTERNET"/>
     <!-- Optional: To provide network connectivity information to Bugsnag -->
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+    
     <!-- Optional: To see which Activity was active at the time of a crash -->
     <uses-permission android:name="android.permission.GET_TASKS"/>
     ```
--- a/README.md
+++ b/README.md
@@ -32,12 +32,6 @@ Add `bugsnag-android` as a dependency in your `pom.xml`:
 </dependency>
 ```
 
-### Using a Jar
-
--   Download the [latest bugsnag-android.jar](https://github.com/bugsnag/bugsnag-android/releases/latest)
--   Place it in your Android app's `libs/` folder
-
-
 Configuring Your AndroidManifest
 --------------------------------
 
--- a/proguard.txt
+++ b/proguard.txt
@@ -1,10 +1,11 @@
--dontobfuscate
+-keeppackagenames com.bugsnag.**
+-keepnames class com.bugsnag.**
+-keepclassmembernames class com.bugsnag.** { *; }
 
--renamesourcefileattribute SourceFile
 -keepattributes Exceptions,InnerClasses,Signature,Deprecated,
                 SourceFile,LineNumberTable,*Annotation*,EnclosingMethod
 
--keep public class * {
+-keep public class com.bugsnag.** {
     public protected *;
 }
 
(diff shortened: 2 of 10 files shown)


src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,15 @@
...
+  android:versionName="1.0">
+
+  <!-- Required: Used to deliver Bugsnag crash reports -->
+  <uses-permission android:name="android.permission.INTERNET" />
+
+  <!-- Optional: To provide network connectivity information to Bugsnag -->
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+
+  <!-- Optional: To see which Activity was active at the time of a crash -->
+  <uses-permission android:name="android.permission.GET_TASKS" />
```
JakeWharton: I wouldn't put these two here as they are automatically added to all consumers requiring opt-out. Opt-in seems a much better strategy for the optional permissions.

README.md
```diff
@@ -49,13 +43,12 @@ Configuring Your AndroidManifest
     </application>
     ```
 
--   Enable the `INTERNET`, `ACCESS_NETWORK_STATE` and `GET_TASKS` permissions:
+-   Optional: Enable the `ACCESS_NETWORK_STATE` and `GET_TASKS` permissions:
```
loopj: I'd rather say "recommended" here, since having these available massively aids in debugging. We can possibly remove `GET_TASKS` since we get this information by default now in Android 5.0+.

larsgrefer: @loopj What is your opinion about this comment by @JakeWharton : https://github.com/bugsnag/bugsnag-android/pull/91#discussion_r48775577

loopj: I agree with Jake, these should not be automatically added (only INTERNET should be automatic, as it is 100% required).

larsgrefer:
The PR is ready to merge now. Please review and merge @kattrali

kattrali:
This looks good to me. Is there any reason we would need to keep jar distribution around? 

/cc @snmaynard

src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
   package="com.bugsnag.android"
   android:versionCode="1"
-  android:versionName="1.0" >
+  android:versionName="1.0">
+
+  <!-- Required: Used to deliver Bugsnag crash reports -->
+  <uses-permission android:name="android.permission.INTERNET" />
```
kattrali: Its nice to be able to guarantee this is enabled as well. :+1:

proguard.txt
```diff
@@ -1,10 +1,11 @@
...
+-keeppackagenames com.bugsnag.**
+-keepnames class com.bugsnag.**
+-keepclassmembernames class com.bugsnag.** { *; }
 
--renamesourcefileattribute SourceFile
 -keepattributes Exceptions,InnerClasses,Signature,Deprecated,
                 SourceFile,LineNumberTable,*Annotation*,EnclosingMethod
 
--keep public class * {
+-keep public class com.bugsnag.** {
```
kattrali: Was there some history behind keeping `*` instead of just `com.bugsnag.*` here, @snmaynard?

larsgrefer: In the current master the proguard rules are only used during the build of the library itself so only classes in this library are affected by `-keep public class *`.

With this PR the proguard rules aren't used during the build of library anymore but during the build of the app which includes the library. With the change above we make sure the rule still only affects the library and not the whole app

src/main/AndroidManifest.xml
larsgrefer: That is one of the reasons for this PR :+1:
see #50

proguard.txt
kattrali: Right, that rings a bell. Thanks. :)

larsgrefer: @kattrali I wonder if the rule is needed at all when proguard runs during the app-build instead of during the library-build

src/main/AndroidManifest.xml
kattrali: Of course, just good to see it in action. :)

larsgrefer:
@kattrali What is the current status of this?

PR #92, opened 2016-01-05
remove redundant modifiers and semicolons

larsgrefer:


kattrali:
Thanks, merged similar changes with #96.

PR #93, opened 2016-01-05
compile with and against api 23

larsgrefer:


kattrali:
:+1: :star: Thanks!

gradle.properties
```diff
@@ -16,6 +16,6 @@ POM_URL=https://bugsnag.com
 POM_ARTIFACT_ID=bugsnag-android
 POM_PACKAGING=jar
 
-ANDROID_BUILD_TARGET_SDK_VERSION=19
-ANDROID_BUILD_TOOLS_VERSION=21.1.1
-ANDROID_BUILD_SDK_VERSION=19
```
larsgrefer: Just because im curious: Why are these 3 values defined here and not in the build.gradle?

/cc @loopj

loopj: ¯\_(ツ)_/¯ - I can't remember the reasoning behind this :)

larsgrefer: I've never seen this before and can't think of any advantages

In fact it has the disadvantage that AndroidStudio can't interprete the values and can't show hints about newer versions

Issue #86, opened 2015-12-19
DeviceData default values.


d4rken:
Example: https://github.com/bugsnag/bugsnag-android/blob/9470323e8e9eababd5dcc04d16a229f429e784e7/src/main/java/com/bugsnag/android/DeviceData.java#L72

If we can't obtain a value, is it not better to omit the value than to send a default value of "0"?
For someone checking stats in the webconsole, it would be better to see that the value isn't available instead of guessing whether bugsnag defaulted to that value or the system provided that value.

If this doesn't conflict with some API requirements of bugsnag, i would submit a pull request to change this.

kattrali:
I agree, @d4rken, that not sending an invalid value is best, though I'd have to double-check the API requirements. It sounds like a good plan to exclude the device data values which are empty (zero value or zero-length string). I'd propose checking for empty values before they are written in `toStream` as a possible solution. 

/cc @snmaynard

snmaynard:
The api has no requirements on this, so we should be able to make any required changes here. Seems like a good idea to me!

d4rken:
About this:
https://github.com/bugsnag/bugsnag-android/blob/97eca13c7aa604a6578a231abb83cdfa5e14bf49/src/main/java/com/bugsnag/android/JsonStream.java#L29
So if we write a null value into the json stream, what does the API do?
Does it output "null" for the value in the webinterface or does it omit the value?

d4rken:
@snmaynard Need some clarification on the API behavior ^

kattrali:
Even if the API properly discarded null values, it is a bit inefficient to send data to be discarded, just as sending zeroes is in the current behavior. Checking whether to send those values and only sending them if the values are non-null before writing them would resolve this.

kattrali:
Closed via #96, released in 3.3.0

PR #96, opened 2016-01-06
Don't send null values

d4rken:
Overall goal: Don't write null values into the output stream. See #86 
To help with that most methods that produce output for the stream have been made static and self contained. The helper methods and parameters (where applicable) have been annotated to indicate whether the value can be null using Androids annotation library.
Fields holding generated data have also been made final (was suggested by LINT checks).

Android Studio also seems to have corrected some javadoc formatting which was not intended, but oh well...

This also includes the same fixes as PR #92


d4rken:
Doesn't look like a code issue:

```
* What went wrong:

Execution failed for task ':connectedDebugAndroidTest'.

> com.android.builder.testing.api.TestException: com.android.builder.testing.api.DeviceException: com.android.ddmlib.ShellCommandUnresponsiveException
```

src/main/java/com/bugsnag/android/Client.java
```diff
@@ -287,22 +287,22 @@ public void setUserName(String name) {
...
-     *
+     * <p/>
      * For example:
+     * <p/>
+     * client.beforeNotify(new BeforeNotify() {
+     * public boolean run(Error error) {
+     * error.setSeverity(Severity.INFO);
+     * return true;
+     * }
+     * })
```
kattrali: This formatting probably needs corrected to the original version. Maybe Android Studio would like a `<code>` block better.

JakeWharton: Multi-line code snippets in Javadoc should use:
```
<pre>{@code
blah blah
}</pre>
```

kattrali:
The changes look good. I got a bit bogged down with the unrelated changes, but looked over everything. :) Next up is doing more integrated testing, just for sanity checks. Thanks for putting work into this, @d4rken.

src/main/java/com/bugsnag/android/Client.java
kattrali: Ah, that's what I was looking for. Thanks, @JakeWharton. :+1:

Issue #23, opened 2014-08-13
Allow configurable number of stored exceptions


snmaynard:
Limit the client to only storing the last n exceptions. This could either be done as part of writeErrorToDisk (write time) or flushErrors (transmit time). We should only transmit the last n exceptions.

duncanhewett:
What would be a sensible number of exceptions for the client to store?  100?

snmaynard:
This one is done!

PR #97, opened 2016-01-19
Limit number of stored errors

duncanhewett:
Limit the number of errors stored for later sending to prevent possible disk space issues (see #23).

Stores up to 100 errors and discards any subsequent errors.  First 100 errors are stored on the basis that they are likely to be the most interesting.  Is this sensible?

### DIFF
--- a/src/main/java/com/bugsnag/android/ErrorStore.java
+++ b/src/main/java/com/bugsnag/android/ErrorStore.java
@@ -12,6 +12,7 @@ import android.content.Context;
  */
 class ErrorStore {
     private static final String UNSENT_ERROR_PATH = "/bugsnag-errors/";
+    private static final int MAX_STORED_ERRORS = 100;
 
     final Configuration config;
     final String path;
@@ -76,6 +77,13 @@ class ErrorStore {
     // Write an error to disk, for later sending
     void write(Error error) {
         if(path == null) return;
+
+        // Limit number of saved errors to prevent disk space issues
+        File exceptionDir = new File(path);
+        if (exceptionDir.isDirectory() && exceptionDir.listFiles().length >= MAX_STORED_ERRORS) {
+            Logger.warn("Discarding error without saving to disk as stored error limit reached");
+            return;
+        }
 
         String filename = String.format("%s%d.json", path, System.currentTimeMillis());
         Writer out = null;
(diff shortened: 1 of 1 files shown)


kattrali:
Looks good. This makes sense, though I think in the case that we reach the limit, we should probably discard the oldest errors instead of the newest ones. I'm not sure on what is a reasonable limit though.

duncanhewett:
Changed to delete the oldest error and store the new one.

100 seems like a reasonable number - if you have that many errors it is most likely the same thing repeatedly so there is probably limited value to reporting additional instances anyway.

kattrali:
Awesome! :+1:

d4rken:
How much space do 100 errors take up? Where are they stored by default?

kattrali:
@d4rken The files are stored in the [cache directory](https://github.com/bugsnag/bugsnag-android/blob/6b7fc3b7c39d981fa73ff538e519ab7ac3092f26/src/main/java/com/bugsnag/android/ErrorStore.java#L24).

duncanhewett:
The size will depend on lots of things like how big/deep the stack traces are and what custom meta-data is being added, but each error could end up being 10s of KB each at least.  

Storing 100 is not going to take up a huge amount of space but seems sensible to be conservative about the number of errors to store.  Happy to change it if anyone has a different opinion though!

d4rken:
Seems reasonable as default number.

d4rken:
I think code comments don't show anymore when the PR is merged?

Are my concerns valid or did i miss something? I must admit I'm not 100% versed in how the offline storage works at the moment.

https://github.com/bugsnag/bugsnag-android/commit/7f0a6261f58dd4aefeb4cce2c35e5f0c74be136f#commitcomment-15562085

https://github.com/bugsnag/bugsnag-android/commit/7f0a6261f58dd4aefeb4cce2c35e5f0c74be136f#commitcomment-15562415

If these are valid i would make an issue ticket and when I find time a PR.

duncanhewett:
@d4rken I thought I'd reply here for visibility:

> What guarantees that this always sorts such that index 0 is the oldest one?

It ultimately uses the default sort from File.compareTo() which should order alphabetically.  As the files are named with a timestamp, the smallest value (oldest time) is always at index 0.

> Maybe we should check how many errors are above the LIMIT and delete multiple at once instead of only always the last .... impact of this would be pretty minor though as eventually the error store should be emptied.

I considered that we may end up with more than 100 if delete fails, but as it's an arbitrary limit I don't think it matters too much (as you suggest).

d4rken:
Ok :+1:

PR #105, opened 2016-04-23
Remove null checks, instruct JsonWriter to ignore them.

JakeWharton:
This also updates JsonWriter to handle the Boolean case which has been [upstreamed into Gson itself][1].

```
$ ./gradlew clean build connectedCheck | tail -3
BUILD SUCCESSFUL

Total time: 31.71 secs
```

 [1]: https://github.com/google/gson/pull/836

### DIFF
--- a/src/main/java/com/bugsnag/android/Error.java
+++ b/src/main/java/com/bugsnag/android/Error.java
@@ -54,8 +54,7 @@ public class Error implements JsonStream.Streamable {
         // Write error basics
         writer.beginObject();
             writer.name("payloadVersion").value(PAYLOAD_VERSION);
-            if(getContext() != null)
-                writer.name("context").value(getContext());
+            writer.name("context").value(getContext());
             writer.name("severity").value(severity);
             writer.name("metaData").value(mergedMetaData);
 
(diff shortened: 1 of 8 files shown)


JakeWharton:
Build failure looks like a flake.

kattrali:
The build failure was a flake. Thanks for the patch, @JakeWharton.

src/main/java/com/bugsnag/android/Error.java
```diff
@@ -75,38 +74,16 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
-            if(config.sendThreads) {
-                writer.name("threads").value(new ThreadState(config));
-            }
+            writer.name("app").value(appData);
+            writer.name("appState").value(appState);
+            writer.name("device").value(deviceData);
+            writer.name("deviceState").value(deviceState);
+            writer.name("breadcrumbs").value(breadcrumbs);
+            writer.name("groupingHash").value(groupingHash);
+            writer.name("threads").value(new ThreadState(config));
```
ndurell: Sorry I meant to comment earlier...but do you want to always send this thread state object even if sendThreads is false?
Is this an option that is no longer used?

JakeWharton: Yep. Looks like an accident. I'll send another PR in a few.

JakeWharton: https://github.com/bugsnag/bugsnag-android/pull/106

PR #106, opened 2016-05-02
Restore erroneously-removed conditional guard on sending thread info.

JakeWharton:


kattrali:
Ack. I totally overlooked it as well.

PR #109, opened 2016-05-10
Device add CPU_ABI property

crossle:
Device need add cpu_abi,  convenient to see cpu type

### DIFF
--- a/src/main/java/com/bugsnag/android/DeviceData.java
+++ b/src/main/java/com/bugsnag/android/DeviceData.java
@@ -61,11 +61,13 @@ class DeviceData implements JsonStream.Streamable {
         writer.name("dpi").value(dpi);
         writer.name("screenResolution").value(screenResolution);
 
+        String cpuAbi = Build.CPU_ABI;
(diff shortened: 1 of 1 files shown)


src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -59,6 +61,12 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
         writer.name("dpi").value(dpi);
         writer.name("screenResolution").value(screenResolution);
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            writer.name("cpuAbi").value(Arrays.toString(Build.SUPPORTED_ABIS));
+        } else {
+            writer.name("cpuAbi").value(Build.CPU_ABI);
```
JakeWharton: This should include `Build.CPU_ABI2`

Pezzah:
Implemented in https://github.com/bugsnag/bugsnag-android/pull/119

PR #111, opened 2016-07-22
Use v2 API format

kattrali:


src/main/java/com/bugsnag/android/Client.java
```diff
@@ -580,6 +585,11 @@ private void notify(final Error error, boolean blocking) {
             return;
         }
 
+        // Add a breadcrumb for this error occurring
+        HashMap<String, String> breadcrumbMetadata = new HashMap<String, String>();
+        breadcrumbMetadata.put("message", error.getExceptionMessage());
```
JakeWharton: Use `Collections.singletonMap` for a much more space-efficient implementation of a single-element `Map`

src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -373,6 +374,18 @@ public static void leaveBreadcrumb(String message) {
...
 
     /**
+     * Leave a "breadcrumb" log message representing an action or event which
+     * occurred in your app, to aid with debugging
+     *
+     * @param name     A short label (max 32 chars)
+     * @param type     A category for the breadcrumb
+     * @param metadata Additional diagnostic information about the app environment
+     */
+    public static void leaveBreadcrumb(String name, BreadcrumbType type, HashMap<String, String> metadata) {
```
JakeWharton: This should take `Map`, not `HashMap`.

src/main/java/com/bugsnag/android/Breadcrumbs.java
```diff
@@ -3,45 +3,89 @@
...
     void add(@NonNull String message) {
-        if (store.size() >= maxSize) {
-            // Remove oldest breadcrumb
-            store.poll();
-        }
-        store.add(new Breadcrumb(message));
+        addToStore(new Breadcrumb(message));
+    }
+
+    void add(@NonNull String name, BreadcrumbType type, HashMap<String, String> metadata) {
```
JakeWharton: `Map` not `HashMap`

src/main/java/com/bugsnag/android/Breadcrumbs.java
```diff
@@ -3,45 +3,89 @@
...
-            this.message = message.substring(0, Math.min(message.length(), MAX_MESSAGE_LENGTH));
+            this.type = BreadcrumbType.MANUAL;
+            HashMap<String, String> metadata = new HashMap<String, String>();
+            metadata.put(MESSAGE_METAKEY,
+                    message.substring(0, Math.min(message.length(), MAX_MESSAGE_LENGTH)));
+            this.metadata = metadata;
+            this.name = DEFAULT_NAME;
+        }
+
+        Breadcrumb(@NonNull String name, BreadcrumbType type, HashMap<String, String> metadata) {
```
JakeWharton: `Map` not `HashMap`

src/main/java/com/bugsnag/android/BreadcrumbType.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    USER ("user");
+
+    private final String type;
+
+    BreadcrumbType(String type) {
+        this.type = type;
+    }
+
+    String serialize() { return type; }
```
JakeWharton: How about making this the implementation of `toString()`?

kattrali: The compiler wasn't a fan, though maybe I missed something:

```
toString() in BreadcrumbType cannot override toString() in Enum
```

foygl: Did you make it public? If you override toString you need to make it public otherwise it will be trying to reduce the access privileges of an inherited method which is not allowed.

kattrali: That makes sense and would've made a good error message. Patched in b63033eec62a644aac9e7fd07a8e9577e6af16ac. Non-breaking change since it was package scoped before.

PR #112, opened 2016-07-25
Add getContext() for logging purposes

nicous:
Its a simple getter that is making me handle this on my application.


kattrali:
Hi @nicous, thank you for the contribution. I think this change looks good. 👍  Merging.

PR #119, opened 2016-09-06
Add cpu abi to device information

Pezzah:
This PR adds the CPU/ABI information to the device information as suggested in https://github.com/bugsnag/bugsnag-android/pull/109

This is done slightly differently to ensure that the tests pass on android v4

### DIFF
--- a/src/main/java/com/bugsnag/android/DeviceData.java
+++ b/src/main/java/com/bugsnag/android/DeviceData.java
@@ -31,7 +31,7 @@ class DeviceData implements JsonStream.Streamable {
     private final Boolean rooted;
     private final String locale;
     private final String id;
-    private final String cpuAbi;
+    private final String[] cpuAbi;
 
     DeviceData(@NonNull Context appContext) {
         screenDensity = getScreenDensity(appContext);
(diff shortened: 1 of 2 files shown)


src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -161,4 +167,40 @@ private static String getAndroidId(Context appContext) {
...
+    private static String getCpuAbi() {
+        String cpuAbi = Build.CPU_ABI;
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            cpuAbi = SupportedAbiWrapper.getSupportedAbis();
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.FROYO) {
+            cpuAbi = Abi2Wrapper.getAbi1andAbi2();
+        }
+
+        return cpuAbi;
```
JakeWharton: This method can be simplified to
```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
  return SupportedAbiWrapper.getSupportedAbis();
}
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.FROYO) {
  return Abi2Wrapper.getAbi1andAbi2();
}
return Build.CPU_ABI;
```

Also is it worth wrapping `Build.CPU_ABI` in an array so that the output format is always consistent regardless of API level?

Issue #71, opened 2015-07-23
setUser() not persisted across sessions.


GuillaumeDelente:
Shouldn't user details be persisted across app sessions?
Right now I have to call `setUser()` at every app start, it would be more convenient if user info were persisted unless manually reset.
This is not specified in the doc either, I had to manually test the behavior in order to find why most of my users were not having a userId set in the crash reports.
Lastly, all the analytics tools I've used so far (segment, google analytics, flurry and mixpanel) persist user info across sessions.

PR #120, opened 2016-09-07
Save user information between sessions

Pezzah:
We have an open issue around persisting user information between app sessions (see https://github.com/bugsnag/bugsnag-android/issues/71) 

This PR stores the user information in an android SharedPreferences file so it can be reloaded the next time the Bugsnag client is created.

- We may want to have an option to turn this on and have it off by default?
- Do we need to think about security on the SharedPreferences file?

### DIFF
--- a/src/androidTest/java/com/bugsnag/android/ClientTest.java
+++ b/src/androidTest/java/com/bugsnag/android/ClientTest.java
@@ -59,7 +59,9 @@ public class ClientTest extends BugsnagTestCase {
             .putString("user.name", "Mr Test")
             .commit();
 
-        Client client = new Client(getContext(), "api-key");
+        Configuration config = new Configuration("api-key");
+        config.setPersistUserBetweenSessions(true);
+        Client client = new Client(getContext(), config);
         final User user = new User();
 
         client.beforeNotify(new BeforeNotify() {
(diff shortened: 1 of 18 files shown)


src/androidTest/java/com/bugsnag/android/ClientTest.java
```diff
@@ -20,4 +23,61 @@ public void testNotify() {
...
+
+        client.notify(new RuntimeException("Testing"));
+
+        // Check the user details have been set
+        assertEquals("123456", user.getId());
+        assertEquals("mr.test@email.com", user.getEmail());
+        assertEquals("Mr Test", user.getName());
+
+        // Tidy up
+        client.clearUser();
```
kattrali: This probably belongs in the `tearDown()` for this class

src/androidTest/java/com/bugsnag/android/ClientTest.java
```diff
@@ -20,4 +23,61 @@ public void testNotify() {
...
+        // Check that the user was store in prefs
+        SharedPreferences sharedPref = getContext().getSharedPreferences("com.bugsnag.android", Context.MODE_PRIVATE);
+        assertEquals("123456", sharedPref.getString("user.id", null));
+        assertEquals("mr.test@email.com", sharedPref.getString("user.email", null));
+        assertEquals("Mr Test", sharedPref.getString("user.name", null));
+
+        // Tidy up
+        client.clearUser();
+    }
+
```
kattrali: There should probably be a test for `clearUser()` as well.

kattrali:
Having an option to enable this doesn’t seem like a bad idea.

PR #123, opened 2016-09-15
Adds support for sending reports with callbacks

kattrali:
This changeset adds support for modifying reports in callbacks/lambdas prior to being sent to Bugsnag. Example:

```java
Bugsnag.notify(new RuntimeException("Unexpected failure"), new Callback() {
    @Override
    public void beforeNotify(Report report) {
        report.getError().setSeverity(Severity.INFO);
        report.getError().setContext("FOO”);
    }
});
```

or using Java 8 lambdas/[retrolambda](https://github.com/evant/gradle-retrolambda):

```
Bugsnag.notify(new RuntimeException("UnknownLambdaFailure"), (report) -> {
    report.getError().setSeverity(Severity.ERROR);
});
```

## Questions/Caveats

The existing permutations of `notify()` with metadata, severity, and context were deprecated as a part of this changeset, though maybe there’s still a place for them?


src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -289,22 +289,64 @@ public static void notify(final Throwable exception) {
...
+    }
+
+    /**
+     * Notify Bugsnag of a handled exception
+     *
      * @param  exception  the exception to send to Bugsnag
      * @param  severity   the severity of the error, one of Severity.ERROR,
      *                    Severity.WARNING or Severity.INFO
+     *
+     * @deprecated Use {@link notify(Throwable,Callback)}
```
Pezzah: I think these deprecated links should be #notify(Throwable, Callback) to get them to work properly in Android Studio, same in Client.java

src/main/java/com/bugsnag/android/Callback.java
```diff
@@ -0,0 +1,7 @@
+package com.bugsnag.android;
+
+
+public interface Callback {
```
Pezzah: This could do with some comments

kattrali:
> I'm not sure if Java 8 is supported by Android yet, it's not possible for me to select as a language version in Android Studio

I set the version in build.gradle and Android Studio picked it up. Then I used callbacks using gradle-retrolambda

JakeWharton:
Also supported natively by the Jack compiler.

src/main/java/com/bugsnag/android/Report.java
```diff
@@ -54,16 +61,23 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
-        this.errors.add(error);
+    public Error getError() {
+        return error;
+    }
+
+    public void setApiKey(@NonNull String apiKey) {
+        this.apiKey = apiKey;
+    }
+
+    public void setNotifierVersion(@NonNull String version) {
```
kattrali: Yeah, as an integration point its useful

src/main/java/com/bugsnag/android/Bugsnag.java
kattrali: Ah, fixed.

kattrali:
> so I think there might be a place for some of the helper methods still.

Yeah, that’s a good point. I pulled the `notify(Throwable,[metadata or severity])` methods out of the icebox.

src/main/java/com/bugsnag/android/Bugsnag.java
loopj: FYI I kept `notify(Throwable, Severity)` as an option on my rewrite of bugsnag-java for ease of use:
https://github.com/bugsnag/bugsnag-java/blob/next/src/main/java/com/bugsnag/Client.java#L262

PR #134, opened 2016-11-21
Add automatic breadcrumbs

Pezzah:
Added some code to automaitcally  create breadcrumbs by using a `BroadcastReceiver` to listen for events in the system.

I've added all broadcast actions and categories from API 9 onwards (apart from time tick and battery changed, which seem too noisy).


kattrali:
Thanks for working on this, @Pezzah. 

The approach makes sense, though I have a few concerns with how all events are reported the same way. Some events seem to better match `STATE` type rather than log, as they reflect changes in the overall state of the app or device. Others seem to be more suited to `USER` as they are user actions, et cetera. I'm not sure though how to best structure this information, but it may not align well to the current layout of having the actions/categories separated by API level.

Do you have some examples of the kind/quantity of breadcrumbs attached using these settings?

Pezzah:
I have refactored this to allow for specifying different breadcrumb types for each action. I've assigned a few to different types, but I think that there are a lot of grey areas around which ones should map to what.

In a quick test I get the following in the simulator
screen shot 2017-01-20 at 11 20 59

and this using an HTC Desire S
screen shot 2017-01-20 at 11 25 35

fcduarte:
Automatic breadcrumbs would be an awesome feature .. Thinking ahead, adding the activities or fragments automatically would be killer! @kattrali @Pezzah any plans to move this PR forward? Thanks!

PR #135, opened 2016-12-02
Updates to allow Android NDK plugin to work

Pezzah:
We are working towards allowing notifies from Android NDK. This will be done using a separate library that depends on the bugsnag-android library.

This PR adds ways for the new NDK library to read and write data from this library:
 - Adds a `NativeInterface` class, which is used as the main access point for JNI
 - Makes the client class `Observable`, and raises `notifyObservers()` when data gets changed

### DIFF
--- a/src/main/java/com/bugsnag/android/Bugsnag.java
+++ b/src/main/java/com/bugsnag/android/Bugsnag.java
@@ -25,7 +25,7 @@ public final class Bugsnag {
      */
     public static Client init(Context androidContext) {
         client = new Client(androidContext);
-        configureClientObservers();
+        NativeInterface.configureClientObservers(client);
         return client;
     }
 
@@ -37,7 +37,7 @@ public final class Bugsnag {
      */
     public static Client init(Context androidContext, String apiKey) {
         client = new Client(androidContext, apiKey);
-        configureClientObservers();
+        NativeInterface.configureClientObservers(client);
         return client;
     }
 
@@ -50,7 +50,7 @@ public final class Bugsnag {
      */
     public static Client init(Context androidContext, String apiKey, boolean enableExceptionHandler) {
         client = new Client(androidContext, apiKey, enableExceptionHandler);
-        configureClientObservers();
+        NativeInterface.configureClientObservers(client);
         return client;
     }
 
@@ -62,30 +62,8 @@ public final class Bugsnag {
      */
     public static Client init(Context androidContext, Configuration config) {
         client = new Client(androidContext, config);
-        configureClientObservers();
+        NativeInterface.configureClientObservers(client);
         return client;
-    }
-
-    private static void configureClientObservers() {
-
-        // Ensure that the bugsnag observer is registered
-        // Should only happen if the NDK library is present
-        try {
-            String className = "com.bugsnag.android.ndk.BugsnagObserver";
-            Class c = Class.forName(className);
-            Observer o = (Observer)c.newInstance();
-            client.addObserver(o);
-        } catch (ClassNotFoundException e) {
-            // ignore this one, will happen if the NDK plugin is not present
-            Logger.warn("Failed to find NDK observer");
-        } catch (InstantiationException e) {
-            Logger.warn("Failed to instantiate NDK observer", e);
-        } catch (IllegalAccessException e) {
-            Logger.warn("Could not access NDK observer", e);
-        }
-
-        // Should make NDK components configure
-        client.notifyBugsnagObservers(NotifyType.ALL);
     }
 
     /**
--- a/src/main/java/com/bugsnag/android/Bugsnag.java
+++ b/src/main/java/com/bugsnag/android/Bugsnag.java
@@ -1,6 +1,7 @@
 package com.bugsnag.android;
 
 import java.util.Map;
+import java.util.Observer;
 
 import android.content.Context;
 
@@ -14,7 +15,7 @@ import android.content.Context;
  * @see Client
  */
 public final class Bugsnag {
-    private static Client client;
+    static Client client;
     private Bugsnag() {}
 
     /**
@@ -24,6 +25,7 @@ public final class Bugsnag {
      */
     public static Client init(Context androidContext) {
         client = new Client(androidContext);
+        configureClientObservers();
         return client;
     }
 
@@ -35,6 +37,7 @@ public final class Bugsnag {
      */
     public static Client init(Context androidContext, String apiKey) {
         client = new Client(androidContext, apiKey);
+        configureClientObservers();
         return client;
     }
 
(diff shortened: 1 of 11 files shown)


src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -58,9 +62,32 @@ public static Client init(Context androidContext, String apiKey, boolean enableE
...
+        // Ensure that the bugsnag observer is registered
+        // Should only happen if the NDK library is present
+        try {
+            String className = "com.bugsnag.android.ndk.BugsnagObserver";
+            Class c = Class.forName(className);
+            Observer o = (Observer)c.newInstance();
+            client.addObserver(o);
+        } catch (ClassNotFoundException e) {
+            // ignore this one, will happen if the NDK plugin is not present
+            Logger.warn("Failed to find NDK observer");
```
kattrali: I think the log level here should be lower, since it would be confusing for non-NDK consumers to see this.

src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -24,6 +25,7 @@ private Bugsnag() {}
      */
     public static Client init(Context androidContext) {
         client = new Client(androidContext);
+        configureClientObservers();
```
loopj: Could (should?) this be moved into the constructor for the `Client` object itself? As far as I remember, a supported way of using Bugsnag on Android is by instantiating a `Client` object directly and not using the `Bugsnag` static class.

src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -386,4 +394,11 @@ protected boolean inProject(String className) {
 
         return false;
     }
+
+    private void notifyObservers(NotifyType type) {
+        if (Bugsnag.client != null) {
```
loopj: This assumes that people are using the static `Bugsnag` class, when they are possibly using a `Client` instance instead. Each `Client` instance will have their own instance of `Configuration`.

src/main/java/com/bugsnag/android/MetaData.java
```diff
@@ -56,13 +58,34 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
         Map<String, Object> tab = getTab(tabName);
 
         if(value != null) {
             tab.put(key, value);
         } else {
             tab.remove(key);
         }
+
+        if (notify && Bugsnag.client != null) {
+            Bugsnag.getClient().notifyBugsnagObservers(NotifyType.META);
```
loopj: Same as above

src/main/java/com/bugsnag/android/MetaData.java
```diff
@@ -72,6 +95,10 @@ public void addToTab(String tabName, String key, Object value) {
      */
     public void clearTab(String tabName) {
         store.remove(tabName);
+
+        if (Bugsnag.client != null) {
+            Bugsnag.getClient().notifyBugsnagObservers(NotifyType.META);
```
loopj: Same as above

src/main/java/com/bugsnag/android/MetaData.java
```diff
@@ -87,17 +114,29 @@ public void clearTab(String tabName) {
 
     void setFilters(String... filters) {
         this.filters = filters;
+
+        if (Bugsnag.client != null) {
+            Bugsnag.getClient().notifyBugsnagObservers(NotifyType.FILTERS);
```
loopj: Same as above

src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Used as the entry point for native code to allow proguard to obfuscate other areas if needed
+ */
+class NativeInterface {
+
+    public static String getContext() {
+        return Bugsnag.getClient().getContext();
```
loopj: Ah - guessing that we are using static `Bugsnag` everywhere to make this whole interface possible. Any workarounds you can think of that would avoid depending on the static class?

src/main/java/com/bugsnag/android/Bugsnag.java
JakeWharton: We use Bugsnag entirely without the `Bugsnag` class so :+1: to this and your greater thoughts about not coupling any functionality to this class.

Pezzah: This can't go in the constructor of the client object because until the end of the constructor there is no client object to add observers to. I think there needs to be another call added if the client object is being used without the Bugsnag class

src/main/java/com/bugsnag/android/Configuration.java
Pezzah: Replicating the observer pattern to make the client observe its config

src/main/java/com/bugsnag/android/MetaData.java
Pezzah: Replicating the Observer pattern to make the config observe its meta data

src/main/java/com/bugsnag/android/MetaData.java
Pezzah: Replicating the Observer pattern to make the config observe its meta data

src/main/java/com/bugsnag/android/NativeInterface.java
Pezzah: Refactored code to try and allow this to work with client objects without the Bugsnag class

Pezzah:
I've refactored this to allow the Client to be used without the Bugsnag class. This needs an extra step to be performed when defining the client:

`NativeInterface.setClient(client);`

This is needed because a client object needs to exist before observers can be added to it, and so that only one client gets used for the native interface. This should also allow the for the scenario where multiple clients are being used by getting the user to specify which one to use for native errors.

There is an outstanding problem with this pattern, which also applies to the existing plugin: If errors get saved to the disk cache (as they do for NDK) then I think they will get sent using which ever client gets constructed first when the application is re-launched. Which means errors that were meant for different projects will all get sent to the same project (or potentially both projects if the files aren't deleted by the time the subsequent clients are constructed).

We could maybe start storing these errors in folders named using the API key to differentiate them

loopj:
Thanks @Pezzah I think this solves my concerns.

Very good point about the fact that multiple clients all store data on-disk in the same place - this is something we might want to address in a separate PR.

Both API key and endpoint would need to be encoded in the file/folder structure, although it is _super_ unlikely that someone would send to two different endpoints from one app.

snmaynard:
I would say so unlikely as to not be worth considering.

kattrali:
> Both API key and endpoint would need to be encoded in the file/folder structure, although it is _super_ unlikely that someone would send to two different endpoints from one app.

Having multiple endpoints seems incredibly unlikely, but for the API key we could potentially add API key as a field on `Error`, and extract it from there at send time. This does not seem pressing though.

PR #137, opened 2016-12-13
Notify blocking on uncaught exception

### LINKED ISSUE #136: Crashes not reported by bugsnag
After application crash if user immediately click "OK" on dialog (having message "Unfortunately, 'Application Name' has stopped) than crash is not getting reported. Device details on which this issue occurs are below:

Manufacturer: Samsung
Model: SM-G313HZ
OS Version- 4.4.2

snmaynard:
Fixes #136. It seems like we should always block on uncaught exception. This is untested for now.


Pezzah:
I was worried about what might happen if the application got stuck in some kind of crash loop, could this just keep sending errors to Bugsnag and prevent the application from closing?

snmaynard:
It should block the thread at that point and then terminate when the uncaught exception handler returns

saurabh0802chd:
any ETA on release of this PR?

PR #141, opened 2017-02-11
Try to avoid using Settings.Secure.Android_ID

### LINKED ISSUE #83: Use Advertising ID instead of Android ID.
https://github.com/bugsnag/bugsnag-android/blob/master/src/main/java/com/bugsnag/android/DeviceData.java#L140

d4rken:
Closes #83.

In some time one could think removing the legacy mode as I doubt there is merit tieing bug reports toegether that are month apart and at that point the Android_ID is stored in preferences and wouldn't need to be accessed directly anymore (making "security" review tools happy that `Settings.Secure.Android_ID` is not accessed).

Testing lacks a bit, maybe at some point the library could be refactored and moved away from instrumentation tests...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,11 @@ android {
 }
 
 dependencies {
+    repositories {
+        jcenter()
+    }
     compile "com.android.support:support-annotations:25.1.1"
+    androidTestCompile "junit:junit:4.12"
 }
 
 // Disable doclint:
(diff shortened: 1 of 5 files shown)


d4rken:
Merge #140 to fix the build issue.

d4rken:
Retrigger travis.

d4rken:
Can anyone tell me why this is failing? I don't see any minimum API issue in the code.

PR #142, opened 2017-02-11
Detect systemless root

d4rken:
`/su/bin` should only exist on rooted device.

### DIFF
--- a/src/main/java/com/bugsnag/android/DeviceData.java
+++ b/src/main/java/com/bugsnag/android/DeviceData.java
@@ -12,7 +12,6 @@ import android.util.DisplayMetrics;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.Locale;
 
 /**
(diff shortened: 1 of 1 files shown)


d4rken:
Merge #140 to fix the build issue.

kattrali:
Thanks, @d4rken! I merged manually to pull out the unrelated changes, but this looks good. 👍

d4rken:
@kattrali Are the other changes against the prefered code style or what's the issue with them? I made those on purpose (wasn't an auto formatter accident).

kattrali:
@d4rken Ah, gotcha. I assumed it was from a formatter. From a maintenance perspective, its hard to look back through commits when a single change can have different purposes. I spend a fair amount of time looking back through rationale for decisions in git history, so generally try to not mix formatting changes in with hot fixes. Outside of that,the changes were a bit different from the existing style, so I the purpose of the change wasn't obvious during review.

PR #150, opened 2017-07-04
Replace android ID with per-install random UUID

fractalwrench:
This replaces the use of Settings.Secure.ANDROID_ID with a random UUID, which is generated on each install of the app.


ivanick87:
Nice

kattrali:
:shipit:

PR #151, opened 2017-07-04
Queue errors to disk

fractalwrench:
Queues errors to disk if the max pool size of the Executor is exceeded.

### DIFF
--- a/src/androidTest/java/com/bugsnag/android/ClientRejectedExecutionTest.java
+++ b/src/androidTest/java/com/bugsnag/android/ClientRejectedExecutionTest.java
@@ -1,18 +1,68 @@
 package com.bugsnag.android;
 
+import org.json.JSONObject;
+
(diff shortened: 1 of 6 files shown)


src/androidTest/java/com/bugsnag/android/ClientRejectedExecutionTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Checks that an exception is not thrown when the max task queue is reached.
+     */
+    public void testRejectedExecution() {
+        Client client = new Client(getContext(), "api-key");
+
+        for (int k = 0; k < MAX_ALLOWED_TASKS * 2; k++) {
+            client.notify(new Throwable());
+        }
```
kattrali: Can we test that after some time, reports are written to disk?

PR #160, opened 2017-07-11
Setup lintoptions with baseline file

fractalwrench:
The CI will now run lint and abort on any errors. A baseline file has been setup that contains pre-existing issues, which can be addressed in the future. The following inspections will be run: http://tools.android.com/tips/lint-checks


kattrali:
This looks good. Do you have any thoughts on whether there should be some alternative storage for a global client (to avoid the issues with leaking `Context`)?

run_lint.sh
```diff
@@ -0,0 +1 @@
+./gradlew lint
```
kattrali: What's this file for?

fractalwrench: This will run the lint inspection locally on the project - we can probably get rid of this or add a note in the Contributing docs?

fractalwrench:
I believe the warning for the static field context leak in Client is just Lint being overzealous, so can be suppressed. The Client constructor calls `context.getApplicationContext()`, so the field stores the global application lifecycle context, which will would always remain in memory anyway until the application process terminates. The inspection doesn't distinguish between Activity/View and application contexts so can lead to false positives like this.

run_lint.sh
kattrali: I think a note in contributing is fine

fractalwrench:
I've updated the PR by addressing all the items in the baseline file, and updating the Contributing docs.

PR #161, opened 2017-07-11
Support custom http stack

fractalwrench:
Updates the current HttpClient implementation to implement an ErrorReportApiInterface. The Bugsnag class has a new public method in its API to set a custom implementation of this interface. This should allow for custom requirements such as certificate pinning to be achieved.


snmaynard:
Are we planning on adding docs for how to do this?

src/androidTest/java/com/bugsnag/android/ErrorReportApiClientTest.java
```diff
@@ -0,0 +1,28 @@
...
+            Bugsnag.setErrorReportApiClient(null);
+            fail("ErrorReportApiClient cannot be null");
+        }
+        catch (Exception ignored) {
+        }
+    }
+
+    public void testBugsnagClient() {
+        Bugsnag.init(getContext(), "123");
+        Bugsnag.setErrorReportApiClient(errorReportApiClient);
```
kattrali: Should also test that its `postReport` method is called after a `notify`

kattrali:
@snmaynard I hope so :)

kattrali:
@fractalwrench Looks good, left a couple of comments.

fractalwrench:
@kattrali I've updated this PR according to the review comments, and created a PR on the Docs repo.

kattrali:
🎉

PR #163, opened 2017-07-12
Setup strictmode notification

fractalwrench:
@kattrali I've investigated the following options for StrictMode and thought it'd be good to discuss what the next step should be.

1. DropboxManager - wasn't feasible as this required the READ_LOGS permission, which is no longer granted to 3rd party apps.
2. Using the Logcat stacktrace by specifying StrictMode.penaltyLog(). There doesn't appear to be any way of reliably intercepting this info.
3. Using StrictMode.penaltyDeath(), which throws an Exception on policy violations. Ultimately this gets propagated up to the DefaultUncaughtExceptionHandler, where we can detect whether the Throwable was caused by a StrictMode violation.

The problem with penaltyDeath() is that it will terminate the thread when a violation occurs, so if we enable inspections by default, it's probable there would be a few false positives from the Android framework/3rd party library code, causing crashes. I suppose we would need some sort of API that allows developers to specify what policies are enabled/disabled.

I think it should be possible to parse the StrictMode exception message for the Policy Violation Type (e.g. File Write, Network on Main Thread). We could then try passing that up in the MetaData when sending the Error Report?


kattrali:
> I think it should be possible to parse the StrictMode exception message for the Policy Violation Type (e.g. File Write, Network on Main Thread). We could then try passing that up in the MetaData when sending the Error Report?

That makes sense to me. Is the formatting consistent?

fractalwrench:
@kattrali it looks like the policy violation codes and exception message format have remained constant since the StrictMode class was first publicly released. Currently the code is setup to include a human-readable description of the policy violation in the Report's MetaData, so if it does change in future that should just default to null.

I think it would be best to document that the SDK records this extra info for debugging, and recommend that developers setup StrictMode in the same way as they would for any other app. The alternative is to enable this behaviour by default, but this could quite easily result in apps crashing due to bugs in 3rd party code/Android framework code that can't be resolved.

kattrali:
> I think it would be best to document that the SDK records this extra info for debugging, and recommend that developers setup StrictMode in the same way as they would for any other app

Agreed. This sounds like it would provide the most value while doing what users expect.

fractalwrench:
This has been added to a PR in the docs repo which has been approved.

src/main/java/com/bugsnag/android/ExceptionHandler.java
```diff
@@ -47,18 +51,30 @@ public ExceptionHandler(UncaughtExceptionHandler originalHandler) {
...
+        for (Client client : clientMap.keySet()) {
+
+            if (strictModeThrowable) { // add strictmode policy violation to metadata
+                String violationDesc = strictModeHandler.getViolationDescription(e.getMessage());
+                MetaData metaData = new MetaData(
+                    Collections.<String, Object>singletonMap(KEY_STRICT_MODE_VIOLATION, violationDesc));
+                client.cacheAndNotify(e, Severity.ERROR, metaData);
+            }
+
+            client.cacheAndNotify(e, Severity.ERROR, null);
```
kattrali: I noticed that reports are being sent in duplicate, one with and one without the strictmode metadata. I assume there should be an `else` clause wrapping this line.

src/main/java/com/bugsnag/android/ExceptionHandler.java
```diff
@@ -47,18 +51,30 @@ public ExceptionHandler(UncaughtExceptionHandler originalHandler) {
...
+      
         // Notify any subscribed clients of the uncaught exception
-        for(Client client : clientMap.keySet()) {
-            client.cacheAndNotify(e, Severity.ERROR);
+        for (Client client : clientMap.keySet()) {
+
+            if (strictModeThrowable) { // add strictmode policy violation to metadata
+                String violationDesc = strictModeHandler.getViolationDescription(e.getMessage());
+                MetaData metaData = new MetaData(
+                    Collections.<String, Object>singletonMap(KEY_STRICT_MODE_VIOLATION, violationDesc));
```
kattrali: This invocation does not specify a tab for the metadata, so its appearing on its own "custom" tab. Since there might be other strictmode-related features we might want to surface in the future, we can instead put this on its own "StrictMode" tab with something like (assuming the appropriate constants for the tab and key names):

```diff
- MetaData metaData = new MetaData(
-     Collections.<String, Object>singletonMap(KEY_STRICT_MODE_VIOLATION, violationDesc));
+ MetaData metaData = new MetaData();
+ metaData.addToTab("StrictMode", "Violation", violationDesc);
```

fractalwrench:
Thanks @kattrali  - I've fixed the missing else clause and moved over to using Tabs in the metadata.

fractalwrench:
Also got rid of some obsolete version checks.

PR #166, opened 2017-07-24
Add Travis-CI deploy action for Javadocs

fractalwrench:
This creates a Travis-CI deploy action that generates Javadocs and uploads them to Github Pages, whenever a change is pushed to the master branch.


kattrali:
Hmm. I don't think we should check the docs in, instead we can run the `javadoc` command in a `before_deploy` step and deploy the docs dir contents generated by that action. It might get a bit noisy otherwise.

fractalwrench:
I've added `docs/` to the gitignore and updated the travis YML accordingly.

.travis.yml
```diff
@@ -32,3 +32,13 @@ before_script:
...
+
+before_deploy: ./gradlew javadoc
+
+deploy:
+  provider: pages
+  local_dir: docs # only include the contents of the docs dir
+  skip_cleanup: true
+  github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
+  on:
+    branch: master # deploy when changes pushed to this branch
```
kattrali: This should be on tags instead of master to ensure only released content is being documented

fractalwrench: This will now only deploy Javadoc when a tag is applied to a commit.

PR #168, opened 2017-07-24
Disable logs on release builds

fractalwrench:


src/main/java/com/bugsnag/android/Logger.java
```diff
@@ -4,16 +4,28 @@
 
 class Logger {
     private static final String LOG_TAG = "Bugsnag";
+    private static boolean enabled = true;
```
JakeWharton: This needs to be `volatile` or other threads might not see updates.

duncanhewett:
Would it make sense to disable logging when the release stage is `production` by default and allow this behaviour to be overridden? This would provide the recommended behaviour without configuration.

JakeWharton:
Pointing at production doesn't mean you're in a release build.

duncanhewett:
True, but the default behaviour of the library is to set the release stage to `production` if the build is not debuggable (specifically [ApplicationInfo.FLAG_DEBUGGABLE is not set](https://github.com/bugsnag/bugsnag-android/blob/84a7e33054e6a57e94b4cf5e3190de89230e7236/src/main/java/com/bugsnag/android/AppData.java#L139])) so defaulting to logging being disabled in that situation could be reasonable?

JakeWharton:
Yep, seems reasonable.

PR #170, opened 2017-07-25
Add Kotlin example app

fractalwrench:
Creates a sample app written in Kotlin. This application is essentially a translation of the Java example - we may want some sort of indicator that the two apps are different (e.g. different icon, color scheme).

I've also setup the CI server to build both the Java/Kotlin example modules and check that they launch.


kattrali:
Looks good. The only thing I might change would be to put the examples into directories within `example`, e.g. `example/java` and `example/kotlin`. That way, there isn't much clutter at the top level, and it mirrors similar situations on other platforms.

fractalwrench:
I've converted the project structure so that the *example* module contains separate productFlavors for Java and Kotlin.

PR #173, opened 2017-08-02
Setup client to send stored reports on disk when connectivity regained

fractalwrench:
Registers a broadcast receiver which detects when network becomes available, and flushes any reports saved to disk.

### DIFF
--- a/src/main/java/com/bugsnag/android/Client.java
+++ b/src/main/java/com/bugsnag/android/Client.java
@@ -1,10 +1,15 @@
 package com.bugsnag.android;
 
 import android.app.Application;
+import android.content.BroadcastReceiver;
 import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
 import android.content.SharedPreferences;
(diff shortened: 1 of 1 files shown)


src/main/java/com/bugsnag/android/Client.java
```diff
@@ -166,14 +171,29 @@ public Client(@NonNull Context androidContext, @NonNull Configuration configurat
...
+
+    private class ConnectivityChangeReceiver extends BroadcastReceiver {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            ConnectivityManager cm = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+            NetworkInfo networkInfo = cm.getActiveNetworkInfo();
+            boolean retryReports = networkInfo != null && networkInfo.isConnectedOrConnecting();
+
+            if (retryReports) {
+                errorStore.flush(errorReportApiClient);
```
kattrali: Is there any risk here for duplicate reports if the error store is being flushed when this is called?

PR #174, opened 2017-08-07
Update README.md

fractalwrench:
product variant -> build variant


kattrali:
Haha I just keep mixing the two!

PR #176, opened 2017-08-08
Shorten event receiver keys

fractalwrench:
Shortens the metadata key if it is originates from an Android package. This prevents the scenario where text such as `android.media.SCO_AUDIO_STATE_CHANGED` is truncated on the dashboard. Also sorts metadata alphabetically in the JSON payload for a more consistent display.


fractalwrench:
@kattrali thinking about it, we may also want to sort the metadata alphabetically on the dashboard as that would avoid having to update the other platforms.

PR #185, opened 2017-08-29
Send synchronous report when uncaught crash happens during app launch

fractalwrench:
Updates the notifier to send reports synchronously when a uncaught exception is thrown during app launch. By default, this is equal to 10,000ms, but is configurable via the Configuration object.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -26,7 +26,7 @@ android {
         targetSdkVersion Integer.parseInt(project.ANDROID_TARGET_SDK_VERSION)
 
         manifestPlaceholders = [
-            bugsnagApiKey: "f35a2472bd230ac0ab0f52715bbdc65d",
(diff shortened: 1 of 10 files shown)


fractalwrench:
Will also need to update the docs for this one if the provided API looks fine.

fractalwrench:
Updated slightly so that the report is sent synchronously, the next time the app is loaded. This also adds a tab to the MetaData so the user can see this on the dashboard.

I'm not 100% sure over whether 10,000ms is a sensible default for this, or whether we should even enable it by default as it could quite easily lead to ANR. It'd be good to get your thoughts on this @kattrali

kattrali:
This works as expected, though I agree that the threshold seems a bit long. Laybe the default threshold should be around 2-3 seconds?

PR #187, opened 2017-08-30
Update example code to use Android Plugin for Gradle 3.0.0

fractalwrench:
This updates our example code to use the 3.0.0 Android gradle build tools plugin. The major changes are:

- Each build variant requires a `flavorDimension`
- `implementation` is used instead of `compile` to speed up build times
- Gradle wrapper bumped to latest release

3.0.0 is currently in Beta, so there may be further changes. This is also a breaking change - I don't believe it would be recommended to compile this from Android Studio 2.X, although it's certainly possible from the command line.


fractalwrench:
The build is broken on Travis as it's still using Maven Central to resolve our gradle plugin. Installing the PR for our gradle plugin to a local maven repo should pass all the tests.

kattrali:
Should we update and merge this in? Things seem pretty stable now.

fractalwrench:
@kattrali probably worth waiting until Android Studio releases a stable version - the plugin is still in beta and is therefore only guaranteed work with a specific build of Android Studio.

PR #188, opened 2017-09-06
Distinguish between handled + unhandled events

fractalwrench:


sdk/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -302,4 +308,48 @@ void setBreadcrumbs(Breadcrumbs breadcrumbs) {
...
+            this.severity = severity;
+            return this;
+        }
+
+        Builder withMetaData(MetaData metaData) {
+            this.metaData = metaData;
+            return this;
+        }
+
+        Builder isUnhandled(boolean unhandled) {
```
kattrali: I'm uncertain about the name of this setter, since it doesn't imply that a value is being set. `withUnhandled`/`withUnhandledValue` are also odd grammatically though 🤔

fractalwrench: Yep I can see how this could be confusing. How about something like: `builder.severity(WARNING).unhandled(false)`

fractalwrench:
@kattrali this one is updated now for the API changes.

sdk/src/main/java/com/bugsnag/android/Severity.java
```diff
@@ -37,4 +37,8 @@ static Severity fromString(String s) {
                 return null;
         }
     }
+
+    public String getName() {
```
kattrali: Should this be `toString()`?

PR #189, opened 2017-09-06
Update to target sdk 26

fractalwrench:
Updates the SDK and example apps to target API Level 26. The SDK doesn't appear to require any dangerous permissions, and I didn't find anything in the Android 6/7/8 behaviour change notes which would obviously affect us.


JakeWharton:
Probably worth removing `targetSdkVersion` from the library. There's (almost) no need for libraries to specify a target SDK.

fractalwrench:
Good point - I'll remove it from the library module. Thanks!

PR #193, opened 2017-09-19
Create TEST_PLAN.md

fractalwrench:
Adds a test plan in a checklist format, which should be run through before and after making a release.

If everything is ok here, we can roll something similar out to the other platform repos.


TEST_PLAN.md
```diff
@@ -0,0 +1,13 @@
...
+## Pre-release
+1. Does the build pass on the CI server?
+1. On a fresh clone of the repository, do the example apps compile and send error reports?
+1. Has the changelog and any documentation been updated?
+1. Have all the version numbers been incremented?
+1. Has a release been created for this version on Github?
+
+## Post-release
+1. Have all Docs PRs been merged?
+1. Do the installation instructions work when creating an example app from scratch?
```
Cawllec: Should these last two be a pre-release test?

kattrali: Yeah. Everything should work before release, and the released version should be tested after imo.

kattrali:
@Cawllec @fractalwrench It might be worth merging this list into the release instructions, since it already mentions [pre-release testing](https://github.com/bugsnag/bugsnag-android/blob/master/CONTRIBUTING.md#2-prepare-for-release) (albeit in less detail). All of these steps (pre-release testing, release, post-release testing) should be done for every release.

fractalwrench:
@kattrali @Cawllec I've updated this so that the checklist is now integrated into the Contributing file.

Personally I think it would make sense to link to a central repo/gist for this release checklist, as the steps are fairly generic and apply to most notifiers. What are your thoughts?

kattrali:
I think the generic list in a document is fine, though I could see room for more specific items on various platforms, like testing changes on different browsers for JS or AppleTV and the mac for cocoa. In both cases, we want to make sure all of those checks happen every release.

CONTRIBUTING.md
```diff
@@ -76,6 +76,23 @@ device/emulator.
...
+- [ ] Has all new functionality been manually tested on a release build?
+- [ ] Do the installation instructions work when creating an example app from scratch?
+- [ ] Have the installation instructions been updated on the dashboard as well as the docs site?
+- [ ] Have all Docs PRs been merged?
+
+
+## Post-release Test Checklist
+- [ ] Do the installation instructions work using the released artefact?
+- [ ] Can a freshly created example app send an error report using the released artefact?
+
```
kattrali: Giving the post-release more thought, we should also run through existing example apps, right?

fractalwrench:
@kattrali I've added a link to a generic list, I can roll this out across other repos with any custom instructions needed.

PR #196, opened 2017-09-27
Allow changing breadcrumbs max size to less than current size

kattrali:


snmaynard:
Can we test setting it to zero disables breadcrumbs?

sdk/src/main/java/com/bugsnag/android/Breadcrumbs.java
```diff
@@ -101,9 +101,8 @@ void clear() {
...
 
     void setSize(int size) {
-        if (size > store.size()) {
-            this.maxSize = size;
-        } else {
+        this.maxSize = size;
+        if (size <= store.size()) {
             // Remove oldest breadcrumbs until reaching the required size
             while (store.size() > size) {
                 store.poll();
```
fractalwrench: This will loop forever if the user sets a negative size.

sdk/src/main/java/com/bugsnag/android/Breadcrumbs.java
```diff
@@ -101,9 +101,8 @@ void clear() {
     }
 
     void setSize(int size) {
-        if (size > store.size()) {
-            this.maxSize = size;
-        } else {
+        this.maxSize = size;
+        if (size <= store.size()) {
```
fractalwrench: We can get rid of this if statement as it's covered in the while loop.

PR #202, opened 2017-10-12
setup module for unity code

fractalwrench:


unity/src/main/java/com/bugsnag/android/unity/UnityCallback.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.bugsnag.android.Callback;
+import com.bugsnag.android.MetaData;
+import com.bugsnag.android.Report;
+
+
+class UnityCallback implements Callback {
+
+    static final String NOTIFIER_NAME = "Bugsnag Unity (Android)";
+    static final String NOTIFIER_VERSION = "3.5.1";
```
kattrali: Every time we need to release unity, we'd need to change bugsnag-android?

PR #203, opened 2017-10-23
Support Library variant mapping file upload

fractalwrench:
Adds a barebones example that will sync with the gradle plugin, and upload a mapping file. It would be good to get thoughts on whether we need to add any other functionality, such as an example app that will allow us to test that the library's mapping file deobfuscates stacktraces correctly.


fractalwrench:
The build won't pass for the `exampleLib` module as it requires the gradle plugin to be installed in a local maven repo. Everything else should be ready for review.

example/build.gradle
```diff
@@ -49,6 +49,25 @@ android {
         }
     }
 
+
+    // Disabled for now - https://github.com/bugsnag/bugsnag-android-gradle-plugin/issues/59
```
kattrali: Don't need this anymore, yeah?

PR #218, opened 2018-01-09
Session tracking amends

fractalwrench:
- Respect notifyReleaseStages by not tracking sessions if this config is set
- When autoCaptureSessions is set via config after initialisation, track an existing session if present (required for downstream notifiers)

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/SessionTracker.java
+++ b/sdk/src/main/java/com/bugsnag/android/SessionTracker.java
@@ -91,7 +91,7 @@ class SessionTracker implements Application.ActivityLifecycleCallbacks {
      */
     private void trackSessionIfNeeded(boolean autoCaptured, Session session) {
         String releaseStage = getReleaseStage();
-        boolean notifyForRelease = client != null && configuration.shouldNotifyForReleaseStage(releaseStage);
+        boolean notifyForRelease = configuration.shouldNotifyForReleaseStage(releaseStage);
 
         if ((configuration.shouldAutoCaptureSessions() || !autoCaptured) && notifyForRelease) {
             sessionQueue.add(session);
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -68,18 +69,50 @@
...
+        Session session = new Session(UUID.randomUUID().toString(), date, user);
+        session.setAutoCaptured(autoCaptured);
+        return session;
+    }
 
-            Session session = new Session(UUID.randomUUID().toString(), date, user);
-            session.setAutoCaptured(autoCaptured);
+    private void trackSessionIfNeeded(boolean autoCaptured, Session session) {
+        String releaseStage = getReleaseStage();
+        boolean notifyForRelease = client != null && configuration.shouldNotifyForReleaseStage(releaseStage);
```
Pezzah: why does this check if the client is not null here? shouldn't you always put a session in the queue/store even if the client is not initialized yet?

sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -68,18 +69,50 @@
...
+    private Session generateSession(@NonNull Date date, @Nullable User user, boolean autoCaptured) {
+        sessionStartMs = date.getTime();
+        Session session = new Session(UUID.randomUUID().toString(), date, user);
+        session.setAutoCaptured(autoCaptured);
+        return session;
+    }
 
-            Session session = new Session(UUID.randomUUID().toString(), date, user);
-            session.setAutoCaptured(autoCaptured);
+    private void trackSessionIfNeeded(boolean autoCaptured, Session session) {
```
Pezzah: could do with a comment on this method

fractalwrench:
Thanks @Pezzah - I've updated the PR with more docs, and removed the null client check (which was a typo).

PR #224, opened 2018-01-16
Parse manifest meta-data for auto capture sessions config

fractalwrench:


sdk/src/androidTest/java/com/bugsnag/android/ClientTest.java
```diff
@@ -197,6 +198,7 @@ public void testFullManifestConfig() {
         data.putBoolean(MF_SEND_THREADS, false);
         data.putBoolean(MF_ENABLE_EXCEPTION_HANDLER, false);
         data.putBoolean(MF_PERSIST_USER_BETWEEN_SESSIONS, true);
+        data.putBoolean(MF_AUTO_CAPTURE_SESSIONS, true);
```
kattrali: Looking at this test as a whole, the public interface is tested more thoroughly if raw string values are used here instead of the internal constants representing the manifest keys. In the case that a typo ends up in `MF_AUTO_CAPTURE_SESSIONS`, this test wouldn't catch it at all.

fractalwrench: Updated this test to use string literals instead.

PR #225, opened 2018-01-17
Enable test coverage with jacoco

fractalwrench:
Adds the `createDebugCoverageReport` task, which generates a Jacoco report under the build folder when run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
   - while [[ `adb shell pm path android` == 'Error'* ]]; do sleep 2; done
   - adb shell input keyevent 82 &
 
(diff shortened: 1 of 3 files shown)


kattrali:
Is there a way we can surface this during CI, or some service we can use for this? Though without checking I'm sure coveralls has some extension for Java/jacoco.

fractalwrench:
Codecov is one option: https://github.com/codecov/example-android

kattrali:
There is this: https://github.com/kt3k/coveralls-gradle-plugin

and we are already using coveralls on other platforms

kattrali:
Updated to get reports. [Here's an example](https://coveralls.io/builds/15171443)

PR #230, opened 2018-01-23
Prevent potential ANRs on latent networks

fractalwrench:


sdk/src/main/java/com/bugsnag/android/ErrorStore.java
```diff
@@ -42,22 +47,47 @@ public int compare(File lhs, File rhs) {
...
-                flushErrorReport(crashReport, errorReportApiClient);
+
+            // Block the main thread for a 2 second interval as the app may crash very soon.
+            // The request itself will run in a background thread and will continue after the 2
+            // second period until the request completes, or the app crashes.
+            completed = false;
+
+            HandlerThread handlerThread = new HandlerThread("Bugsnag Launch Crash Thread");
+            handlerThread.start();
+            Handler handler = new Handler(handlerThread.getLooper());
```
kattrali: Why is this a looper? As far as I can tell, it only runs once.

PR #233, opened 2018-01-25
Validate that report files are correct JSON

fractalwrench:


fractalwrench:
Preliminary validation that a File contains parseable JSON.

TODO:
- Add similar validation for Sessions stored in Files
- Update filename to contain a unique ID

Issue #221, opened 2018-01-10
Automatic Breadcrumbs and sensitive data


kenumir:
Automatic Breadcrumbs are grat but log user sensitive data like WiFi network name or incoming call phone numer if application has permissions.

It would be great if you could cut out this data - privacy policy.

fractalwrench:
Thanks for the suggestion @kenumir - this makes a lot of sense. We've recently added a [configuration option](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/Configuration.java#L466) to disable automatic breadcrumbs entirely.

We're also planning on adding an API which allows you to filter breadcrumbs, which would work similarly to how [beforeNotify does currently](https://docs.bugsnag.com/platforms/android/sdk/#attaching-custom-diagnostics).

PR #237, opened 2018-02-02
Add a callback for filtering breadcrumbs

fractalwrench:
This adds a callback which allows a developer to specify whether a breadcrumb should be captured or discarded. This also exposes the `Breadcrumb` class as part of our public API, and allows manipulation of the `MetaData`. Some duplication within the Breadcrumbs class has also been removed.

Addresses #221

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/BeforeBreadcrumb.java
+++ b/sdk/src/main/java/com/bugsnag/android/BeforeBreadcrumb.java
@@ -2,8 +2,31 @@ package com.bugsnag.android;
 
 import android.support.annotation.NonNull;
 
+/**
+ * Add a "before breadcrumb" callback, to execute code before every
+ * breadcrumb captured by Bugsnag.
+ * <p>
+ * You can use this to modify breadcrumbs before they are stored by Bugsnag.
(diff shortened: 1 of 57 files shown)


sdk/src/main/java/com/bugsnag/android/BeforeBreadcrumb.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Runs the "before breadcrumb" callback. If the callback returns
+     * <code>false</code> any further BeforeBreadcrumb callbacks will not be called
+     * and the breadcrumb will not be captured by Bugsnag.
+     *
+     * @param breadcrumb the breadcrumb to be captured by Bugsnag
+     * @see Breadcrumb
+     */
+    boolean send(@NonNull Breadcrumb breadcrumb);
```
bengourley: is `send` the best name for it? If I understand correctly, this method decides if it will "save" not "send" the breadcrumb. I'd call it something like `shouldSave(breadcrumb)`.

fractalwrench:
Updated names in response to feedback from Duncan

PR #243, opened 2018-02-07
Address further checkstyle warnings

fractalwrench:
Addresses further checkstyle warnings, including:

- Operator should be on newline
- Should only be 1 class per source file
- Empty catch clause
- Distance from variable declaration and use
- Max count of 1 for Abbreviation chars
- Suppressed all remaining warnings in sdk module

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -34,6 +34,7 @@ import java.util.concurrent.RejectedExecutionException;
  *
  * @see Bugsnag
  */
+@SuppressWarnings("checkstyle:JavadocTagContinuationIndentation")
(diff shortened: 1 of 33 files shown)


sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -404,6 +402,7 @@ public void setEndpoint(String endpoint) {
      *
      * @param buildUUID the buildUUID.
      */
+    @SuppressWarnings("checkstyle:AbbreviationAsWordInName")
```
Pezzah: The parameter to the function can change, even if the function name has to stay the same

fractalwrench: Thanks @Pezzah - I've updated the param name

PR #250, opened 2018-02-22
Add unusual crash scenarios

fractalwrench:
Adds more unusual crash scenarios:

- Empty StackTrace
- OutOfMemoryError
- StackOverflow
- Other crash handler installed


kattrali:
All of these are good candidates for full payload checks.

PR #251, opened 2018-02-22
Add scenarios for user enabled/disabled

fractalwrench:


fractalwrench:
Added additional permutations to the scenarios

PR #253, opened 2018-02-22
NotifyRelease stage scenarios

fractalwrench:
Adds scenarios for when an exception is captured inside + outside of a notifyReleaseStage


fractalwrench:
Added additional test cases.

PR #261, opened 2018-03-01
Add metadata scenario

fractalwrench:
Adds a scenario which adds metadata in a callback

### DIFF
--- a/features/meta_data_scenario.feature
+++ b/features/meta_data_scenario.feature
@@ -1,0 +1,7 @@
+Feature: Android support
+
+Scenario: Test handled Android Exception
+    When I run "MetaDataScenario" with the defaults
+    Then I should receive a request
+    And the request is a valid for the error reporting API
+    And the event "metaData.Custom.foo" equals "Hello World!"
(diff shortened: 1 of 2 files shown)


features/meta_data_scenario.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Android support
+
+Scenario: Test handled Android Exception
```
kattrali: The scenario name should state what's actually being tested here, e.g. 
"Add metadata in a notify() callback". It also reveals what other cases might be needed here, like changing or removing metadata.

features/meta_data_scenario.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Android support
```
kattrali: All of the handled scenarios should go into a single file and have "Sending handled exceptions" as the feature name.

fractalwrench:
Updated the scenario name. I feel it would be easier to move all the handled exceptions into one feature once they're merged into the [WIP PR](https://github.com/bugsnag/bugsnag-android/pull/246)

PR #263, opened 2018-03-01
add tests for setting metadata filters by default

fractalwrench:
MetaData filters are currently left as null unless they are explicitly set. This adds a fix and test coverage to prevent this scenario.

Additionally I've added coverage in 2 Maze Runner scenarios: https://github.com/bugsnag/bugsnag-android/pull/262/files


snmaynard:
Why not just set the default in the metadata class?

fractalwrench:
You're right, that's a much better way of doing it - I've updated the PR.

Issue #127, opened 2016-09-26
beforeNotify() gets called multiple times


jermainedilao:
beforeNotify() gets called multiple times. Is it on my side only?

kattrali:
Thanks for the report, @jermainedilao. Are you observing `beforeNotify` being called multiple times for a single error report? What version of bugsnag-android are you using?

jermainedilao:
Yes, for a single error report. However I didnt specify it here. I only
placed "com.bugsnag:bugsnag-android:+" could this be relevant to whats
happening? Thanks for the response.

PR #266, opened 2018-03-09
Prevent ConcurrentModificationException in Before notify/breadcrumb callbacks

fractalwrench:
Prevents a `ConcurrentModificationException` from occurring in the `BeforeNotify` and `BeforeRecordBreadcrumb` callbacks, if an element is added to the collection during or at the same time as `notify` is called.

A `ConcurrentLinkedQueue` is used in place of a `LinkedHashSet` as this retains the order of the callbacks. There's a bit of a tradeoff here with [#127](https://github.com/bugsnag/bugsnag-android/issues/127) - I've added a check to the client method to avoid adding duplicate callbacks, and assumed that anyone adding to the Collection directly knows what they're doing.

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/ConcurrentCallbackTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ConcurrentCallbackTest.java
@@ -1,5 +1,6 @@
 package com.bugsnag.android;
 
+import android.support.annotation.NonNull;
 import android.support.test.filters.SmallTest;
 import android.support.test.runner.AndroidJUnit4;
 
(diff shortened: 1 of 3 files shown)


sdk/src/androidTest/java/com/bugsnag/android/ConcurrentCallbackTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void testClientNotifyModification() throws Exception {
+        final Collection<BeforeNotify> beforeNotifyTasks = client.config.getBeforeNotifyTasks();
+        client.beforeNotify(new BeforeNotify() {
+            @Override
+            public boolean run(Error error) {
+                beforeNotifyTasks.add(new BeforeNotifySkeleton());
+                // modify the Set, when iterating to the next callback this should not crash
+                return true;
```
kattrali: This test fails without the change?

fractalwrench: Yes, as it would modify the contents of the `LinkedHashSet` when an iterator has already been implicitly created in [`runBeforeNotifyTasks`](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/Client.java#L1274)

PR #267, opened 2018-03-09
Setup checkstyle for test folder

fractalwrench:
Updates the project to run checkstyle on any java file under src. This PR also addresses a warning regarding the ordering of static imports


fractalwrench:
Note: this will fail for the testharness module, which is being superseded by mazerunner

Issue #265, opened 2018-03-05
Access user-id?


d4rken:
In some cases I would like users to be able to send me their user-id such that I can specifically look into their issues.

AFAIK the user-id is currently not accessible, is there a reason not to add a getter for this?

fractalwrench:
@d4rken you should be able to set the user ID via: `Bugsnag.setUser(final String id, final String email, final String name)`

This will attach a user ID to a report. Would you be able to explain the usecase on why a getter would be useful?

d4rken:
Setting the user-id would overwrite the current one that Bugsnag generates by default, right?

I don't want to change the user-id, just to make it visible in the app, the current user-id.

Setting my own user-id would not enable me to find and view reports that were already send with the old (bugsnag default) ID.

Looking at the code I'd just add `client.getUser()` and then either make the user fields `public final` or add getters.

Theoretically I think we should also unclutter the API a bit there and just add a `setUser(user)` and remove the many overlapping api methods, but that may be another issue/PR.

fractalwrench:
@d4rken that makes sense. I think uncluttering the API would be a good idea, maybe something like the following would make sense, and we could deprecate older methods:

`Bugsnag.setUser(new User("abc", null, null))`
`User user = Bugsnag.getUser()`

d4rken:
Yeah something like that. Need to check the init call chain a bit more...
I could make a PR sometime this week, should I?

For general API uncluttering a deprecation would probably not be sufficient, rather require a breaking change major release. Less setter methods, more builder patterns to modify "config" objects... etc. anyways, different ticket for that 😉

fractalwrench:
Sure :) if you're happy to make a PR then I can review it, otherwise we'll schedule this internally.

If you have other thoughts for decluttering the API I'd be quite interested to hear about them in a different ticket. There are a few things I'd like to do personally, such as reducing mutable state, and sharing some common code with our Java Notifier, but I'd be interested to hear an outside perspective on this.

d4rken:
> Sure :) if you're happy to make a PR then I can review it, otherwise we'll schedule this internally.

Done

> If you have other thoughts for decluttering the API I'd be quite interested to hear about them in a different ticket.

Nothing concrete, mostly related to pulling details into their own classes like I did in #268.
I've made a ticket, #269.

> There are a few things I'd like to do personally, such as reducing mutable state, and sharing some common code with our Java Notifier, but I'd be interested to hear an outside perspective on this.

[...]

PR #268, opened 2018-03-10
Refactoring `User´

d4rken:
* Introduced builder pattern for `User`, `User` is now immutable.
* Reduced duplicate/simplified code related to `User` and moved `User` related implementation details from `Client` to `User´
* Removed methods that were just used internally and migrated them to `Client.setUser(user)`
* Tagged/deprecated all public APIs that set specific `User` fields. In the next semvar major release these  could be removed.
* Added unit tests for the new code and adapted the existing tests such that both the new and the now deprecated methods are covered.

1. Cleaner API after the next major release removes the deprecated methods
2. Easier future SDK updates. Want to add new `User` information? Just update `User`.
3. `User` is now exposed via `Client.getUser()` which solves my issue in #265

@fractalwrench

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
@@ -1,0 +1,53 @@
+package com.bugsnag.android;
+
+import android.support.test.filters.SmallTest;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.UUID;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.core.Is.is;
+
+@RunWith(AndroidJUnit4.class)
+@SmallTest
+public class UserTest {
+    @Test
+    public void testBuilder_defaults() {
+        User user = User.builder().build();
+
+        assertThat(user.getId(), is(nullValue()));
+        assertThat(user.getEmail(), is(nullValue()));
+        assertThat(user.getName(), is(nullValue()));
+    }
+
+    @Test
+    public void testBuilder_values() {
+        final String id = UUID.randomUUID().toString();
+        final String email = UUID.randomUUID().toString();
+        final String name = UUID.randomUUID().toString();
+
+        User user = User.builder().name(name).id(id).email(email).build();
+
+        assertThat(user.getId(), is(id));
+        assertThat(user.getEmail(), is(email));
+        assertThat(user.getName(), is(name));
+    }
+
+    @Test
+    public void testBuilder_fromPrevious() {
+        final String id = UUID.randomUUID().toString();
+        final String email = UUID.randomUUID().toString();
+        final String name = UUID.randomUUID().toString();
+
+        User previous = User.builder().name(name).id(id).email(email).build();
+        User successor = User.builder(previous).build();
+
+        assertThat(successor.getId(), is(id));
+        assertThat(successor.getEmail(), is(email));
+        assertThat(successor.getName(), is(name));
+    }
+}
--- a/sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
@@ -1,5 +1,7 @@
 package com.bugsnag.android;
 
+import android.content.SharedPreferences;
+import android.support.test.InstrumentationRegistry;
 import android.support.test.filters.SmallTest;
 import android.support.test.runner.AndroidJUnit4;
 
@@ -8,7 +10,9 @@ import org.junit.runner.RunWith;
 
 import java.util.UUID;
 
+import static com.bugsnag.android.BugsnagTestUtils.getSharedPrefs;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.core.Is.is;
 
@@ -49,5 +53,29 @@ public class UserTest {
         assertThat(successor.getId(), is(id));
         assertThat(successor.getEmail(), is(email));
         assertThat(successor.getName(), is(name));
+    }
+
+    @Test
+    public void testRepo() {
+        SharedPreferences sharedPref = getSharedPrefs(InstrumentationRegistry.getContext());
+
+        User pre = User.builder()
+            .name(UUID.randomUUID().toString())
+            .id(UUID.randomUUID().toString())
+            .email(UUID.randomUUID().toString())
+            .build();
+
+        User.Repo repo = new User.Repo(sharedPref);
+        repo.set(pre);
+
+        User post = repo.get();
+        assertThat(post, is(not(nullValue())));
+
+        assertThat(post.getId(), is(pre.getId()));
+        assertThat(post.getEmail(), is(pre.getEmail()));
+        assertThat(post.getName(), is(pre.getName()));
+
+        repo.set(null);
+        assertThat(repo.get(), is(nullValue()));
     }
 }
--- a/sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
@@ -19,22 +19,14 @@ import static org.hamcrest.core.Is.is;
 @RunWith(AndroidJUnit4.class)
 @SmallTest
 public class UserTest {
-    @Test
-    public void testBuilder_defaults() {
-        User user = User.builder().build();
-
-        assertThat(user.getId(), is(nullValue()));
-        assertThat(user.getEmail(), is(nullValue()));
-        assertThat(user.getName(), is(nullValue()));
-    }
 
     @Test
-    public void testBuilder_values() {
+    public void testToBuilder() {
         final String id = UUID.randomUUID().toString();
         final String email = UUID.randomUUID().toString();
         final String name = UUID.randomUUID().toString();
 
-        User user = User.builder().name(name).id(id).email(email).build();
+        User user = new User.Builder().name(name).id(id).email(email).build().toBuilder().build();
 
         assertThat(user.getId(), is(id));
         assertThat(user.getEmail(), is(email));
@@ -42,13 +34,35 @@ public class UserTest {
     }
 
     @Test
-    public void testBuilder_fromPrevious() {
+    public void testBuilderDefaults() {
+        User user = new User.Builder().build();
+
+        assertThat(user.getId(), is(nullValue()));
+        assertThat(user.getEmail(), is(nullValue()));
+        assertThat(user.getName(), is(nullValue()));
+    }
+
+    @Test
+    public void testBuilderDefaultConstructor() {
         final String id = UUID.randomUUID().toString();
         final String email = UUID.randomUUID().toString();
         final String name = UUID.randomUUID().toString();
 
-        User previous = User.builder().name(name).id(id).email(email).build();
-        User successor = User.builder(previous).build();
+        User user = new User.Builder().name(name).id(id).email(email).build();
+
+        assertThat(user.getId(), is(id));
+        assertThat(user.getEmail(), is(email));
+        assertThat(user.getName(), is(name));
+    }
+
+    @Test
+    public void testBuilderUserArgument() {
+        final String id = UUID.randomUUID().toString();
+        final String email = UUID.randomUUID().toString();
+        final String name = UUID.randomUUID().toString();
+
+        User previous = new User.Builder().name(name).id(id).email(email).build();
+        User successor = new User.Builder(previous).build();
 
         assertThat(successor.getId(), is(id));
         assertThat(successor.getEmail(), is(email));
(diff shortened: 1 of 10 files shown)


d4rken:
* `testMultipleSessionFiles` passes locally, is this a flaky test?

fractalwrench:
Yep, it's a bit of a flaky test on Travis.

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -1,28 +1,23 @@
...
+import android.content.SharedPreferences;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 
 import java.io.IOException;
 
 /**
  * Information about the current user of your application.
  */
 class User implements JsonStream.Streamable {
```
fractalwrench: This class would need to be public if exposed as part of the API

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -1,28 +1,23 @@
...
-    private String id;
-    private String email;
-    private String name;
+    @Nullable private String id;
+    @Nullable private String email;
+    @Nullable private String name;
```
fractalwrench: These fields can now be final

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -1,28 +1,23 @@
...
 
-    User() {
-    }
-
-    User(String id, String email, String name) {
```
fractalwrench: I think for now it's worth keeping this as a public convenience constructor that builds a Builder.

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -1,28 +1,23 @@
...
-        this.id = id;
-        this.email = email;
-        this.name = name;
-    }
-
-    User(@NonNull User user) {
-        this(user.id, user.email, user.name);
+    User(Builder builder) {
```
fractalwrench: Can be private

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -36,27 +31,131 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
+    }
+
+    /**
+     * Create a new user object based on a previous user object.
+     *
+     * @param user the previous user object.
+     */
+    public static Builder builder(User user) {
+        return new Builder(user);
+    }
```
fractalwrench: I think we can get rid of the two methods above, and favour one way of creating a builder:

`User user = new User.Builder().id("Foo").build();`

The second method could be replaced by something like the following:

`User.Builder builder = user.toBuilder();`

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -36,27 +31,131 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
+
+    public static class Builder {
+        @Nullable String id;
+        @Nullable String email;
+        @Nullable String name;
```
fractalwrench: Fields can be private

sdk/src/main/java/com/bugsnag/android/User.java
```diff
@@ -36,27 +31,131 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
+
+        public Builder() {
+        }
+
+        public Builder(User user) {
+            id(user.getId());
+            email(user.getEmail());
+            name(user.getName());
```
fractalwrench: A test case for null users would be useful here

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -511,124 +506,75 @@ public void setAutoCaptureSessions(boolean autoCapture) {
...
-
-        SharedPreferences sharedPref =
-            appContext.getSharedPreferences(SHARED_PREF_KEY, Context.MODE_PRIVATE);
-        sharedPref.edit()
-            .remove(USER_ID_KEY)
-            .remove(USER_EMAIL_KEY)
-            .remove(USER_NAME_KEY)
-            .apply();
-        notifyBugsnagObservers(NotifyType.USER);
+    public User getUser() {
```
fractalwrench: Would be nice to add `@see #setUser(User)` to the JavaDoc

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -511,124 +506,75 @@ public void setAutoCaptureSessions(boolean autoCapture) {
...
-     * You can search for this information in your Bugsnag dashboard.
+     * Set the user information that the Bugsnag {@link Client} supplies with every error.<p>
+     * Pass NULL to clear the current user data, i.e. reset it to defaults.
      *
-     * @param id a unique identifier of the current user
+     * @param user a {@link User} object made by {@link User#builder()} or NULL
      */
-    public void setUserId(String id) {
-        setUserId(id, true);
+    public void setUser(@Nullable User user) {
```
fractalwrench: The JavaDocs for this method should include as much information as before, although we'll need to change how the params are handled:

```
Set details of the user currently using your application.
-     * You can search for this information in your Bugsnag dashboard.
-     * <p/>
-     * For example:
-     * <p/>
-     * client.setUser("12345", "james@example.com", "James Smith");
-     *
-     * @param id    a unique identifier of the current user (defaults to a unique id)
-     * @param email the email address of the current user
-     * @param name  the name of the current user
```

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -234,56 +234,56 @@ public static void setAutoCaptureSessions(boolean autoCapture) {
...
      * <p>
      * For example:
      * <p>
-     * Bugsnag.setUser("12345", "james@example.com", "James Smith");
+     * Bugsnag.setUser(User.builder().id("12345").email("james@example.com").name("James Smith"));
      *
-     * @param id    a unique identifier of the current user (defaults to a unique id)
-     * @param email the email address of the current user
-     * @param name  the name of the current user
+     * @param user The user information to set or NULL to reset to defaults.
```
fractalwrench: Needs to include same information as before about id/email/name

sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
```diff
@@ -0,0 +1,81 @@
...
+        User successor = User.builder(previous).build();
+
+        assertThat(successor.getId(), is(id));
+        assertThat(successor.getEmail(), is(email));
+        assertThat(successor.getName(), is(name));
+    }
+
+    @Test
+    public void testRepo() {
+        SharedPreferences sharedPref = getSharedPrefs(InstrumentationRegistry.getContext());
```
fractalwrench: Would it be worth `sharedPref.edit().clear().commit()` beforehand here?

sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
```diff
@@ -0,0 +1,81 @@
...
+    public void testBuilder_defaults() {
+        User user = User.builder().build();
+
+        assertThat(user.getId(), is(nullValue()));
+        assertThat(user.getEmail(), is(nullValue()));
+        assertThat(user.getName(), is(nullValue()));
+    }
+
+    @Test
+    public void testBuilder_values() {
```
fractalwrench: Our codestyle uses camelcase, looks like I've got a bit of work to do configuring our checkstyle as it should have caught this D=

fractalwrench:
@d4rken left a few minor comments, mostly this looks very good and I'm quite excited to get it merged :)

@kattrali will also have a look through the PR, but feel free to ping me in the meantime if you have any questions/clarifications

sdk/src/main/java/com/bugsnag/android/User.java
d4rken: Of course!

sdk/src/main/java/com/bugsnag/android/User.java
d4rken: I disagree. This was never public, so why would we want to introduce something new, "for now", that would be deprecated along with the other APIs?

Exposing only the builder to create `User` objects is what will make it easier to add additional user fields in the future, while keeping the API clean.

sdk/src/main/java/com/bugsnag/android/User.java
d4rken: Making this private will generate a synthetic accessor call due to access from `User.Builder.build()` as there is no advantage to making private I'd prefer to avoid increasing the method count unnecessarily by one (once deprecated methods are removed, bugsnags method count be reduced even more which is always nice for a library).

sdk/src/main/java/com/bugsnag/android/User.java
d4rken: As there is no advantage to making them private, we should keep them package protected as this would save another 3 methods (synthetic accessors).

sdk/src/main/java/com/bugsnag/android/User.java
d4rken: Good idea, this would also simplify internal use as `user.toBuilder()...` is nicer than `User.builder(user)...`.

d4rken: I will also drop the `User.builder()` method too to save another method as the builder constructors are public anyways.

sdk/src/main/java/com/bugsnag/android/User.java
d4rken: I agree, also found a mistake in `Error` that could lead to NPE, fixing that now.

sdk/src/main/java/com/bugsnag/android/Client.java
d4rken: 👍

sdk/src/main/java/com/bugsnag/android/Client.java
d4rken: That information is available in the javadoc of `User` though? Which the dev will see when they construct a `User` object. By duplicating java doc here, everytime `User` is changed, the java doc has to be updated in 3 places instead of just in `User` where it belongs.

sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
d4rken: Have you considered changing it? I think
`testMethodA_defaults` and `testMethodA_nullArgument`
is more readable than
`testMethodADefaults` and `testMethodANullArguments`
because it's immediately clear that it's two tests for the same method, just differents aspects.

I'll change it though, I don't have strong feelings about this.

sdk/src/androidTest/java/com/bugsnag/android/UserTest.java
d4rken: Definitely, to guard against any future changes that could make this test flaky.

d4rken:
What's the status on this?

fractalwrench:
@kattrali is planning to review this with a second pair of eyes, but is currently at a conference for the rest of the week.

anastasia-zaitsewa:
Would like to see this merged as well ➕

d4rken:
Is there no interest in merging this?

d4rken:
@fractalwrench @kattrali What's the issue here?

fractalwrench:
@d4rken I'm really sorry about how this was handled - it doesn't live up to our standards and I can only apologise that it happened. Shortly after this PR was opened, our policy on reviewing external PRs changed so that no longer accept changesets with feature additions or refactors of existing functionality. Coupled with an automation change that syncs GitHub issues to an internal ticketing system, I'm sorry to say that this PR slipped through the cracks.

As a consequence of this, we will:

- Update the Contributing guidelines across all our repositories to make the difference between bug fix/feature PRs much clearer.
- Create an internal ticket for providing accessors on every field within the Report payload.
- Raise this at our next engineering retrospective meeting, to avoid this happening again to others in future.

Again, I'd like to say how sorry we are - I've been in this position and know there's nothing more frustrating than spend time opening a PR and having it hang around for months. If there are any additional actions on our part that you think might be helpful in making this right, please let us know.

d4rken:
😦 

> Shortly after this PR was opened, our policy on reviewing external PRs changed so that no longer accept changesets with feature additions or refactors of existing functionality.

What caused this change?

> As a consequence of this, we will:
> * Update the Contributing guidelines across all our repositories to make the difference between bug fix/feature PRs much clearer.
> * Create an internal ticket for providing accessors on every field within the Report payload.
> * Raise this at our next engineering retrospective meeting, to avoid this happening again to others in future.

👍 

> If there are any additional actions on our part that you think might be helpful in making this right, please let us know.


Don't worry, I'm still a happy customer.

kattrali:
> What caused this change?

There's a few causes I can chime in on:

1. We have an internal roadmap which may be in conflict with features developed externally, which is why it often faster to send feedback about parts of the library or Bugsnag interface which are difficult to use or introduce friction to your workflow. We can pull together everyone's feedback and think about how all of these pieces should work together.
2. Generally, pull requests which include significant code changes are difficult to review. Error reporters have to work in crashy, multi-threaded environments and overlooking small adjustments can mean failure to capture reports, deadlocking when calling a function from a background thread unexpectedly, or worse. Features and refactors developed internally go through a design documentation and review process prior to code being written. Its not a burden we want to put on our customers as the process may sometimes seem opaque.
3. While we want each Bugsnag library to be an example of idiomatic use of both language and the development environment, we also want terminology and core concepts to be the same across different Bugsnag libraries, which makes integration easier for customers using multiple platforms and providing support easier for teams at Bugsnag.

All that being said, we absolutely welcome feedback about using Bugsnag via either GitHub issues or support@bugsnag.com. Thanks again.

PR #270, opened 2018-03-14
Add semaphore to error store when flushing cached reports

fractalwrench:
Adds a semaphore to the `ErrorStore` that is used when flushing cached reports. This prevents a scenario where a report could potentially be sent multiple times.


sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreAsyncTest.java
```diff
@@ -0,0 +1,101 @@
...
+        errorStore.flushOnLaunch(apiClient);
+        assertEquals(1, requestCount.get());
+    }
+
+    @Test
+    public void testLaunchThenConnectivity() throws Exception {
+        writeFakeError();
+        errorStore.flushOnLaunch(apiClient);
+        errorStore.flushAsync(apiClient);
+        assertEquals(1, requestCount.get());
```
Cawllec: Referring to a desk conversation: This assertion was failing inconsistently when the emulator was first launched with the error:
```
java.lang.AssertionError: expected:<1> but was:<0>
at org.junit.Assert.fail(Assert.java:88)
at org.junit.Assert.failNotEquals(Assert.java:834)
at org.junit.Assert.assertEquals(Assert.java:645)
at org.junit.Assert.assertEquals(Assert.java:631)
at com.bugsnag.android.ErrorStoreAsyncTest.testLaunchThenConnectivity(ErrorStoreAsyncTest.java:90)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
```

fractalwrench:
Unit test removed as this will be covered by mazerunner instead.

fractalwrench:
@kattrali I have run the 3 new mazerunner scenarios against this and they pass, whereas on master they would fail.

PR #274, opened 2018-03-15
Add scenarios for asynchronously flushing error store

fractalwrench:
Adds scenarios which ensure that only 1 request is sent for the following scenarios:

- Connectivity changes immediately before onLaunch flush
- Connectivity changes immediately after onLaunch flush
- Connectivity changes in quick succession


fractalwrench:
@snmaynard these scenarios fail on the current branch, the fix is in a separate PR: https://github.com/bugsnag/bugsnag-android/pull/270

PR #284, opened 2018-04-10
Avoid adding extra comma separator if File input is empty or null

fractalwrench:
Previously, if the file input was null or empty, an additional comma was added by the JSON serialiser as a separator, leading to invalid JSON, e.g. `[{}, , {}]`. This can lead to session/error payloads being dropped as invalid.

This fixes JSON serialisation so that a comma is not included in this case, e.g. `[{}, {}]`

This doesn't fix the case where a file is non-empty but isn't valid JSON, which will be addressed separately.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/JsonStream.java
+++ b/sdk/src/main/java/com/bugsnag/android/JsonStream.java
@@ -49,6 +49,10 @@ public class JsonStream extends JsonWriter {
      * Writes a File (its content) into the stream
      */
     public void value(@NonNull File file) throws IOException {
+        if (file == null || file.length() <= 0) {
+            return;
+        }
+
         super.flush();
         beforeValue(false); // add comma if in array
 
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/bugsnag/android/JsonStream.java
```diff
@@ -49,6 +49,10 @@ public void value(@Nullable Streamable streamable) throws IOException {
      * Writes a File (its content) into the stream
      */
     public void value(@NonNull File file) throws IOException {
+        if (file == null || file.length() <= 0) {
```
Cawllec: Do you need the `null` check if `@NonNull` is used above?

fractalwrench: `@NonNull` doesn't enforce anything in Java unfortunately. The IDE would highlight any null parameters with a warning, but developers could choose to ignore this.

Cawllec: Ah, fair enough.

Issue #285, opened 2018-04-13
Cannot use two client instances with different API keys


pyricau:
### Expected behavior
Let's say I create two distinct Bugsnag clients, like so:

```
Client client1 = new Client(application, "key1", false);
Client client2 = new Client(application, "key2", false)
```

These two clients use two distinct keys, and are expected to upload errors to two distinct bugsnag projects.

### Observed behavior

If there is no network when we try to notify and the errors get stored on the filesystem, the API keys stored in the errors will be ignored and the errors will be uploaded to the project of whichever client flushes first.

### Steps to reproduce

* Turn of Wifi
* Create a client instance with API KEY 1
* Send an error to client (it gets stored because no wifi)
* Kill the app
* Start the app
* Create a client instance with API KEY 2, get it to flush
* => the crash 1 gets uploaded to project 2 instead of project 1.

### Version
Still happening on master, all releases have this bug.

### Additional information

At this line: https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/ErrorStore.java#L125

```java
    private void flushErrorReport(File errorFile, ErrorReportApiClient errorReportApiClient) {
        try {
            Report report = new Report(config.getApiKey(), errorFile);
            errorReportApiClient.postReport(config.getEndpoint(), report,
                config.getErrorApiHeaders());
// ...
```

The report is uploaded with the API Key from the config of the error store, not the API Key from that config that was used when the error was created (that config is actually stored in the error).

### Possible solutions

1) Allow customizing the error store and session store folders. This would allow each client to store data in a distinct folder and avoid clients stepping over each other
2) Read the content of the error before uploading it, to retrieve the proper API key. This is potentially risky.
3) Fix this server side, check for Error file content with config.apiKey set to something different than the API Key that was used to upload.

fractalwrench:
Thanks for the detailed report @pyricau - I can see that the current `ErrorStore` implementation is problematic for the apiKey, endpoint, and headers, if multiple clients are instantiated. I'll start writing some test cases for these scenarios and discuss with the rest of the team what the best solution is.

felipehjcosta:
Hi guys! 
Any plan to add this feature in the future?

alexjlockwood:
Anyone have any potential workarounds for this issue?

mattdyoung:
Hi

[...]

PR #286, opened 2018-04-13
Add multi-client scenario

fractalwrench:
Adds a mazerunner scenario where multiple clients with different api keys are initialised, an unhandled exception is thrown and stored, then delivered by both clients to the server.

This scenario is currently expected to fail due to one client using the wrong API key, further explanation can be found in #285 

Some other scenarios we may wish to consider covering as part of this include:

- `config.getEndpoint()`, which is accessed in the same way as `apiKey`
- `config.getSessionEndpoint()`, for sessions
- Error API + Session API headers, although these are unlikely to change regularly


features/multi_client.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Initializing nultiple clients sends payloads with correct API key to each
+
+Scenario: Unhandled error reported by two clients with different API keys
```
kattrali: I initially didn't understand the expectations of this scenario, since there are two cases rolled into one. Perhaps the name could be amended or additional feature description added to describe the test? Any non-"Scenario" lines leading up to each scenario is treated as a comment and printed on the console during the test run.

As I understand it, the intended effect is:
1. Two clients are configured
2. The device is offline
3. An unhandled error occurs, and a report is cached on disk
4. The app is restarted with delivery enabled
5. One request is sent by each client, each using the correct API key

If that's correct, I'd rename the scenario to something like:

> An unhandled error captured while offline is detected by two clients with different API keys

And maybe add a bit of preamble:

```gherkin
Feature: Multi-client support

    Multiple Bugsnag clients can be configured, each with its own API key and
    configuration options. A report which is captured by a given client should
    use the correct key in report.

    Scenario: An unhandled error captured while offline is detected by two clients with different API keys
```

It also surfaces what additional scenarios should be added (two clients sending unhandled errors while online? two clients sending handled errors offline? etc)

features/multi_client.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Initializing nultiple clients sends payloads with correct API key to each
```
kattrali: "nultiple"

mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/HandledExceptionSessionScenario.kt
```diff
@@ -12,7 +13,7 @@ internal class HandledExceptionSessionScenario(config: Configuration,
 
     override fun run() {
         super.run()
-        disableSessionDelivery()
+        disableSessionDelivery(Bugsnag.getClient())
```
kattrali: In the future it would be helpful to separate the refactoring parts of this changeset (moving the internal functions to a separate file) into a separate commit. It wasn't initially clear whether these changes were otherwise related to the new scenario.

mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/MultiClientApiKeyScenario.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.bugsnag.android.Client
+import com.bugsnag.android.Configuration
+import com.bugsnag.android.disableAllDelivery
+
+/**
+ * Configures two Bugsnag clients with different API keys. The error should be captured by each,
+ * and the correct API key should be used for both.
+ */
+internal class MultiClientApiKeyScenario(config: Configuration,
+                                         context: Context) : MultiClientScenario(config, context) {
```
kattrali: Some of my initial confusion was from this scenario being a subclass, unlike the others, and that two `Scenario` classes are combined into a single test case. While this approach reuses code, it obscures the intent as thus far a `Scenario` class corresponded one-to-one to a `Scenario` in a feature. It would be semantically clearer to have a single `Scenario` class which does different things in the first run vs a later run, perhaps by setting additional extended data on the intent.

Rough example where Scenarios can load `intent.getStringExtra` via `getRuntimeVariable`:

```kotlin
    override fun run() {
        // configure two clients
        firstClient = Client(context, config)
 
        Thread.sleep(10) // enforce request order
        val secondConfig = Configuration("abc123")
        secondConfig.endpoint = config.endpoint
        secondConfig.sessionEndpoint = config.sessionEndpoint
        secondClient = Client(context, secondConfig)

        // cause an exception while offline
        if getRuntimeVariable("OfflineException") == "on" {
            disableAllDelivery(firstClient!!)
            disableAllDelivery(secondClient!!)
            throw IllegalArgumentException("MultiClientApiKeyScenario")
        }
    }
```

features/multi_client.feature
fractalwrench: fxied all teh typos :)

fractalwrench:
Updated in response to feedback.

This PR now relies on the following addition to mazerunner: https://github.com/bugsnag/maze-runner/pull/27

fractalwrench:
Changes are now available for review in #290

PR #290, opened 2018-04-19
Support multi client storage

fractalwrench:
Adds support for storage of files on a per-client basis, by using a different directory for each client based on a hash of their configured API key and endpoint.

This PR alters the `FileStore` class and its descendants so that new errors/sessions are stored in the appropriate location. Previously cached errors/sessions will still be flushed from disk from the old location - JUnit test coverage is added for this.

Also see #286 and #285

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
@@ -23,6 +23,8 @@ import org.junit.runner.RunWith;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.nio.file.Paths;
+import java.util.List;
 
 @RunWith(AndroidJUnit4.class)
 @SmallTest
@@ -121,6 +123,21 @@ public class ErrorStoreTest {
         assertTrue(errorStore.isStartupCrash(5345));
         assertTrue(errorStore.isStartupCrash(9999));
         assertFalse(errorStore.isStartupCrash(10000));
+    }
+
+    @Test
+    public void testFindOldFiles() throws Throwable {
+        new File(errorStorageDir, "foo.json").createNewFile();
+
+        File dir = new File(errorStorageDir, "1059309/52903");
+        dir.mkdirs(); // api/endpoint dirs
+        new File(dir, "foo.json").createNewFile();
+
+        List<File> files = errorStore.findStoredFiles();
+        for (File file : files) {
+            assertTrue(file.getAbsolutePath().endsWith("foo.json"));
+        }
+        assertEquals(2, files.size());
     }
 
     /**
--- a/sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
@@ -44,7 +44,7 @@ public class ErrorStoreTest {
         config = client.config;
         errorStore = client.errorStore;
         errorStorageDir = errorStore.storageDir;
-        FileUtils.clearFilesInDir(errorStorageDir);
+        FileUtils.clearFiles(errorStore);
     }
 
     /**
@@ -54,7 +54,7 @@ public class ErrorStoreTest {
      */
     @After
     public void tearDown() throws Exception {
-        FileUtils.clearFilesInDir(errorStorageDir);
+        FileUtils.clearFiles(errorStore);
     }
 
     @Test
--- a/sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
@@ -23,7 +23,6 @@ import org.junit.runner.RunWith;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
-import java.nio.file.Paths;
 import java.util.List;
 
 @RunWith(AndroidJUnit4.class)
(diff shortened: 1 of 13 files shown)


fractalwrench:
>If the app crashes, then is updated with a new version of Bugsnag, is the "old" state of the filesystem compatible with this new logic?

Yes. `findStoredFiles` should find any Files in either the new location, or previous location.

>Could the state of the filesystem cause Bugsnag to crash?

I'm not aware of anything within these changes that would cause an unhandled crash within the notifier, although the risky areas would be around:

- ensuring the old/new location actually exists
- handling IOErrors when Files cannot be created
- null/File#exists() checks

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -16,8 +17,11 @@
 
     @NonNull
     protected final Configuration config;
+
     @Nullable
-    final String storeDirectory;
+    final String oldDirectory;
+
+    File storageDir;
```
kattrali: Why should this property be renamed?

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -86,22 +89,50 @@ String write(@NonNull T streamable) {
...
+                    files.add(value);
                 }
             }
         }
-        return files;
     }
+
+    // support multiple clients in the same app by using a unique directory path
+
+    private File getStorageDir(String path, @NonNull Configuration config) {
```
kattrali: If a project rotates API keys or the on-premise endpoint changes, will stored files ever be sent (or cleared)?

sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
```diff
@@ -123,6 +123,19 @@ public void isStartupCrash() throws Exception {
...
+    public void testFindOldFiles() throws Throwable {
+        new File(errorStore.oldDirectory, "foo.json").createNewFile();
+        File dir = errorStore.storageDir;
+        new File(dir, "foo.json").createNewFile();
+
+        List<File> files = errorStore.findStoredFiles();
+        for (File file : files) {
+            assertTrue(file.getAbsolutePath().endsWith("foo.json"));
+        }
+        assertEquals(2, files.size());
```
kattrali: If there are multiple clients, will all attempt to find and send files from the "old" directory?

sdk/src/main/java/com/bugsnag/android/FileStore.java
fractalwrench: I wanted it to be clear that there is a difference between the old directory we previously used, and the client-specific directory that we will use in future. Maybe the nomenclature is a bit off here?

sdk/src/main/java/com/bugsnag/android/FileStore.java
fractalwrench: Not currently. Is this something we need to address?

sdk/src/androidTest/java/com/bugsnag/android/ErrorStoreTest.java
fractalwrench: Yes. I've added a mazerunner scenario which covers this.

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -29,31 +33,30 @@
...
+        if (storageDir == null) {
             return null;
         }
 
         // Limit number of saved errors to prevent disk space issues
-        File exceptionDir = new File(storeDirectory);
-        if (exceptionDir.isDirectory()) {
-            File[] files = exceptionDir.listFiles();
+        if (storageDir.isDirectory()) {
+            File[] files = storageDir.listFiles();
```
fractalwrench: This needs fixing as it doesn't take the previous directory into account.

fractalwrench:
@kattrali I wanted to check whether the remaining inline comments require any actions, namely:

- Should `oldDirectory` be renamed to something else?
- Do we need to handle the case of API key/endpoint rotation, and if so, how would that be achieved?

sdk/src/main/java/com/bugsnag/android/FileStore.java
kattrali: Yeah - in the worst case, files shouldn't be left on disk indefinitely. In the best case, everything is delivered according to the configuration at the time the report was captured. Splitting directories by endpoint and key works until  either component needs to change. A better approach would be to write the entire request to disk including the endpoint and the headers. For example, a (nearly) raw HTTP request (format amended to encode protocol and port):

```http
POST https://notify.bugsnag.com
Bugsnag-Api-Key: aaaabaaaabaaaabaaaa
Content-Type: application/json

{"events": [{"severity": "warning", "exceptions":[]}]
```

on premise example for contrast:

```http
POST https://bugsnag.example.internal:79002
Bugsnag-Api-Key: aaaabaaaabaaaabaaaa
Content-Type: application/json

{"events": [{"severity": "warning", "exceptions":[]}]
```

Which could then be parsed until encountering the double newline for the endpoint and headers, then stream the rest as the body directly.

This should:

* Avoid cycling directories, ensuring everything is delivered and cleared from disk
* Tie API key to a request even if cached on disk
* Not affect reports not cached to disk
* Have the same integration tests
* Have the same/similar logic for handling reports in the old location

kattrali:
> Should `oldDirectory` be renamed to something else?

Seems fine, perhaps should be annotated for removal in the next major

> Do we need to handle the case of API key/endpoint rotation, and if so, how would that be achieved?

Ah, added detail [inline](https://github.com/bugsnag/bugsnag-android/pull/290#discussion_r185384048) for this.

sdk/src/main/java/com/bugsnag/android/FileStore.java
fractalwrench: I think this approach makes sense.

fractalwrench: https://developer.android.com/reference/android/content/Context#getcachedir

fractalwrench:
Closing this PR off as the intended approach has changed somewhat.

PR #293, opened 2018-04-20
Adds additional coverage for Session class serialisation

fractalwrench:
Adds missing test coverage as identified by static analysis

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/SessionTrackingPayloadTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/SessionTrackingPayloadTest.java
@@ -4,7 +4,9 @@ import static com.bugsnag.android.BugsnagTestUtils.generateSession;
 import static com.bugsnag.android.BugsnagTestUtils.generateSessionTracker;
 import static com.bugsnag.android.BugsnagTestUtils.streamableToJson;
 import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
 import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertTrue;
 
 import android.content.Context;
 import android.support.test.InstrumentationRegistry;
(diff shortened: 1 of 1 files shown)


sdk/src/androidTest/java/com/bugsnag/android/SessionTrackingPayloadTest.java
```diff
@@ -93,4 +101,17 @@ public void testMultipleSessionFiles() throws Exception {
...
+        session = new Session("id", new Date(), null, false);
+        Context context = InstrumentationRegistry.getContext();
+        payload = generatePayloadFromSession(context, session);
+        assertFalse(session.isAutoCaptured());
+        session.setAutoCaptured(true);
+        assertTrue(session.isAutoCaptured());
+
+        JSONObject rootNode = streamableToJson(payload);
+        JSONObject sessionNode = rootNode.getJSONArray("sessions").getJSONObject(0);
+        assertFalse(sessionNode.has("user"));
```
kattrali: This seems to imply there should be some other test for `user` contents. Also, shouldn't the fields which exist in the payload be tested?

sdk/src/androidTest/java/com/bugsnag/android/SessionTrackingPayloadTest.java
```diff
@@ -93,4 +101,17 @@ public void testMultipleSessionFiles() throws Exception {
         assertEquals(2, sessions.length());
     }
 
+    @Test
+    public void testDefaultOverride() throws Exception {
```
kattrali: Does this mean "test auto captured override"?

sdk/src/androidTest/java/com/bugsnag/android/SessionTrackingPayloadTest.java
fractalwrench: Yep - I believe they're mostly covered in the `testPayloadSerialisation` method, and the serialisation tests for individual classes (e.g. User).

PR #294, opened 2018-04-20
Add test coverage for client.notifyBlocking

fractalwrench:
Adds missing test coverage for non-deprecated overloads of notifyBlocking in the client class


fractalwrench:
Renamed test methods.

PR #295, opened 2018-04-23
Filestore thread safety fixes

fractalwrench:
We sometimes receive invalid payloads with an empty/malformed array where an error/session should be - e.g. `[]` or `[,]`. My hypothesis is that this occurs when the max limit of Error/Sessions are cached on disk. When the app is relaunched, if `write` is called just before the payload is serialised, then this can lead to the oldest File being deleted, which means there is no data to serialise.

This attempts to address this scenario by retaining a reference to any Files which are currently being flushed, and not deleting them.

A potential downside to this implementation is that it currently requires manually calling `cancelQueuedFiles` and `deleteStoredFiles`, which could be forgotten.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/FileStore.java
+++ b/sdk/src/main/java/com/bugsnag/android/FileStore.java
@@ -9,6 +9,8 @@ import java.io.FileWriter;
 import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
@@ -59,9 +61,7 @@ abstract class FileStore<T extends JsonStream.Streamable> {
                 Arrays.sort(files, comparator);
                 Logger.warn(String.format("Discarding oldest error as stored "
                     + "error limit reached (%s)", files[0].getPath()));
-                if (!files[0].delete()) {
-                    files[0].deleteOnExit();
-                }
+                deleteStoredFiles(Collections.singleton(files[0]));
             }
         }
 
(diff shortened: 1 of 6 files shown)


fractalwrench:
@bugsnag/platforms this is now ready for review.

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -82,26 +100,63 @@ String write(@NonNull T streamable) {
                 filename), exception);
         } finally {
             IOUtils.closeQuietly(out);
+            lock.unlock();
```
Cawllec: Is there a chance this can fail if IOUtils throws?

Cawllec: NVM, I just looked up what `IOUtils.closeQuietly` does

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -82,26 +100,63 @@ String write(@NonNull T streamable) {
...
         }
         return null;
     }
 
-    @NonNull abstract String getFilename(T streamable);
+    @NonNull
+    abstract String getFilename(T streamable);
 
     List<File> findStoredFiles() {
```
Pezzah: I think this can be called multiple times in the current design, which might cause bad things to happen. Potentially this should only return files that are not already in the queued files set?

fractalwrench: Good point. I've updated this method so that it only returns stored files which aren't in the `queuedFiles` collection.

PR #296, opened 2018-04-23
test: add async notify test coverage

fractalwrench:
Improves code coverage by adding tests for async notify methods

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/ClientNotifyTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ClientNotifyTest.java
@@ -1,0 +1,87 @@
+package com.bugsnag.android;
+
+import static org.junit.Assert.assertEquals;
+
+import android.support.test.filters.SmallTest;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Before;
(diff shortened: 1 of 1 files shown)


sdk/src/androidTest/java/com/bugsnag/android/ClientNotifyTest.java
```diff
@@ -72,7 +77,55 @@ public void beforeNotify(Report report) {
         assertEquals("Message", error.getExceptionMessage());
     }
 
+    @Test
+    public void testNotifyAsyncMetadata() throws Exception {
+        MetaData metaData = new MetaData();
+        metaData.addToTab("animals", "dog", true);
+
+        client.notify(new RuntimeException("Foo"), metaData);
+        apiClient.awaitReport();
```
Cawllec: Are you testing purpose of the Async functionality i.e. The call to `client.notify` is non-blocking anywhere?

fractalwrench: That's a good point, I'm not aware of a direct test for that anywhere. I can add in an assertion to check that the report is null after calling notify, but before calling the countdown latch.

PR #299, opened 2018-04-24
Add scenarios for setting custom API clients

fractalwrench:
Adds mazerunner scenarios which confirm that custom API clients are always used for sending of errors/sessions, and fixes edge cases where that may not be the case. Scenarios added include:

- Handled error
- Unhandled error cached
- Manual session
- Manual session cached

Scenarios are verified using a custom API client that writes an extra header. If this is not present, the wrong API client was used.

The fixes to the notifier include:

- Delaying flushing of stored error reports until the error API client can be overridden
- Ignoring connectivity changes if they occur before the error API client can be overriden
- Writing sessions to disk if they occur before the session API client can be overriden

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -59,7 +59,7 @@ public class Client extends Observable implements Observer {
     static final String MF_AUTO_CAPTURE_SESSIONS =
         BUGSNAG_NAMESPACE + ".AUTO_CAPTURE_SESSIONS";
 
-    private volatile boolean clientInitialised = false;
+    volatile boolean clientInitialised = false;
 
     @NonNull
     protected final Configuration config;
@@ -217,13 +217,13 @@ public class Client extends Observable implements Observer {
             public void run() {
                 try {
                     Thread.sleep(10); // allow users to set custom API clients
-                    clientInitialised = true;
                     errorStore.flushOnLaunch(errorReportApiClient);
                 } catch (InterruptedException exception) {
                     Logger.warn("Failed to flush errors on launch", exception);
                 }
             }
         });
+        clientInitialised = true;
     }
 
     private class ConnectivityChangeReceiver extends BroadcastReceiver {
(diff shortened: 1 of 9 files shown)


sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -206,12 +207,23 @@ public void run() {
...
+            public void run() {
+                try {
+                    Thread.sleep(10); // allow users to set custom API clients
+                    errorStore.flushOnLaunch(errorReportApiClient);
+                } catch (InterruptedException exception) {
+                    Logger.warn("Failed to flush errors on launch", exception);
+                }
+            }
+        });
+        clientInitialised = true;
```
kattrali: While this fixes the immediate issue, the longer term problem to solve here is that some configuration options should be set correctly when the client is created. Properties like `errorReportingApiClient` should be on `Configuration` so that the client can then immediately use what has been configured if its available, and then we should phase out the convenience setters for those properties in the `Bugsnag` interface.

Adding a delay here increases the risk of reports not being sent if the app enters a crash loop.

fractalwrench:
I think those points make a lot of sense. `BadResponseException` only appears to log a message throughout the Notifier, so I think it would make sense to use a single exception type of `DeliveryFailureException`, to signify that sending the payload should be attempted again at a future date.

Therefore, I think the following actions are required to get this PR into the final state of review:

- Change the `deliver` method signature to throw `DeliveryFailureException` only.
- Update the docs to signify that throwing this exception indicates that the request should be retained and sent at a later date.
- Refactor internal delivery code so that a bad response logs a message within `DefaultDelivery`, then deletes the request if written to disk.
- Map `NetworkException` onto `DeliveryFailureException` for the deprecated *ApiClients, swallow/log `BadResponseException`.

@kattrali if I've missed any additional required amends or one of these is wrong, can you let me know before I start implementation?

kattrali:
Yep, that plan addresses my concerns perfectly, thanks @fractalwrench

fractalwrench:
Updated PR with actions as discussed. I also altered the logging so that an exception is always logged, and updated the Docs PR: https://github.com/bugsnag/docs.bugsnag.com/pull/719

PR #300, opened 2018-04-25
fix: prevent potential lifecycle breadcrumb crash

fractalwrench:
Change `SessionTracker` to ignore exceptions when creating lifecycle breadcrumb before a client is fully constructed. This uses the same logic as `EventReceiver`.


fractalwrench:
@Cawllec Github seems to automatically dismiss approvals whenever there's a push - I've added a JavaDoc comment to address a checkstyle warning in the test. May be best to wait until Travis completes

PR #302, opened 2018-04-27
Add client breadcrumbs code coverage

fractalwrench:
Adds tests which cover client breadcrumb methods directly - other tests already verify the behaviour of the breadcrumbs class itself.


fractalwrench:
Addressed all inline feedback.

PR #303, opened 2018-04-27
chore: add git hooks for checkstyle/lint

fractalwrench:
Runs checkstyle and lint before pushing code, enabling us to catch issues earlier than CI.

This file should be symlinked to `.git/hooks/pre-push`

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,3 @@
+echo "Running lint + checkstyle before push";
+./gradlew lint checkstyle
+


fractalwrench:
I'm not aware of equivalent functionality built-in to gradle (although it could certainly be added as a task).

I'm thinking it would be best to add _something_ to automate setting up hooks before this is merged, as I think it would make sense to roll out what we do here to other repos. It could be as simple as adding a bash script:

`ln -s -f ../../.githooks/pre-push .git/hooks/pre-push`

Or maybe we could use Husky, some similar tool, or a completely different approach. What are your thoughts?

bengourley:
I dunno, I feel slightly uneasy about adding another _entire_ toolchain for this, since we already have the java/android stuff, plus ruby (gems, bundler) for maze runner. We'd need node/npm and a package.json. It maybe seems overkill? On the other hand, we're all going to have node/npm on our machines already… so maybe it's not so bad.

If you could just make the symlink with a command that will be run by everyone, that might be ok, I'm assuming you can do that with gradle?

Cawllec:
If gradle is required by the repo and has copying capabilities included, could we add a task there without disrupting the build tasks already defined in script?

fractalwrench:
Closing this for now as I believe it'd be best to set this up as a gradle plugin, that automatically copies hooks to the correct directory.

Issue #291, opened 2018-04-19
Breadcrumbs uploaded with new SDK display weirdly in web UI


edenman:
### Expected behavior
Breadcrumbs display with the full message

### Observed behavior
https://app.bugsnag.com/recharge-labs/android/errors/5ad8db56c0ebd600182ae1ea has breadcrumbs but the message is crammed into the left side and the full message is not shown

### Steps to reproduce
1. Upgrade to com.bugsnag:bugsnag-android:4.3.3 (from 4.1.5, if that matters)
2. Set some breadcrumbs using `Bugsnag.leaveBreadcrumb(message)`
3. Force the app to crash

### Version
4.3.3

edenman:
Thanks!  I couldn't tell if it was a dashboard thing or a client thing.

edenman:
@fractalwrench any update on this?  it makes breadcrumb analysis entirely useless.

fractalwrench:
@edenman there appears to be an issue with the 4.3.3 client where the breadcrumb message is serialised as the title instead, leading to truncated text on the dashboard.

Additionally, there's a layout issue with long breadcrumbs that our dashboard team is working to address separately (example below)

<img width="232" alt="screen shot 2018-05-01 at 10 30 03" src="https://user-images.githubusercontent.com/11800640/39468386-a4caad68-4d2a-11e8-860e-43dac822e954.png">

edenman:
That's unfortunate.  Let me know when there's a fix for the client serialization issue?

fractalwrench:
A fix for this issue should now be available in [v4.3.4](https://github.com/bugsnag/bugsnag-android/releases/tag/v4.3.4).

PR #306, opened 2018-05-01
Fix: serialise breadcrumb message correctly

fractalwrench:
The breadcrumb message should be serialised as part of the metadata, not as the name attribute.

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/BreadcrumbsTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/BreadcrumbsTest.java
@@ -1,5 +1,7 @@
 package com.bugsnag.android;
 
+import static com.bugsnag.android.BreadcrumbType.MANUAL;
+import static com.bugsnag.android.BugsnagTestUtils.generateClient;
 import static com.bugsnag.android.BugsnagTestUtils.streamableToJsonArray;
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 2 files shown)


fractalwrench:
Partially addresses #291

sdk/src/androidTest/java/com/bugsnag/android/BreadcrumbsTest.java
```diff
@@ -172,4 +175,21 @@ public void testPayloadType() throws JSONException, IOException {
...
+        Queue<Breadcrumb> store = client.breadcrumbs.store;
+        int count = 0;
+
+        for (Breadcrumb breadcrumb : store) {
+            if (MANUAL == breadcrumb.getType() && "manual".equals(breadcrumb.getName())) {
+                count++;
+                assertEquals("Hello World", breadcrumb.getMetadata().get("message"));
+            }
+        }
+        assertEquals(1, count);
```
Cawllec: Is the count necessary for this test? And if so, is the for loop the best way to test for it?  Would testing the length of the queue be more clear?

fractalwrench: The size and order of the breadcrumb queue cannot be guaranteed when calling `client#leaveBreadcrumb`, as automatic breadcrumbs may occur at any time (around 6-7 tend to be recorded in practice).

PR #307, opened 2018-05-01
Perf: use buffered streams for IO

fractalwrench:
`FileReader` and `FileWriter` are not buffered, meaning they are slow. This change wraps all IO streams in a buffered reader/writer which in an (unscientific) unit test of the ErrorStore, reduced the write of an error from ~230ms to ~115ms.

This also allows us to specify the charset encoding explicitly, which is a potential (but unproven) cause for File serialisation issues on devices from certain manufacturers.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
+++ b/sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
@@ -3,11 +3,13 @@ package com.bugsnag.android;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 
+import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.nio.charset.Charset;
 import java.util.Map;
 
 class DefaultHttpClient implements ErrorReportApiClient, SessionTrackingApiClient {
@@ -69,7 +71,9 @@ class DefaultHttpClient implements ErrorReportApiClient, SessionTrackingApiClien
 
             try {
                 out = conn.getOutputStream();
-                JsonStream stream = new JsonStream(new OutputStreamWriter(out));
+                Charset charset = Charset.forName("UTF-8");
+                BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(out, charset));
+                JsonStream stream = new JsonStream(writer);
                 streamable.toStream(stream);
                 stream.close();
             } finally {
--- a/sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
+++ b/sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
@@ -67,17 +67,16 @@ class DefaultHttpClient implements ErrorReportApiClient, SessionTrackingApiClien
                 conn.addRequestProperty(entry.getKey(), entry.getValue());
             }
 
-            OutputStream out = null;
+            JsonStream stream = null;
 
             try {
-                out = conn.getOutputStream();
+                OutputStream out = conn.getOutputStream();
                 Charset charset = Charset.forName("UTF-8");
                 BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(out, charset));
-                JsonStream stream = new JsonStream(writer);
+                stream = new JsonStream(writer);
                 streamable.toStream(stream);
-                stream.close();
             } finally {
-                IOUtils.closeQuietly(out);
+                IOUtils.closeQuietly(stream);
             }
 
 
--- a/sdk/src/main/java/com/bugsnag/android/FileStore.java
+++ b/sdk/src/main/java/com/bugsnag/android/FileStore.java
@@ -70,22 +70,20 @@ abstract class FileStore<T extends JsonStream.Streamable> {
 
         String filename = getFilename(streamable);
 
-        Writer out = null;
+
+        JsonStream stream = null;
         try {
             FileOutputStream fos = new FileOutputStream(filename);
-            out = new BufferedWriter(new OutputStreamWriter(fos, "UTF-8"));
-
-            JsonStream stream = new JsonStream(out);
+            Writer out = new BufferedWriter(new OutputStreamWriter(fos, "UTF-8"));
+            stream = new JsonStream(out);
             stream.value(streamable);
-            stream.close();
-
             Logger.info(String.format("Saved unsent payload to disk (%s) ", filename));
             return filename;
         } catch (Exception exception) {
             Logger.warn(String.format("Couldn't save unsent payload to disk (%s) ",
                 filename), exception);
         } finally {
-            IOUtils.closeQuietly(out);
+            IOUtils.closeQuietly(stream);
         }
         return null;
     }
--- a/sdk/src/main/java/com/bugsnag/android/FileStore.java
+++ b/sdk/src/main/java/com/bugsnag/android/FileStore.java
@@ -4,8 +4,11 @@ import android.content.Context;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
+import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.FileWriter;
+import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


fractalwrench:
@kattrali it's very much a long shot but I thought it might be possible that some manufacturers don't use UTF-8 encoding by default, which is mandated by the Android docs. This change makes the use of UTF-8 explicit. However, thinking about it, if this were the cause, we would probably see various other issues as a result.

sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
```diff
@@ -69,7 +71,9 @@ private int makeRequest(String urlString,
 
             try {
                 out = conn.getOutputStream();
-                JsonStream stream = new JsonStream(new OutputStreamWriter(out));
+                Charset charset = Charset.forName("UTF-8");
+                BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(out, charset));
+                JsonStream stream = new JsonStream(writer);
```
martin308: do any of these other objects need to be `closed`?

sdk/src/main/java/com/bugsnag/android/JsonStream.java
```diff
@@ -53,9 +56,10 @@ public void value(@NonNull File file) throws IOException {
         beforeValue(false); // add comma if in array
 
         // Copy the file contents onto the stream
-        FileReader input = null;
+        Reader input = null;
         try {
-            input = new FileReader(file);
+            FileInputStream fis = new FileInputStream(file);
+            input = new BufferedReader(new InputStreamReader(fis, "UTF-8"));
```
martin308: do any of these objects also need closing?

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -69,7 +72,8 @@ String write(@NonNull T streamable) {
 
         Writer out = null;
         try {
-            out = new FileWriter(filename);
+            FileOutputStream fos = new FileOutputStream(filename);
+            out = new BufferedWriter(new OutputStreamWriter(fos, "UTF-8"));
```
martin308: do any of these objects also need closing?

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -69,7 +72,8 @@ String write(@NonNull T streamable) {
...
 
             JsonStream stream = new JsonStream(out);
             stream.value(streamable);
```
martin308: not related to this PR but should we be closing this stream on the next line in the `finally` block?

sdk/src/main/java/com/bugsnag/android/JsonStream.java
fractalwrench: I believe calling `BufferedReader#close` also closes the `InputStreamReader`, which in turn closes the `FileInputStream`, which means that everything is automatically closed.

sdk/src/main/java/com/bugsnag/android/FileStore.java
fractalwrench: Addressed by other inline comment on JsonStream

sdk/src/main/java/com/bugsnag/android/FileStore.java
fractalwrench: Yes, that would make more sense than closing two separate streams. I've updated the PR so that we do this.

fractalwrench:
@martin308 thanks for the feedback! I've responded to the comments inline, and have also made a slight modification so that we close the `JsonStream` in a `finally` block.

fractalwrench:
Copied from #307 (addressing flaky tests) for extra context:

Thanks for the review comments. I'm going to close this PR for now as I've noticed that #307 drastically reduces the flakiness of the tests on Travis (builds 3 times without incident), and I have another hypothesis about what may have caused the flakiness.

I managed to intermittently reproduce the behaviour on an old emulator (API 16), and the log output suggests that other tests are constructing `Client` objects, which triggers an asynchronous flush of Files. I believe this leads to a race where the File is flushed before `ConcurrentSkipListSet` finishes calculating the `size` of Files, leading to an output of 0, where we would expect N. The performance improvements in #307 halve the time it takes to write a File, which could mean that `ConcurrentSkipListSet` now nearly always wins the race.

Passing output:
```
05-09 12:22:35.064 5313-5325/com.bugsnag.android.test I/TestRunner: started: testCancelQueuedFiles(com.bugsnag.android.ErrorStoreTest)
05-09 12:22:35.064 5313-5313/com.bugsnag.android.test I/MonitoringInstrumentation: Activities that are still in CREATED to STOPPED: 0
05-09 12:22:35.074 5313-5325/com.bugsnag.android.test I/Bugsnag: Saved unsent payload to disk (/data/data/com.bugsnag.android.test/cache/bugsnag-errors/1525868555077_d2da4d31-71ac-4fbc-a1d6-15ccdb6e07cd_startupcrash.json)
05-09 12:22:35.074 5313-5325/com.bugsnag.android.test I/TestRunner: finished: testCancelQueuedFiles(com.bugsnag.android.ErrorStoreTest)
```

Failing output:
```
05-09 12:23:27.134 5605-5617/com.bugsnag.android.test I/TestRunner: started: testCancelQueuedFiles(com.bugsnag.android.ErrorStoreTest)
05-09 12:23:27.134 5605-5605/com.bugsnag.android.test I/MonitoringInstrumentation: Activities that are still in CREATED to STOPPED: 0
05-09 12:23:27.144 5605-5622/com.bugsnag.android.test I/Bugsnag: Sending 1 saved error(s) to Bugsnag
    Deleting sent error file 1525868607149_b1034799-9497-46f1-8072-d1c4b4cdb04f_startupcrash.json
05-09 12:23:27.144 5605-5617/com.bugsnag.android.test I/Bugsnag: Saved unsent payload to disk (/data/data/com.bugsnag.android.test/cache/bugsnag-errors/1525868607149_b1034799-9497-46f1-8072-d1c4b4cdb04f_startupcrash.json)
05-09 12:23:27.144 5605-5617/com.bugsnag.android.test I/TestRunner: failed: testCancelQueuedFiles(com.bugsnag.android.ErrorStoreTest)
```

I think this is more of an issue with invalid assumptions made within our tests rather than anything sinister in the notifier. I'll create a ticket to investigate further which tests exactly are initialising Clients and causing this behaviour.

sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
```diff
@@ -65,15 +67,16 @@ private int makeRequest(String urlString,
...
-                JsonStream stream = new JsonStream(new OutputStreamWriter(out));
+                OutputStream out = conn.getOutputStream();
+                Charset charset = Charset.forName("UTF-8");
+                BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(out, charset));
+                stream = new JsonStream(writer);
                 streamable.toStream(stream);
-                stream.close();
             } finally {
-                IOUtils.closeQuietly(out);
+                IOUtils.closeQuietly(stream);
```
Cawllec: Will closing the `JsonStream` close the `OutputStream` as well?  There's no explicit closing of the `OutputStream`.

sdk/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -83,23 +86,21 @@ String write(@NonNull T streamable) {
...
-            stream.close();
-
             Logger.info(String.format("Saved unsent payload to disk (%s) ", filename));
             return filename;
         } catch (Exception exception) {
             Logger.warn(String.format("Couldn't save unsent payload to disk (%s) ",
                 filename), exception);
         } finally {
-            IOUtils.closeQuietly(out);
+            IOUtils.closeQuietly(stream);
```
Cawllec: Same Q as above.

sdk/src/main/java/com/bugsnag/android/DefaultHttpClient.java
fractalwrench: Yes - also answered in previous comments in a bit more depth.

fractalwrench: Specifically: https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/JsonWriter.java#L423

Cawllec: Ok, is this verifiable or testable at all?  It seems like something that could change across implementations/versions.

fractalwrench: I've added a test to check that JsonWriter closes its underlying stream - beyond that we'd just be testing the behaviour of the Java standard library.

PR #309, opened 2018-05-02
refactor: potential fix for flaky async client notify tests

fractalwrench:
Uses a countdown latch rather than sleeping for 1ms to check whether a method is async. This seems to fail rather consistently at the moment on Travis API 21 emulators, which are faster than the 16/19 ones.

This implementation splits the test into two classes - one with blocking calls, and one with async.

### DIFF
--- a/sdk/src/androidTest/java/com/bugsnag/android/ClientNotifyAsyncTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ClientNotifyAsyncTest.java
@@ -1,5 +1,10 @@
 package com.bugsnag.android;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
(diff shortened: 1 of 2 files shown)


sdk/src/androidTest/java/com/bugsnag/android/ClientNotifyAsyncTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        CountDownLatch nullCheckLatch = new CountDownLatch(1);
+
+        @Override
+        public void postReport(String urlString,
+                               Report report,
+                               Map<String, String> headers)
+            throws NetworkException, BadResponseException {
+            try {
+                nullCheckLatch.await(20, TimeUnit.MILLISECONDS);
```
Cawllec: Is it worth extending this timeout?  AFIACT it won't effect the correct working of the latch, but could prevent issues if the tests are running on an especially slow device/simulator.

PR #311, opened 2018-05-03
[WIP] Delivery api implementation

fractalwrench:


sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -126,12 +124,13 @@ public Client(@NonNull Context androidContext, @NonNull Configuration configurat
 
         ConnectivityManager cm =
             (ConnectivityManager) appContext.getSystemService(Context.CONNECTIVITY_SERVICE);
-        DefaultHttpClient defaultHttpClient = new DefaultHttpClient(cm);
-        errorReportApiClient = defaultHttpClient;
-        sessionTrackingApiClient = defaultHttpClient;
+
+        if (configuration.getDelivery() == null) {
```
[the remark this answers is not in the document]
fractalwrench: The default `Delivery` implementation currently requires a `Context` for construction, which isn't available in `Configuration`. Therefore at this point, the return value would always be null unless a developer had specified a custom implementation.

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -632,20 +629,23 @@ void setUserName(String name, boolean notify) {
     }
 
     @SuppressWarnings("ConstantConditions")
+    @Deprecated
     void setErrorReportApiClient(@NonNull ErrorReportApiClient errorReportApiClient) {
         if (errorReportApiClient == null) {
             throw new IllegalArgumentException("ErrorReportApiClient cannot be null.");
         }
-        this.errorReportApiClient = errorReportApiClient;
+        DeliveryCompat compat = (DeliveryCompat) config.getDelivery();
```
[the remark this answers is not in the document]
fractalwrench: It won't crash in the [current implementation](https://github.com/bugsnag/bugsnag-android/pull/311/files#diff-18f38731c79b15200b4f49b6eb495a3cR521) - but I agree we could use an `instanceof` check here.

fractalwrench:
Updated in response to feedback.

sdk/src/main/java/com/bugsnag/android/Client.java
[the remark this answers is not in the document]
kattrali: It would crash if someone called `setDelivery(new MyCustomDelivery())` and then `setErrorReportApiClient()`, as `getDelivery()` wouldn't have contained the property `errorReportApiClient`. The fix looks good though.

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -1237,16 +1250,21 @@ public void disableExceptionHandler() {
...
-        } catch (NetworkException exception) {
-            Logger.info("Could not send error(s) to Bugsnag, saving to disk to send later");
-
-            // Save error to disk for later sending
-            errorStore.write(error);
-        } catch (BadResponseException exception) {
-            Logger.info("Bad response when sending data to Bugsnag");
+        } catch (DeliveryFailureException exception) {
+            switch (exception.reason) {
+                case CONNECTIVITY:
```
loopj: Style nitpick: the Java convention is to use an exception hierarchy in these situations, any reason we are using an enum instead?

fractalwrench: @loopj it was suggested in the design phase that we use a single exception instead, rather than the previous `NetworkException` and `BadResponseException`. I think it'd make sense to continue using these as we were before with the new `Delivery` interface, as it would get rid of these switch statements and exception mapping code. @kattrali are you happy with this change?

bengourley: Specifically the implementation went in this direction as a response to [this comment](https://bugsnag.atlassian.net/browse/PLAT-1056?focusedCommentId=27929&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27929):

> Should there be a generic `DeliveryFailureException` for outside use? Alternately, should the expected exception types be further documented and/or made generic?

@kattrali I'd be interested to know what the motiviation behind this was, or what your ideal implementation would like like?

I made the suggestion to @fractalwrench that there could be a hierarchy as follows:

```

kattrali: My thought was that we would had a single class for external implementors of `Delivery` to use to rethrow exceptions, not necessarily that we wouldn't throw a subclass of it internally and handle them differently. 

The intent was to reduce the concepts we need to explain for implementors, for example 

> if your `deliver` method fails, throw a `DeliveryFailureException` to retry

that way, they necessarily need to check different types of failures when rethrowing to trigger a later delivery.

fractalwrench: We would need to expose any subclasses of DeliveryFailureException so that the notifier can either save or discard the event.

fractalwrench: I prefer the second option as it feels more intuitive and will reduce the length of the catch block, but I can equally see that throwing a single exception type may make things easier to document.

kattrali: Exposing the subclasses is fine - the distinction just needs to be clear for implementors what will happen when throwing the different types so people know which will retry and which will be discarded.

fractalwrench:
Merging into #299 and resolving exception changes in there.

PR #314, opened 2018-05-10
Enable automatic session tracking by default

fractalwrench:
## Goal

Enables automatic tracking of sessions by default, and adds a new API for setting the notify/session endpoints which deprecates the old method.

## Design

This should prevent the unintentional scenario where a user sets the notify endpoint with a sessions endpoint.

### Added
`config.setEndpoints(String notify, String sessions)`

### Changed
Deprecation notice for:
`config.setEndpoint`
`config.setSessionsEndpoint`
`client.setEndpoint`

Manifest meta-data parsing logic

## Tests

An additional Mazerunner test for ensuring a session is automatically tracked on app launch, and unit tests for ensuring:

- Setting autoCaptureSessions to false prevents a session from being sent at app load
- Manually calling sendSession() sends a session regardless of the state of autoCaptureSessions
- Manually calling sendSession() with endpoint configured but not sessionsEndpoint issues a warning and does not send a session

This doesn't automatically test that configuring endpoint without session endpoint logs a warning, as this is difficult to achieve on Android.

### Outstanding Questions

The deprecated options are not aliased onto the new API - I believe this is ok.

## Review

  - [ ] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -312,21 +312,21 @@ public class Client extends Observable implements Observer {
         config.setAppVersion(data.getString(MF_APP_VERSION));
         config.setReleaseStage(data.getString(MF_RELEASE_STAGE));
 
-        String endpoint = data.getString(MF_ENDPOINT);
-
-        if (endpoint != null) {
-            config.setEndpoint(endpoint);
-        }
-        String sessionEndpoint = data.getString(MF_SESSIONS_ENDPOINT);
-
-        if (sessionEndpoint != null) {
-            config.setSessionEndpoint(sessionEndpoint);
+        if (data.containsKey(MF_ENDPOINT)) {
+            String endpoint = data.getString(MF_ENDPOINT);
+            String sessionEndpoint = data.getString(MF_SESSIONS_ENDPOINT);
+            //noinspection ConstantConditions (pass in null/empty as this function will warn)
+            config.setEndpoints(endpoint, sessionEndpoint);
         }
 
         config.setSendThreads(data.getBoolean(MF_SEND_THREADS, true));
         config.setPersistUserBetweenSessions(
             data.getBoolean(MF_PERSIST_USER_BETWEEN_SESSIONS, false));
-        config.setAutoCaptureSessions(data.getBoolean(MF_AUTO_CAPTURE_SESSIONS, false));
+
+        if (data.containsKey(MF_AUTO_CAPTURE_SESSIONS)) {
+            config.setAutoCaptureSessions(data.getBoolean(MF_AUTO_CAPTURE_SESSIONS));
+        }
+
         config.setEnableExceptionHandler(
             data.getBoolean(MF_ENABLE_EXCEPTION_HANDLER, true));
         return config;
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -390,7 +390,10 @@ public class Client extends Observable implements Observer {
      * endpoint.
      *
      * @param endpoint the custom endpoint to send report to
+     * @deprecated use {@link com.bugsnag.android.Configuration#setEndpoints(String, String)}
+     * instead.
      */
+    @Deprecated
     public void setEndpoint(String endpoint) {
         config.setEndpoint(endpoint);
     }
--- a/sdk/src/main/java/com/bugsnag/android/Configuration.java
+++ b/sdk/src/main/java/com/bugsnag/android/Configuration.java
@@ -136,7 +136,10 @@ public class Configuration extends Observable implements Observer {
      * endpoint.
      *
      * @param endpoint the custom endpoint to send report to
+     * @deprecated use {@link com.bugsnag.android.Configuration#setEndpoints(String, String)}
+     * instead.
      */
+    @Deprecated
     public void setEndpoint(String endpoint) {
         this.endpoint = endpoint;
     }
@@ -175,7 +178,10 @@ public class Configuration extends Observable implements Observer {
      * endpoint.
      *
      * @param endpoint the custom endpoint to send session data to
+     * @deprecated use {@link com.bugsnag.android.Configuration#setEndpoints(String, String)}
+     * instead.
      */
+    @Deprecated
     public void setSessionEndpoint(String endpoint) {
         this.sessionEndpoint = endpoint;
     }
--- a/sdk/src/main/java/com/bugsnag/android/Configuration.java
+++ b/sdk/src/main/java/com/bugsnag/android/Configuration.java
@@ -144,8 +144,23 @@ public class Configuration extends Observable implements Observer {
         this.endpoint = endpoint;
     }
 
+    /**
+     * Set the endpoints to send data to. By default we'll send error reports to
+     * https://notify.bugsnag.com, and sessions to https://session.bugsnag.com, but you can
+     * override this if you are using Bugsnag Enterprise to point to your own Bugsnag endpoint.
+     *
+     * Please note that it is recommended that you set both endpoints. If the notify endpoint is
+     * missing, an exception will be thrown. If the session endpoint is missing, a warning will be
+     * logged and sessions will not be sent automatically.
+     * 
+     * @param notify the notify endpoint
+     * @param sessions the sessions endpoint
+     *
+     * @throws IllegalArgumentException if the notify endpoint is empty or null
+     */
     @SuppressWarnings("ConstantConditions")
-    public void setEndpoints(@NonNull String notify, @NonNull String sessions) {
+    public void setEndpoints(@NonNull String notify, @NonNull String sessions)
+        throws IllegalArgumentException {
         boolean invalidNotify = TextUtils.isEmpty(notify);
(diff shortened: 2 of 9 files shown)


sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -134,11 +136,46 @@ public String getEndpoint() {
...
+        if (invalidNotify) {
+            throw new IllegalArgumentException("Notify endpoint cannot be empty or null.");
+        }
+        if (invalidSessionsEndpoint) {
+            Logger.warn("The session tracking endpoint has not been set. "
+                + "Session tracking is disabled");
+            autoCaptureSessions = false;
+        }
+        this.endpoint = notify;
+        this.sessionEndpoint = sessions;
```
kattrali: if `invalidSessionsEndpoint` is true, then shouldn't this be set to null? Whether to send a session could then be determined by whether the endpoint is null rather than adding a new property.

It would also reduce the odds of unexpected behavior if `Bugsnag.setAutoCaptureSessions()` is called after configuration.

fractalwrench:
Refactored to check that `sessionEndpoint` is null rather than an adding an additional boolean flag, when starting a new sessions.

Also, here is the related Docs PR: https://github.com/bugsnag/docs.bugsnag.com/pull/720

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -312,21 +312,21 @@ static Configuration populateConfigFromManifest(@NonNull Configuration config,
...
+            //noinspection ConstantConditions (pass in null/empty as this function will warn)
+            config.setEndpoints(endpoint, sessionEndpoint);
         }
 
         config.setSendThreads(data.getBoolean(MF_SEND_THREADS, true));
         config.setPersistUserBetweenSessions(
             data.getBoolean(MF_PERSIST_USER_BETWEEN_SESSIONS, false));
-        config.setAutoCaptureSessions(data.getBoolean(MF_AUTO_CAPTURE_SESSIONS, false));
+
+        if (data.containsKey(MF_AUTO_CAPTURE_SESSIONS)) {
```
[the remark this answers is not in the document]
fractalwrench: This method no longer requires any knowledge of whether sessions should be automatically captured or not, and relies on `Configuration` instead to specify a sensible default.

sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -134,11 +135,50 @@ public String getEndpoint() {
...
+        } else {
+            this.endpoint = notify;
+        }
+
+        boolean invalidSessionsEndpoint = TextUtils.isEmpty(sessions);
+
+        if (invalidSessionsEndpoint) {
+            Logger.warn("The session tracking endpoint has not been set. "
+                + "Session tracking is disabled");
+            autoCaptureSessions = false;
```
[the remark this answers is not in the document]
fractalwrench: It's not necessarily required but I think that's a good suggestion as it makes the semantics clearer that `autoCaptureSessions` is a field, and would be more consistent with the rest of the method.

sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -134,11 +135,50 @@ public String getEndpoint() {
...
+            this.sessionEndpoint = null;
+        } else {
+            this.sessionEndpoint = sessions;
+        }
+        this.autoCaptureSessions = !invalidSessionsEndpoint;
```
Cawllec: If this is being set in the if/else above, does it need to be set again here?  It may end up being set to `true` after it's been manually disabled.

fractalwrench: Both are not required, I've removed `autoCaptureSessions = false`

Cawllec: There could be an issue where someone sets `autoCaptureSessions` to false, then inputs a valid sessions endpoint.  The:
```java
this.autoCaptureSessions = !invalidSessionsEndpoint;
```
would then incorrectly set the `autoCaptureSessions` back to true, overriding the user's configuration.  Is that a possibility?

fractalwrench: Good point - I've added an explicit test case for this scenario.

sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -64,6 +64,11 @@
      * @param user the session user (if any)
      */
     void startNewSession(@NonNull Date date, @Nullable User user, boolean autoCaptured) {
+        if (configuration.getSessionEndpoint() == null) {
+            Logger.warn("The session tracking endpoint has not been set. "
```
Cawllec: Should have mentioned this earlier, sorry - Should this warning be a one-time thing? We could end up adding a lot of logger noise.

fractalwrench: The `Logger` class will automatically disable itself in production if that's a concern? The default implementation of session tracking waits at least 30 seconds before starting a new session on Android, so for the default use case I'd imagine this message would only be logged once or twice.

Cawllec: I guess that's ok, are there any similar cases already present in the notifier?

sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -134,11 +135,49 @@ public String getEndpoint() {
...
+     * Please note that it is recommended that you set both endpoints. If the notify endpoint is
+     * missing, an exception will be thrown. If the session endpoint is missing, a warning will be
+     * logged and sessions will not be sent automatically.
+     *
+     * @param notify the notify endpoint
+     * @param sessions the sessions endpoint
+     *
+     * @throws IllegalArgumentException if the notify endpoint is empty or null
+     */
+    @SuppressWarnings("ConstantConditions")
```
martin308: Might be worth adding a comment on the reasoning behind suppressing specific warnings

sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -134,11 +135,49 @@ public String getEndpoint() {
...
+    public void setEndpoints(@NonNull String notify, @NonNull String sessions)
+        throws IllegalArgumentException {
```
martin308: This seems like a big change in behavior? The other methods on configuration just accept whatever was set

fractalwrench: IIRC the design doc for this change states that we should throw an exception if `notify` is not a valid URL, and log a warning if `sessions` is not. The rationale being that the notifier can't really do anything useful if the notify endpoint has been misconfigured, but can still send errors if session tracking is not in use. I think it makes sense to fail fast in this instance if a programmer supplies an invalid value.

sdk/src/main/java/com/bugsnag/android/SessionTracker.java
fractalwrench: @Cawllec it's not particularly easy to test log messages in Android unfortunately

sdk/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Added an inline comment

Cawllec:
Looks good, just the CI that's not working ATM.

fractalwrench:
CI has passed now (the test flakiness seems to be back on Travis)

sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -135,11 +136,49 @@ public String getEndpoint() {
...
+     * Please note that it is recommended that you set both endpoints. If the notify endpoint is
+     * missing, an exception will be thrown. If the session endpoint is missing, a warning will be
+     * logged and sessions will not be sent automatically.
+     *
+     * @param notify the notify endpoint
+     * @param sessions the sessions endpoint
+     *
+     * @throws IllegalArgumentException if the notify endpoint is empty or null
+     */
+    @SuppressWarnings("ConstantConditions") // ignore checkstyle nullability warnings
```
martin308: I was thinking more along the lines of the reasoning behind suppressing it. I'm not sure what this does exactly but here is a possibly terrible example "Suppress this warning as we know that the "notify" parameter can be null due to this amazing reason... but we can ignore it because of some other scenario..."

fractalwrench: I've removed this annotation as it appears it's no longer needed. I believe it was originally included because a null check was performed on `notify` and `sessions`, which are annotated as non-null. Checkstyle picks up the nullability annotation, but unfortunately this isn't enforced by the compiler for Java, so it made sense to perform a null check.

This code now uses `TextUtils.isEmpty`, which also checks for an empty string along with null, which isn't a constant condition.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 4.X.X (TBD)
+
+**Note**: this release alters the behaviour of the notifier to track session automatically. If you
```
bengourley: I'm reusing this copy for the js changelog and noticed a little typo:

```diff
+ track session automatically
- track sessions automatically
```

PR #317, opened 2018-05-18
build: Automatically update dex / size / count badge during release

kattrali:
Invoke ~arcane magic~ awk to update the badge for library sizes when running the release and bump tasks. Also added more checks to ensure the git status is clean before running.


kattrali:
Its even more magical than that, running `make VERSION=[number] release` will run it automatically ✨

Issue #318, opened 2018-05-23
Crashes aren't being reported immediately on Android P


mattprecious:
### Expected behavior
Crashes on P are reported

### Observed behavior
Crashes on P are not reported

### Steps to reproduce
Crash the app on P

### Version
4.4.0

### Additional information
It seems that P is being more aggressive in killing processes after a crash.

The error is persisted to disk successfully [here](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/Client.java#L929), but the process is killed before `flushReports` is called asynchronously [here](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/ErrorStore.java#L100).

When the app is started again, [`flushOnLaunch`](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/ErrorStore.java#L53) is called but `flushAsync` is never invoked because the stored report is being filtered out from `findLaunchCrashReports()` [here](https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/ErrorStore.java#L54).

It seems that, as implemented, the only way to successfully report persisted crashes is to disable the launch crash threshold.

mattprecious:
It eventually came in... I updated the app and it reported the crash from 40m prior. Not sure what hook was invoked for that.

kattrali:
Thanks for the report, @mattprecious. It looks like we'll need to make some adjustments to account for reports not being sent at crash time.

fractalwrench:
Fixed in v4.4.1

PR #319, opened 2018-05-24
Android P FileStore flush fix

fractalwrench:
## Goal

Some unhandled error reports aren't being sent immediately on Android P and in situations with no connectivity.

Internally `FileStore` retains a `Collection` of files which have previously been found for a request, to prevent duplicate reports being sent. This puts the onus on the calling code to cancel or delete the files once the request has been completed. 

This changeset ensures that these files are only requested once when calling `flushOnLaunch`, then cancels all files after delivery has taken place (successfully or not), which resets the `FileStore` state.

After this, `flushAsync` is then called, which will flush any remaining reports that have not yet been delivered.

### Changed

`FileStore` (see above description)

## Tests

Manual test on Android P simulator of unhandled crashes. Mazerunner scenario that fails prior to the applied fix.

### Alternative Approaches

The delivery/flushing code is fairly complex, it may be worth rethinking our implementation at some future point.

### Linked issues

#318 

## Review

  - [x] Final review

### DIFF
--- a/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/StartupCrashFlushScenario.kt
+++ b/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/StartupCrashFlushScenario.kt
@@ -1,0 +1,42 @@
+package com.bugsnag.android.mazerunner.scenarios
+
+import android.content.Context
+import android.os.Handler
+import com.bugsnag.android.*
+
+/**
+ * Generates an uncaught exception, catches it, and persists it to disc, preventing any delivery.
+ *
+ * To generate a startup crash, set "eventMetaData" to "StartupCrash", otherwise the default
+ * behaviour is to report a normal crash.
+ *
+ * The mazerunner scenario that tests this works in 3 parts:
+ *
+ * 1. Generate and persist a normal crash on disk, by waiting longer than the threshold for
+ *    a startup crash
+ * 2. Generate and persist a startup crash on disk, by crashing as soon as the scenario starts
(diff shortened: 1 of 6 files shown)


fractalwrench:
I've updated the mazerunner scenario to throw exceptions rather than writing errors to the store.

I've also made a slight adjustment to the `ErrorStore` logic so that if `config.launchCrashThresholdMs == 0`, all reports will be flushed async, as launch crash delivery behaviour is disabled. This addresses an internal ticket.

mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/StartupCrashFlushScenario.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+internal class StartupCrashFlushScenario(config: Configuration,
+                                         context: Context) : Scenario(config, context) {
+
+    override fun run() {
+        super.run()
+        disableAllDelivery()
+
+        if ("StartupCrash" == eventMetaData) {
+            config.launchCrashThresholdMs = Long.MAX_VALUE
```
snmaynard: I think this needs some rethinking. I think you dont want to set `config.launchCrashThresholdMs` in here. Also both of these crashes will count as launch crashes on subsequent boots.

fractalwrench:
Updated the docs to make it a bit clearer what this scenario is doing, and also updated mazerunner to disable delivery before Bugsnag is initialised.

kattrali:
Reworked the test to better illustrate the user scenario here, and introduced a delay to separate the launch crash from a regular one.

PR #324, opened 2018-06-04
Trim very long stacktraces

### LINKED ISSUE #322: StackOverflowException fails to be reported because of SSLException writing it
### Expected behavior
StackOverflowException that occurs is reported to bugsnag

### Observed behavior
Bugsnag crashes silently when trying to report the issue

### Steps to reproduce
Cause a StackOverflowException with an enormous stacktrace (mine has 506,000 StackTraceelements

### Version
4.4.0

### Additional information
The exception happens here: https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/Stacktrace.java#L37

This is the exception: `javax.net.ssl.SSLException: Write error: ssl=0x8c2bd1c0: I/O error during system call, Broken pipe`

And it is likely caused because the `stacktrace` variable in that code has 506,000 elements and writing all of those is itself causing an out of space issue. I think ideally you max out how many elements we try to print?

Short of that, printing to syster err doesnt so anything to help us on Android, as I dont think those get printed to logcat or anything. If possible bugsnag should itself crash if theres a problem I would think, so its easier to catch and report issues like this

fractalwrench:
## Goal

Trims very long stacktraces by ensuring that only the first 200 stack frames are serialised. This should ensure that for a stacktrace with 10^6 stackframes, the notifier will still deliver unhandled errors, and the pipeline won't drop the request.

Additionally, `Logger` is used rather than `printStacktrace`. The former is redirected by Android to logcat, but a message limit can result in
an SSLException being thrown for large messages. `Logger` uses logcat internally and is disabled in production by default, which should prevent these issues affecting users.

### Changed

- Stacktrace serialisation now ends after 200 frames have been processed.
- Throwables are put through the `Logger` rather than `printStacktrace`.

## Tests

Added a mazerunner scenario and JUnit test.

### Outstanding Questions

I've picked 200 frames as it seemed like a reasonable number. We may want to determine whether this sounds sensible for all our notifiers, if we want the same trimming behaviour across platforms.

### Linked issues
Fixes #322 

## Review

  - [ ] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/ExceptionHandler.java
+++ b/sdk/src/main/java/com/bugsnag/android/ExceptionHandler.java
@@ -2,6 +2,7 @@ package com.bugsnag.android;
 
 import android.os.StrictMode;
 import android.support.annotation.NonNull;
+import android.util.Log;
 
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.util.Map;
@@ -91,7 +92,7 @@ class ExceptionHandler implements UncaughtExceptionHandler {
             originalHandler.uncaughtException(thread, throwable);
         } else {
             System.err.printf("Exception in thread \"%s\" ", thread.getName());
-            throwable.printStackTrace(System.err);
+            Logger.warn("Exception", throwable);
         }
     }
 }
--- a/sdk/src/main/java/com/bugsnag/android/Stacktrace.java
+++ b/sdk/src/main/java/com/bugsnag/android/Stacktrace.java
@@ -38,7 +38,7 @@ class Stacktrace implements JsonStream.Streamable {
 
                 writer.endObject();
             } catch (Exception lineEx) {
-                Logger.warn("Failed to serialise stacktrace", lineEx);
+                Logger.warn("Failed to serialize stacktrace", lineEx);
             }
         }
 
--- a/sdk/src/main/java/com/bugsnag/android/Stacktrace.java
+++ b/sdk/src/main/java/com/bugsnag/android/Stacktrace.java
@@ -9,6 +9,9 @@ import java.io.IOException;
  * where appropriate.
  */
 class Stacktrace implements JsonStream.Streamable {
+
+    private static final int STACKTRACE_TRIM_LENGTH = 200;
+
     final Configuration config;
     final StackTraceElement[] stacktrace;
 
@@ -21,7 +24,8 @@ class Stacktrace implements JsonStream.Streamable {
     public void toStream(@NonNull JsonStream writer) throws IOException {
         writer.beginArray();
 
-        for (StackTraceElement el : stacktrace) {
+        for (int k = 0; k < stacktrace.length && k < STACKTRACE_TRIM_LENGTH; k++) {
+            StackTraceElement el = stacktrace[k];
             try {
                 writer.beginObject();
                 writer.name("method").value(el.getClassName() + "." + el.getMethodName());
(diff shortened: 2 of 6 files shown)


sdk/src/main/java/com/bugsnag/android/Stacktrace.java
```diff
@@ -21,7 +24,8 @@
     public void toStream(@NonNull JsonStream writer) throws IOException {
         writer.beginArray();
 
-        for (StackTraceElement el : stacktrace) {
+        for (int k = 0; k < stacktrace.length && k < STACKTRACE_TRIM_LENGTH; k++) {
```
bengourley: my understanding of this change is that it omits anything beyond the first 200 values in that stacktrace when serializing. How does it work if the report fails to send? Does the serialized report get saved to storage or does the original report entity get saved? What I'm trying to get at here is whether an unruly stacktrace will be reallocated some memory and/or persistent storage… or is it just passed in by reference and always operated on in its initially allocated memory?

sdk/src/main/java/com/bugsnag/android/Stacktrace.java
```diff
@@ -9,6 +9,9 @@
  * where appropriate.
  */
 class Stacktrace implements JsonStream.Streamable {
+
+    private static final int STACKTRACE_TRIM_LENGTH = 200;
```
bengourley: 200 seems reasonable. Have you done any size analysis to see what the typical payload size an error with 200 stackframes would be? If it's consistently under 1MB then I think that's good, but if it's over 200 then maybe we should bring it down a little to ensure the reports don't get rejected by the API?

sdk/src/main/java/com/bugsnag/android/ExceptionHandler.java
```diff
@@ -91,7 +92,7 @@ public void uncaughtException(@NonNull Thread thread, @NonNull Throwable throwab
             originalHandler.uncaughtException(thread, throwable);
         } else {
             System.err.printf("Exception in thread \"%s\" ", thread.getName());
-            throwable.printStackTrace(System.err);
+            Logger.warn("Exception", throwable);
```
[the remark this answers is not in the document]
fractalwrench: Android redirects stderr to its own logger, but there's a limit on the message size which means larger methods won't be logged. `Logger` is also disabled by default in production, where it might not make sense to log this information.

sdk/src/main/java/com/bugsnag/android/Stacktrace.java
fractalwrench: If a report fails to send it will be persisted to disk as a file using the same `toStream` method. That file will then be resent without any knowledge of its contents.

sdk/src/main/java/com/bugsnag/android/Stacktrace.java
fractalwrench: The stacktrace from the new mazerunner scenario weighs in at about 30kb, and the payload as a whole at about 40kb. Note: that's pretty-printed JSON collected from mazerunner, so in reality we can expect those sizes to be about 25% smaller.

fractalwrench:
I believe this is ready for further review.

sdk/src/main/java/com/bugsnag/android/Stacktrace.java
```diff
@@ -34,7 +38,7 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
 
                 writer.endObject();
             } catch (Exception lineEx) {
-                lineEx.printStackTrace(System.err);
+                Logger.warn("Failed to serialise stacktrace", lineEx);
```
Pezzah: Not American enough

PR #330, opened 2018-06-18
WIP: Android Unity Changes

fractalwrench:
This is the branch where all changes to Android required to support the Unity C# layer will land.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/InternalApi.java
+++ b/sdk/src/main/java/com/bugsnag/android/InternalApi.java
@@ -1,0 +1,19 @@
+package com.bugsnag.android;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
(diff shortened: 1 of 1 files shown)


fractalwrench:
@kattrali I'm not 100% clear on what that last bit means - did you have an example in mind?

kattrali:
`Client.getAppData()` returns an object with various fields on it but no way to get a key/value mapping of all of the included values. To fetch all values from another library would be something like:

```java
AppData appData = client.getAppData();
// Create a Map map
map.put("buildUUID", appData.getBuildUuid());
map.put("duration", appData.getDuration());
map.put("releaseStage", appData.getReleaseStage());
// etc
```

It would be easier to use this interface if `Client.getAppData()` returned a `Map` of all values or have `AppData.toMap()` or similar:

```java
Map appData = client.getAppData().toMap();
```

Using the existing interface, I'd be concerned that if we add a new app field, it wouldn't be picked up unless the other library's code explicitly updated to fetch the new field.

kattrali:
Is this still `wip`?

fractalwrench:
It's ready as far as I'm concerned - I'm assuming no further changes are required from the Unity side.

PR #331, opened 2018-06-18
Add public accessors to AppData and AppDataSummary classes

fractalwrench:
## Goal

Adds public accessors to AppData and AppDataSummary, for use by 3rd party developers and internal API (denoted with marker annotation).

Note: the intention is for further work to be carried out in a separate PR (e.g. migrating legacy fields to metadata, extracting the data capture to a separate class, etc)

### Changed

Converted fields/methods to properties (getter/setter backed with a field) in AppData and AppDataSummary.

## Tests

Adapted unit tests for JSON serialisation of each class, and added a test for the default value + override value of each accessor.

## Review

  - [ ] Final review

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -28,7 +28,7 @@ android {
         targetSdkVersion Integer.parseInt(project.ANDROID_TARGET_SDK_VERSION)
 
         manifestPlaceholders = [
-            bugsnagApiKey                    : "your-api-key",
+            bugsnagApiKey                    : "572dd1154e6c5b291e5d1b42e506ae34",
             bugsnagBuildUUID                 : "abc123",
             bugsnagAppVersion                : "1.0.0",
             bugsnagEndpoint                  : "https://notify.bugsnag.com",
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -28,7 +28,7 @@ android {
         targetSdkVersion Integer.parseInt(project.ANDROID_TARGET_SDK_VERSION)
 
         manifestPlaceholders = [
-            bugsnagApiKey                    : "572dd1154e6c5b291e5d1b42e506ae34",
+            bugsnagApiKey                    : "your-api-key",
             bugsnagBuildUUID                 : "abc123",
             bugsnagAppVersion                : "1.0.0",
             bugsnagEndpoint                  : "https://notify.bugsnag.com",
--- a/sdk/src/main/java/com/bugsnag/android/AppData.java
+++ b/sdk/src/main/java/com/bugsnag/android/AppData.java
@@ -53,21 +53,22 @@ public class AppData extends AppDataSummary {
     public void toStream(@NonNull JsonStream writer) throws IOException {
         writer.beginObject();
         serialiseMinimalAppData(writer);
-
         writer.name("id").value(packageName);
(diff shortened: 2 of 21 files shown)


fractalwrench:
@martin308 I made a start with these two classes. Let me know if you're happy with the general approach to tests/docs, as I can split things apart into separate PRs if it makes easier for review.

My intention at the minute is to split out further work in separate PRs (e.g. adding accessors on the client object, refactoring the data capture)

sdk/src/main/java/com/bugsnag/android/AppData.java
```diff
@@ -34,7 +41,12 @@
         this.appContext = appContext;
         this.sessionTracker = sessionTracker;
         appName = getAppName(appContext);
-        packageName = getPackageName(appContext);
+
+        packageName = calculatePackageName(appContext);
+        buildUuid = config.getBuildUUID();
+        durationMs = getDurationMs();
+        foregroundMs = sessionTracker.getDurationInForegroundMs(System.currentTimeMillis());
+        inForeground = sessionTracker.isInForeground();
```
martin308: is there an issue with pre calculating these fields that they won't be true at the time we read the data?

martin308:
@fractalwrench thanks for the preview! I added one question around pre calculating some of the fields but apart from that it looks good to me

sdk/src/main/java/com/bugsnag/android/AppData.java
fractalwrench: For the purposes of Unity, I don't think so. I believe the intention is that the Unity layer will call something like: `client.generateAppData()` which will instantiate a new object each time.

For Android, I'd anticipate that these values could be slightly different from previous versions, as they should now be calculated at the time of report capture, rather than at the point of serialisation. I think on average it would be different by the order of milliseconds, and worst case, seconds. We can add a general note about this - the values will be more correct IMO.

example/build.gradle
```diff
@@ -28,7 +28,7 @@ android {
         targetSdkVersion Integer.parseInt(project.ANDROID_TARGET_SDK_VERSION)
 
         manifestPlaceholders = [
-            bugsnagApiKey                    : "your-api-key",
+            bugsnagApiKey                    : "572dd1154e6c5b291e5d1b42e506ae34",
```
martin308: Should this be checked in?

sdk/src/main/java/com/bugsnag/android/AppData.java
```diff
@@ -34,31 +41,36 @@
...
     }
 
+    // TODO migrate metadata fields to separate class
+    void addAppMetaData(MetaData metaData) {
+        metaData.addToTab("app", "name", appName);
+        metaData.addToTab("app", "packageName", packageName);
+        metaData.addToTab("app", "versionName", getVersionName());
+        metaData.addToTab("app", "activeScreen", getActiveScreenClass());
+        metaData.addToTab("app", "memoryUsage", getMemoryUsage());
+        metaData.addToTab("app", "lowMemory", isLowMemory(appContext));
```
martin308: How would the unity notifier access these values?

fractalwrench: Comment addressed in #333

PR #333, opened 2018-06-19
Device data Android Unity changes

fractalwrench:
## Goal

Adds public accessors to DeviceData and DeviceDataSummary, for use by 3rd party developers and internal API (denoted with marker annotation).

Note: the intention is for further work to be carried out in a separate PR (e.g. migrating legacy fields to metadata, extracting the data capture to a separate class, etc)

### Changed

Converted fields/methods to properties (getter/setter backed with a field) in DeviceData and DeviceDataSummary.

## Tests

Adapted unit tests for JSON serialisation of each class, and added a test for the default value + override value of each accessor.

## Review

  - [ ] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/DeviceData.java
+++ b/sdk/src/main/java/com/bugsnag/android/DeviceData.java
@@ -34,6 +34,18 @@ class DeviceData extends DeviceDataSummary {
 
     private static final String INSTALL_ID_KEY = "install.iud";
 
+
+    @Nullable
+    private String id;
+    private long freeMemory;
+    private long totalMemory;
+    private Long freeDisk;
+
+    @Nullable
+    private String orientation;
+
+
+
     @Nullable
     final Float screenDensity;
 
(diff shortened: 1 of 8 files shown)


sdk/src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -70,46 +83,115 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
...
+        metaData.addToTab("device", "brand", Build.BRAND);
+        metaData.addToTab("device", "apiLevel", Build.VERSION.SDK_INT);
+        metaData.addToTab("device", "osBuild", Build.DISPLAY);
+        metaData.addToTab("device", "locale", locale);
+        metaData.addToTab("device", "screenDensity", screenDensity);
+        metaData.addToTab("device", "dpi", dpi);
+        metaData.addToTab("device", "emulator", isEmulator());
+        metaData.addToTab("device", "screenResolution", screenResolution);
+        metaData.addToTab("device", "cpuAbi", cpuAbi);
+    }
```
martin308: How were you thinking that the unity notifier would access all of these properties?

fractalwrench: For the Android notifier, these need to be added to `error.metadata`. My thinking is that we could expose an internal method on `Client` that allows for the population of this metadata. Something like:

```

PR #338, opened 2018-06-19
feat: add public accessors to error class

fractalwrench:
## Goal

Adds public accessors to Error, for use by 3rd party developers and internal API, and nullability annotations.

Note: the intention is for further work to be carried out in a separate PR (extracting the data capture to a separate class, etc)

### Changed

Converted fields/methods to properties (getter/setter backed with a field) in Error.

## Tests

Adapted unit tests for the default value + override value of each accessor.

## Review

  - [ ] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Error.java
+++ b/sdk/src/main/java/com/bugsnag/android/Error.java
@@ -2,7 +2,6 @@ package com.bugsnag.android;
 
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
-import android.text.TextUtils;
 
 import java.io.IOException;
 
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -112,7 +134,7 @@ public void setContext(String context) {
      */
     @Nullable
     public String getContext() {
-        if (!TextUtils.isEmpty(context)) {
+        if (context != null) {
```
martin308: what's the intent of this change?

fractalwrench: This was unintentional - I've reverted back to the TextUtils.isEmpty approach which is more robust.

PR #339, opened 2018-06-20
Refactor App data collection

fractalwrench:
## Goal

To separate the collection of application data from its serialisation, supporting the Unity C# layer changes.

## Changeset

- Remove all methods from `AppData` and `AppDataSummary` which aren't accessors or serialisers
- Create `AppDataCollector`, which collects data for `AppData`, `AppDataSummary`, and app `MetaData`.
- Reduce number of packageManager lookups + cache other suitable values in `AppDataCollector` compared to `AppData`
- Generate a new `AppData` instance for each error report.
- Remove error's dependency on `AppData#getActiveScreenClass()`
- Add `Client#populateAppMetaData(MetaData metaData)` method for accessing extra app data as internal API

## Tests

Ran mazerunner and existing unit test suite. Added unit tests for new methods.

## Review

  - [ ] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/AppData.java
+++ b/sdk/src/main/java/com/bugsnag/android/AppData.java
@@ -1,10 +1,5 @@
 package com.bugsnag.android;
 
-import android.app.ActivityManager;
-import android.content.Context;
-import android.content.pm.ApplicationInfo;
-import android.content.pm.PackageManager;
-import android.os.SystemClock;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/com/bugsnag/android/AppData.java
```diff
@@ -15,15 +10,7 @@
...
-    private static final long startTimeMs = SystemClock.elapsedRealtime();
-
-    @Nullable
-    final String appName;
-
-    @NonNull
-    private final Context appContext;
-    private final SessionTracker sessionTracker;
-
+    @SuppressWarnings("NullableProblems") // initialised after construction
```
fractalwrench: Warnings are suppressed as we're anticipating that this field will always be set to a non-null value after construction. This will allow users to access the value in a Kotlin-esque way, i.e.:

`appData.packageName`

Rather than requiring the safe call operator each time to unwrap the optional:

`appData?.packageName`

PR #343, opened 2018-06-28
Disable BuildConfig generation

JakeWharton:
This contains a few fields and a class which is useless to consumers and unused by the library. It's removed by ProGuard and R8, but by not even generating it those bytes are saved for all consumers.

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -29,6 +29,11 @@ android {
         warningsAsErrors true
         checkAllWarnings true
     }
+
+    // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
+    libraryVariants.all {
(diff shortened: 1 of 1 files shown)


f2prateek:
I can see one potential use for the BuildConfig here, e.g. https://github.com/bugsnag/bugsnag-android/blame/7cc8a23577300da02a327cd5b1ed7ca4910325c6/sdk/src/main/java/com/bugsnag/android/Notifier.java#L12 could reference `BuildConfig.VERSION_NAME` instead of hardcoding the version.

JakeWharton:
True. I only looked for existing usages. It's a shame it results in a public type that pollutes consumers instead of something private to the module.

PR #344, opened 2018-07-03
Unity serialisation changes

fractalwrench:
## Goal

Simplify bridging of metadata for Unity and other notifiers.

## Changeset

- Update serialisation unit tests (removed setter tests as this would just be testing `java.util.Map`)
- Add `mapToJson` utility for constructing JSON in tests
- Combine `AppData` + `AppDataSummary` into one class
- Combine `DeviceData` + DeviceDataSummary` into one class
- Remove public getters/setters/fields on above classes, and return a `Map` instead

Reduce the number of accessors on `Client` to:

```
AppData getAppData();
DeviceData getDeviceData();
```

Add the following accessors on AppData/DeviceData:

```
Map<String, Object> getAppData()
Map<String, Object> getAppDataSummary()
Map<String, Object> getDeviceData()
Map<String, Object> getDeviceDataSummary()
```

- Update `Error` and `SessionTrackingPayload` to serialise a map
- Update `NativeInterface`, `Client`, other misc classes to access information from map
- Combined SessionTrackingPayload constructors to reduce method count

Existing JSON serialisation code in `MetaData` has been refactored into a separate class called `ObjectJsonStreamer`. This serialises the `Map` fields in the Error and Session payloads.

## Tests

Ran unit tests and mazerunner scenario. Inspected error payload manually in dashboard.

## Review

  - [ ] Final review


sdk/src/androidTest/java/com/bugsnag/android/ClientTest.java
```diff
@@ -303,28 +303,17 @@ public void testBreadcrumbStoreNotModified() {
...
-        assertNotNull(appDataSummary);
-        assertNotEquals(appData.getAppDataSummary(), appData);
-    }
-
     @Test
     public void testAppDataMetaData() {
         Client client = generateClient();
         Map<String, Object> app = client.getAppData().getAppDataMetaData();
-        assertEquals(5, app.size());
+        assertEquals(6, app.size());
```
kattrali: What is the 6th item? Should there be an assertion for that?

fractalwrench: The `codeBundleId` (which is null by default). I can add an assertion for this.

fractalwrench: I was thinking of the wrong object - it's actually `activeScreen`, which will be null in the unit test. I've added a check to ensure the key is present in the map.

PR #345, opened 2018-07-04
Specify consumer Proguard files

fractalwrench:
## Goal

Adds ProGuard keep rules to the AAR artefact. This means that the library will work correctly when minification is enabled, without relying on the bugsnag gradle plugin to edit the library's proguard files.

## Design

This allows users to avoid using the bugsnag gradle plugin if they so wish, without any additional configuration of ProGuard on their part.

## Changeset

Added a proguard file with keep rules for the relevant classes.

## Tests

Used Android Studio's APK analyzer to verify that an example project without the gradle plugin applied retained the correct classes in its Dex file.

## Discussion

- We may want to consider the task in our gradle plugin which adds these rules automatically. The risk of this is that it could break builds for anyone who doesn't 
read the release notes.
- Do we need to update [the Docs](https://docs.bugsnag.com/build-integrations/gradle/)? They seemed ok to me.

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
  - [x] Final review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Disable BuildConfig generation [#343](https://github.com/bugsnag/bugsnag-android/pull/343)
 
+* Add consumer proguard rules for automatic ProGuard configuration without the Bugsnag gradle plugin [#345](https://github.com/bugsnag/bugsnag-android/pull/345)
+
 ## 4.5.0 (2018-06-18)
 
 This release alters the behaviour of the notifier to track sessions automatically. 
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -16,6 +16,7 @@ android {
     defaultConfig {
         minSdkVersion Integer.parseInt(project.ANDROID_MIN_SDK_VERSION)
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles 'proguard-rules.pro'
     }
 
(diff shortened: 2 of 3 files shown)


kattrali:
> We may want to consider the task in our gradle plugin which adds these rules automatically. The risk of this is that it could break builds for anyone who doesn't
read the release notes.

What can we do to make this transition smoother? Can the Gradle plugin be changed to conditionally add the Bugsnag rules (based on the version of bugsnag-android in use or something)?

> Do we need to update the Docs? They seemed ok to me.

Seems ok here too.

fractalwrench:
>What can we do to make this transition smoother? Can the Gradle plugin be changed to conditionally add the Bugsnag rules (based on the version of bugsnag-android in use or something)?

It should be possible to check the dependency at runtime within the gradle plugin, using something along the following lines:

```
    project.configurations.forEach { conf ->
        conf.allDependencies.forEach { dep ->
            if (dep.group == "com.bugsnag" && dep.name == "bugsnag-android") {
                println("Dep: " + dep)
            }
        }
    }
```

If it makes sense we could grab the lowest version from the dependencySet (after de-duping etc), and only write to the ProGuard file if the version is less than 4.X.X.

kattrali:
Sounds good to me 🤔

fractalwrench:
Added in: https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/116

PR #346, opened 2018-07-16
feat: update proguard rules to keep line number/source info

fractalwrench:
## Goal

Automatically keep source file and line number information with ProGuard, as otherwise grouping is likely to differ for existing projects.

### Linked issues

Related to #345 

## Review

  - [ ] Final review


kattrali:
Is there a way for consumers of the library to disable this rule, such as when more concerned with obfuscation than error reporting?

fractalwrench:
Yes, using the `renamesourcefileattribute` attribute can bypass this.

kattrali:
That may be something to document. LGTM.

PR #354, opened 2018-08-01
v4.6.0

fractalwrench:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## 4.6.0 (2018-08-02)
+
+* Android P compatibility fixes - ensure available information on StrictMode violations is collected [#350](https://github.com/bugsnag/bugsnag-android/pull/350)
+
+* Disable BuildConfig generation [#343](https://github.com/bugsnag/bugsnag-android/pull/343)
+
+* Add consumer proguard rules for automatic ProGuard configuration without the Bugsnag gradle plugin [#345](https://github.com/bugsnag/bugsnag-android/pull/345)
+
+* Internal refactor of app/device data serialisation
```
kattrali: This isn't a user-facing change

fractalwrench: I felt that due to the scope of these changes it would be sensible to include a note - if that's not something we want to do then I can change that in future.

PR #355, opened 2018-08-02
Capture trace of error reporting thread and identify with boolean flag

fractalwrench:
## Goal

Captures the trace of `Thread.currentThread()`, which was previously omitted, and to identify it with a boolean flag in the serialised JSON.

## Changeset

- Alter `ThreadState` to stop removing the current thread from serialisation
- Conditionally write `errorReportingThread` boolean flag by checking thread ID

## Tests

Added unit tests for general ThreadState serialisation, and for checking that the current thread is now captured. Ran mazerunner locally, tested with dashboard.

## Review

  - [x] Final review


sdk/src/main/java/com/bugsnag/android/ThreadState.java
```diff
@@ -15,34 +15,26 @@
 class ThreadState implements JsonStream.Streamable {
     private static final String THREAD_TYPE = "android";
 
-    final Configuration config;
+    private final Configuration config;
     private final Thread[] threads;
     private final Map<Thread, StackTraceElement[]> stackTraces;
+    private final long currentThreadId;
```
snmaynard: Why do we now store this?

fractalwrench: It's not guaranteed that `toStream` will be called on the same thread as the constructor. As we need to identify the thread where the error was originally captured, the solution I chose was to store the capturing thread ID during construction, then use it to conditionally serialise `errorReportingThread` during serialisation.

kattrali: The constructor is not guaranteed to run on the error reporting thread either. In the case of unhandled errors, the exception handler can be on a different thread altogether and this will only report the `cacheAndNotify` call and generating the `ThreadState`. The `uncaughtException()` hook provides the crashed thread which should instead be used to generate the information for the error reporting thread.

sdk/src/main/java/com/bugsnag/android/ThreadState.java
```diff
@@ -15,34 +15,26 @@
...
     /**
      * Returns an array of threads excluding the current thread, sorted by thread id
      *
-     * @param currentThreadId the current thread id
-     * @param liveThreads     all live threads
+     * @param liveThreads all live threads
      */
-    private Thread[] sanitiseThreads(long currentThreadId,
-                                     Map<Thread, StackTraceElement[]> liveThreads) {
+    private Thread[] sanitiseThreads(Map<Thread, StackTraceElement[]> liveThreads) {
```
kattrali: The doc comment (and possibly method name) should be updated to reflect the current behavior.

fractalwrench:
Thanks for the review @kattrali - I've updated `ThreadState` so that it now requires a `Thread` parameter in its constructor, which is populated via the `thread` value from `uncaughtException`, and `Thread.currentThread` for `notify` calls. This should address the wrong thread being selected for unhandled exceptions.

>For handled exceptions, I never see any thread marked as the reporting thread on the Threads tab.

I was not able to reproduce this behaviour, did you have a code example that will trigger this?

fractalwrench:
I've updated the thread serialisation to capture the threads at the point of report generation, rather than serialisation. I've also made sure that we pass in the current `Thread` to `ThreadState` rather than assuming `Thread.currentThread`, as this value will often be wrong.

kattrali:
> I was not able to reproduce this behaviour, did you have a code example that will trigger this?

Nothing special, just fired up the example app with the handled exception buttons. It still happens sometimes.

sdk/src/main/java/com/bugsnag/android/ExceptionHandler.java
```diff
@@ -77,12 +77,12 @@ public void uncaughtException(@NonNull Thread thread, @NonNull Throwable throwab
                 StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.LAX);
 
                 client.cacheAndNotify(throwable, Severity.ERROR,
-                    metaData, severityReason, violationDesc);
+                    metaData, severityReason, violationDesc, thread);
```
kattrali: This bit needs adjustment; since the thread has crashed, it may get recycled in the pool to do something else, depending on which thread crashed. Other threads are in an approximate state relative to when the exception occurred given that some may have started or died in the interim, which is expected. 

However, all of the required elements are here to construct the expected result. For unhandled reports, the `errorReportingThread` should have:

1. `thread.getId()`
2. `thread.getName()`
3. `exception.getStackTrace()`

The easiest way to accomplish this may be to replace the stack for the error reporting thread to use the exception stack if the exception is unhandled.

fractalwrench:
>Nothing special, just fired up the example app with the handled exception buttons. It still happens sometimes.

I've tried pressing each button in the example app 10x on an API 27 emulator. For all of the 70 received events _"Error reported from this thread"_ is present on one entry in the threads tab. Do I need to increase the frequency to reproduce this issue, or is there some sort of code change required?

One potential cause I can think of would be that `Thread.getAllStackTraces()` only returns alive threads, and that the `thread` parameter passed into `uncaughtException` might be terminated, and therefore not in the map.

If this is the case then we might be able to record the thread ID, name, and a zero-element array stacktrace by adding the terminated thread into the set of live threads. However, I'm still not sure how this could explain why this behaviour was observed for a blocking `notify` call - I suspect there's something subtly different in our environments and it would be good to get to the bottom of this.

>However, all of the required elements are here to construct the expected result. For unhandled reports, the errorReportingThread should have: exception.getStackTrace()

I discussed this with Duncan at project kick-off and believe the requested behaviour was to capture the thread trace for both handled/unhandled errors, rather than the exception stacktrace. I'll run this past the necessary Slack channel before making the change

fractalwrench:
This behaviour is reproducible on API 25 not API 27, investigating further

sdk/src/androidTest/java/com/bugsnag/android/ThreadStateTest.kt
```diff
@@ -0,0 +1,160 @@
...
+            expectedTrace.forEachIndexed { index, element ->
+                val jsonObject = serialisedTrace.getJSONObject(index)
+                assertEquals(element.lineNumber, jsonObject.getInt("lineNumber"))
+            }
+        }
+    }
+
+    private fun verifyCurrentThreadStructure(json: JSONArray,
+                                             currentThreadId: Long,
+                                             action: ((thread: JSONObject) -> Unit)? = null) {
```
fractalwrench: Allows passing a function that will be invoked when the errorReportingThread json is found. This means assertions can be run with less boilerplate finding the necessary thread

sdk/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -399,8 +399,13 @@ HandledState getHandledState() {
...
 
-        Builder(@NonNull Configuration config, @NonNull Throwable exception, Session session) {
-            this.threadState = new ThreadState(config);
+        Builder(@NonNull Configuration config,
+                @NonNull Throwable exception,
+                @Nullable Session session,
+                @NonNull Thread thread,
+                boolean unhandled) {
+            Throwable exc = unhandled ? exception : null;
+            this.threadState = new ThreadState(config, thread, Thread.getAllStackTraces(), exc);
```
fractalwrench: If an exception is passed in (default for unhandled errors), the exception stacktrace will be serialised rather than the thread stacktrace

fractalwrench:
There appears to be [a bug](https://issuetracker.google.com/issues/64122757) in API levels 24/25 where `Thread.getAllStacktraces()` does not actually capture all threads, which meant the `errorReportingThread` would not be serialised at all on those levels. This led to the observed behaviour with handled exceptions not showing the expected information in the dashboard.

I have applied a workaround for this by capturing the stacktrace of the `errorReportingThread` directly, via `Thread.getStackTrace()`. However, it's worth noting that the rest of the affected threads will _not be serialised_ on Android 7 devices. I'm not aware of a suitable workaround for this at the present time.

I have also altered the thread serialisation so that the exception stacktrace is used for the errorReportingThread, for unhandled errors only.

Testing included:

- Full Mazerunner suite on API 21
- Addition of new unit tests and running on API 16, 19, 21, 25, 27
- Manual tests on API 25/27 (inspected thread tab on dashboard):
  - Handled + unhandled exception button in app
  - Wrapping handled + unhandled exception button code with `Thread {}.start()` and confirming `errorReportingThread` differed
- Self-review of changeset

fractalwrench:
Added an additional mazerunner scenario, which has passed locally (CI build is blocked again but I believe this should be ready for review.)

sdk/src/main/java/com/bugsnag/android/ThreadState.java
simonbowring: Just checking, is this resolved now?

fractalwrench: @simonbowring yes, this is resolved.

features/error_reporting_thread.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Error Reporting Thread
+
+Scenario: Only 1 thread is flagged as the error reporting thread
+    When I run "HandledExceptionScenario" with the defaults
```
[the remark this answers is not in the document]
fractalwrench: yes, good point.

fractalwrench:
@simonbowring this is the requested behaviour for handled events - unhandled events will show the exception stacktrace rather than the thread trace.

PR #356, opened 2018-08-03
refactor: Moves configuration creation for manifest to separate class

fractalwrench:
## Goal

This changeset is intended to reduce the complexity of the `Client` class, which is currently 757 SLOC. This is partially achieved by extracting the logic to populate `Configuration` from the AndroidManifest to a separate class.

## Changeset

- Moves code which populates a configuration object from the AndroidManifest into a separate `ConfigFactory` class
- Update test imports to pass build

## Tests

Ran unit tests, mazerunner

## Review

  - [x] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -1,5 +1,6 @@
 package com.bugsnag.android;
 
+import static com.bugsnag.android.ConfigFactory.MF_BUILD_UUID;
 import static com.bugsnag.android.MapUtils.getStringFromMap;
 
 import android.app.Activity;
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -258,94 +242,6 @@ public void update(Observable observable, Object arg) {
...
-
-    /**
-     * Populates the config with meta-data values supplied from the manifest as a Bundle.
-     *
-     * @param config the config to mutate
-     * @param data   the manifest bundle
-     * @return the updated config
-     */
-    @NonNull
-    static Configuration populateConfigFromManifest(@NonNull Configuration config,
```
martin308: what's the default access modifier that's applied here? Does moving this have any external implications?

fractalwrench: The package-level modifier will be applied, so anything with the `com.bugsnag.android` namespace can access this. I don't believe this has any external implications as I'm not aware of any other notifiers which call this method.

PR #357, opened 2018-08-06
Consolidate kotlin/bugsnag plugin versions

fractalwrench:
Consolidates the kotlin/bugsnag plugin versions in one place, and updates dependencies to the latest expected by Android Studio stable. This prevents compiler warnings about the build tools being out of date, and will also make it easier to update the bugsnag/kotlin plugins from one place in future.


gradle.properties
```diff
@@ -29,6 +29,5 @@ POM_ARTIFACT_ID=bugsnag-android
 POM_PACKAGING=aar
 ANDROID_MIN_SDK_VERSION=14
 ANDROID_TARGET_SDK_VERSION=27
-ANDROID_BUILD_TOOLS_VERSION=27.0.0
+ANDROID_BUILD_TOOLS_VERSION=27.0.3
```
fractalwrench: min supported version for current stable version of Android Studio

mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyStacktraceScenario.kt
```diff
@@ -15,6 +16,7 @@ internal class EmptyStacktraceScenario(config: Configuration,
         Bugsnag.notify(EmptyException("EmptyStacktraceScenario"))
     }
 
+    @SuppressLint("NewApi")
```
fractalwrench: AGP uses a newer version of android lint which detects this Throwable constructor is only available from API 19 onwards, however, we don't currently run mazerunner on that API level.

PR #359, opened 2018-08-15
Set maxBreadcrumbs via Configuration rather than Client

fractalwrench:
## Goal

Provides `config.setMaxBreadcrumbs` in favour of `client.setMaxBreadcrumbs`, which is now deprecated.

## Changeset

- Deprecates `client.setMaxBreadcrumbs` and `Bugsnag.setMaxBreadcrumbs`
- Add `maxBreadcrumbs` property to `Configuration`
- Convert `BreadcrumbsTest` to Kotlin, documenting what each test case attempts to verify, and adding new tests

## Tests

Ran unit tests, mazerunner locally, performed manual testing on dashboard around breadcrumb limits. Added new unit tests for accessor behaviour, and documented existing tests.

## Review

  - [x] Final review


sdk/src/main/java/com/bugsnag/android/Breadcrumbs.java
```diff
@@ -33,32 +38,25 @@ void clear() {
...
+            pruneBreadcrumbs();
         } catch (IOException ex) {
             Logger.warn("Dropping breadcrumb because it could not be serialized", ex);
         }
     }
+
+    private void pruneBreadcrumbs() {
+        int maxBreadcrumbs = configuration.getMaxBreadcrumbs();
+
+        // Remove oldest breadcrumb
```
simonbowring: This comment just needs to be tweaked to state that the oldest breadcrumbs will be removed until the max size is reached, not just one breadcrumb

PR #360, opened 2018-08-17
Catch Exception within DefaultDelivery class

fractalwrench:
## Goal

Certain devices such as Blackberries seem to throw exceptions other than IOException within `DefaultDelivery`. This change catches the unexpected failures and drops the payload, while logging a warning, to prevent bugsnag-android causing crashes.

## Changeset

Added a catch clause

## Tests

Ran existing unit tests, mazerunner locally

## Discussion

This behaviour has not been reproduced locally, and appears to be a device-specific issue. It's possible that this could be some inherent issue with the Blackberry repackaging process.

## Review

  - [x] Final review

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/DefaultDelivery.java
+++ b/sdk/src/main/java/com/bugsnag/android/DefaultDelivery.java
@@ -79,6 +79,9 @@ class DefaultDelivery implements Delivery {
             return conn.getResponseCode();
         } catch (IOException exception) {
             throw new DeliveryFailureException("IOException encountered in request", exception);
+        } catch (Exception exception) {
+            Logger.warn("Unexpected error delivering payload", exception);
+            return 400;
         } finally {
             IOUtils.close(conn);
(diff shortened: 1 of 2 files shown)


kattrali:
As this is just a fix, it should be against master rather than `next`.

sdk/src/main/java/com/bugsnag/android/DefaultDelivery.java
```diff
@@ -79,6 +79,9 @@ int deliver(String urlString,
             return conn.getResponseCode();
         } catch (IOException exception) {
             throw new DeliveryFailureException("IOException encountered in request", exception);
+        } catch (Exception exception) {
+            Logger.warn("Unexpected error delivering payload", exception);
+            return 400;
```
kattrali: This should be made a constant, to make it more declarative why `400` is being used here.

fractalwrench:
Moved to #361 to change base branch against master

PR #371, opened 2018-09-04
Run mazerunner scenarios on CI

fractalwrench:
## Goal

Run mazerunner scenarios on CI.

## Changeset

- Adds mazerunner scenarios to Travis CI environment matrix, running on API 21 emulator
- Alter scenarios to relaunch the app rather than start the activity manually. This adds a pre-requisite wait of X seconds that if not present would cause scenarios to fail.
- Add wait time of 15s after launching activity when running on Travis CI, due to slowness of emulator
- Reduce specificity of StrictMode scenarios (format of exception message differs across API levels)
- Make session scenarios more deterministic by flushing sessions (`SessionSetAutoCaptureScenario` tests that a session is delivered naturally without these test hooks)

## Tests

Ran tests on CI.

### Alternative Approaches

Running suite locally, or only on specific builds, rather than on every commit.

### Outstanding Questions

Running mazerunner scenarios on Travis-CI is really slow as we are limited to ARM emulators. I'm quite concerned that in the long-term this approach will prevent us from running against different API targets, and in the short-term may lead to further congestion of our build queue.

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
  - [x] Final review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
 source "https://rubygems.org"
 
-gem 'bugsnag-maze-runner', git: 'git@github.com:bugsnag/maze-runner'
+gem 'bugsnag-maze-runner', :git => 'https://github.com/bugsnag/maze-runner'
 gem 'pry'
--- a/features/strict_mode.feature
+++ b/features/strict_mode.feature
@@ -5,7 +5,6 @@ Scenario: StrictMode DiscWrite violation
     Then I should receive a request
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "android.os.StrictMode$StrictModeViolation"
-    And the exception "message" equals "policy=262145 violation=1"
     And the event "metaData.StrictMode.Violation" equals "DiskWrite"
 
 
@@ -14,5 +13,4 @@ Scenario: StrictMode Network on Main Thread violation
     Then I should receive a request
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "android.os.StrictMode$StrictModeViolation"
-    And the exception "message" equals "policy=262148 violation=4"
     And the event "metaData.StrictMode.Violation" equals "NetworkOperation"
--- a/mazerunner/src/main/java/com/bugsnag/android/TestHarnessHooks.kt
+++ b/mazerunner/src/main/java/com/bugsnag/android/TestHarnessHooks.kt
@@ -2,12 +2,18 @@ package com.bugsnag.android
 
 import android.content.Context
 import android.net.ConnectivityManager
+import android.os.Handler
+import android.os.HandlerThread
 
 /**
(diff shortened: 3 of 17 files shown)


Gemfile
```diff
@@ -1,4 +1,4 @@
 source "https://rubygems.org"
 
-gem 'bugsnag-maze-runner', git: 'git@github.com:bugsnag/maze-runner'
+gem 'bugsnag-maze-runner', :git => 'https://github.com/bugsnag/maze-runner'
```
fractalwrench: allows Travis to clone the repo

features/strict_mode.feature
```diff
@@ -5,7 +5,6 @@ Scenario: StrictMode DiscWrite violation
     Then I should receive a request
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "android.os.StrictMode$StrictModeViolation"
-    And the exception "message" equals "policy=262145 violation=1"
```
fractalwrench: The policy value differs on the Travis CI emulator as it's a different API level, and all we really care about here is that the metadata value was set

mazerunner/src/main/java/com/bugsnag/android/TestHarnessHooks.kt
```diff
@@ -2,12 +2,18 @@ package com.bugsnag.android
...
  * Accesses the session tracker and flushes all stored sessions
  */
-internal fun flushAllSessions() {
-    Bugsnag.getClient().sessionTracker.flushStoredSessions()
+internal fun flushAllSessions(client: Client) {
+    val thread = HandlerThread("HandlerThread")
+    thread.start()
+    Handler(thread.looper).post {
+        client.sessionTracker.flushStoredSessions()
+    }
```
fractalwrench: Avoids performing network operations on the main thread, that could have caused scenarios to fail intermittently

mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/CustomClientSessionFlushScenario.kt
```diff
@@ -14,14 +14,16 @@ internal class CustomClientSessionFlushScenario(config: Configuration,
                                                 context: Context) : Scenario(config, context) {
 
     override fun run() {
```
fractalwrench: This scenario now only creates one session, as it is supposed to test that a manual session is cached on disk then delivered.

mazerunner/src/main/java/com/bugsnag/android/TestHarnessHooks.kt
kattrali: I don't entirely understand what this function is for; aren't sessions delivered automatically?

fractalwrench:
Closing as superseded by work in #369

PR #374, opened 2018-09-28
Fix Travis CI build

fractalwrench:
In the bugsnag-android-ndk repository, the `NEXUS_USERNAME` and `NEXUS_PASSWORD` credentials were supplied as an [envar on Travis](https://github.com/bugsnag/bugsnag-android-ndk/blob/master/build.gradle#L96). This changeset alters the SDK + NDK build script to use the previous method of supplying the username/password, which passes the lint/unit test steps.

The mazerunner scenarios seem to have passed one job, but are having a bit of trouble with timeouts. I will try restarting the jobs, but this would be worth keeping an eye on.

### DIFF
--- a/ndk/build.gradle
+++ b/ndk/build.gradle
@@ -85,8 +85,8 @@ publishing {
                 url 'https://oss.sonatype.org/service/local/staging/deploy/maven2/'
             }
             credentials {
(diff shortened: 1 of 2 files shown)


kattrali:
Balance has been restored to the force.

PR #375, opened 2018-10-01
v4.8.2

fractalwrench:
- Increments the version to 4.8.2 for next release
- Update changed instructions for publishing artefact to local maven repository
- Fix Kotlin example app by loading NDK library and calling into native functions


examples/sdk-app-example/src/main/res/layout/main.xml
```diff
@@ -35,7 +35,6 @@
       style="@style/ButtonTheme"
       android:layout_width="match_parent"
       android:layout_height="wrap_content"
-      android:onClick="crashWithNativeSignal"
```
fractalwrench: this is redundant as a click listener is set programmatically within the mainactivity

PR #378, opened 2018-10-17
fix(signal): Fix stacktrace resolution for native crashes on ARM32

kattrali:
## Goal

When a signal is raised on 32-bit ARM, the current context is the signal
stack rather than the crash stack. To work around this, set the register
state before unwinding the first frame (using the program counter as the
first frame). Then the stack can be unwound normally.

## Changeset

* Add `signal_context` to `bsg_libunwind_state`
* Set the `signal_context` to the unwind function signal context (which may be null, as it is in the event of using `notify()`
* Use `state.signal_context` in the unwind callback to set the initial register state

## Tests

Manually tested against:

* Nexus 4 (ARM32) running API 21
* Nexus 5P (ARM32) on API 24
* Pixel 2 (ARM64) on API 28
* Tested against various x86/x64 emulators to make sure the architecture macros are correct

Automated test verifying the correct stacktrace (file & method) for a given crash

## Review

The following items are needed:

- [ ] Reproducing the initial problem - Signal-based crashes on ARM32 using v4.8.2 (not using `abort()`) should produce an uninformative stacktrace without the original frames
- [ ] Confirm consistency between the changeset and the goal stated above.
- [ ] Explore and discover any crash cases where this change results in a worse stacktrace (or no stacktrace 😱 )


ndk/src/main/jni/utils/string.h
```diff
@@ -27,6 +28,12 @@ void bsg_strncpy(char *dst, char *src, size_t len) __asyncsafe;
  * Copy a string from src to dst, null padding the rest
  */
 void bsg_strncpy_safe(char *dst, char *src, int dst_size);
+
+/**
+ * True if first and second are the same length and contain the same characters
+ * in the same order.
+ */
+bool bsg_strequal(char *first, char *second) __asyncsafe;
```
fractalwrench: I can't find any usage of this function in the project. We should remove it if it's not used here, or add it in a separate PR if it's needed more generally.

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXExternalStackElementScenario.java
```diff
@@ -0,0 +1,31 @@
...
+        config.setAutoCaptureSessions(false);
+    }
+
+    @Override
+    public void run() {
+        super.run();
+        String metadata = getEventMetaData();
+        if (metadata != null && metadata.equals("non-crashy")) {
+            return;
+        }
```
fractalwrench: ~30 files in the `scenario` package contain the same 4 lines as above, and nearly every scenario contains `config.setAutoCaptureSessions(false);`.

I think it'd be good to think about whether there are any abstractions/sensible defaults we could use to reduce this duplication. This could be addressed as a separate ticket and shouldn't hold up the work in this PR.

features/native_crash_handling.feature
```diff
@@ -190,3 +190,22 @@ Feature: Native crash reporting
         And the event "severity" equals "error"
         And the event "unhandled" is true
 
+    Scenario: Causing a crash in a separate library
```
fractalwrench: This scenario fails on an x86, API 27 emulator, with the following error message:

>And the "method" of stack frame 0 equals "something_innocuous"                                                                 # maze-runner-24f53756011b/lib/features/steps/error_reporting_steps.rb:103
      Expected: "something_innocuous"
        Actual: "bsg_unwind_stack_libunwind" (Minitest::Assertion)
      features/native_crash_handling.feature:208:in `And the "method" of stack frame 0 equals "something_innocuous"'

ndk/src/main/jni/utils/stack_unwinder_libunwind.c
```diff
@@ -25,11 +26,21 @@ bsg_libunwind_callback(struct _Unwind_Context *context, void *arg) __asyncsafe {
   bsg_libunwind_state *state = (bsg_libunwind_state *)arg;
 
 #if defined(__arm__)
```
fractalwrench: I found the intro to the PR quite helpful in understanding how this works. It'd be good to include the explanation as a comment so that future travellers have the same context about the workaround.

ndk/src/main/jni/utils/stack_unwinder_libunwind.c
```diff
@@ -1,4 +1,5 @@
 #include "../utils/build.h"
+#include "../utils/string.h"
```
fractalwrench: this include doesn't appear to be used?

fractalwrench:
I've left a few comments inline. A couple of other items that may need addressing include:

- CI is not passing for this PR
- The new mazerunner scenario does not pass on an x86, API 27 emulator

I've tested that this works by running the mazerunner scenario on an ARMv7, API 22 emulator. With the changes the scenario passes and the frames are as expected; without the changes, the stackframe is equal to the signal stack (and does contain a frame referencing libmonochrome, around 4 frames down).

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXExternalStackElementScenario.java
kattrali: Its still helpful to know for every test case exactly what is going on, and then not fight it if for some reason its needed in a more complex case down the line.

kattrali: To add some additional context, say we refactor this into a method, what then would it save? Maybe a line at most?

```java
if (shouldNotCrash()) {
  return;
}
```

fractalwrench: Personally I think a method like `shouldNotCrash` would improve readability. The reader doesn't have to know what event metadata is, why it needs a null-check, or where 'non-crashy' has been set - they only need to know that the author didn't want the scenario to crash at this point.

Additionally, we wouldn't have to worry about updating 30 separate source files if we need to make a change to this feature, and could also add some brief JavaDoc to explain the convention of setting 'non-crashy' as an Environment Variable within a feature.

This doesn't need to be addressed as part of this PR but I do feel it's worth creating a ticket to address this at some future point.

snmaynard: I have taken a look at this and I think that my main problem with it is the name of the variable its stored in. It's too easily confused with event metadata. If we made this called something obviously environment variable like then this code seems fine to me.

PR #381, opened 2018-10-31
feat: Allow setting context to null from beforeSend

kattrali:
## Changeset

* Moved automatic context detection from during serialization to capture-time, which is in `Client`
* Deleted a test for the context logic in Error

## Tests

* [x] Add new test scenario for overriding context
* (Other scenarios cover automatic context assignment and setting context to a new value)

## Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Allow setting context to null from callbacks
+  [#381](https://github.com/bugsnag/bugsnag-android/pull/381)
+
 ## 4.9.0 (2018-10-29)
 
 ### Enhancements
--- a/sdk/src/androidTest/java/com/bugsnag/android/ErrorTest.java
+++ b/sdk/src/androidTest/java/com/bugsnag/android/ErrorTest.java
@@ -315,25 +315,10 @@ public class ErrorTest {
     }
 
     @Test
-    public void testConfigContext() throws Exception {
-        String expected = "Junit test suite";
-        error.setContext(null);
-        config.setContext(expected);
-        assertEquals(expected, error.getContext());
-    }
-
-    @Test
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Allow setting context to null from callbacks
+  [#380](https://github.com/bugsnag/bugsnag-android/pull/380)
```
fractalwrench: The PR should be 381

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -838,6 +838,10 @@ void notify(@NonNull Error error,
         // Attach user info to the error
         error.setUser(user);
 
+        // Attach default context from active activity
+        String context = config.getContext();
+        error.setContext(context != null ? context : appData.getActiveScreenClass());
```
fractalwrench: This needs to check whether the context has already been set on the error object, and avoid setting the field if this is the case. For example, the context will be wrong for the following `notify` method on `Client`:

```
Bugsnag.getClient().notify("Test", "exc_message", "CustomContext", new StackTraceElement[]{}, Severity.ERROR, new MetaData());
```

sdk/src/androidTest/java/com/bugsnag/android/ErrorTest.java
```diff
@@ -314,28 +314,13 @@ public void testBugsnagExceptionName() throws Exception {
         assertEquals("Busgang", err.getExceptionName());
     }
 
-    @Test
-    public void testConfigContext() throws Exception {
-        String expected = "Junit test suite";
-        error.setContext(null);
-        config.setContext(expected);
```
fractalwrench: This test could be reworked so that `config.setContext` is called before the error is built.

Issue #380, opened 2018-10-28
NDK stacktrace is not detailed enough


peterdk:
### Expected behavior
I get NDK stacktraces from crashes within the JNI code, but unfortunately it doesnt show the correct line of the crash. It only shows the linenumber where the method itself ends, and doesnt show where in the method something goes wrong.

I would expect to pinpoint the exact line where the crash happens

### Observed behavior
It shows the line where the method itself ends, not the crash location exactly. Also it says 'unknown symbol'.

### Steps to reproduce
I added a test nullpointer crash inside the method

### Version
3.8.2
armeabi-v7a 
Gradle uploads mapping and all .so file mappings correctly.

### Additional information
```
SIGILL: Illegal instruction
        at bsg_unwind_stack_libunwind(/data/app/FOO/lib/arm/libbugsnag-ndk.so:27562)
        at bsg_unwind_stack(/data/app/FOO/lib/arm/libbugsnag-ndk.so:26862)
        at bsg_handle_signal(/data/app/FOO/lib/arm/libbugsnag-ndk.so:26316)
        at 0xf4c6996c(/system/lib/libc.so:100716)
        at unknown_symbol(C:/Users/Peter/git/FOO/src/main/jni/FileCache.cpp:62)
        at unknown_symbol(C:/Users/Peter/git/FOO/src/main/jni/SamplerSynth.cpp:32)
```
```
MemoryStream * FileCache::getBackedStream(int note)
{
	if (note < 0 || note >= cacheSize)
	{
		ALOGD("Note %d is larger then stream cache size %d", cacheSize);
		return NULL;
	}

	MemoryStream * test = NULL;
	__umito_log_print(LOG_INFO, "FileCache","%d", test->getBytesOutputted()); //CRASH
    if (backedStreamCaches != NULL && backedStreamCaches[note] != 0)
    {
        int streamLength = backedStreamLengths[note];
		return new MemoryStream(backedStreamCaches[note], streamLength, false);
    }else{
        __umito_log_print(LOG_INFO, "FileCache", "Returning NULL stream for note %d: no cached stream available", note);
        return NULL;
    }
}//Line 62 is this line
```
I also tried compiling without -O2, and with -fno-omit-frame-pointer -funwind-tables -Wl,--no-merge-exidx-entries. No different results there.

peterdk:
Same issue on x86.

kattrali:
[...]

PR #383, opened 2018-11-03
fix(ndk): Ensure 64-bit devices running 32-bit binaries do not use 32-bit device unwinding logic

### LINKED ISSUE #382: Regression: 4.9.0 NDK stacktraces are 1 line only
### Expected behavior
Get full stacktrace  when having a native crash.
```
SIGSEGV: Segmentation violation (invalid memory reference)
        at bsg_unwind_stack_libunwind(/data/app/FOO/lib/arm/libbugsnag-ndk.so:27562)
        at bsg_unwind_stack(/data/app/FOO/lib/arm/libbugsnag-ndk.so:26862)
        at bsg_handle_signal(/data/app/FOO/lib/arm/libbugsnag-ndk.so:26316)
        at art::FaultManager::HandleFault(int, siginfo*, void*)(/system/lib/libart.so:3534022)
        at 0xb6c73514(/system/lib/libc.so:95508)
        at unknown_symbol(C:/Users/Peter/git/FOO/src/main/jni/FileCache.cpp:45)
        at unknown_symbol(C:/Users/Peter/git/FOO/src/main/jni/NativeSampler.cpp:112)
        at unknown_symbol(C:/Users/Peter/git/FOO/src/main/jni/SamplerSynth.cpp:32)
```
### Observed behavior
I get for the same crashes now a single line stacktrace. (all my crashes have single lines now)
```
SIGSEGV: Segmentation violation (invalid memory reference)
        at unknown_symbol(C:/Users/Peter/git/FOO/src/main/jni/FileCache.cpp:45)
```
### Steps to reproduce
Use 4.9.0 and crash in NDK code.

### Version
4.9.0
Previous version used: 4.8.2

### Additional information

kattrali:
Small refactor of the API 21+ stack unwinding logic to restrict 32-bit
unwinding style from 64-bit devices. 

While the stack enhancements added in #378 are intended only to be run
on 32-bit devices and gated to the __arm__ build, when 64-bit devices
are running the 32-bit library, the stacktraces for those devices were
unintentionally affected. 

This is the case when an app is configured to only run 32-bit binaries 
(Android cannot run both 64-bit and 32-bit binaries at once), and many
popular libraries (e.g. React Native) only distribute 32-bit libraries.

Added additional improvement to the 32-bit stacktrace resolution by
switching to nongnu libunwind only for those devices. This significantly
improves the number of frames which can be unwound without risk of a
segfault.

## Tests

* Tested against ARM 32-/64-bit devices against `raise(signum)`, `abort()`, floating point exceptions, and invalid casts. With the exception of `abort()` the stacktrace is improved.
* Sanity checked against x86/x64 to ensure the logic is restricted correctly

To reproduce the original problem add ABI filters to an apps gradle config to only load 32-bit binaries:

```groovy
android {
  // ...
  defaultConfig {
    // Ensure no 64-bit libs are loaded/preferred over the 32-bit counterparts
    packagingOptions {
        exclude "lib/arm64-v8a/libbugsnag-ndk.so"
        // Add `exclude` lines for any other included library
    }

    ndk {
      abiFilters "armeabi-v7a", "x86"
    }
  }
}
```

### Linked issues

Fixes #382 
Related to #380

## Review

- [ ] Testing additional esoteric C crashes to compare the unwinding results across devices/architectures

### DIFF
--- a/ndk/src/main/CMakeLists.txt
+++ b/ndk/src/main/CMakeLists.txt
@@ -24,6 +24,7 @@ add_library( # Specifies the name of the library.
 include_directories(
              jni/
              jni/deps
+             jni/external/libunwind/include
              )
 
 target_include_directories(bugsnag-ndk PRIVATE ${BUGSNAG_DIR}/assets/include)
@@ -47,3 +48,9 @@ set_target_properties(bugsnag-ndk
                       COMPILE_OPTIONS
                       -Werror -Wall -pedantic)
 
+if(${ANDROID_ABI} STREQUAL "armeabi" OR ${ANDROID_ABI} STREQUAL "armeabi-v7a")
+    add_library(libunwind STATIC IMPORTED)
+    set_target_properties(libunwind PROPERTIES IMPORTED_LOCATION
+        ${ANDROID_NDK}/sources/cxx-stl/llvm-libc++/libs/${ANDROID_ABI}/libunwind.a)
+    target_link_libraries(bugsnag-ndk libunwind)
+endif()
--- a/ndk/src/main/jni/utils/stack_unwinder_libunwind.c
+++ b/ndk/src/main/jni/utils/stack_unwinder_libunwind.c
@@ -4,6 +4,10 @@
 #include <report.h>
 #include <unwind.h>
 
+#if defined(__arm__)
+#include <libunwind.h>
+#endif
+
 typedef struct {
   size_t frame_count;
   uintptr_t frame_addresses[BUGSNAG_FRAMES_MAX];
--- a/ndk/src/main/jni/utils/stack_unwinder_libunwind.c
+++ b/ndk/src/main/jni/utils/stack_unwinder_libunwind.c
@@ -11,16 +11,14 @@
 typedef struct {
   size_t frame_count;
   uintptr_t frame_addresses[BUGSNAG_FRAMES_MAX];
-#if defined(__arm__)
-  void *signal_context;
-#endif
-  siginfo_t *signal_info;
 } bsg_libunwind_state;
 
 bsg_libunwind_state *bsg_global_libunwind_state;
+bool bsg_libunwind_global_is32bit = false;
 
-bool bsg_configure_libunwind(void) {
+bool bsg_configure_libunwind(bool is32bit) {
   bsg_global_libunwind_state = calloc(1, sizeof(bsg_libunwind_state));
+  bsg_libunwind_global_is32bit = is32bit;
   return true;
 }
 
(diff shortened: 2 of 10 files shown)


ndk/src/main/jni/utils/stack_unwinder_libunwind.c
```diff
@@ -1,75 +1,26 @@
...
   siginfo_t *signal_info;
 } bsg_libunwind_state;
 
-bsg_libunwind_state *bsg_global_libunwind_state;
-
-bool bsg_configure_libunwind(void) {
-  bsg_global_libunwind_state = calloc(1, sizeof(bsg_libunwind_state));
-  return true;
-}
+bool bsg_configure_libunwind(void) { return true; }
```
fractalwrench: is this needed if we're linking the library?

ndk/src/main/jni/utils/stack_unwinder_libunwind.c
```diff
@@ -83,19 +34,69 @@ bsg_libunwind_callback(struct _Unwind_Context *context, void *arg) __asyncsafe {
...
+    unw_set_reg(&cursor, UNW_ARM_R8, signal_mcontext->arm_r8);
+    unw_set_reg(&cursor, UNW_ARM_R9, signal_mcontext->arm_r9);
+    unw_set_reg(&cursor, UNW_ARM_R10, signal_mcontext->arm_r10);
+    unw_set_reg(&cursor, UNW_ARM_R11, signal_mcontext->arm_fp);
+    unw_set_reg(&cursor, UNW_ARM_R12, signal_mcontext->arm_ip);
+    unw_set_reg(&cursor, UNW_ARM_R13, signal_mcontext->arm_sp);
+    unw_set_reg(&cursor, UNW_ARM_R14, signal_mcontext->arm_lr);
+    unw_set_reg(&cursor, UNW_ARM_R15, signal_mcontext->arm_pc);
+    unw_set_reg(&cursor, UNW_REG_IP, signal_mcontext->arm_pc);
+    unw_set_reg(&cursor, UNW_REG_SP, signal_mcontext->arm_sp);
```
fractalwrench: The previous implementation had a block comment that was fairly helpful at understanding what this code was doing - I'm having a harder time following this method without that sort of documentation. 

It's not clear to me why we now need to set the register values for `UNW_REG_IP` and `UNW_REG_SP`, and why the single-frame mode has been removed if there's a risk of unwinding causing a segfault.

ndk/src/main/CMakeLists.txt
```diff
@@ -47,3 +52,9 @@ set_target_properties(bugsnag-ndk
                       COMPILE_OPTIONS
                       -Werror -Wall -pedantic)
 
+if(${ANDROID_ABI} STREQUAL "armeabi" OR ${ANDROID_ABI} STREQUAL "armeabi-v7a")
```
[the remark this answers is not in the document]
kattrali: This library only applies to 32-bit and doesn’t work for 64-bit.

ndk/src/main/jni/utils/stack_unwinder_libunwind.c
kattrali: All of the unwinders have a config method for consistency.

ndk/src/main/jni/utils/stack_unwinder_libunwind.c
kattrali: > It's not clear to me why we now need to set the register values for UNW_REG_IP and UNW_REG_SP,

This is now using a different library which provides the option of setting additional registers for IP and SP, so I set those as well. 

> if there’s a risk of unwinding causing a segfault

There’s no longer a risk so it’s not needed. The underlying implementation is more stable.

kattrali:
Reworked the implementation a bit to:

1. Add additional inline documentation
2. Use the ABI list we already have to switch unwinding logic. Its better than linking an additional dependency.

ellsworthrw:
This version no longer sends reports for native crashes.

PR #393, opened 2018-12-11
Improve kotlin support by allowing property access

fractalwrench:
## Goal

If public getters/setters are available for a Java field, then this code can be called from Kotlin using the [property access syntax](https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters). This gives users the choice of using a more Kotlin-esque syntax:

```
config.autoCaptureSessions = false
```

while retaining backward compatibility with the previous interface:

```
config.setAutoCaptureSessions(false)
```

## Changeset

- A public getter and setter were added for the `autoCaptureSessions`, `codeBundleId`, and `notifierType` properties.
- `shouldAutoCaptureSessions()` was deprecated and forward onto `getAutoCaptureSessions()`
- Internal use of `shouldAutoCaptureSessions()` was changed to `getAutoCaptureSessions()` to avoid using deprecated code
- All `KotlinPropertyAccess` checks were removed from the lint_baseline file, which suppresses warnings.
- The lint_baseline file was regenerated to account for new violations added in #389, and changes in line numbers.

## Tests

Ran existing unit tests + mazerunner scenarios on CI.

### DIFF
--- a/sdk/lint-baseline.xml
+++ b/sdk/lint-baseline.xml
@@ -2,39 +2,6 @@
 <issues format="4" by="lint 3.2.1" client="gradle" variant="all" version="3.2.1">
 
     <issue
-        id="KotlinPropertyAccess"
-        message="This method should be called `getAutoCaptureSessions` such that `autoCaptureSessions` can be accessed as a property from Kotlin; see https://android.github.io/kotlin-guides/interop.html#property-prefixes"
-        errorLine1="    public boolean shouldAutoCaptureSessions() {"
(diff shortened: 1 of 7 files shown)


fractalwrench:
To summarise internal discussion, there are 4 options:

`getAutoCaptureSessions`, `isAutoCaptureSessions`, renaming the Java method to `isAutoCapturingSessions` and losing the property access syntax, or doing nothing.

Of these, `getAutoCaptureSessions` is the preferred option.

PR #394, opened 2018-12-11
Fix calculation of durationInForeground when autoCaptureSessions is false

fractalwrench:
## Goal
If `autoCaptureSessions` is set to false, the `app.durationInForeground` value always equals 0 in reports.

Additionally, if the app is put into the background and then brought back into the foreground within a 30 second threshold, the `durationInForeground` value is not reset, and can be erroneously large.

This changeset fixes these two issues.

## Design
Please see the design doc for a detailed overview.

## Tests

Added a mazerunner scenario to verify that `durationInForeground` is > 0 when `autoCaptureSessions` is false. Additionally, manual testing checked the following scenarios, for both potential values of `autoCaptureSessions`:

- Handled/Unhandled exceptions have non-zero value for `durationInForeground`
- Putting the activity in the background (by clicking the home button then recent tasks button) immediately resets the `durationInForeground` field


Pezzah:
Can there be a test for the application sending an error while it is not in the foreground, and check that durationInForeground is set to zero

PR #398, opened 2018-12-11
Add nullability annotations to Configuration

fractalwrench:
See #395 for an in-depth explanation.


sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -336,7 +344,7 @@ public void setNotifyReleaseStages(@Nullable String[] notifyReleaseStages) {
      *
      * @param projectPackages a list of package names
      */
-    public void setProjectPackages(String[] projectPackages) {
+    public void setProjectPackages(@NonNull String[] projectPackages) {
```
bengourley: the getter for this is `@Nullable` but the setter is `@NonNull`. As is `get|setNotifyReleaseStages(…)`. Is this intentional? Can they be set to `null` elsewhere internally, but not via the setter?

fractalwrench:
Comments/remaining work will be addressed in #395

PR #399, opened 2018-12-12
Add nullability annotations to public NativeInterface methods 

fractalwrench:
See #395 for an in-depth explanation.


sdk/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -318,57 +330,59 @@ public static String getEndpoint() {
...
      */
+    @NonNull
     public static String getAppVersion() {
         return getClient().getConfig().getAppVersion();
     }
 
     /**
      * Return which release stages notify
      */
+    @NonNull
```
bengourley: It looks like me like this can be `null` as this annotation was not changed in the config branch:
https://github.com/bugsnag/bugsnag-android/pull/398/files#diff-18f38731c79b15200b4f49b6eb495a3cR305

fractalwrench:
Comments/remaining work will be addressed in #395

PR #403, opened 2018-12-20
Prevent Bugsnag.init from instantiating more than one client

fractalwrench:
## Goal

If `Bugsnag.init` is called multiple times, multiple `Client` objects will be instantiated. This is generally not the desired behaviour, could lead to undesired side-effects, and is also a very expensive operation (approx 100ms for each initialisation on a test device).

For a real world example, if `Bugsnag.init` is called from within `onCreate` in an `Activity` rather than an `Application` subclass, it's possible that multiple clients could be instantiated over the [app's lifecycle](https://developer.android.com/guide/components/activities/intro-activities#mtal).

This changeset alters `Bugsnag.init` so that it will only instantiate a `Client` if one hasn't been created already. For exceptional cases where a user needs to instantiate multiple `Client` objects, it remains possible to do so by invoking the `Client` constructor directly.

## Changeset

- Synchronized initialisation code within the `Bugsnag.init` methods, so that only one thread can attempt initialisation at a time.
- Only created a new `Client` object if the existing field is null, and otherwise return the existing instance.
- Log a warning if the existing client field is not null, as this most likely indicates the integration is not following our latest recommendations.

## Tests

- Ran the example app with two `Bugsnag.init` calls, and verified that a warning was logged.
- Created a mazerunner test which repeatedly invokes all method variants of `Bugsnag.init` on multiple threads, and verifies only one client was created.

### DIFF
--- a/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/BugsnagInitScenario.kt
+++ b/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/BugsnagInitScenario.kt
@@ -1,0 +1,39 @@
+package com.bugsnag.android.mazerunner.scenarios
+
+import android.content.Context
+import com.bugsnag.android.Bugsnag
+import com.bugsnag.android.Client
+import com.bugsnag.android.Configuration
+import java.lang.RuntimeException
+import java.util.concurrent.Callable
+import java.util.concurrent.Executors
+
+internal class BugsnagInitScenario(
+    config: Configuration,
+    context: Context
+) : Scenario(config, context) {
+
+    init {
+        config.setAutoCaptureSessions(false)
+    }
+
+    override fun run() {
+        val threadPool = Executors.newFixedThreadPool(8)
+        val callables = mutableListOf<Callable<Client?>>()
+
+        IntRange(1, 25).forEach {
+            callables.add(Callable { Bugsnag.init(context) })
+            callables.add(Callable { Bugsnag.init(context, config.apiKey) })
+            callables.add(Callable { Bugsnag.init(context, config.apiKey, false) })
+            callables.add(Callable { Bugsnag.init(context, Configuration(config.apiKey)) })
+        }
+
+        val futures = threadPool.invokeAll(callables)
+        val uniqueClients = futures.map { it.get() }.distinct()
+
+        val bugsnag = uniqueClients.first()!!
+        bugsnag.addToTab("client", "count", uniqueClients.size)
+        bugsnag.notify(RuntimeException())
+    }
+
+}
--- a/sdk/src/main/java/com/bugsnag/android/Bugsnag.java
+++ b/sdk/src/main/java/com/bugsnag/android/Bugsnag.java
@@ -106,9 +106,9 @@ public final class Bugsnag {
     }
 
     private static void logClientInitWarning() {
-        Logger.warn("It appears that Bugsnag.init() was called more than once. Subsequent " +
-            "calls have no effect, but may indicate that Bugsnag is not integrated in an" +
-            " Application subclass, which can lead to undesired behaviour.");
+        Logger.warn("It appears that Bugsnag.init() was called more than once. Subsequent "
+            + "calls have no effect, but may indicate that Bugsnag is not integrated in an"
+            + " Application subclass, which can lead to undesired behaviour.");
     }
 
     /**
--- a/sdk/src/main/java/com/bugsnag/android/Bugsnag.java
+++ b/sdk/src/main/java/com/bugsnag/android/Bugsnag.java
@@ -19,9 +19,10 @@ import java.util.Map;
 @SuppressWarnings("checkstyle:JavadocTagContinuationIndentation")
 public final class Bugsnag {
 
-    @Nullable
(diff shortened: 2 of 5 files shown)


features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/BugsnagInitScenario.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        IntRange(1, 25).forEach {
+            callables.add(Callable { Bugsnag.init(context) })
+            callables.add(Callable { Bugsnag.init(context, config.apiKey) })
+            callables.add(Callable { Bugsnag.init(context, config.apiKey, false) })
+            callables.add(Callable { Bugsnag.init(context, Configuration(config.apiKey)) })
+        }
+
+        val futures = threadPool.invokeAll(callables)
+        val uniqueClients = futures.map { it.get() }.distinct()
```
fractalwrench: This blocks until all the `Callables` have finished execution, then gets the returned `Client` objects and removes any duplicates from the Collection (there should only ever be 1 item in this list).

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -19,7 +19,8 @@
 @SuppressWarnings("checkstyle:JavadocTagContinuationIndentation")
 public final class Bugsnag {
 
-    @Nullable
```
bengourley: why did the client lose its `@nullable` annotation?

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -33,8 +34,14 @@ private Bugsnag() {
      */
     @NonNull
     public static Client init(@NonNull Context androidContext) {
-        client = new Client(androidContext);
-        NativeInterface.configureClientObservers(client);
+        synchronized (lock) {
```
bengourley: why did you choose to make a `synchronized (lock) {}` block rather than label the entire static method `public static synchronized Client init()`?

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -75,11 +94,23 @@ public static Client init(@NonNull Context androidContext,
      */
     @NonNull
     public static Client init(@NonNull Context androidContext, @NonNull Configuration config) {
-        client = new Client(androidContext, config);
-        NativeInterface.configureClientObservers(client);
+        synchronized (lock) {
```
snmaynard: Why is this code duplicated in all the init methods rather than calling one implementation init method from all the other ones?

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
fractalwrench: That would use the intrinsic lock on the `Bugsnag` class object, rather than a lock we've specified explicitly and made private. Even though it's a terrible idea, users could theoretically synchronise 3rd party code using the intrinsic lock on the Bugsnag class.

This approach is also more granular. We don't have to synchronise the entire contents of the method, and it's possible to use more than 1 lock in the same class, if we need to synchronise some other unrelated code at some future date.

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
fractalwrench: I considered putting the `Client` initialisation into a `Callable` (or equivalent) and passing that into a common method, but felt that this was outside the scope of changes originally agreed during design review. If you're happy with that approach I can alter the PR.

snmaynard: I was just thinking in the other init methods creating a configuration object and then calling down to the other init function with that config object?

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
fractalwrench: The main reason we add nullability annotations is to provide information on our public APIs - as this field isn't public and will nearly always be non-null, I felt this annotation is somewhat misleading.

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -75,11 +94,23 @@ public static Client init(@NonNull Context androidContext,
...
+                NativeInterface.configureClientObservers(client);
+            } else {
+                logClientInitWarning();
+            }
+        }
         return client;
     }
 
+    private static void logClientInitWarning() {
+        Logger.warn("It appears that Bugsnag.init() was called more than once. Subsequent "
```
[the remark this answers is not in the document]
fractalwrench: Logcat can automatically handle any message up to 4Kb in size, so we should be ok to leave this as-is.

sdk/src/main/java/com/bugsnag/android/Bugsnag.java
fractalwrench: I've updated the `init` methods so that they now do this.

fractalwrench:
I've addressed all outstanding feedback and have reduced duplication in the `init` methods.

PR #415, opened 2019-01-16
Update vendored GSON dependency to latest available version

fractalwrench:
## Goal

The `JsonWriter` class was originally vendored in 2014, and has not been updated since then despite bug fixes to the GSON library.

## Design

It was originally suggested that we use the framework version of `JsonWriter`. However, this approach has several issues:

1. The framework class is `final`, therefore cannot be extended in Java, meaning we would need to make substantial modifications to our implementation of `JsonStream`
2. `setSerializeNulls` is not present, meaning we would always serialize null values if we used the framework version
3. The `beforeValue` method is private and therefore cannot be accessed, meaning we would need to make further alterations to `JsonStream` to support serialising a file.

Therefore the lowest risk approach seemed to be updating the vendored GSON dependency, rather than relying on the framework version.

## Changeset

Updated the `JsonWriter` and `JsonScope` classes to the latest available version on the [GSON repository](https://github.com/google/gson/tree/da5cae371e8b739fe63a6c6d16debf7b297dea0e/gson/src/main/java/com/google/gson/stream).

`SuppressAllWarnings` was used to suppress sites in the GSON code which violate our existing checkstyle/lint rules. 

## Tests

Ran existing tests on CI.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/JsonStream.java
+++ b/sdk/src/main/java/com/bugsnag/android/JsonStream.java
@@ -69,7 +69,7 @@ public class JsonStream extends JsonWriter {
         }
 
         super.flush();
-        beforeValue(false); // add comma if in array
+        beforeValue(); // add comma if in array
 
         // Copy the file contents onto the stream
         Reader input = null;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 4.X.X (TBD)
+
+* Update vendored GSON dependency to latest available version
+ [#415](https://github.com/bugsnag/bugsnag-android/pull/415)
+
 ## 4.10.0 (2019-01-07)
 
 * Improve kotlin support by allowing property access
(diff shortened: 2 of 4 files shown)


sdk/src/main/java/com/bugsnag/android/JsonStream.java
```diff
@@ -69,7 +69,7 @@ public void value(@NonNull File file) throws IOException {
         }
 
         super.flush();
-        beforeValue(false); // add comma if in array
+        beforeValue(); // add comma if in array
```
bengourley: why did this need to change?

fractalwrench: The boolean parameter for this method has been [removed](https://github.com/bugsnag/bugsnag-android/pull/415/files#diff-f33a35f88c8e8350f363fd92e8e25e0cL679) since we last updated. Passing in a parameter which hasn't been explicitly specified would lead to a compilation error in Java.

fractalwrench:
Thanks @bengourley - I've added a note to each class that states where it was retrieved from.

PR #416, opened 2019-01-16
[Discussion] add dependency on the Kotlin standard library

fractalwrench:
## Goal

This changeset adds the [Kotlin language](https://kotlinlang.org/) as a dependency to bugsnag-android, which allows us to write Kotlin code that would be compiled into the final artefact.

I'm opening this PR so that we can discuss the advantages and disadvantages of this change, then decide on a course of action.

## Background

Kotlin has enjoyed first-class support on Android since [2017](https://android-developers.googleblog.com/2017/05/google-io-2017-empowering-developers-to.html), and has seen rapid growth in that time. Recent surveys have estimated that between [27%](https://android-developers.googleblog.com/2018/10/kotlin-momentum-for-android-and-beyond.html) to [60%](https://pusher.com/state-of-kotlin) of Android apps have adopted Kotlin, and a greater number depend on libraries which are written in Kotlin.

Jake Wharton's [evaluation of the language](https://docs.google.com/document/d/1ReS3ep-hjxWA8kZi0YqDbEhCqTt29hG8P44aA9W0DM8/preview) for Square is a good starting point for understanding why Kotlin has been successful in the Android ecosystem, as is JetBrains' [Kotlin Tutorial](https://kotlinlang.org/docs/reference/basic-syntax.html). I'll attempt to cover some of the advantages and disadvantages of Kotlin in a way that's specific to bugsnag-android.

#### Null Safety

Null pointer exceptions are the most common cause of crashes in Java applications, and even our notifier has [not been immune to this](https://github.com/bugsnag/bugsnag-android/pulls?utf8=%E2%9C%93&q=is%3Apr+npe+sort%3Aupdated-desc+). Kotlin [enforces null checks](https://kotlinlang.org/docs/reference/null-safety.html) at compile-time with its type system.

For example, if we wished to make it impossible to set the notify endpoint to null, then we could enforce this using the following code:

```
class Configuration(notifyEndpoint: String) // defines a Configuration class with a non-null parameter

val config = Configuration(null) // will not compile from Kotlin
Configuration config = new Configuration(null); // will warn at compile time then immediately crash when invoked from Java
```

We currently attempt to enforce this from Java using [nullability annotations](https://developer.android.com/studio/write/annotations#adding-nullness). However, this approach does not automatically add a [null check](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/runtime/kotlin/jvm/internal/Intrinsics.java#L24), so can lead to NPEs occuring deep within the notifier, rather than immediately upon misuse of our API.

#### Deficiencies of Java on Android

Android currently supports Java language level 1.7 and a [limited subset of 1.8](https://developer.android.com/studio/write/java8-support). This lags considerably behind the desktop/server ecosystem, and means we are unable to use newer Java features such as lambdas, streams, local type inference, and others. Although it's always possible to achieve the same results with an older language level, it tends to require more lines of code, which increases the scope for introducing bugs.

It is highly unlikely that Android will add these features in the near future, and if that happens, our usage of new features will still be restricted by backwards compatibility concerns. In the words of Jake Wharton:

>"Java 10 will push Optional onto the stack, but we'll all be dead before Android sees that."

#### Java Interoperability
Kotlin has two-way interoperability with Java code. This makes it very simple to add Kotlin to a project incrementally, without bugsnag-android users even needing to know the library is written in Kotlin.

```
// 1. calling Kotlin from Java


class AnrOptions {
    val detectThresholdMs: Long
}

Long l = new AnrOptions().getDetectThresholdMs();


// 2. Calling Java from Kotlin


public class AnrOptions {
    Long getDetectThresholdMs() { ... }
    void setDetectThresholdMs() { ... }
}

val l = AnrOptions().detectThresholdMs // calling Java from Kotlin
```

Further information on interoperability is available on the [Kotlin website](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html).

#### Lambdas and streams
Kotlin offers lambdas, which are syntactic sugar over Java's anonymous inner classes:

```
// java
Async.run(new Runnable() {
    @Override
    public void run() {
        // perform background task
    }
});

// kotlin
Async.run {
    // perform background task
}
```

Additionally, the Kotlin standard library offers a Collections API which provides useful utilities such as `filter` and `map`. While these operations are possible in vanilla Java, they are much more verbose at our current language level:

```
val lineNumbers: List<Int> = listOf<StackTraceElement>()
        .filter { it.className == "IgnoreClass" }
        .map { it.lineNumber }
```

Lambdas can also be used to create [DSLs](https://kotlinlang.org/docs/reference/type-safe-builders.html). For example, a Kotlin-esque way of creating a new bugsnag instance and configuring it may look like the following:

```
Bugsnag("api-key") {
    endpoint = "http://example.com"
    ... // configure other fields
}
```

Where the constructor for Bugsnag takes a `String` as its first parameter, and a lambda as its second parameter.

#### Properties
Kotlin supports properties, which are much simpler to read than getters/setters. Consider the Configuration class, which could be implemented something like the following, drastically reducing the amount of code required:

```
class Configuration(
    val apiKey: String, // API key is guaranteed to be a final, non-null String
    var buildUuid: String, // Build UUID is guaranteed to be a non-null String, with a mutable field
    var context: String? // Context is a nullable, mutable String
    ...
)
```

It is also possible to specify [custom setters](https://kotlinlang.org/docs/reference/properties.html#getters-and-setters). Combined with [extension functions](https://kotlinlang.org/docs/reference/extensions.html), I believe there's potential to simplify the syntax for relaying Observable messages to the bugsnag-android-ndk, without altering the architecture.

#### Data Classes
Some classes only exist to hold information - in Kotlin, these can be denoted with the `data` keyword. In addition to having immutable properties which reduces the need to worry about state changes, a data class will automatically generate the following:

- Getters/setters
- Equals/hash code
- copy
- toString

```
data class Breadcrumb(val name: String, val type: BreadcrumbType)
listOf<Breadcrumb>( ... ).distinct() // remove duplicates from collection using comparative equality
```

Implementing the same class in Java could easily require 50-100 lines of code, and would be brittle to additions/removals of fields.

#### Multiplatform
Kotlin offers the option of [multiplatform libraries](https://kotlinlang.org/docs/reference/multiplatform.html), which may come in useful if we ever decide to share core business logic between notifiers.

#### Rolling back to Java
Kotlin is compiled down to Java bytecode. This makes it straightforward for us to dip our toe in the water, as we can easily convert Kotlin classes back to Java if we decide the language is not for us.

#### Another Dependency
Adding Kotlin would increase the size of an app that does not already have a direct or transitive dependency on the language. My testing in an example app with ProGuard puts this at around 80Kb. 

I don't think this is a particularly common scenario as most Android apps these days will have at the very least a transitive dependency on Kotlin. Additionally, new minifiers such as R8 should minify this footprint even further.

#### Onboarding
Kotlin is a new language and will be unfamiliar to Java developers, meaning some extra time may be required for education. In practice we've already been using Kotlin in mazerunner and unit tests for months, and have not encountered any major problems as a result of this.

#### Static analysis
Java has been released for over 20 years, and as a consequence has a mature ecosystem with static analysis tools. The Kotlin ecosystem now does offer tools such as Android Lint, KtLint, and Detekt, so this is becoming less of an issue.

## Summary
I would suggest that we add the dependency and try writing new code in Kotlin, and keeping existing code in Java. After a period of a few months, we should evaluate whether Kotlin has made a positive impact on the development of bugsnag-android, and decide whether to continue or roll back to Java.

I'm curious to hear everyone else's thoughts on this.

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -52,7 +52,7 @@ coveralls {
 dependencies {
     api "com.android.support:support-annotations:$supportLibVersion"
 
-    androidTestImplementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    api "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
 
     testImplementation "junit:junit:$junitVersion"
     testImplementation "org.mockito:mockito-core:$mockitoVersion"


bengourley:
Quick question… is there a way to author in Kotlin without increasing the size of the notifier?

For example, could you generate `.java` code in a build step from the Kotlin compiler so that we have it as a development dependency but don't force that on our users? And if so, would that be a terrible thing to do?

fractalwrench:
@bengourley I think it might be possible to compile Kotlin down to bytecode and ship that without adding a dependency. However, this would assume that the code doesn't reference any of the standard library - which is one of the main benefits of using the language in the first place.

It's also worth noting that the APK size will only increase if a project has no dependency on Kotlin (i.e. if the project is entirely in Java, and only relies on 3rd party libraries written entirely in Java). If a project is already using Kotlin in some way, it's effectively zero cost.

kattrali:
What does this mean for things like the class/method count? Do Kotlin classes compile the same way?

Does it matter what version of Kotlin we are using vs the host application? e.g. does it affect app size or compatibility?

fractalwrench:
The impact on class/method count largely depends on minification settings, and which parts of the Kotlin standard library we use. The general rule is that it tends to minify quite well. As an example, this PR has zero effect on the Dex Count of our Java example app, because we don't yet reference any Kotlin classes so ProGuard can optimise them away. Using Kotlin in an app in some form also adds a `kotlin` directory to the APK, which uses 88Kb at the time of writing:

screenshot 2019-01-17 at 09 14 31

I'm not immediately aware of any issues that would arise from using different versions of Kotlin, but the documentation is a little hazy on this point so it'd be worth investigating a bit further. An example of a widely used library written in Kotlin is [Okio](https://github.com/square/okio) - from what I can see this currently uses Kotlin 1.2.60.

JakeWharton:
As long as you're not using kotlin-reflect in your app, you should ideally add
```
    exclude 'kotlin/**'
    exclude '**/*.kotlin_metadata'
    exclude 'META-INF/*.kotlin_module'
```
to your `packagingOptions` to eliminate that from the APK. Sadly it can't be added by default because they might be needed for people using Kotlin reflection.

fractalwrench:
Closing in favour of #501

PR #420, opened 2019-01-21
fix(ndk): Switch unwinders based on required async signal-safety

kattrali:
When capturing a stacktrace from a signal handler or C++ terminate 
handler different unwinders deadlock on different architectures and/or
Android API levels. This change uses different unwinders based on
whether async signal safety is required or not, by supported API levels.

It introduces libunwindstack as the signal-safe unwinder for API 15+ on
ARM64 and x86/x64 devices. For notify()/C++ exceptions, use libunwind,
nongnu libunwind, or libcorkscrew depending on the API level.

* libcorkscrew hangs when unwinding from a C++ exception on ARM 32-bit.
* libunwind hangs on ARM64 from a signal handler
* nongnu libunwind hangs on ARM32 from a signal handler if raise() is
  called directly

## Design

The `bsg_environment` object now stores two configured unwinders for a given device/OS:
* `signal_unwind_type`: unwinder to use in a signal handler
* `unwind_type`: unwinder to use everywhere else (`notify()`, C++ handler, etc)

These options are configured during the native `install()` phase

## Changeset

* Added `libunwindstack` as a submodule and `stack_unwinder_libunwindstack()` as a preferred unwinder on not ARM32.
* Added additional sanity checks in case any configured state is null in the libunwind checker

## Tests

The existing tests worked well for determining if a valid stack/report is gathered on every device and I ran them against API 16, 21, 27, and 28 on ARM32, ARM64, and x86. In addition I ran the scenarios in `features/native*` manually in the example app on Pixel 2/API 28 and Nexus 5X/API 16 and Nexus 5X/API 21 to examine the stack traces.

### DIFF
--- a/ndk/src/main/jni/utils/stack_unwinder_libunwindstack.cpp
+++ b/ndk/src/main/jni/utils/stack_unwinder_libunwindstack.cpp
@@ -1,0 +1,65 @@
+#include "../utils/string.h"
+#include "stack_unwinder_libunwindstack.h"
+#include <stdlib.h>
+#include <ucontext.h>
+#include <unwindstack/Elf.h>
+#include <unwindstack/MapInfo.h>
+#include <unwindstack/Maps.h>
+#include <unwindstack/Memory.h>
+#include <unwindstack/Regs.h>
+
+ssize_t
+bsg_unwind_stack_libunwindstack(bsg_stackframe stacktrace[BUGSNAG_FRAMES_MAX],
+                                siginfo_t *info, void *user_context) {
+  if (user_context == NULL) {
+    return 0; // only handle unwinding from signals
+  }
+
+  // Fetch register values from signal context. To get registers without
+  // a context, use unwindstack::Regs::CreateFromLocal()
+  const std::unique_ptr<unwindstack::Regs> regs(
+      unwindstack::Regs::CreateFromUcontext(unwindstack::Regs::CurrentArch(),
+                                            user_context));
+
+  std::string unw_function_name;
+  unwindstack::LocalMaps maps;
+
+  if (!maps.Parse()) {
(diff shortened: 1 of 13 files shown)


ndk/src/main/jni/utils/stack_unwinder_libunwindstack.cpp
```diff
@@ -0,0 +1,68 @@
...
+                                siginfo_t *info, void *user_context) {
+  if (user_context == NULL) {
+    return 0; // only handle unwinding from signals
+  }
+
+  const std::unique_ptr<unwindstack::Regs> regs(
+      unwindstack::Regs::CreateFromUcontext(unwindstack::Regs::CurrentArch(),
+                                            user_context));
+
+  // if (user_context != NULL) {
```
fractalwrench: This can probably be removed and accessed from VCS if we need some record of it.

ndk/src/main/jni/utils/stack_unwinder_libunwindstack.cpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  const std::shared_ptr<unwindstack::Memory> memory(
+      new unwindstack::MemoryLocal);
+
+  int frame_count = 0;
+  for (int i = 0; i < BUGSNAG_FRAMES_MAX; i++) {
+    unwindstack::MapInfo *const map_info = maps.Find(regs->pc());
+    unwindstack::Elf *const elf = map_info->GetElf(memory, false);
+    if (!map_info || !elf) {
```
fractalwrench: If `map_info` can be null at this point then I believe we'd also need to perform a null check at `map_info->GetElf`?

PR #429, opened 2019-02-13
Add stopSession() and resumeSession() to Client

fractalwrench:
## Goal
Adds the ability to stop and resume sessions. This prevents the handled/unhandled error count from incrementing when a session is in the stopped state, which may be desirable if a user manually tracks sessions and does not care about crashes that occur in the background.

## Changeset

Added the public interface for stopping/resuming a session, along with the initial implementation, unit test coverage, and docs.

__NOTE__: NDK support was added in #432 and mazerunner scenarios have been added in #430.

### DIFF
--- a/ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
+++ b/ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
@@ -54,7 +54,9 @@ public class NativeBridge implements Observer {
 
     public static native void removeMetadata(String tab, String key);
 
-    public static native void startedSession(String sessionID, String key);
+    public static native void startedSession(String sessionID, String key, int handledCount);
+
+    public static native void stoppedSession();
 
     public static native void updateAppVersion(String appVersion);
 
@@ -130,6 +132,9 @@ public class NativeBridge implements Observer {
                 break;
             case START_SESSION:
                 handleStartSession(arg);
+                break;
+            case STOP_SESSION:
+                stoppedSession();
                 break;
             case UPDATE_APP_VERSION:
                 handleAppVersionChange(arg);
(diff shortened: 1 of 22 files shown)


sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -65,7 +66,9 @@
      * @param date the session start date
      * @param user the session user (if any)
      */
-    @Nullable Session startNewSession(@NonNull Date date, @Nullable User user,
+    @Nullable
+    @VisibleForTesting
```
fractalwrench: This documents that the method has been left package-visible for existing unit tests, and would otherwise be private.

sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -78,6 +81,29 @@
...
+    }
+
+    boolean resumeSession() {
+        Session session = currentSession.get();
+
+        if (session == null) {
+            startSession(false);
+            return false;
+        } else {
+            return session.isStopped.compareAndSet(true, false);
```
Pezzah: Does this cope with the case "The previous session ended in a crash. In other words, one session can only have 1 crash at most." in the project definition? Should we also check if there have been unhandled errors?

sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -289,7 +289,15 @@ public void update(@NonNull Observable observable, @NonNull Object arg) {
      * session everytime the app enters the foreground.
      */
     public void startSession() {
-        sessionTracker.startNewSession(new Date(), user, false);
+        sessionTracker.startSession(false);
+    }
+
+    public final void stopSession() {
```
Pezzah: There is a static method the Bugsnag object to call `startSession` can we add ones for stop and resume as well?

sdk/src/main/java/com/bugsnag/android/SessionTracker.java
snmaynard: I think @duncanhewett said we don't need to enforce this now.

ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -54,7 +54,9 @@ public static native void addBreadcrumb(String name, String type, String timesta
 
     public static native void removeMetadata(String tab, String key);
 
-    public static native void startedSession(String sessionID, String key);
+    public static native void startedSession(String sessionID, String key, int handledCount);
```
Pezzah: I don't think that we can always be sure that unhandled count will be zero, so do we also need to pass that through here

fractalwrench: Passing through the `unhandledCount` will need to be done when we implement ANR detection. Doing this would add an additional field to the `bugsnag_report` struct, so would require a migration - therefore we'll wait until the ANR project to implement this.

PR #430, opened 2019-02-13
Add mazerunner scenarios for stopping/resuming a session

fractalwrench:
## Goal

Verifies that no session information is present in an error payload when the session is stopped, and that the previous session information is used when the session is resumed.

This PR adds mazerunner test coverage. See #429 for the full implementation and explanation.

### DIFF
--- a/features/session_stopping.feature
+++ b/features/session_stopping.feature
@@ -1,0 +1,19 @@
+Feature: Stopping and resuming sessions
+
+Scenario: When a session is stopped the error has no session information
+    When I run "StoppedSessionScenario"
+    Then I should receive 3 requests
+    And the request 0 is valid for the session tracking API
+    And the request 1 is valid for the error reporting API
+    And the request 2 is valid for the error reporting API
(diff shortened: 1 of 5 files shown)


features/session_stopping.feature
```diff
@@ -0,0 +1,19 @@
...
+    And the payload field "events.0.session" is null for request 2
+
+Scenario: When a session is resumed the error uses the previous session information
+    When I run "ResumedSessionScenario"
+    Then I should receive 3 requests
+    And the request 0 is valid for the session tracking API
+    And the request 1 is valid for the error reporting API
+    And the request 2 is valid for the error reporting API
+    And the payload field "events.0.session.events.handled" equals 1 for request 1
+    And the payload field "events.0.session.events.handled" equals 2 for request 2
```
Cawllec: I'm going to add a comparison between UUID's and start times here, but other than that it LGTM

Cawllec:
What would happen if you paused sessions _then_ called `startSession`?

PR #432, opened 2019-02-14
Support stopSession() and resumeSession() in NDK

fractalwrench:
## Goal

Adds the ability to stop and resume sessions in the NDK layer. When a session is stopped, session information will not be added to error payloads and handled/unhandled error counts will not be incremented.

For fatal NDK errors, we cannot call the JNI without crashing the process, so we write an error report to disk using C. This PR updates the serialisation code to handle the concept of a stopped session.

See #429 for the Java implementation, and further details.

## Changeset

- Emit a `SESSION_STARTED` observable event from `SessionTracker.java` when a session is started or resumed, which ultimately sets session information on the NDK payload.
- Add a `SESSION_STOPPED` observable event, which sets a boolean flag on `bugsnag_report` that prevents serialisation of session information in fatal NDK error reports
- Update the `SESSION_STARTED` event to also include the handled count, rather than assuming it will always equal 0.

## Tests
Added mazerunner scenarios to verify that session information is not present in error payloads of a stopped session.

### DIFF
--- a/ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
+++ b/ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
@@ -134,7 +134,7 @@ public class NativeBridge implements Observer {
                 handleStartSession(arg);
                 break;
             case STOP_SESSION:
-                handleStopSession();
+                stoppedSession();
                 break;
             case UPDATE_APP_VERSION:
                 handleAppVersionChange(arg);
--- a/ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
+++ b/ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
@@ -56,6 +56,8 @@ public class NativeBridge implements Observer {
 
     public static native void startedSession(String sessionID, String key);
 
+    public static native void stoppedSession();
+
     public static native void updateAppVersion(String appVersion);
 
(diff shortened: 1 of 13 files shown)


sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -81,27 +81,42 @@ Session startNewSession(@NonNull Date date, @Nullable User user,
...
+        return startNewSession(new Date(), client.getUser(), autoCaptured);
     }
 
     void stopSession() {
         Session session = currentSession.get();
 
         if (session != null) {
             session.isStopped.set(true);
+            setChanged();
+            notifyObservers(new NativeInterface.Message(
```
kattrali: There should be an `ObserverInterfaceTest` covering `STOP_SESSION` events.

ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -326,6 +334,10 @@ private void handleStartSession(Object arg) {
         warn("START_SESSION object is invalid: " + arg);
     }
 
+    private void handleStopSession() {
```
kattrali: This method, while consistent with many of the others, is redundant. Could instead just call `stoppedSession()` above, similar to `clearBreadcrumbs()` and `addHandledEvent()`, which also don't need additional processing before crossing to the native layer.

PR #434, opened 2019-02-19
Fix incorrect session handledCount when notifying in quick succession

fractalwrench:
## Goal

By default `bugsnag.notify()` captures error reports synchronously, then delivers them asynchronously. The delivery process serialises information about the current session into the error, including counts of handled/unhandled errors that have occurred in the session.

These counts were serialised in the async delivery, from a shared `session` object reference. This meant if `bugsnag.notify()` was called twice, the session object `handledCount` would be incremented twice, _then_ serialised with the incorrect counts.

This PR fixes this behaviour to ensure the counts are correct.

## Design

A defensive copy has been made of the session object when populating the error field, meaning mutations made by other errors do not affect the state of other reports.

## Changeset

- Updated `Error.Builder` (and calling code) to take a `SessionTracker` rather than `Session` (a lot of sites use this method, so it may be easier to inspect the PR by viewing individual commits)
- Added `copySession` method which copies a `Session` object
- Made `incrementHandledError` and `incrementUnhandledError` return a `Session` object which is a copy of `SessionTracker#currentSession`
- Move incrementing of handled/unhandled counts from `Client#notify` to `Error.Builder#build`, as this more accurately reflects where the error capture occurs.

## Tests

- Added a unit test for copying sessions

- This can be manually reproduced by notifying twice, and verifying that the `handledCount` equals 1 for one report, and 2 for the other report, rather than 2 for both.

```
bugsnag.notify(err)
bugsnag.notify(err)
```

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -863,7 +863,7 @@ public class Client extends Observable implements Observer {
             callback.beforeNotify(report);
         }
 
-        if (sessionTracker.getCurrentSession() != null) {
+        if (!error.getHandledState().isUnhandled() && error.getSession() != null) {
             setChanged();
             notifyObservers(new NativeInterface.Message(
                 NativeInterface.MessageType.NOTIFY_HANDLED, error.getExceptionName()));
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -657,7 +657,7 @@ public class Client extends Observable implements Observer {
      * @param exception the exception to send to Bugsnag
      */
     public void notify(@NonNull Throwable exception) {
-        Error error = new Error.Builder(config, exception, sessionTracker.getCurrentSession(),
+        Error error = new Error.Builder(config, exception, sessionTracker,
             Thread.currentThread(), false)
             .severityReasonType(HandledState.REASON_HANDLED_EXCEPTION)
             .build();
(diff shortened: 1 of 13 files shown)


sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -863,17 +863,10 @@ void notify(@NonNull Error error,
...
-        } else {
-            sessionTracker.incrementHandledError();
-            if (sessionTracker.getCurrentSession() != null) {
-                setChanged();
-                notifyObservers(new NativeInterface.Message(
-                            NativeInterface.MessageType.NOTIFY_HANDLED, error.getExceptionName()));
-            }
+        if (sessionTracker.getCurrentSession() != null) {
+            setChanged();
+            notifyObservers(new NativeInterface.Message(
```
Pezzah: previously we would only notify observers for handled errors (presumably because of unpredictable behaviour after an unhandled error?)

sdk/src/main/java/com/bugsnag/android/Session.java
```diff
@@ -55,12 +63,14 @@ int getHandledCount() {
         return handledCount.intValue();
     }
 
-    void incrementHandledErrCount() {
+    Session incrementHandledErrCount() {
```
Pezzah: I don't think it's obvious that these methods increment the current session and also return a copy of the incremented session.

Perhaps `copySession` should just be called independently

sdk/src/main/java/com/bugsnag/android/Client.java
fractalwrench: I've updated this to only notify if the error object is `handled` and has a session.

sdk/src/main/java/com/bugsnag/android/Session.java
fractalwrench: I've renamed this `incrementHandledAndCopy`, and also added some JavaDoc.

PR #435, opened 2019-02-26
Wait for session scenarios to send session before crashing

fractalwrench:
The E2E test job failed for Android 24 on CI only. I believe this is due to the native code killing the app before the session can be delivered, which is only an issue on Travis CI's slow ARM emulators.

The fix uses a handler to delay this crash for a reasonable period, meaning the session is always delivered in time. This mirrors the approach used in [previous](https://github.com/bugsnag/bugsnag-android/blob/master/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXSessionInfoCrashScenario.java) [scenarios](https://github.com/bugsnag/bugsnag-android/blob/9f69368dd606d6522371b0150fa3ea260e41fb0b/features/native_api.feature).

### DIFF
--- a/features/native_session_tracking.feature
+++ b/features/native_session_tracking.feature
@@ -2,6 +2,8 @@ Feature: NDK Session Tracking
 
 Scenario: Stopped session is not in payload of unhandled NDK error
     When I run "CXXStopSessionScenario"
+    And I wait a bit
+    And I wait a bit
     And I configure the app to run in the "non-crashy" state
     And I relaunch the app
     Then I should receive 2 requests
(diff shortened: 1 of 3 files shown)


features/native_session_tracking.feature
```diff
@@ -2,6 +2,8 @@ Feature: NDK Session Tracking
 
 Scenario: Stopped session is not in payload of unhandled NDK error
     When I run "CXXStopSessionScenario"
+    And I wait a bit
```
Cawllec: What does this do, and why do we required two of them?  Is it possible to detect if an app is running through CLI somehow?

fractalwrench: This step will wait for a set amount of time that is lower locally, but increased on CI. I believe it requires 2 as that's what was used in previous scenarios involving sessions and native errors.
https://github.com/bugsnag/bugsnag-android/blob/31d75900a89f1de6f67047fb98ba80eead2a4943/features/steps/build_steps.rb#L1

I believe it might be possible to detect if an app is running through adb but we don't currently have that setup.

Cawllec: Is it relied on somewhere else as well, or could we just bump the time up in a single step?

fractalwrench: I believe the timing is relied on in other places.

PR #439, opened 2019-03-12
fix: Add helper for syncing a unity session to native reports

kattrali:
## Goal

Allow registering a session with the `SessionTracker` to be propagated to crash reports. This allows the session to be tracked separately via bugsnag-unity while preserving the stability score.

Related to bugsnag/bugsnag-unity#138

## Changeset

Added a helper to `NativeInterface` to allow updating the current session.

## Tests

Tested manually as a part of bugsnag/bugsnag-unity#138


fractalwrench:
This will need a changelog entry when the changes are complete.

PR #442, opened 2019-03-14
Add ANR detection to bugsnag-android

fractalwrench:
## Goal

[ANRs](https://developer.android.com/topic/performance/vitals/anr) occur when an Android application has been blocking the main thread for a substantial amount of time, and typically means the user is about to close the app. Bugsnag should capture and report these events as unhandled errors.

## Changeset

- Added public boolean flag `config.detectAnrs`, which is enabled by default if the app does not have a debugger attached (this would cause false positives)
- Added `config.anrThresholdMs` which allows developers to control the threshold at which ANRs are detected (5000ms by default)
- Added "ANRError" as a severity reason
- Initialised a blocked thread detector within the normal `Client` initialisation, which reports an unhandled error when the main thread is blocked for 5000ms+

### ANR detection implementation

The `BlockedThreadDetector` uses a `Handler` to periodically post a `Runnable` on the main thread. This `Runnable` updates a timestamp obtained from `SystemClock.elapsedRealTime()`.

A separate watcher thread checks whether the timestamp has been updated in the last 5000ms. If this is not the case, then the main thread has been unable to process messages for 5 seconds and this is interpreted as an ANR.

## Discussion
ANR detection requires additional work that will be addressed in separate PRs:

- Performance of ANR detection should be benchmarked against the previous version of bugsnag-android to ensure there is no unacceptable degradataion
- Unity requires a public interface for configuration options
- The NDK requires alterations to how the `unhandledCount` is serialised.
- React Native requires a release with the latest version of bugsnag-android

## Tests

- Manually verified that an ANR is detected using the example app
- Added unit tests
- Added mazerunner scenarios which were reviewed in #440

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
+++ b/sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
@@ -1,0 +1,106 @@
+package com.bugsnag.android;
+
+import android.os.Handler;
+import android.os.Looper;
+import android.os.SystemClock;
+
+/**
+ * Detects whether a given thread is blocked by continuously posting a {@link Runnable} to it
+ * from a watcher thread, invoking a delegate if the message is not processed within
+ * a configured interval.
+ */
+final class BlockedThreadDetector {
+
+    private static final int DEFAULT_CHECK_INTERVAL_MS = 1000;
+
+    interface Delegate {
+
+        /**
+         * Invoked when a given thread has been unable to execute a {@link Runnable} within
+         * the {@link #blockedThresholdMs}
+         *
+         * @param thread the thread being monitored
+         */
+        void onThreadBlocked(Thread thread);
+    }
+
+    final Looper looper;
+    final long checkIntervalMs;
+    final long blockedThresholdMs;
+    final Handler handler;
+    final Delegate delegate;
+
+    volatile long lastUpdateMs;
+    volatile boolean isAlreadyBlocked = false;
+
+    BlockedThreadDetector(long blockedThresholdMs,
+                          Looper looper,
+                          Delegate delegate) {
+        this(blockedThresholdMs, DEFAULT_CHECK_INTERVAL_MS, looper, delegate);
+    }
+
+    BlockedThreadDetector(long blockedThresholdMs,
+                          long checkIntervalMs,
+                          Looper looper,
+                          Delegate delegate) {
+        if ((blockedThresholdMs <= 0 || checkIntervalMs <= 0
+            || looper == null || delegate == null
+            || checkIntervalMs > blockedThresholdMs)) {
+            throw new IllegalArgumentException();
+        }
+        this.blockedThresholdMs = blockedThresholdMs;
+        this.checkIntervalMs = checkIntervalMs;
+        this.looper = looper;
+        this.delegate = delegate;
+        this.handler = new Handler(looper);
+    }
+
+    void updateLivenessTimestamp() {
+        lastUpdateMs = SystemClock.elapsedRealtime();
+    }
+
+    void start() {
+        updateLivenessTimestamp();
+        handler.post(livenessCheck);
+        watcherThread.start();
+    }
+
+    final Runnable livenessCheck = new Runnable() {
+        @Override
+        public void run() {
+            updateLivenessTimestamp();
+            handler.postDelayed(this, checkIntervalMs);
+        }
+    };
+
+    final Thread watcherThread = new Thread() {
+        @Override
+        public void run() {
+            while (!isInterrupted()) {
+                // when we would next consider the app blocked if no timestamp updates take place
+                long now = SystemClock.elapsedRealtime();
+                long nextCheckIn = Math.max(lastUpdateMs + blockedThresholdMs - now, 0);
+
+                try {
+                    Thread.sleep(nextCheckIn); // throttle checks to the configured threshold
+                } catch (InterruptedException exc) {
+                    interrupt();
+                }
+                checkIfThreadBlocked();
+            }
+        }
+
+        private void checkIfThreadBlocked() {
+            long delta = SystemClock.elapsedRealtime() - lastUpdateMs;
+
+            if (delta > blockedThresholdMs) {
+                if (!isAlreadyBlocked) {
+                    delegate.onThreadBlocked(looper.getThread());
+                }
+                isAlreadyBlocked = true; // prevents duplicate reports for the same ANR
+            } else {
+                isAlreadyBlocked = false;
+            }
+        }
+    };
+}
--- a/sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
+++ b/sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
@@ -5,7 +5,11 @@ import android.os.Build;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Looper;
+import android.os.Process;
 import android.os.SystemClock;
+import android.support.annotation.Nullable;
+
+import java.util.List;
 
 /**
  * Detects whether a given thread is blocked by continuously posting a {@link Runnable} to it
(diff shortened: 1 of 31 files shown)


sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -641,6 +645,58 @@ public void setDelivery(@NonNull Delivery delivery) {
...
+    /**
+     * Sets the time in milliseconds at which an
+     * <a href="https://developer.android.com/topic/performance/vitals/anr">ANR</a> is detected
+     * by Bugsnag. By default, Bugsnag will record an ANR whenever the main thread has been blocked
+     * for 5000 milliseconds or longer.
+     * <p/>
+     * If you wish to disable ANR detection completely, you should set the
+     * {@link #setDetectAnrs(boolean)} property to false.
+     * <p/>
+     * Attempting to set this property to any value below 100ms will result in the anrThresholdMs
```
bengourley: why do you want to adjust this value if an invalid one is supploed, rather than throw an `IllegalArgumentException`?

fractalwrench: @duncanhewett did you have any opinions on this one? I'd be happy with either setting the value to 100ms, or throwing an exception immediately.

Pezzah: I think if it's less than 1000ms it will throw an `IllegalArgumentException` here anyway https://github.com/bugsnag/bugsnag-android/blob/android-anr-detection/sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java#L49

duncanhewett: Either is fine with me.

fractalwrench: In that case I will leave the current implementation, which sets the anrThresholdMs as 100ms.

fractalwrench:
>My only comment is that when the device is online, the error report from a detected ANR get saved to disk and then immediately sent. This seems like a whole bunch of unnecessary file operations when the app is in a potentially very busy state. I suspect it's because the error is considered "unhandled" and so it goes down the route "the app is crashing, serialise it asap". Do you think it would be possible and/or a good idea to check the network state and attempt to send first if the device is online?

The ANR dialog allows the user to kill the app process by clicking "Close". Therefore we write the report to disk immediately, which is the same processing used for when an uncaught exception occurs and the process is about to die.

sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
```diff
@@ -0,0 +1,106 @@
...
+import android.os.SystemClock;
+
+/**
+ * Detects whether a given thread is blocked by continuously posting a {@link Runnable} to it
+ * from a watcher thread, invoking a delegate if the message is not processed within
+ * a configured interval.
+ */
+final class BlockedThreadDetector {
+
+    private static final int DEFAULT_CHECK_INTERVAL_MS = 1000;
```
Pezzah: As far as I can tell this interval is always used (apart from for testing).

This is slightly confusing because the name implies that it is only the default (and it perhaps could be configured by the user).

It also means that any threshold less than `DEFAULT_CHECK_INTERVAL_MS` will cause an `IllegalArgumentException` (which is not obvious)

sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
```diff
@@ -0,0 +1,105 @@
...
+            handler.postDelayed(this, checkIntervalMs);
+        }
+    };
+
+    final Thread watcherThread = new Thread() {
+        @Override
+        public void run() {
+            while (!isInterrupted()) {
+                // when we would next consider the app blocked if no timestamp updates take place
+                long now = SystemClock.elapsedRealtime();
```
pyricau: "uptimeMillis" probably works better here, as time in deep sleep shouldn't be accounted for as "application not responding"

sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
```diff
@@ -0,0 +1,105 @@
...
+                long nextCheckIn = Math.max(lastUpdateMs + blockedThresholdMs - now, 0);
+
+                try {
+                    Thread.sleep(nextCheckIn); // throttle checks to the configured threshold
```
pyricau: Create a HandlerThread and corresponding Handler and use postDelay instead?

pyricau:
We've already implemented this in Square apps, and there are a few gotchas.

ANRs aren't actually triggered by the main thread being blocked for 5 seconds. They are triggered when a touch event has been enqueued to an app and hasn't been consumed for 5 seconds.

AFAIK there's no easy way to read the event queue and see how long events have been waiting in there (if someone knows, I'd love to learn more). So the closest proxy is usually "main thread has been blocked for 5 seconds AND the app is in foreground (foreground being a proxy for "can receive touch events). Even foreground is hard to capture as an activity might be stuck on onCreate or onStop so activity lifecycle callbacks don't have the right information in due time. A likely better way is to capture the running process importance, which indicates whether the system considers the app to be foreground or not.

Also, posting to the main thread works fine in general but when the queue gets a lot of messages I believe the system will prioritize consuming touch events based on vsync, so you could have a really large message queue and yet no ANR. So you don't want to be too eager when detecting ANRs, and one way to do that is, if you're getting close to the ANR deadline, use Handler.sendMessageAtFrontOfQueue on the main thread.

fractalwrench:
Thanks for the detailed review comments @pyricau - they've been really helpful. We plan on making the following changes to our implementation:

1. use `uptimeMillis` rather than `elapsedRealTimeMillis`
2. use a `HandlerThread` with `postDelayed`, so that detection is paused when the system enters deep sleep
3. calculate whether the app is in the foreground, and only report an ANR when this is true

The intent of this would be to change our definition of an ANR as something that only occurs when the main thread is blocked for 5s _and_ the app is in the foreground.

fractalwrench:
@bengourley I was able to reproduce that behaviour when I put the app in the background, then quickly brought it to the foreground again, but if I left it for >10 seconds an ANR was not detected. The example app's ANR blocks the main thread for 10s, so I believe what we're seeing is expected behaviour.

kattrali:
An immediate ANR when the app has been in the foreground for less than the ANR threshold doesn't make sense though. This edge case could be avoided though by checked that the time in foreground is greater than the threshold instead of only whether in the foreground at all [here](https://github.com/bugsnag/bugsnag-android/pull/442/files#diff-b1f18acaf3a7e035b99aa8d37a2872acR101) though.

sdk/src/main/java/com/bugsnag/android/BlockedThreadDetector.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private boolean isInForeground() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+            ActivityManager.RunningAppProcessInfo info
+                = new ActivityManager.RunningAppProcessInfo();
+            ActivityManager.getMyMemoryState(info);
+            return info.importance <= ActivityManager.RunningAppProcessInfo.IMPORTANCE_VISIBLE;
+        } else {
+            return true;
```
pyricau: You can get it pre jelly bean too:

```
 internal fun retrieveProcessInfo(): RunningAppProcessInfo? {
    val activityManager = application.getSystemService(
        Context.ACTIVITY_SERVICE
    ) as ActivityManager
    val runningProcesses: List<RunningAppProcessInfo>?
    try {
      runningProcesses = activityManager.runningAppProcesses
    } catch (exception: SecurityException) {
      // SecurityException: Isolated process not allowed to call getRunningAppProcesses
      return null
    }
    if (runningProcesses != null) {
      val myPid = android.os.Process.myPid()
      for (process in runningProcesses) {
        if (process.pid == myPid) {
          return process
        }
      }
    }
    return null
  }
```

pyricau:
As a reminder, an ANR is *not* shown every time the main thread is blocked for 5 seconds. Blocking of the main thread can cause an ANR, but not always.

ANRs are triggered when touch events in the input queue aren't consumed after 5 seconds. That's why a common trick is to ignore time the main thread spends being blocked when the app is in background, because you can't get touch events in background.

The main thread is a serial queue but some things (like rendering & touch events) get prioritized ahead of the rest.

Another idea: listen to touch events. If a touch event was received within the last 5 seconds, then that touch event was definitely consumed and therefore an ANR won't show. To listen to all touch events you need an activity lifecycle listener, then you get the window and you wrap the existing window callback with your own and replace.

fractalwrench:
Thanks for all the helpful review comments @pyricau. I've made one change to the implementation so that the Process Info is collected on API <16. After internal discussion we've decided that we'd like to stick with the current implementation, as wrapping the existing window and listening to all touch events could be considered quite invasive for a 3rd party library.

pyricau:
Makes sense.

One last thing: on [twitter](https://twitter.com/mullender/status/1112727291153440768) Matthijs Mullender mentioned the OS sends a sigquit to the app. If you already have a native signal handler in place for bugsnag maybe there's something to look at here.

fractalwrench:
Using a signal handler sounds like a very interesting approach as it would be a bit more definitive that the app actually crashed. We're going to investigate using that longer-term, and go with the current changes for now.

kenumir:
The idea is like in https://github.com/SalomonBrys/ANR-WatchDog
I used it in my apps but generates lots of false positive ANRs, this should be treated as an indication of a possible error.

PR #448, opened 2019-04-02
chore: update example app build.gradle to conditionally use local artefact

fractalwrench:
## Goal

The example app has sporadically been unable to build due to being unable to find SO files from bugsnag-android-ndk. This alters the example app so that a previously commented out task can be switched on by a flag in the build.gradle.

## Tests

I could reproduce this behaviour by running a clean build without the bugsnag gradle plugin applied. After applying this task to the example project, the behaviour could not be reproduced.

### DIFF
--- a/examples/sdk-app-example/build.gradle
+++ b/examples/sdk-app-example/build.gradle
@@ -73,13 +73,21 @@ android {
     }
 }
 
+// set this value to true if you wish to test with a local artefact
+// it will instruct gradle to copy the SO files from the AAR to the correct build directory
+def usingLocalArtefact = false
+
 dependencies {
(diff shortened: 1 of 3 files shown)


examples/sdk-app-example/build.gradle
```diff
@@ -92,73 +100,9 @@ dependencies {
     }
 }
 
+apply from: 'checkstyle.gradle'
 apply plugin: 'com.bugsnag.android.gradle'
-bugsnag {
-    ndk true
```
Pezzah: Is the bugsnag config being intentionally removed here?

fractalwrench: The plugin automatically detects whether a project uses the NDK so this isn't strictly required.

PR #449, opened 2019-04-02
chore: simplify example app to use one build variant

fractalwrench:
## Goal

The example app currently uses separate buildVariants for the Java/Kotlin language. This has caused confusion internally in the past so it makes sense to combine them and simplify the mental overhead of the example app.

## Changeset

- Updates Kotlin to 1.3.21 (version bundled with stable Android Studio release)
- Combined the java/kotlin build variants so that there is now only 1 target, rather than 2, which uses both Java/Kotlin code
- Removed unused test dependencies
- Updated gradle dependency versions
- Suppressed warning in AndroidManifest.xml which showed a big scary warning about app indexing
- Altered package of `CrashyClass` so that it will be obfuscated by ProGuard rules, allowing easier testing of obfuscation

## Tests

Ran the example app manually and verified that each button still sends an error.


Pezzah:
Do we have any idea about how many of our customers use kotlin vs java? It might be better to have more of the example code in java if that is the main language we expect our customers to be using

fractalwrench:
Around 27% of the top apps in the Google Play store used Kotlin in [October 2018](https://android-developers.googleblog.com/2018/10/kotlin-momentum-for-android-and-beyond.html) - this number is likely much higher now. In addition, most tutorials and new libraries are now written in Kotlin - for example, [Google's starter guide](https://developer.android.com/training/basics/firstapp/starting-activity?hl=en) defaults to Kotlin.

PR #454, opened 2019-04-03
Add detekt static analysis plugin

fractalwrench:
## Goal

Our Java code uses checkstyle, but our Kotlin code is not statically analysed with the exception of Android Lint. Using [detekt](https://github.com/arturbosch/detekt) should allow us to detect if there are an excessive number of non-conventional Kotlin code smells.

## Changeset

- Added detekt to gradle build classpath
- Enabled detekt for all Kotlin source files in the sdk module

## Tests

Ran manually and verified that there are no existing issues with our Kotlin code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ script:
 jobs:
   include:
   - stage: lint
-    script: ./gradlew lint checkstyle
+    script: ./gradlew lint checkstyle detekt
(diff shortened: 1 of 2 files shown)


fractalwrench:
The task won't fail unless the number of issues exceed a certain threshold, so that behaviour is intentional.

PR #456, opened 2019-04-08
Enable C++ linting on CI, fix violations

kattrali:
## Goal

Ensure common issues in C/C++ code is caught and handled.

## Design

`cppcheck` is commonly available tool and easy to install despite the NDK being an unusual environment. It also found things! 😄

### DIFF
--- a/ndk/src/main/jni/handlers/signal_handler.c
+++ b/ndk/src/main/jni/handlers/signal_handler.c
@@ -161,7 +161,10 @@ void bsg_invoke_previous_signal_handler(int signum, siginfo_t *info,
 
 void bsg_handle_signal(int signum, siginfo_t *info,
                        void *user_context) __asyncsafe {
-  if (bsg_global_env == NULL || bsg_global_env->handling_crash) {
+  if (bsg_global_env == NULL) {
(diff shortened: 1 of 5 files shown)


ndk/src/main/jni/handlers/signal_handler.c
```diff
@@ -161,7 +161,10 @@ void bsg_invoke_previous_signal_handler(int signum, siginfo_t *info,
 
 void bsg_handle_signal(int signum, siginfo_t *info,
                        void *user_context) __asyncsafe {
-  if (bsg_global_env == NULL || bsg_global_env->handling_crash) {
+  if (bsg_global_env == NULL) {
```
fractalwrench: I'm not sure I fully understand the distinction here, what was the issue with the previous code?

kattrali: The following line didn't make sense if `env` is null and was a potential npe.

fractalwrench: Ah, that makes sense now! I initially thought the first condition checked `bsg_global_env != NULL`.

PR #459, opened 2019-04-11
Refactor language level migration

fractalwrench:
This PR makes 3 changes as reported via Android Studio inspection:

- Removing redundant information from type brackets
- Replacing large if statement with switch statement

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/ErrorReader.java
+++ b/sdk/src/main/java/com/bugsnag/android/ErrorReader.java
@@ -291,15 +291,19 @@ class ErrorReader {
         String attributeValue = null;
         while (reader.hasNext()) {
             String name = reader.nextName();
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/bugsnag/android/ErrorReader.java
```diff
@@ -325,31 +329,41 @@ private static Session readSession(JsonReader reader) throws IOException {
...
-            } else if (name.equals("user")) {
-                user = readUser(reader);
-            } else {
-                reader.skipValue();
+                    break;
+                case "events":
+                    reader.beginObject();
+                    while (reader.hasNext()) {
+                        String eventName = reader.nextName();
+                        switch (eventName) {
```
simonbowring: Same as above

PR #461, opened 2019-04-11
Prefer using zero-based array when converting collection

fractalwrench:
There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).

In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/ErrorReader.java
+++ b/sdk/src/main/java/com/bugsnag/android/ErrorReader.java
@@ -245,7 +245,7 @@ class ErrorReader {
             frames.add(readStackFrame(reader));
         }
         reader.endArray();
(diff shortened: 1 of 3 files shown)


JakeWharton:
For what it's worth the performance note was only proved on the JVM, not on Android. Usually it's irrelevant since this rarely occurs in hot code paths anyway.

PR #466, opened 2019-04-11
Alter In foreground calculation

fractalwrench:
## Goal

We should update the collected value for `app.inForeground` to use the process importance rather than lifecycle callbacks. This is because if an ANR occurs on application startup, the activity may not enter the `STARTED` state due to its main thread being blocked, but the app process would still be considered in the foreground.

It's important to note that the `inForeground` value is still only updated in the NDK layer on initialisation, and when an [activity lifecycle event occurs](https://github.com/bugsnag/bugsnag-android/blob/20f176105c37f3a0a70cc91a613cc85bce7fd63c/sdk/src/main/java/com/bugsnag/android/SessionTracker.java#L374). If we wished to inform the NDK layer of each time the process importance altered, using something like [`ProcessLifecycleOwner`](https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner) would be a suitable API for doing so. However, this would require a migration to androidx and would add another dependency to our SDK, so has not been implemented in this PR.

## Changeset

- Updated `SessionTracker` to calculate `inForeground` on initialisation, so that the value is accurate if an ANR occurs immediately on startup
- Created `ForegroundDetector` class which encapsulates the inForeground checks for process importance
- Updated `BlockedThreadDetector` to use `ForegroundDetector` exclusively
- Updated `SessionTracker` to use `ForegroundDetector` primarily and also the existing implementation as a secondary source of information
- 

## Tests

- Triggered a crash (JVM and native) in the foreground and background, and confirmed that `inForeground` was the correct value for each.
- Updated existing unit tests
- Ran in foreground mazerunner feature


sdk/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -390,13 +393,17 @@ void updateForegroundTracker(String activityName, boolean activityStarting, long
             }
         }
         setChanged();
+        notifyNdkInForeground();
```
kattrali: I might have the wrong end of the elephant here, but since this is only triggered by activity lifecycle callbacks, does this change improve the value of `isInForeground` for the NDK?

fractalwrench: I've updated the PR description to explain why this approach is better following an in-person conversation, which I believe resolved this comment.

sdk/src/main/java/com/bugsnag/android/ForegroundDetector.java
```diff
@@ -0,0 +1,62 @@
...
+            (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
+    }
+
+    boolean isInForeground() {
+        ActivityManager.RunningAppProcessInfo info = getProcessInfo();
+
+        if (info != null) {
+            return info.importance <= ActivityManager.RunningAppProcessInfo.IMPORTANCE_VISIBLE;
+        } else { // prefer a potential false positive if process info not available
+            return true;
```
kattrali: Why default rather than falling back to the old implementation?

fractalwrench: We made the decision when implementing ANR detection that we should prefer a false positive when we're unable to determine process importance. This is because otherwise the notifier might discard an actual error, since the old implementation can be inaccurate if the main thread is blocked and lifecycle callbacks cannot be invoked.

kattrali: I see. To rephrase, there are actually three states for whether the app is in the foreground, `yes`, `no`, and `unknown`. In the `unknown` case, ANR errors should still be reported, so this uses `yes` instead of `unknown` or an implementation based on main thread activity. Is that a fair assessment? 

It would be clearer (and delineate responsibilities better) to actually return a value which indicates that the state is currently unknown and let the ANR reporter handle that appropriately, but barring that, the comment should state for future travelers why a potential false positive is preferable and how it affects reporting. "If this was false, some ANRs might get ignored" was not what I expected, and I can see how a future PR to tweak to how this value is computed could unintentionally break this.

PR #471, opened 2019-04-16
Use packageInfo.longVersionCode to collect versionCode on API 28+

fractalwrench:
## Goal
API 28 added an additional `android:versionCodeMajor` attribute, which can be used if an app has run out of space in the `android:versionCode` int, which must always be incremented on the Google Play store. 

The `PackageInfo` API we used to collect the version code previously has been deprecated in favour of [`longVersionCode`](https://developer.android.com/reference/android/content/pm/PackageInfo.html#getLongVersionCode()), which contains an amalgam of the versionCode + versionCodeMajor.

## Changeset

Altered the source of the versionCode collected in error reports, in API 28+.

## Tests

Verified that when running in an example app, the versionCode matches the value in the AndroidManifest for:

- API 28 with versionCodeMajor defined
- API 28 without versionCodeMajor defined
- API 26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## TBD
 
+* Use packageInfo.longVersionCode to collect versionCode on API 28+
+[#471](https://github.com/bugsnag/bugsnag-android/pull/471)
+
 ### Bug fixes
 
 * [NDK] Fix possible null pointer dereference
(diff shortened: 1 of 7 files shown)


JakeWharton:
Are there plans to actually support long version codes?

fractalwrench:
It would probably make sense to include `android:versionCodeMajor` in reports, thinking about it. I'll double-check internally that this is something we want to do, then amend the PR - thanks @JakeWharton

fractalwrench:
We've taken the decision to collect the raw value of `packageInfo.longVersionCode` on API 28+. We'll also update the bugsnag gradle plugin to collect this value when reporting version code information.

PR #472, opened 2019-04-17
Migrate version information to device.runtimeVersions

fractalwrench:
## Goal
We should collect the API level in `device.runtimeVersions.apiLevel` and `device.runtimeVersions.osBuild` for all error reports and sessions.

## Changeset
- Moved `device.apiLevel` to `device.runtimeVersions.androidApiLevel`
- Moved `device.osBuild` to `device.runtimeVersions.osBuild`
- Updated NDK JNI calls to retrieve api level and osBuild from new location

## Tests

Ran existing tests and created new mazerunner scenarios, which verify that a JVM error/session and NDK fatal error include `device.runtimeVersions.androidApiLevel`.


sdk/src/androidTest/java/com/bugsnag/android/DeviceDataSummaryTest.java
```diff
@@ -63,6 +63,9 @@ public void testJsonSerialisation() throws JSONException {
         assertNotNull(deviceDataJson.getString("model"));
         assertNotNull(deviceDataJson.get("cpuAbi"));
         assertTrue(deviceDataJson.has("jailbroken"));
+
+        JSONObject versions = deviceDataJson.getJSONObject("runtimeVersions");
+        assertNotNull(versions.get("androidApiLevel"));
```
bengourley: is "not null" the best assertion you can make? Is this because the tests can run on any version of Android? Could this not be an int greater than `n`, where `n` is the minimum supported api level?

ndk/src/main/jni/metadata.c
```diff
@@ -145,6 +145,25 @@ int bsg_get_map_value_int(JNIEnv *env, bsg_jni_cache *jni_cache, jobject map,
...
 }
 
+int bsg_populate_api_level_from_map(JNIEnv *env, bsg_jni_cache *jni_cache, jobject device_data) {
+    int value = -1;
+    jstring map_key = (*env)->NewStringUTF(env, "runtimeVersions");
+    jstring version_key = (*env)->NewStringUTF(env, "androidApiLevel");
+    jobject _map = (*env)->CallObjectMethod(env, device_data, jni_cache->hash_map_get, map_key);
+
+    if (_map != NULL) {
+        jobject _versions =
```
bengourley: `_versions` could have a more appropriate name here – it's the value of "androidApiLevel", so I guess it could just be `_apiLevel` or something.

ndk/src/main/jni/metadata.c
```diff
@@ -145,6 +145,25 @@ int bsg_get_map_value_int(JNIEnv *env, bsg_jni_cache *jni_cache, jobject map,
   return 0;
 }
 
+int bsg_populate_api_level_from_map(JNIEnv *env, bsg_jni_cache *jni_cache, jobject device_data) {
+    int value = -1;
```
bengourley: This could be initialised to `0` to be consistent with the `bsg_get_map_value_from_int` function above. But it's a massive edge case to not have an api level.

ndk/src/main/jni/metadata.c
```diff
@@ -145,6 +145,25 @@ int bsg_get_map_value_int(JNIEnv *env, bsg_jni_cache *jni_cache, jobject map,
...
+    jstring map_key = (*env)->NewStringUTF(env, "runtimeVersions");
```
bengourley: can you confirm that `jstring`s returned by `NewStringUTF()` don't need to have `DeleteLocalRef()` called on them?

sdk/src/androidTest/java/com/bugsnag/android/DeviceDataSummaryTest.java
fractalwrench: I've changed this to >= 14, which is the lowest SDK version we support.

ndk/src/main/jni/metadata.c
fractalwrench: I've renamed the variables in this method so they're a little less confusing.

ndk/src/main/jni/metadata.c
fractalwrench: Yes - these values will be automatically freed when the native method finishes execution and returns to the Java layer: https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#global_local

fractalwrench:
Thanks for the review, I've responded to the comments and updated the PR.

Moving the `osBuild` is not in the product or engineering design so no change has been made - @duncanhewett is this ok?

fractalwrench:
I have updated the PR so that `device.osBuild` is now serialised in `device.runtimeVersions.osBuild` instead.

As part of these changes I created `bsg_get_map_value_obj` to reduce duplication in retrieving an object from a Map, via the JNI. I've also refactored the existing `bsg_get_map_value_*` functions to use this, which has cut down on repetition.

PR #474, opened 2019-04-25
Add internal api for mutating session payload before sending

fractalwrench:
## Goal

We require an internal API for mutating a session payload before it is sent, for notifiers such as React Native/Unity which need to write runtime version info.

## Changeset

- Added `BeforeSendSession` functional interface
- Invoked collection of send session callbacks prior to delivering a session for the first time

## Tests

Added a unit test that verifies the callback is invoked when sending a session

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Configuration.java
+++ b/sdk/src/main/java/com/bugsnag/android/Configuration.java
@@ -58,6 +58,8 @@ public class Configuration extends Observable implements Observer {
     private final Collection<BeforeSend> beforeSendTasks = new ConcurrentLinkedQueue<>();
     private final Collection<BeforeRecordBreadcrumb> beforeRecordBreadcrumbTasks
         = new ConcurrentLinkedQueue<>();
+    private final Collection<BeforeSendSession> sessionCallbacks = new ConcurrentLinkedQueue<>();
+
     private String codeBundleId;
     private String notifierType;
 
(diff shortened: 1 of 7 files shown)


sdk/src/main/java/com/bugsnag/android/SessionTrackingPayload.java
```diff
@@ -49,4 +49,8 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
     Session getSession() {
         return session;
     }
+
+    Map<String, Object> getDevice() {
```
bengourley: Nit pick, but I don't think the change to this file belongs in this PR.

sdk/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -826,4 +828,12 @@ protected boolean inProject(@NonNull String className) {
     protected Collection<BeforeRecordBreadcrumb> getBeforeRecordBreadcrumbTasks() {
         return beforeRecordBreadcrumbTasks;
     }
+
+    void addBeforeSendSession(BeforeSendSession beforeSendSession) {
+        sessionCallbacks.add(beforeSendSession);
+    }
+
+    Collection<BeforeSendSession> getSessionCallbacks() {
```
bengourley: is it possible to return an iterator that represents this list? The fact you can get hold of the collection and mutate it means its privacy is leaked and it may as well just be a public field rather than a private one with a getter?

sdk/src/main/java/com/bugsnag/android/SessionTrackingPayload.java
fractalwrench: I probably should have mentioned that there have been some changes in this [react native PR](https://github.com/bugsnag/bugsnag-react-native/pull/345) which mean that this accessor is required in order to alter the payload.

sdk/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: It would be possible to wrap this with something like `Collections.unmodifiableList`, or even convert the class to Kotlin and use truly immutable collections. Other [similar callbacks](https://github.com/bugsnag/bugsnag-android/blob/6f6c88b1a416e02f0bf3fea0cdf3d5cd34fabc4c/sdk/src/main/java/com/bugsnag/android/Configuration.java#L494) don't seem to do this though, so for now I'd lean in favour of staying consistent with that implementation.

PR #478, opened 2019-05-03
Begin removing sleeps from mazerunner scenarios

fractalwrench:
## Goal

Several mazerunner scenarios use `Thread.sleep` to block, when they could use alternative non-blocking APIs that will allow the scenario to run faster. This will improve test run times when running locally and on CI.

## Changeset

- Removed `Thread.sleep` where it had no effect on the scenario result
- Made scenario implement `ActivityLifecycleCallbacks`
- Altered scenarios which rely on foreground to use lifecycle callbacks rather than arbitrary sleep

## Tests

Ran altered mazerunner scenarios locally.


features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/AutoContextScenario.kt
```diff
@@ -18,12 +20,11 @@ internal class AutoContextScenario(config: Configuration,
...
+        registerActivityLifecycleCallbacks()
         context.startActivity(Intent(context, SecondActivity::class.java))
-
-        val a = context as Activity
-        a.window.decorView.postDelayed({
-            Bugsnag.notify(generateException())
-        }, 2000)
     }
 
+    override fun onActivityStarted(activity: Activity) {
```
fractalwrench: This will be invoked as soon as the activity is in the foreground. The context should be automatically set to the activity's name at this time, so there's no point waiting for 2s.

features/in_foreground.feature
```diff
@@ -1,7 +1,7 @@
 Feature: In foreground field populates correctly
 
 Scenario: Test handled exception after delay
-    When I run "InForegroundScenario"
+    When I run "HandledExceptionScenario"
```
fractalwrench: For this scenario, we just need to verify that when a regular exception is thrown in the foreground - this is achievable by using `HandledExceptionScenario` since we altered `InForegroundScenario` elsewhere.

PR #482, opened 2019-05-15
Reduce sensitivity of Foreground Detection, reducing potential ANR false positives

fractalwrench:
## Goal

When detecting whether the application is in the foreground, we should try using
`IMPORTANCE_FOREGROUND_SERVICE` rather than `IMPORTANCE_VISIBLE`. We have observed false positives in ANRs for some specific device manufacturers and tightening the threshold for what is considered the foreground may reduce the signal-noise ratio.

## Tests

Ran ANR mazerunner scenarios locally.


fractalwrench:
Thanks @bengourley - I've updated that method to return `false` if that information isn't available.

PR #489, opened 2019-05-31
refactor: combine sdk + ndk modules

fractalwrench:
## Goal

The SDK + NDK modules should be combined so that we can deploy `bugsnag-android` (aliased with `bugsnag-android-ndk`) as a single artefact that detects JVM + C errors.

## Changeset

- Removed the `ndk` gradle module and moved native code to the `sdk` module
- Updated the git submodule path (`git submodule sync` may be necessary if already checked out)
- Updated tests, mazerunner, CI script, etc to maintain previous behaviour
- Updated outdated commands in the contributing guide

### Releasing

Added a `releaseNdkArtefact` gradle property which when passed by the command line builds the `bugsnag-android-ndk` artefact, rather than the default `bugsnag-android artefact`. The notifier can now be deployed with the following commands, then releasing on sonatype/bintray in the usual manner:

```
./gradlew assembleRelease publishSDKPublicationToMavenLocal && ./gradlew assembleRelease publishSDKPublicationToMavenLocal -PreleaseNdkArtefact=true
```

### General cruft removal

- Added nullability annotations to `NativeBridge` to pass lint checks
- Updated DexCount to remove Provider API warning being logged each gradle sync
- Replaced single quotes with double quotes (which are compatible with Kotlin Gradle DSL)
- Disabled example app failing bugsnag upload by default, which would prevent `./gradlew assemble` from succeeding
- Split gradle script into logical sections (e.g. separate file for release script configuration)

## Tests

I verified the following:

- `sdk` module compiles
- Checkstyle, detekt, lint, dexcount, and coveralls tasks pass
- A mazerunner feature passes
- The generated POM is correct for both artefacts
- The example app runs
- Local artifact can be used in example app (bugsnag-android-ndk only)
- Android instrumentation tests pass
- C tests pass
- Contributing guide is up to date
- Pass lint/checkstyle

## Known issues

- The example app won't work with a locally installed `bugsnag-android` artefact. This is because the gradle plugin won't extract the shared object files, and will be addressed with a separate PR
- The CI is likely to break (there's a long feedback loop waiting for Travis CI)


kattrali:
This changeset handles creating the two artifacts but switching the behavior for detecting C/C++ crashes by default. Is that going into a separate changeset? If so, I'd feel more comfortable reviewing this one and that one side by side so we know the release strategy in this PR will work.

fractalwrench:
@kattrali yes - that work has been split into a separate ticket so will be delivered separately

fractalwrench:
The previously mentioned changes are now ready for review in #491

PR #491, opened 2019-06-03
Add detectNdkCrashes property to control NDK crash reporting

fractalwrench:
## Goal

Add `config.detectNdkCrashes` property, which controls whether the NDK integration is enabled or not. In the bugsnag-android artefact, this value should be disabled by default; in bugsnag-android-ndk it should be enabled.

## Changeset

- Built the NDK lib for the mazerunner fixture by default
- Add `BuildConfig.DETECT_NDK_CRASHES` which is set via a gradle command line parameter
- Add `detectNdkCrashes` property to `Configuration`
- Move NDK initialisation to within `Client` constructor
- Simplify `NativeInterface` to avoid complexity + cost of reflection, by instantiating `NativeBridge` directly

## Tests

Added mazerunner scenario for when NDK crash detection is explicitly enabled and disabled. The default value is covered by existing tests.

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/Client.java
+++ b/sdk/src/main/java/com/bugsnag/android/Client.java
@@ -4,6 +4,7 @@ import static com.bugsnag.android.ConfigFactory.MF_BUILD_UUID;
 import static com.bugsnag.android.MapUtils.getStringFromMap;
 
 import com.bugsnag.android.NativeInterface.Message;
+import com.bugsnag.android.ndk.NativeBridge;
 
 import android.app.Activity;
 import android.app.ActivityManager;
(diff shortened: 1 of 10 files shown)


sdk/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -240,6 +241,10 @@ public void onOrientationChanged(int orientation) {
 
         // Flush any on-disk errors
         errorStore.flushOnLaunch();
+
+        if (config.getDetectNdkCrashes()) {
```
kattrali: Rather than making this conditional, the native interface should always be configured since it will be required by ANR detection.

`NativeBridge.install()` should instead receive `config.getDetectNdkCrashes()` as the `autoNotify` argument.

kattrali: _Insert in-person conversation_

This can stay as-is provided ANR detection is also added as a condition.

fractalwrench: If `detectNdkCrashes` and `detectAnrs` are both disabled, we can avoid initialising the NDK layer entirely. This is because it costs around 100ms to initialise the NDK layer and we should skip this if a user doesn't want any of that information. A potential risk is that this might introduce some new state to the notifier if users somehow trigger install of the NDK directly.

PR #493, opened 2019-06-05
Update example for new module, instructions

kattrali:


fractalwrench:
We should remember to add a note to the changelog about updating to the latest bugsnag-android-gradle-plugin for anyone using bugsnag-android, otherwise the SO files won't be found

PR #495, opened 2019-06-07
Add detection manifest options

kattrali:
## Goal

Make detection options configurable via App Manifest

## Changeset

* Update ConfigFactory to load detectNdkCrashes and detectAnrs from manifest if provided

## Tests

* Added new test assertions for the default values and overrides via manifest


sdk/src/main/java/com/bugsnag/android/ConfigFactory.java
```diff
@@ -100,7 +102,8 @@ static void populateConfigFromManifest(@NonNull Configuration config,
         config.setSendThreads(data.getBoolean(MF_SEND_THREADS, true));
         config.setPersistUserBetweenSessions(
             data.getBoolean(MF_PERSIST_USER_BETWEEN_SESSIONS, false));
-
+        config.setDetectNdkCrashes(data.getBoolean(MF_DETECT_NDK_CRASHES, false));
+        config.setDetectAnrs(data.getBoolean(MF_DETECT_ANRS, false));
```
fractalwrench: We should only set the field when the key is present in the manifest, otherwise we will have to alter code in two places when the default behaviour for `detectNdkCrashes` changes to `true`.

PR #500, opened 2019-06-13
Send minimal error report if cached file is corrupted/empty

fractalwrench:
## Goal

If a report cannot be delivered due to delivery failure, it is cached on disk. If this cached report cannot be deserialised, or is only partially written, e.g. due to lack of disk space, then currently no bugsnag report will be sent.

This changeset alters bugsnag-android to send a minimal error report for cached JVM exceptions only, by encoding basic error information in the filename.

## Changeset

- Updated `ErrorStore` to pass a `Delegate` that is invoked whenever a cached file cannot be deserialised
- Implemented `Delegate` in `Client` by sending a report with `DeliveryStyle.NO_CACHE` (this prevents minimal errors from generating other minimal errors if delivery fails and there's a serialisation bug in the notifier)
- Updated `Error` to include `incomplete` field that is set when a report is incomplete
- In the previous implementation cached files were only deserialised from JSON if a `beforeSend` callback was set; files are now always deserialised, which negates the need for `Report#errorFile`
- Added functions to the `ErrorStore` which encode and decode basic information about the error from the filename

## Tests

- Added a mazerunner scenario for a cached handled and unhandled JVM exception, where the file is either corrupted or entirely empty
- Adapted and added to existing unit test coverage for encoding/decoding of basic error info in the report's filename
- Manually verified that when a cached report is edited on disk and delivered to the dashboard:

Screenshot 2019-06-13 at 14 07 47

### DIFF
--- a/sdk/src/main/java/com/bugsnag/android/ErrorStore.java
+++ b/sdk/src/main/java/com/bugsnag/android/ErrorStore.java
@@ -23,7 +23,14 @@ import java.util.concurrent.Semaphore;
 class ErrorStore extends FileStore<Error> {
 
     interface Delegate {
-        void onErrorReadFailure(String filename);
+
+        /**
+         * Invoked when a cached error report cannot be read, and a minimal error is
+         * read from the information encoded in the filename instead.
+         *
+         * @param minimalError the minimal error, if encoded in the filename
+         */
+        void onErrorReadFailure(Error minimalError);
     }
 
     private static final String STARTUP_CRASH = "_startupcrash";
@@ -162,7 +169,8 @@ class ErrorStore extends FileStore<Error> {
                 + " to Bugsnag, will try again later", exception);
         } catch (Exception exception) {
             if (delegate != null) {
-                delegate.onErrorReadFailure(errorFile.getName());
+                Error minimalError = generateErrorFromFilename(errorFile.getName());
+                delegate.onErrorReadFailure(minimalError);
             }
             deleteStoredFiles(Collections.singleton(errorFile));
         }
@@ -212,7 +220,8 @@ class ErrorStore extends FileStore<Error> {
             String encodedErr = filename.substring(errorInfoStart, errorInfoEnd);
 
             char sevChar = encodedErr.charAt(0);
-            Severity severity = calculateSeverity(sevChar);
+            Severity severity = Severity.fromChar(sevChar);
+            severity = severity == null ? Severity.ERROR : severity;
 
             boolean unhandled = encodedErr.charAt(2) == 'u';
             HandledState handledState = HandledState.newInstance(unhandled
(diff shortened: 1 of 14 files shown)


fractalwrench:
Note to self: we should probably truncate the error class beyond a certain size to prevent the filename from being too large

sdk/src/main/java/com/bugsnag/android/ErrorStore.java
```diff
@@ -194,10 +262,14 @@ String getFilename(Object object) {
                 }
             }
         } else {
+            // NDK should always be U + E, as these errors are always fatal
```
bengourley: it says in the description of this PR that this feature only works for JVM errors, so is this code path never decended?

sdk/src/main/java/com/bugsnag/android/ErrorStore.java
```diff
@@ -133,22 +145,19 @@ void flushReports(Collection<File> storedReports) {
 
     private void flushErrorReport(File errorFile) {
         try {
-            Report report = null;
-            if (config.getBeforeSendTasks().size() > 0) {
```
bengourley: You mentioned in the PR description that "files are now always deserialised" but I don't know why this change was made… could you elaborate?

fractalwrench:
Thanks for the review! The best way to edit files on an emulator/device would be to set the device into airplane mode so that the crash is definitely stored, then use the [Device File Explorer](https://developer.android.com/studio/debug/device-file-explorer) to edit the file before relaunching the app. The file will be stored under `data/data/com.bugsnag.android.example/cache/bugsnag-errors, and you should be able to edit it like a regular file.

I'll have a read through the other comments and address them inline.

sdk/src/main/java/com/bugsnag/android/ErrorStore.java
fractalwrench: Previously there were two codepaths here:

1. If zero `beforeSendCallbacks` were set on the notifier, then we'd stream the error directly from disk to a network connection.
2. If `beforeSendCallbacks` are set on the notifier we'd deserialise the file into an `Error` object, then serialise that as JSON when making the request.

If the file has been corrupted then deserialisation of JSON should fail - at which point we need to send a minimal error report.

sdk/src/main/java/com/bugsnag/android/ErrorStore.java
fractalwrench: That's correct - I can remove this bit.

I originally thought that it would make sense to capture minimal error reports for NDK crashes, but realised that we can't encode the error class at the time of the crash, and as the severity/handled state will always be the same a minimal error report wouldn't be very useful.

fractalwrench:
I've addressed the remaining comments and also truncated the error class to a max of 40 chars - this is now ready for re-review.

PR #510, opened 2019-07-01
fix(ndk): Skip null metadata values from serialization

kattrali:
In the event that a metadata value is null, the patched section would
crash in `GetStringUTFChars()`. This is most easily reproduced by calling
notify/throwing on an exception with no message.

## Tests

* Manually tested by throwing an exception without a message. This would propagate null through `leaveErrorBreadcrumb()` which would then be serialized by the native layer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Fix abort() in native code when storing breadcrumbs with null values in
+  metadata
+  [#510](https://github.com/bugsnag/bugsnag-android/pull/510)
+
 ## 4.15.0 (2019-06-10)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Fix abort() in native code when storing breadcrumbs with null values in
+  metadata
```
fractalwrench: This is missing a link to the PR

features/handled_exception.feature
```diff
@@ -11,6 +11,17 @@ Scenario: Test handled Kotlin Exception
     And the exception "message" equals "HandledExceptionScenario"
     And the payload field "events.0.device.cpuAbi" is a non-empty array for request 0
 
+Scenario: Report a handled exception without a message
```
fractalwrench: This scenario won't be run against browserstack as it's in the legacy mazerunner directory. It'd be worth discussing with @Cawllec whether we need to write new scenarios to target both browserstack _and_ the local emulator.

PR #514, opened 2019-07-05
Remove deprecated interfaces from API

fractalwrench:
## Goal

Removes interfaces which have been deprecated for a long time from our API. This should reduce the size impact of the SDK, simplify the implementation, and encourage users to upgrade to superior APIs.

## Changeset

Removes the following methods:

- Obsolete `notify()` variants
- `SessionTrackingApiClient`, `ErrorReportApiClient` and associated classes in favour of `Delivery`
- Removed `endpoint`, `projectPackages` and  `maxBreadcrumbs` from `Bugsnag/Client`
- Removed `endpoint`, `sessionEndpoint`, `shouldAutoCaptureSessions`, `anrThresholdMs`, `inProject` from `Configuration`
- Remove individual `setNotifier*` methods on `Report` in favour of `Report#getNotifier()`
- Updated unit tests where appropriate, otherwise deleted where no longer necessary
- Updated SDK implementation to use new APIs, where the old API was still used


fractalwrench:
The mazerunner deprecations aren't related to the public API, so I've left them as-is for this PR, although it would be possible to update the scenarios to use the newer method.

`NativeInterface#configureClientObservers()` has only recently been deprecated in v4.17.0, so I was of the opinion that it would make sense to remove it whenever we release v6.0.0, rather than v5.0.0. Although this is more of an internal API so it might make sense to go for the breaking change - what are your thoughts?

bengourley:
> Although this is more of an internal API so it might make sense to go for the breaking change - what are your thoughts?

Do you think anybody is using this api other than us? If "no" then I think we can ditch it completely in the v5 release. It's still a major bump and this has already been marked with the deprecation annotation.

fractalwrench:
I feel the answer to that question is no, so I'll update the PR to remove that method.

fractalwrench:
@bengourley `configureClientObservers` has now been removed.

PR #522, opened 2019-07-11
Modularise bugsnag-android into JVM, NDK, and ANR artefacts

fractalwrench:
## Goal

The current version of `bugsnag-android` includes JVM, NDK, and ANR error detection by default. We should modularise these into 3 separate artefacts, so that it's possible for users to exclude code/resources they do not use from their APK.

This should be an invisible change to anyone using `bugsnag-android` and `bugsnag-android-ndk`.

## Changeset

- Created 3 modules which publish a JVM, NDK, and ANR artefact
- Created bugsnag-android and bugsnag-android-ndk module which publish the artefact intended for public consumption
- Updated CI/mazerunner to install to mavenLocal rather than copying AARs
- Updated contributing/readme instructions to account for new artefact structure
- Removed AAR size badge
- Altered ProGuard rules to retain `BugsnagPlugin` implementations
- Extracted shared gradle logic for library projects (e.g. lintoptions) to root-level build.gradle
- Disabled NDK crash detection by default for all artefacts
- Updated example app to use mavenLocal rather than compiling module directly, as this was somewhat error-prone in location SO files
- Migrated ANR code to separate module and SO file
- Add an publicly accessible `BugsnagPluginInterface` that allows for registration of plugins
- Create plugins for ANR + NDK module using `BugsnagPluginInterface`, which the `Client` attempts to load if the right `Configuration` option has been set

## Tests

Installed the artefacts into a local maven repository, and tested:

- depending on `bugsnag-android` allows JVM, NDK, and ANR errors to be captured
- both `bugsnag-plugin-android-anr` and `bugsnag-plugin-android-ndk` can be excluded from the APK if their respective configuration options are disabled
- When ANR/NDK detection are disabled using the `bugsnag-android` dependency, no shared libraries are loaded (improves init time from around 80ms to 50ms on a Nexus 5x)

## Remaining work

- The following bugsnag gradle plugin needs to be merged for SO files to be extracted from the new AARs. For review purposes, this can be installed locally. https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/168
- The artefact pages need to be setup on Sonatype/Bintray.
- We may want to consider smoke tests to verify that when the ANR/NDK artefacts are excluded, Bugsnag still initialises.


bugsnag-android-anr/README.md
```diff
@@ -0,0 +1,9 @@
+# bugsnag-android
```
kattrali:
```suggestion
# bugsnag-android-anr
```

bugsnag-android-core/README.md
```diff
@@ -0,0 +1,9 @@
+# bugsnag-android
```
kattrali:
```suggestion
# bugsnag-android-core
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -49,7 +49,7 @@
     private boolean automaticallyCollectBreadcrumbs = true;
 
     private boolean detectAnrs = false;
-    private boolean detectNdkCrashes = BuildConfig.DETECT_NDK_CRASHES;
+    private boolean detectNdkCrashes = false;
```
kattrali: Does this mean the NDK config option is no longer supported via the manifest?

bugsnag-android-ndk/README.md
```diff
@@ -0,0 +1,9 @@
+# bugsnag-android
```
kattrali:
```suggestion
# bugsnag-android-ndk
```

gradle.properties
```diff
@@ -11,7 +11,7 @@ org.gradle.jvmargs=-Xmx1536m
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
-VERSION_NAME=4.16.0
+VERSION_NAME=9.9.9
```
kattrali: Was this intentional?

examples/sdk-app-example/build.gradle
```diff
@@ -47,39 +35,25 @@ android {
...
+    lintOptions.disable "OnClick"
+    externalNativeBuild.cmake.path "CMakeLists.txt"
 }
 
 // set this value to true if you wish to test with a local artefact
 // it will instruct gradle to copy the SO files from the AAR to the correct build directory
 def usingLocalArtefact = false
 
 dependencies {
-    if (usingLocalArtefact) {
```
kattrali: This was kind of handy (and `usingLocalArtefact` is still defined above)

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Nope, just that we always default it to false.

examples/sdk-app-example/build.gradle
fractalwrench: This has caused quite a bit of confusion in other people who have used the notifier's example app. This method also occasionally fails to find the SO files and fails to build the entire project, which hasn't been reproducible yet but requires invalidating the entire gradle cache.

I'd therefore be in favour of just pushing people towards using `mavenLocal` and `./gradlew assembleRelease publishToMavenLocal` to test local changes

fractalwrench:
As discussed I will adapt this PR to do the following:

`bugsnag-android`: meta package that depends on all 3 artefacts
`bugsnag-android-ndk`: meta package that depends on all 3 artefacts (identical to bugsnag-android)

`bugsnag-plugin-android-jvm`: JVM exception handling
`bugsnag-plugin-android-anr`: ANR detection
`bugsnag-plugin-android-ndk`: NDK handlers

CONTRIBUTING.md
```diff
@@ -224,12 +198,11 @@ If you are a project maintainer, you can build and release a new version of
   - [ ] Update the version number and dex count badge by running `make VERSION=[number] bump`
   - [ ] Inspect the updated CHANGELOG, README, and version files to ensure they are correct
 - Once merged:
-  - Pull the latest changes (checking out master if necessary) and build by running `./gradlew sdk:assembleRelease`
+  - Pull the latest changes (checking out master if necessary) and build by running `./gradlew assembleRelease`
   - Release to GitHub:
     - [ ] Run `git tag vX.X.X && git push origin --tags`
     - [ ] Create a release from your new tag on [GitHub Releases](https://github.com/bugsnag/bugsnag-android/releases)
-    - [ ] Upload the generated `.aar` file from `{sdk,ndk}/build/outputs/aar/bugsnag-android-*.aar`
```
kattrali: Is there some discussion about removing the artifacts from GitHub releases?

fractalwrench: Manual installation has already been removed from our docs - it seems reasonable to ask anyone integrating an AAR manually to download it themselves from mavenCentral/bintray.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/AppDataTest.java
```diff
@@ -52,7 +52,7 @@ public void setUp() throws Exception {
 
     @Test
     public void testAccessors() {
-        assertEquals("com.bugsnag.android.test", appData.get("packageName"));
+        assertEquals("com.bugsnag.android.jvm.test", appData.get("packageName"));
```
kattrali: Is this still correct? There's a number of references to `com.bugsnag.android.jvm` hanging around.

bugsnag-android-core/src/main/AndroidManifest.xml
```diff
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-  package="com.bugsnag.android"
-  android:versionCode="1"
-  android:versionName="1.0">
+  package="com.bugsnag.android.jvm">
```
kattrali:
```suggestion
  package="com.bugsnag.android.core">
```
?

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.h
```diff
@@ -2,7 +2,7 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <unistd.h>
-#include <string.h>
+#include "string.h"
```
kattrali: This import used angle brackets intentionally to reference the system string utility rather than a local file. [GCC docs on search paths](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html).

```suggestion
#include <string.h>
```

gradle.properties
```diff
@@ -11,7 +11,7 @@ org.gradle.jvmargs=-Xmx1536m
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
-VERSION_NAME=4.16.0
+VERSION_NAME=4.16.1
```
kattrali: Was this intended?

features/fixtures/mazerunner/build.gradle
```diff
@@ -68,7 +76,7 @@ android {
 }
 
 dependencies {
-    implementation(name:'bugsnag-android-ndk', ext:'aar')
```
kattrali: There is the possibility with this change (even considering the exclusions above) that the tests could end up running against a release version instead of the local version. Is it now required to test the artifact this way?

examples/sdk-app-example/src/main/java/com.bugsnag.android/example/ExampleActivity.kt
```diff
@@ -139,13 +146,14 @@ class ExampleActivity : AppCompatActivity() {
...
+    @Suppress("UNUSED_PARAMETER")
     fun crashWithCallback(view: View) {
         val e = RuntimeException("Customized Error Report")
 
-        Bugsnag.notify(e, Callback { report ->
+        Bugsnag.notify(e) { report ->
             // modify the report
-            report.error?.metaData = generateUserMetaData()
-        })
+            report.error.metaData = generateUserMetaData()
```
kattrali: Did the nullability of `Report.error` change?

bugsnag-plugin-android-anr/src/main/jni/bugsnag_anr.h
```diff
@@ -0,0 +1,16 @@
+/**
+ * JNI interface between bugsnag-plugin-android-anr JVM and C++
+ */
+#ifndef BUGSNAG_ANR_H
+#define BUGSNAG_ANR_H
```
kattrali: Nothing is defined in this file. It should be deleted altogether, moving the one import to `bugsnag_anr.c`

bugsnag-plugin-android-anr/src/main/jni/utils/string.c
```diff
@@ -0,0 +1,13 @@
+#include <string.h>
```
kattrali: This should instead reference the relative string header rather than the standard system directory. With the addition of another function in this file, or ordering changes, it could become a confusing bug later. [GCC docs on search paths](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html)

```suggestion
#include "string.h"
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
kattrali: Wait but why?

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -12,6 +10,7 @@
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Observer;
```
kattrali: What is this import for?

bugsnag-plugin-android-anr/src/main/jni/bugsnag_anr.c
```diff
@@ -0,0 +1,18 @@
...
+
+#include <jni.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+JNIEXPORT void JNICALL Java_com_bugsnag_android_AnrPlugin_installAnrDetection(
+    JNIEnv *env, jobject _this, jobject byteBuffer) {
+    bsg_handler_install_anr((*env)->GetDirectBufferAddress(env, byteBuffer));
```
kattrali: There should be a null check for the buffer, just for sanity. If for whatever reason it could not be created, it would be an instant crash.

examples/sdk-app-example/src/main/java/com.bugsnag.android/example/ExampleActivity.kt
fractalwrench: Yes, in a recent release of bugsnag-android. Updating the example app's dependency means we can get rid of the unnecessary null check.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/AppDataTest.java
fractalwrench: Nope, this was missed, I've updated the tests and manifest package ID.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.h
fractalwrench: Android Studio seems to automatically suggest this as an import - I'll revert this change.

features/fixtures/mazerunner/build.gradle
fractalwrench: That shouldn't be possible as we [prevent gradle from fetching](https://github.com/bugsnag/bugsnag-android/pull/522/files/0720ad34d7a4298e6d0965d46f6196ebea10b18c#diff-c1c6e096014df07e7dd9bed50f7d2628R5) `com.bugsnag:bugsnag-android` from both mavenCentral and jcenter. This integration method also allows us to verify that the POM isn't malformed.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: This made sense with the original design where only `bugsnag-android` would have been published. Due to the fact we're now publishing bugsnag-android-ndk this will need to be updated.

features/fixtures/mazerunner/build.gradle
kattrali: I noted the exclusions in the original question, but I can definitely see cases where artifacts are incorrectly pulled from an unexpected source, causing the tests to pass (or fail) when they should not.

fractalwrench: I'm not 100% clear on what cases this could happen in, did you have an example in mind?

kattrali: bad caching :)

fractalwrench: I don't think this is a likely scenario as:

- mazerunner will exit before the scenarios run if `publishToMavenLocal` fails
- `mavenLocal` is specified as the first repository in the order for resolving dependencies
- `jcenter` and `mavenCentral` have been configured to stop resolving the `bugsnag-android` and `bugsnag-android-ndk` artefacts

This also improves our test coverage as it verifies the POM isn't malformed, which cannot be tested by using an AAR alone. Therefore I am leaving this as-is unless there is a concrete issue with this approach

PR #529, opened 2019-07-19
Attempt to fix mazerunner scenarios by disabling automatic session tracking

fractalwrench:
##Goal

The scenarios under `tests/features` were out of sync with the scenarios under `features`. Updating the scenarios should fix issues where the tests fail on buildkite due to an unexpected request.


fractalwrench:
Discussed in-person and this is ok to merge.

PR #530, opened 2019-07-22
Immutable config: add autoNotify property

fractalwrench:
## Goal

Adds config.autoNotify in favour of constructor/enable/disable methods. This value should be set
before Bugsnag.init, and if true will automatically capture JVM, NDK, and ANR errors.

### Deprecations
```
init(context, string, boolean)
enableExceptionHandler(boolean)
disableExceptionHandler(boolean)
```

- Added `autoNotify` property
- Updated existing implementation to use `config.autoNotify` where possible, and suppressed warnings where not possible

## Tests

Manually verified behaviour in example app, updated unit tests


fractalwrench:
We can't change the implementation to run _entirely_ off the `config.autoNotify` flag without breaking backwards compatibility, because `config.autoNotify` will be treated as immutable after initialisation. 
This is because it's only possible to call `ExceptionHandler.disable(this)` after initialisation, at which point changes to `config.autoNotify` won't be respected.

We could make the decision to break compatibility entirely, but it feels like serving a deprecation notice  is the nicer choice as it will allow people time to migrate and doesn't take too much effort on our part.

I will alter `config.getEnableExceptionHandler()` so that it returns the value of `config.getAutoNotify()`, which should hopefully make things somewhat clearer.

fractalwrench:
@bengourley `config.getEnableExceptionHandler()` has now been aliased onto `config.getAutoNotify()`. I've maintained backwards compatibility for now in this PR.

PR #532, opened 2019-07-22
immutable config: create Endpoints class, deprecate existing variants

fractalwrench:
## Goal

Creates an `Endpoints` class which encapsulates the `notify` and `sessions` fields, and automatically rejects any `null` values for either of these parameters. 

## Changeset

- Added `Configuration#setEndpoints(Endpoints)` method
- Deprecated existing methods, added backwards compatibility
- Removed redundant unit tests which were added when we supported setting the `notify` and `sessions` fields non-atomically

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
@@ -33,8 +33,6 @@ public class Configuration extends Observable implements Observer {
     private String buildUuid;
     private String appVersion;
     private String context;
-    private volatile String endpoint = "https://notify.bugsnag.com";
-    private volatile String sessionEndpoint = "https://sessions.bugsnag.com";
 
     private String[] ignoreClasses;
     @Nullable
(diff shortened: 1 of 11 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -154,86 +153,69 @@ public void setContext(@Nullable String context) {
...
-
-        if (invalidSessionsEndpoint) {
-            Logger.warn("The session tracking endpoint has not been set. "
-                + "Session tracking is disabled");
-            this.sessionEndpoint = null;
-            this.autoCaptureSessions = false;
-        } else {
-            this.sessionEndpoint = sessions;
-        }
+    public void setEndpoints(@NonNull Endpoints endpoints) {
```
bengourley: why is this better than `setEndpoints(@NonNull String notify, @NonNull String sessions)`?

bengourley: I think I'm going to answer my own question. It more closely matches the data structure you'd load from the config file.

PR #534, opened 2019-07-23
Fix NPE causing crash when reporting a minimal error

fractalwrench:
## Goal

v4.16.1 added reporting of minimal error information, by encoding key information such as handled state in the filename. If the JSON payload was corrupted in any way, this means an incomplete report could be delivered with at least some information.

If an older corrupted report was cached on disk this would return a `null` error as the information was not encoded in the filename. This changeset adds a null check to prevent an NPE in the `notify()` call.

## Tests

Added a mazerunner scenario that creates an empty file with an old filename that does not contain encoded information. The notifier should not generate a report from this file and should also not crash.


fractalwrench:
>the offending file isn't removed from the device's filesystem

I think this may be an issue with adb not synchronising the file directory, which we observed last time when testing the original minimal error implementation. The file is deleted here: https://github.com/bugsnag/bugsnag-android/pull/534/files#diff-5fcb9931669c81ccf180176a177f999eR180

I will investigate why that message shows up twice (and also make a small fix to the mazerunner tests)

bengourley:
> I think this may be an issue with adb not synchronising the file directory

I'm not sure about that… if repeatedly open/close the app (without making it crash again) it says this every time:

```
Sending 1 saved error(s) to Bugsnag
Sending 1 saved error(s) to Bugsnag
```

This suggests to me that the file isn't removed? (in addition to seeing it in the device file explorer)

fractalwrench:
I've done the following testing:

- Added an empty file via the Device File Explorer named '"1504255147933_683c6b92-b325-4987-80ad-77086509ca1e.json", which did not send a minimal error report or crash the app
- Added a partial + empty file via adb on a rooted emulator, which sent a minimal error report without crashing the app.

In all cases the file was deleted after the app was launched, and I did not observe `'Sending 1 saved error(s) to Bugsnag'` being displayed more than once at any point.

bengourley:
> > I think this may be an issue with adb not synchronising the file directory
> 
> I'm not sure about that… if repeatedly open/close the app (without making it crash again) it says this every time:
> 
> ```
> Sending 1 saved error(s) to Bugsnag
> Sending 1 saved error(s) to Bugsnag
> ```
> 
> This suggests to me that the file isn't removed? (in addition to seeing it in the device file explorer)

This was down to some weirdness with my emulator state. I tested on a different one (with a different api level) and it worked.

@fractalwrench tested on an emulator with the same api level as my weird on (26) and it also worked.

PR #536, opened 2019-07-23
Tests/revise test structure

Cawllec:
## Goal

Enable more control over when longer running tests run.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -76,14 +76,14 @@ steps:
 
   - wait
 
-  - label: ':android: Android 5 end-to-end tests'
+  - label: ':android: Android 9 end-to-end tests'
     plugins:
       artifacts#v1.2.0:
         download: "build/fixture.apk"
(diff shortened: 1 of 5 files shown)


fractalwrench:
Waiting for a few hours before merging just to make sure we have consensus on this change.

.buildkite/pipeline.yml
```diff
@@ -99,50 +99,52 @@ steps:
     concurrency: 5
     concurrency_group: 'browserstack-app'
 
-  - label: ':android: Android 6 end-to-end tests'
+  - block: "Trigger full test suite"
```
snmaynard: I think that block will mean that this doesn't technically pass from a github p of v until it is unblocked?

Why do the tests take 30 minutes to run?

fractalwrench: This is desired behaviour, as we can merge to non-protected branches but should run the full suite when merging to next/master.

Reducing CI job time is being discussed separately and is out-of-scope for this PR

Issue #533, opened 2019-07-22
Add config option to disable the orientation sensor


M66B:
[FairEmail](https://github.com/M66B/FairEmail/) is a privacy friendly email app and has a quite critical user base. It appeared that Bugsnag is using the accelerometer via the Android [OrientationEventListener](https://developer.android.com/reference/android/view/OrientationEventListener), which is unwanted in this app. I have disabled the listener [with an ugly hack](https://github.com/M66B/FairEmail/commit/6e56003ece73bbebbf90e2a112ce03f21e8d5c3c), but IMHO this should be possible by using the client configuration.

Can this please be added?

mattdyoung:
Hi @M66B 

Thanks for the suggestion. This sounds like a reasonable thing to allow in configuration but we've not had any other customers asking for it so it's unlikely to be prioritized for implementation in the near future unless that changes.

M66B:
Would you accept a pull request for this?

skvalex:
+1 for ability to disable it. Got plenty emails from users asking why my call recording app is using accelerometer all the time.

mattdyoung:
Hi @M66B 

Apologies for the delay in getting back to you. We've been having some internal discussions on this and we're now looking at a broader change to remove usage of the accelerometer altogether as there are alternative approaches we can use to determine whether the app is displayed in landscape / portrait without relying on device orientation changes from the accelerometer.

So we're not planning to accept your PR, but this concern should be addressed by our changes in the near future.

M66B:
Thanks for your response @mattdyoung 
Do you have any idea when this will be changed?

mattdyoung:
We're planning on doing a proof-of-concept next week. If all goes well we should be able to release something soon after that so we expect to have something in the order of weeks, rather than days or months.

M66B:
Great, thanks!

intmainreturn00:
What is the current status of this issue? Is it done already? 🙏
If there is some other way other than fork and fix for now? 

I want to add my 5 cents on why this could be important.
If someone uses [accelerometer batching](https://source.android.com/devices/sensors/batching) in android (it's basically a low energy mode for obtaining frequent accelerometer data) it only works if no one else is using accelerometer. 
In my case after many hours of research, I figured out that battery savers killed my service with accelerometer, because it was draining the battery, because bugsnag crash report utility was using accelerometer too. 😞

phillipsam:
[...]

PR #539, opened 2019-07-26
Added config option to disable capturing the device orientation

M66B:
## Goal

To disable sending the device orientation for people with privacy concerns about this.

Refs #533

## Design

Add and check a new property to enable the orientation listener.

## Changeset

A property was added to the Configuration class, default set to true, and methods were added to the Configuration class to change the default and to query to current property state.

The configuration property is used is check if the orientation listener should be enabled on creating the Client object.

## Tests

Since the change is very straightforward, no tests were performed nor added.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -239,7 +239,8 @@ public class Client extends Observable implements Observer {
             }
         };
         try {
-            orientationListener.enable();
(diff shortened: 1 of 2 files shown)


mattdyoung:
Hi @M66B 

Thanks for the PR, but we're not going to merge for the reasons mentioned here:
https://github.com/bugsnag/bugsnag-android/issues/533#issuecomment-516497004

Disabling via your proposed solution will no longer be necessary once we've implemented our planned upcoming changes.

Thanks, Matt

PR #541, opened 2019-07-26
refactor: extract user management code to separate class

fractalwrench:
## Changeset

This refactors the User implementation into a separate class, which encapsulates knowledge about shared preferences, and reduces the complexity of the Client class.

## Tests

Relied on existing tests


fractalwrench:
@simonbowring the `ArgumentContextScenario` test used a [deprecated API](https://github.com/bugsnag/bugsnag-android/pull/514/files#diff-9f3bb9fb69ec33c03c48cde0d51cbdb3L516) that was removed as part of #514, which means the E2E tests won't pass unless it's deleted.

simonbowring:
@fractalwrench, ok, are the two currently failing checks a separate problem that affects all PRs?

fractalwrench:
@simonbowring yes - this is an issue with the Browserstack devices on Android 5 that Alex is investigating.

PR #543, opened 2019-07-28
Set autoCapture as false by default

AllanWang:
Following the docs:

https://github.com/bugsnag/bugsnag-android/blob/8f490e66a287d0b55a3c114f7c3b7e45c5c4dcae/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java#L449-L459

it appears that auto capture should be false by default.

Even if you have some other function that sets it false upon init, it would be worth having the actual variable as false by default.


mattdyoung:
Hi @AllanWang 

Thanks for the PR, but it's the comment about the behaviour being disabled by default that's wrong.

Session tracking should be enabled by default, but can be disabled as described here in our docs:
https://docs.bugsnag.com/platforms/android/#session-tracking

I'm closing the PR as not required, but we'll make sure the comment gets corrected to avoid confusion in our next release.

PR #546, opened 2019-07-31
Fix potential segfaults when adding breadcrumb with NDK

fractalwrench:
## Goal

We have noticed infrequent crashes when adding a breadcrumb and NDK error detection is enabled. This changeset fixes some potential causes of this behaviour.

## Changeset

- Made `Breadcrumb` immutable by taking a defensive copy of the map parameter, and returning a decorated `Collection` from the getter that is unmodifiable.
- Corrected `sizeof` calculation when iterating through breadcrumb metadata
- Added null check for metadata and metadata keys in breadcrumbs

## Tests

- Added unit tests to verify `Breadcrumb` metadata immutability
- Tested leaving breadcrumbs with metadata in example app

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
@@ -7,6 +7,7 @@ import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -40,10 +41,7 @@ public final class Breadcrumb implements JsonStream.Streamable {
     Breadcrumb(@NonNull String name,
                @NonNull BreadcrumbType type,
                @NonNull Map<String, String> metadata) {
-        this.timestamp = DateUtils.toIso8601(new Date());
-        this.type = type;
-        this.metadata = metadata;
-        this.name = name;
+        this(name, type, new Date(), metadata);
     }
 
     Breadcrumb(@NonNull String name,
@@ -52,8 +50,8 @@ public final class Breadcrumb implements JsonStream.Streamable {
                @NonNull Map<String, String> metadata) {
         this.timestamp = DateUtils.toIso8601(captureDate);
         this.type = type;
-        this.metadata = metadata;
         this.name = name;
+        this.metadata = new HashMap<>(metadata);
     }
 
     @NonNull
@@ -68,7 +66,7 @@ public final class Breadcrumb implements JsonStream.Streamable {
 
     @NonNull
     public Map<String, String> getMetadata() {
-        return metadata;
+        return Collections.unmodifiableMap(metadata);
     }
 
     @NonNull
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
@@ -66,7 +66,7 @@ public final class Breadcrumb implements JsonStream.Streamable {
 
     @NonNull
     public Map<String, String> getMetadata() {
-        return Collections.unmodifiableMap(metadata);
+        return metadata;
     }
 
     @NonNull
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
```diff
@@ -68,7 +66,7 @@ public BreadcrumbType getType() {
 
     @NonNull
     public Map<String, String> getMetadata() {
-        return metadata;
+        return Collections.unmodifiableMap(metadata);
```
kattrali: Aren't there cases where people might be changing/removing breadcrumb metadata in a callback? The case which comes to mind is intent extras from the `EventReceiver`.

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -178,18 +178,24 @@ int bsg_populate_cpu_abi_from_map(JNIEnv *env, bsg_jni_cache *jni_cache,
 
 void bsg_populate_crumb_metadata(JNIEnv *env, bugsnag_breadcrumb *crumb,
                                  jobject metadata) {
+  if (metadata == NULL) {
+    return;
+  }
+
   bsg_jni_cache *jni_cache = bsg_populate_jni_cache(env);
-  int size = (int)(*env)->CallIntMethod(env, metadata, jni_cache->map_size);
+  int mapSize = (int)(*env)->CallIntMethod(env, metadata, jni_cache->map_size);
```
kattrali: The names are clear though this is mixing camel case and snake case.

bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
fractalwrench: While it's technically possible people are using the API in this way, I don't think we'd be breaking the contract as we haven't specified anything about what the Map is that's returned. Has the intent extras/other examples come up regularly before?

Also stepping back a bit, I think ideally `Breadcrumb` should return a `MetaData` here as it would be more consistent with the rest of our API. Although this would be a breaking change that should be addressed separately.

kattrali: > Has the intent extras/other examples come up regularly before?

I don't know about regularly, but there have been support requests in the past about removing metadata from automatic breadcrumbs. Since reports from the NDK pass through `beforeSend` callbacks, the breadcrumbs for native reports could also be reasonably customized. 

Making this collection unmodifiable doesn't make breadcrumb usage safer since the interface to leaving a breadcrumb doesn't expose the breadcrumb instance.

fractalwrench: I've altered this to return the backing map directly for now, and will raise a ticket to improve the API in future.

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -178,18 +178,24 @@ int bsg_populate_cpu_abi_from_map(JNIEnv *env, bsg_jni_cache *jni_cache,
...
+
   bsg_jni_cache *jni_cache = bsg_populate_jni_cache(env);
-  int size = (int)(*env)->CallIntMethod(env, metadata, jni_cache->map_size);
+  int map_size = (int)(*env)->CallIntMethod(env, metadata, jni_cache->map_size);
   jobject keyset =
       (*env)->CallObjectMethod(env, metadata, jni_cache->map_key_set);
   jobject keylist = (*env)->NewObject(
       env, jni_cache->arraylist, jni_cache->arraylist_init_with_obj, keyset);
-  for (int i = 0; i < size && i < sizeof(crumb->metadata); i++) {
+  size_t metadataSize = sizeof(crumb->metadata) / sizeof(bsg_char_metadata_pair);
```
kattrali:
```suggestion
  size_t metadata_size = sizeof(crumb->metadata) / sizeof(bsg_char_metadata_pair);
```

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -178,18 +178,24 @@ int bsg_populate_cpu_abi_from_map(JNIEnv *env, bsg_jni_cache *jni_cache,
...
+
+  for (int i = 0; i < map_size && i < metadataSize; i++) {
```
kattrali:
```suggestion
  for (int i = 0; i < map_size && i < metadata_size; i++) {
```

PR #551, opened 2019-08-01
Remove deprecated interfaces

fractalwrench:
## Goal

Rather than deprecating methods, we should remove the signatures entirely as part of the next major version bump. This changeset removes these methods and updates tests which used them.

## Tests

Ran existing instrumentation and mazerunner tests.


bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -34,7 +34,8 @@ private Bugsnag() {
      */
     @NonNull
     public static Client init(@NonNull Context androidContext) {
-        return init(androidContext, null, true);
+        String apiKey = null;
```
bengourley: what's the purpose of this change?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -392,73 +375,6 @@ public void setContext(@Nullable String context) {
...
-    @Deprecated
-    public void setAutoCaptureSessions(boolean autoCapture) {
-        config.setAutoCaptureSessions(autoCapture);
-
-        if (autoCapture) { // track any existing sessions
-            sessionTracker.onAutoCaptureEnabled();
-        }
-    }
-
-    private void setUserInternal(User user) {
```
bengourley: Why was this removed? Was it not used anywhere? Is user still observable if you set it via `setUser()`?

fractalwrench: This was removed unintentionally and has been added back in.

bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
fractalwrench: The change should be functionally equivalent - `autoNotify` is true by default so can be removed. We need to specify that `apiKey` is a `String` explicitly, as `init` has two methods that would otherwise be ambiguous if we passed null:

```
init(context, string)
init(context, configuration)
```

PR #552, opened 2019-08-02
Immutable config: make string[] properties use Collection instead

fractalwrench:
## Goal

Converts the `projectPackages`, `notifyReleaseStages`, `ignoreClasses`, and `filters` properties to use a `Collection<String>` rather than a `String[]`. This will reduce the surface area for NPEs, and allow us to change the implementation of the `Collection` behind the scenes if required which is not possible with an array.

A comparison of the old and new API is presented below:

```
client.setFilters("password", "auth") // old api
client.setFilters(listOf("password", "auth")) // new api
```

## Changeset

- Converted `String[]` properties to use `Collection<String>` throughout the codebase
- Updated `Configuration` implementation to use a `Set` as a backing field, which prevents duplicate entries
- Made getters return an unmodifiable `Set`, to prevent mutation

## Tests

Updated existing tests, and added a unit test that verifies direct modification of the properties throws an exception.


bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -345,7 +347,7 @@ public static String getAppVersion() {
      */
     @Nullable
     public static String[] getNotifyReleaseStages() {
-        return getClient().getConfig().getNotifyReleaseStages();
+        return getClient().getConfig().getNotifyReleaseStages().toArray(new String[]{});
```
bengourley: is this an array by necessity of the way items are passed from JVM->native?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -234,9 +237,9 @@ public void setFilters(@Nullable String[] filters) {
      *
      * @return Ignore classes
      */
-    @Nullable
-    public String[] getIgnoreClasses() {
-        return ignoreClasses;
+    @NonNull
+    public Collection<String> getIgnoreClasses() {
+        return Collections.unmodifiableSet(ignoreClasses);
```
bengourley: why is this made immutable upon `get()`? Why not make it an unmodifiable set when it is received in `setIgnoreClasses()` ? Same goes for `notifyReleaseStages` and `projectPackages`

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: It's been kept as an array as this will reduce rework required in the NDK/Unity, and as we're creating a copy of an array rather than exposing the original, the caller won't be able to mutate this state.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Java doesn't really have a truly immutable collection. `Collections.unmodifiableSet` provides a view of an existing collection to an end-user, that will throw an exception when operations such as element insertion/deletion are attempted. Either approach would be valid - arguably doing it in the setter would be better as it would prevent `Configuration` mutating the value, but #555 makes a truly immutable copy of this property anyway.

PR #553, opened 2019-08-06
Improve ANR error message information

fractalwrench:
## Goal

Android records a cause for an ANR in the `ProcessRecord` for the app process. Adding the message to error reports will improve context and differentiate timed out input dispatching from failing to respond to a broadcast event.

## Changeset

When `SIGQUIT` is caught bugsnag creates an `Error` which captures the stacktrace. Rather than sending this report immediately, the notifier now polls the `ActivityManager` for 30 seconds in an attempt to access [process error information](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/services/core/java/com/android/server/am/AppErrors.java#1088).

Once this information is available the notifier updates the error message to the value of `ProcessErrorStateInfo#shortMsg`, resulting in a stacktrace that looks like this:

```
ANR:  Input dispatching timed out (Waiting to send non-key event because the touched window has not finished processing certain input events that were delivered to it over 500.0ms ago.  Wait queue length: 24.  Wait queue head age: 6396.4ms.)
        at java.lang.Thread.sleep(Thread.java:-2)
        at java.lang.Thread.sleep(Thread.java:373)
        at java.lang.Thread.sleep(Thread.java:314)
        at com.bugsnag.android.example.ExampleActivity$onCreate$3.onClick(ExampleActivity.kt:38)
        at android.view.View.performClick(View.java:6256)
```

*Important Note*: Android has a concept of 'silent ANRs' which occur when the process is in the background. For silent ANRs, the process is [killed immediately](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/services/core/java/com/android/server/am/AppErrors.java#1082) without showing a dialog to the user. With this changeset Bugsnag will no longer detect silent ANRs, which may reduce 'overreporting' that has been raised as feedback.

### Other changes

- Made it possible to set the `name` and `message` on a `BugsnagException`, and added public methods for doing this via the `Error` object also
- Updated `Error` and `Exceptions` to always use a `BugsnagException` field. Regular `Throwables` are now wrapped into a `BugsnagException`

## Tests

- Added unit test coverage
- Ran existing tests and manually tested ANR detection in example app


bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -16,12 +22,45 @@ internal class AnrPlugin : BugsnagPlugin {
...
     private fun handleAnr(thread: Thread, client: Client) {
+        // generate a full report as soon as possible, then wait for extra process error info
         val errMsg = "Application did not respond to UI input"
         val exc = BugsnagException("ANR", errMsg, thread.stackTrace)
+        val error = Error.Builder(client.config, exc, client.sessionTracker, thread, true)
+            .severity(Severity.ERROR)
+            .severityReasonType(HandledState.REASON_ANR)
+            .build()
+
+        // wait for the ANR dialog to show, reporting once the information is available
```
kattrali: The [info is available](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/services/core/java/com/android/server/am/AppErrors.java#1088) before [the dialog shows](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/services/core/java/com/android/server/am/AppErrors.java#1093), but just barely. The delay is caused by dumping various stats about the running processes.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -16,12 +22,45 @@ internal class AnrPlugin : BugsnagPlugin {
...
+
+                    if (anrDetails == null) {
+                        Logger.warn("Looping in attempt to capture ANR details")
+                        Async.run(this)
+                    } else {
+                        Logger.warn("Captured ANR details: $anrDetails")
+
+                        // TODO determine how to populate error report
+                        val metaData = error.metaData
+                        metaData.addToTab("ANR", "Message", anrDetails.shortMsg)
```
kattrali: This should replace the `errMsg` above.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -16,12 +22,45 @@ internal class AnrPlugin : BugsnagPlugin {
...
+                        metaData.addToTab("ANR", "Component", anrDetails.tag)
```
kattrali: Isn't this the same as the default context, since it shows the activity name? If not, is it a better context for what is there currently?

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -4,8 +4,14 @@ import java.nio.ByteBuffer
 
 internal class AnrPlugin : BugsnagPlugin {
 
+    companion object {
+        private const val INFO_POLL_THRESHOLD_MS: Long = 2500
```
kattrali: This might need to be a shorter threshold. I've seen the dialog pop in ~1 second to as many as 12 seconds (when the device was overheating from other processes).

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
fractalwrench: I don't think this is the same as the default context. I believe if an ANR occurs in a `BroadcastReceiver` this value will be different - although we'll need to test that hypothesis.

kattrali: That would be pretty cool if so!

fractalwrench: This does not appear to be the case in my testing with a broadcastReceiver.

bugsnag-plugin-android-anr/src/androidTest/java/com/bugsnag/android/AnrDetailsCollectorTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun emptyListReturnsNull() {
+        Mockito.`when`(am.processesInErrorState).thenReturn(listOf())
+        assertNull(collector.captureProcessErrorState(am, 0))
+    }
+
+    @Test
+    fun differentPidReturnsNull() {
+        Mockito.`when`(am.processesInErrorState).thenReturn(listOf(stateInfo))
+        val captureProcessErrorState = collector.captureProcessErrorState(am, PID_EXAMPLE)
+        assertEquals(stateInfo, captureProcessErrorState)
```
kattrali: This test and the one below seem to be identical. Was that intentional? Nothing here is returning null.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrDetailsCollector.kt
```diff
@@ -0,0 +1,73 @@
...
+    internal fun captureProcessErrorState(am: ActivityManager, pid: Int): ProcessErrorStateInfo? {
+        return try {
+            val processes = am.processesInErrorState ?: emptyList()
+            processes.firstOrNull { it.pid == pid }
+        } catch (exc: RuntimeException) {
+            null
+        }
+    }
+
+    internal fun mutateError(error: Error, anrState: ProcessErrorStateInfo) {
```
kattrali: This function could use a more descriptive name about how the error is changed. `addErrorStateInfo()`, etc.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrDetailsCollector.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.os.Handler
+import android.os.HandlerThread
+import android.os.Process
+import android.support.annotation.VisibleForTesting
+import java.util.concurrent.atomic.AtomicInteger
+
+internal class AnrDetailsCollector {
+
+    companion object {
+        private const val INFO_POLL_THRESHOLD_MS: Long = 1000
```
kattrali: Did you try this at a shorter duration? 100ms should be acceptable for the short period of time this poll needs to run.

bugsnag-plugin-android-anr/src/androidTest/java/com/bugsnag/android/AnrDetailsCollectorTest.kt
fractalwrench: This was unintentional - I've updated the assertion for this test

fractalwrench:
@kattrali thanks for the review comments, I've updated the implementation and also added an `ANR` tab with a `CPU Info` key. The poll interval in `AppNotRespondingMonitor` has been left at 5ms.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BugsnagExceptionTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private fun jsonObjectFromException(bugsnagException: BugsnagException): JSONObject {
+        val exceptions = Exceptions(config, bugsnagException)
+        val json = streamableToJsonArray(exceptions)
+        val exc = json.get(0) as JSONObject
+        return exc
+    }
+
+    // TODO convert JavaScriptException to BugsnagException and create Stacktrace
```
tomlongridge: Any reason this can't be done now? Is it lengthy to do?

fractalwrench: I added this comment early on in the implementation of this fix and evidently forgot to remove it. It turns out that `JavaScriptException` adds in an additional `columnNumber` field in its serialisation code, so it can't use `BugsnagException` unless we update the implementation of that class to handle this. I also felt that converting `JavaScriptException` to `BugsnagException` would make this changeset larger than it needed to be to fix the issue, and risk introducing some other issue.

PR #555, opened 2019-08-09
Immutable Config: perform a deep copy of Configuration

fractalwrench:
## Goal

The Client constructor currently takes a `Configuration` parameter. Presently it's possible for a user to retain the reference to that parameter and mutate `Configuration` after initialisation. For example, like thus:

```
val config = Configuration("api-key")
Client(config)
config.releaseStage = "foo"
```

This changeset updates the notifier to make a deep copy of any immutable properties into an `ImmutableConfig` class. It also returns an implementation of `BugsnagConfiguration`, which contains mutators for properties such as metadata which can be altered after initialisation.

## Design

`ImmutableConfig` is not publicly accessible, and is implemented as a [data class](https://kotlinlang.org/docs/reference/data-classes.html) meaning that it is effectively immutable and therefore thread-safe. One alternative approach would have been to create a copy of a `Configuration` object. This approach was not chosen as it would still be possible to mutate fields from within the notifier.

`BugsnagConfiguration` _is_ publicly accessible and returns methods for mutating fields that can change after initialisation, e.g. `context`. It is accessible by calling `client.getConfiguration()`. This attempts to be consistent with the design for bugsnag-cocoa.

## Changeset

- Added `ImmutableConfig` which is used internally within notifier
- Added `BugsnagConfiguration` which is returned to anyone mutating `Configuration`

Due to the large scope of the changeset, other change have been annotated with inline comments to make review easier.

## Tests

- Added new unit test coverage for converting `Configuration` to `ImmutableConfig`
- Updated existing tests to use `ImmutableConfig` as needed
- Ran example app and sent example errors to dashboard


fractalwrench:
One more note: I'm not particularly convinced anymore of the need for `BugsnagConfiguration`, as it's possible to achieve the same using methods on the `Client`. It would be good to get thoughts on this.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/AppDataOverrideTest.kt
```diff
@@ -22,13 +22,13 @@ class AppDataOverrideTest {
 
     @Before
     fun setUp() {
-        val config = Configuration("api-key")
+        val config = BugsnagTestUtils.generateConfiguration()
```
fractalwrench: This is a test helper method that produces a `Configuration` object - `convert` maps that to an `ImmutableConfig` object. Several of the unit tests have been updated to use this.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbsTest.kt
```diff
@@ -22,13 +22,11 @@ import java.util.Locale
...
 
     private lateinit var breadcrumbs: Breadcrumbs
-    private lateinit var config: Configuration
     private var client: Client? = null
 
     @Before
     fun setUp() {
-        config = generateConfiguration()
-        breadcrumbs = Breadcrumbs(config)
+        breadcrumbs = Breadcrumbs(32)
```
fractalwrench: Breadcrumbs now takes an int directly rather than config.maxBreadcrumbs

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ConcurrentCallbackTest.java
```diff
@@ -33,7 +33,8 @@ public void tearDown() throws Exception {
 
     @Test
     public void testClientNotifyModification() throws Exception {
-        final Collection<BeforeNotify> beforeNotifyTasks = client.config.getBeforeNotifyTasks();
+        Configuration config = (Configuration) client.getConfiguration();
```
fractalwrench: We cast `BugsnagConfiguration` to `Configuration` for the purposes of testing only.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ConfigurationTest.java
```diff
@@ -42,31 +43,31 @@ public void testEndpoints() {
...
 
         // Should notify if releaseStage in notifyReleaseStages
-        config.setNotifyReleaseStages(new String[]{releaseStage});
-        assertTrue(config.shouldNotifyForReleaseStage(releaseStage));
+        immutableConfig = createConfigWithReleaseStages(config, stages, "production");
+        assertTrue(immutableConfig.shouldNotifyForReleaseStage());
     }
 
-    @Test
-    public void testShouldIgnore() {
```
fractalwrench: This test has been moved to `ErrorTest`

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ConfigurationTest.java
```diff
@@ -88,22 +89,6 @@ public void testAutoCaptureSessions() throws Exception {
         assertFalse(config.getAutoCaptureSessions());
     }
 
-    @Test
```
fractalwrench: These have been moved to `ImmutableConfigTest`

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
```diff
@@ -129,15 +129,6 @@ public void testAddNullMetadataToMetaDataSendsMessage() {
         assertEquals("bar", metadataItem.get(1));
     }
 
-    @Test
```
fractalwrench: These tests are no longer needed as immutable properties don't need to sync with the native layer.

bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -50,9 +54,9 @@
     private final ThreadState threadState;
     private boolean incomplete = false;
 
-    Error(@NonNull Configuration config, @NonNull Throwable exception,
+    Error(@NonNull ImmutableConfig config, @NonNull Throwable exception,
           HandledState handledState, @NonNull Severity severity,
-          Session session, ThreadState threadState) {
+          Session session, ThreadState threadState, MetaData globalMetaData) {
```
fractalwrench: `globalMetaData` is populated from `BugsnagConfiguration`

bengourley:
> One more note: I'm not particularly convinced anymore of the need for BugsnagConfiguration, as it's possible to achieve the same using methods on the Client. It would be good to get thoughts on this.

If I understand this correctly, I think I agree with it.

It seems stange to me that there's an `immutableConfig` and `mutableConfig` passed around. I don't think the the stuff which is mutable is config and so mutating it shouldn't happen on a thing called config _after_ bugsnag has been initialised.

fractalwrench:
I've renamed all variable usage of `mutableConfig` to `clientState`.

I will raise my second thoughts around `BugsnagConfiguration` in an appropriate Slack channel to get some discussion going, but I think any potential change can be addresed in another PR.

fractalwrench:
@bengourley would you mind taking a look at this one again now that `mutableConfig` has been renamed to `clientState`?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -23,11 +23,8 @@
  * User-specified configuration storage object, contains information
  * specified at the client level, api-key and endpoint configuration.
  */
-public class Configuration extends Observable implements Observer {
```
Pezzah: I'm concerned that this change might introduce errors for people upgrading if they have code that looks like

```
val config = Configuration("my-api-key")
config.endpoints = Endpoints("http://localhost:8000", "http://localhost:10000")
Bugsnag.init(this, config)

config.releaseStage = "test" // Will no longer apply the release stage to the running bugsnag
```
Then it will start to fail silently after the change

I wonder if it would be better to have a builder pattern for the new configuration, where the immutable config is only available as part of the builder:

```
val configBuilder = ConfigurationBuilder("my-api-key")
configBuilder.endpoints = Endpoints("http://localhost:8000", "http://localhost:10000")
val config = configBuilder.build()
Bugsnag.init(this, config)

config.releaseStage = "test"  // Error cannot set releaseStage
```

That would mean that errors would be discovered at compile time

bengourley:
> @bengourley would you mind taking a look at this one again now that mutableConfig has been renamed to clientState?

Retrospectively yes it looks good to me.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
bengourley: I think you make a good point. With the builder interface would that be a compile or a runtime error? If runtime you could do implement this with the current interface – `Configuration` has a `private boolean consumed` field which gets set to `true` when passed to `Bugsnag.init()` and subsequent setters check `consumed==false` and throw otherwise.

If this would be a compile error, arguably that's much better.

Pezzah: The way I've written that example I think it would be a compile time error (but that is a bit contrived). In reality it would probably be written more like:

```
Bugsnag.init(this, new ConfigBuilder()
            .setEndpoints("http://localhost:8000", "http://localhost:10000")
            .setReleaseStage("test")
            .build()
)
```
which means that the config object can't be accessed directly after bugsnag has been initialised anyway

fractalwrench: I'm not sure the builder route makes as much sense on Kotlin as it would Java, since the convention there is to use named parameters with default values.

We could certainly throw an exception from all setters once `Configuration` has been consumed. We'd have to be sure that this would be worth the complexity/maintenance overhead for whenever a new property is added

snmaynard: What does the java interface look like if you use named and default parameters in a kotlin class? The builder pattern in kotlin isn't that bad if it makes the java side much better - but we really want to pick the lesser of two evils here I think.

snmaynard: Would offering both make sense? We could rely on proguard to strip the builder class if someone is using kotlin and not referencing the builder? I don't know if thats a sensible/possible suggestion - just throwing out ideas to try and get the best of both rather than the lesser of two evils.

PR #561, opened 2019-08-15
Unify mazerunner scenario fixtures

fractalwrench:
## Goal

The code for each mazerunner scenario class is currently duplicated under `tests/features` and `features`. This increases the amount of maintenance required when changing tests, or adding new ones, as we have to edit files in two places and keep them both in sync with each other.

This changeset adds a `mazerunner-scenarios` module in the project that is capable of containing `Scenario` classes. The two separate fixtures have been configured to depend on this module when mazerunner is invoked, meaning the duplication is removed.

An additional benefit is that this detects compilation/style issues with scenario code much earlier on in the CI job, reducing the feedback loop.

## Changeset

- Added detekt + lint baseline files to suppress existing warnings for each module
- Added a module `mazerunner-scenarios`, which is compiled as part of the main project
- Moved `Scenario.kt` and 3 other scenarios to `mazerunner-scenarios`, as a POC
- Updated the fixture projects to also depend on the `mazerunner-scenarios` module (these still remain separate gradle projects, and we still test against an AAR)

Eventually we will want to spend a bit more time moving more mazerunner scenarios over - this PR only aims to provide a good starting point rather than completing the task in one go.

## Tests

Ran existing tests


fractalwrench:
We can disable the linter for the `mazerunner-scenarios` if needed, I think the major benefit is getting quicker feedback about whether a scenario class compiles or not. What's your preference for linting the scenarios?

Cawllec:
It's worth running it as is, and we'll see over the course of moving the other scenarios into the `mazerunner-scenarios` package whether it's an issue or not.

PR #564, opened 2019-08-20
Reduce information collected in minimal error report

fractalwrench:
## Goal

Avoids collecting misleading information in minimal error reports, as these are generated on a
separate app launch and can cause user confusion. This prevents collection of breadcrumbs and dynamic app/device data, all of which can change between sessions. We now only collect a subset of static app/device data, that is equivalent to what we send in our session payloads.

## Changeset

- Disabled breadcrumb collection in incomplete error reports
- Only collected app/device summary data, which consists of static information that does not change between sessions
- Avoided collecting session/context information in payload

## Tests

- Updated mazerunner scenario
- Verified that a rooted emulator collected suitable information (see screenshots for example error)

Screenshot 2019-08-20 at 13 38 05
Screenshot 2019-08-20 at 13 38 01


fractalwrench:
I have updated `failureReason` so that is named as `notifierFailureDetails` instead, as suggested in a review in #566

I have also updated `flushErrorReport()` so that it always sends minimal reports rather than passing them through `notify()`, as suggested by @kattrali

bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
```diff
@@ -170,14 +174,23 @@ private void flushErrorReport(File errorFile) {
...
-            if (delegate != null) {
-                Error minimalError = generateErrorFromFilename(errorFile.getName());
+            deleteStoredFiles(Collections.singleton(errorFile));
+        }
+    }
 
-                if (minimalError != null) {
-                    delegate.onErrorReadFailure(minimalError);
-                }
+    @Nullable
```
kattrali: This should not be nullable. In the event that the filename metadata cannot be parsed fully, it should make a default error selecting from `Exception`:`error`:`unhandled` for whichever parts are missing.

bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
```diff
@@ -170,14 +174,23 @@ private void flushErrorReport(File errorFile) {
             Logger.warn("Could not send previously saved error(s)"
                 + " to Bugsnag, will try again later", exception);
         } catch (Exception exception) {
-            if (delegate != null) {
-                Error minimalError = generateErrorFromFilename(errorFile.getName());
+            deleteStoredFiles(Collections.singleton(errorFile));
```
kattrali: What error case is this `catch` block handling?

fractalwrench: Any unchecked exceptions which can be thrown from within the `try` block, e.g. in the unlikely case that an `IllegalStateException` is thrown from within `Report#toStream()`.

PR #566, opened 2019-08-20
Add failureReason array to enable diagnostics in error reports

fractalwrench:
## Goal

Adds an array for internal use when debugging notifier issues. As an additional diagnostic tool we now record the exception message from when a minimal report is created, along with its file length, which should tell us whether the report was an empty file or invalid JSON.

## Changeset

Adds a `failureReason` array of strings to the error payload.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
@@ -4,7 +4,9 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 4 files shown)


fractalwrench:
Closing in favour of #564

bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -108,6 +112,15 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
             writer.name("threads").value(threadState);
         }
 
+        if (!failureReasons.isEmpty()) {
+            writer.name("failureReason").beginArray();
```
bullmo: I think that it may be better to name this something like `reportFailureDetails` or `notifierFailureDetails` so that it:
- Is clearer that it is an array (as it is pluralized)
- Is more generic than `failureReason` (as we may want to include diagnostic details that aren't the reason for failure)
- Is a bit clearer that it's not to do with the error that is being reported but is regarding the generation of the report itself

fractalwrench: Can do. I'll be raising this in another PR and will rename the fields/JSON `notifierFailureDetails`.

PR #570, opened 2019-08-22
Report internal SDK errors to bugsnag

fractalwrench:
## Goal

Rather than sending minimal error reports, the SDK now reports internal SDK errors to bugsnag
instead in a way that does not route them to customer dashboards. A minimal report is generated
which is not altered in callbacks and is not cached to disk.

## Changeset

- Updated `ErrorStore` to inform a delegate of an exception, rather than creating an error from the filename
- Removed sending of minimal error reports and associated tests
- Updated `onErrorReadFailure()` to add metadata to a generated error about the file being read, and then report an internal error to bugsnag
- Added `reportInternalBugsnagError()`, which sends an error report that is picked up by Bugsnag's internal monitoring
- Send minimal app/device data in internal error reports, along with some metadata about the exception
- Avoid running callbacks, caching internal error reports to disk, or reattempting delivery

Note: in order to alter the headers on a per-request basis, `DefaultDelivery` must be used as the delivery mechanism, as otherwise we'd need to make a breaking change to the delivery signature. If a user has specified a custom `Delivery` mechanism we're unable to send these reports. This API limitation has already been rectified in the immutable config work where we can specify different headers for individual requests

## Tests

- Manually inspected payload
- Sent payload to dashboard (by temporarily altering changeset to add an API key)
- Added a mazerunner scenario to verify internal error reports are sent


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -972,6 +982,51 @@ void notify(@NonNull Error error,
...
+                @Override
+                public void run() {
+                    try {
+                        Delivery delivery = config.getDelivery();
+
+                        // can only modify headers if DefaultDelivery is in use
+                        if (delivery instanceof DefaultDelivery) {
+                            Map<String, String> headers = config.getErrorApiHeaders();
+                            headers.put("Bugsnag-Internal-Error", "true");
+                            headers.remove(Configuration.HEADER_API_KEY);
```
snmaynard: I cant think why this needs to be removed?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -972,6 +982,51 @@ void notify(@NonNull Error error,
...
+                            DefaultDelivery defaultDelivery = (DefaultDelivery) delivery;
+                            defaultDelivery.deliver(config.getEndpoint(), report, headers);
+                        }
+
+                    } catch (Exception exception) {
+                        Logger.warn("Failed to report minimal error to Bugsnag", exception);
```
snmaynard: switch minimal to internal

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: The original proposal was to remove the API key rather than add in the `Bugsnag-Internal-Error` header - I'll update so that we don't remove the API key.

fractalwrench: We don't send the API key in the payload or header because if on-prem users upgrade their notifier without upgrading on-prem, these internal error reports will be delivered to their dashboard.

tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/InternalReportScenario.kt
```diff
@@ -0,0 +1,17 @@
...
+
+internal class InternalReportScenario(config: Configuration,
+                                      context: Context) : Scenario(config, context) {
+    init {
+        config.setAutoCaptureSessions(false)
+    }
+
+    override fun run() {
+        super.run()
+        sendInternalReport(RuntimeException("Whoops"), config, Bugsnag.getClient())
```
kattrali: Rather than literally calling the internal method for sending an internal report, a good test would create a malformed file in the cache dir and see what happens when it is flushed through the system. This current implementation shows that `reportInternalBugsnagError()` works correctly when called directly but not what would happen if a report file was broken.

fractalwrench:
I've updated the mazerunner scenario so that it now creates an empty file in the reports directory, which will trigger an internal error report.

bengourley:
I tested the following permutations and all exhibited the expected outcomes:

| Case | Outcome | Expected |
|------|-----------|----------|
| Empty file in the `bugsnag-errors` directory | Sent `java.io.EOFException` to internal dashboard | ✅ |
| Invalid JSON file in `bugsnag-errors` directory | Sent `android.util.MalformedJsonException` to internal dashboard | ✅ |
| Valid JSON with no errors in `bugsnag-errors` directory | Sent `java.io.IOException` to internal dashboard | ✅ |
| Missing stacktrace | Sent the stack-less exception to normal dashboard | ✅ |

PR #574, opened 2019-08-27
fix(config): Avoid NPE if setting metadata to null

kattrali:
I came across this when testing something else - if config.metadata is set to null, it causes an NPE. I then realized there was a test for this prior to #513, which I completely glossed over. So I added back the test and fixed the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


kattrali:
Agreed that there should have been a warning though I didn't see anything. I happened to do it while testing something unrelated, and my metadata variable became null in a non-obvious way. The annotations don't force anyone to code in a particular way.

PR #576, opened 2019-08-30
Fix deserialization of custom stackframe fields in cached error reports

fractalwrench:
## Goal

This changeset fixes the deserialization of custom fields in a stackframe by `ErrorReader`, when an error report has been cached on disk. This caused several issues:

- Prevented delivery of JS exceptions in React Native where the method was null for some stackframes
- Missing `columnNumber` in React Native JS exceptions, reducing the available information for source mapping
- Missing `loadAddress`, `frameAddress`, and `symbolAddress` fields for NDK errors
- Stacktrace length was not trimmed for JS/NDK exceptions at the JVM deserialisation layer, meaning this could cause recursive stacktraces to be dropped by our API

## Changeset

If no `beforeSend` callback has been specified, skip deserialization of the error report and stream the file directly, to reduce unnecessary work and mitigate the impact of any other deserialisation bugs.

Generally speaking, the serialization has been changed to allow an arbitrary `List<Map<String, Object>>`, which contains the full stacktrace. Broken down across classes, this has had the following effects:

`CachedThread`:
- Instantiates a `Stacktrace` object eagerly in the constructor, rather than in the stream method
- Adds a secondary constructor which takes an arbitrary list of stackframes.

`ErrorReader`:
- Deserializes stackframes as a `List<Map>` rather than mapping data into a `StackTraceElement`. 

`Exceptions`:
- The `toStream` method now delegates to the `BugsnagException` class. The pre-existing implementation has been split across `BugsnagException` and `Stacktrace` respectively.

`Stacktrace`:
- Trim stacktrace length in the constructor rather than in the `toStream` method
- Add secondary constructor which provides a `List<Map>` for arbitrary stackframes
- Convert `StackTraceElement` to `List<Map>` by adapting the existing serialization code

`BugsnagException`:
- Implement `Streamable`. If the wrapped exception is of type `Streamable`, delegate the streaming to it
- Add a constructor which accepts `List<Map>` for serializing stackframes
- If a `List<Map>` has been supplied, construct the `Stacktrace` using that, rather than the throwable's stacktrace.

## Tests
- Added mazerunner scenarios with custom streamable exceptions
- Added and updated existing unit test coverage

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StacktraceTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StacktraceTest.java
@@ -16,7 +16,9 @@ import org.junit.Test;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 @FlakyTest(detail = "Checks a stacktrace's line number, so fails when lines are added/deleted.")
 @SmallTest
@@ -41,7 +43,7 @@ public class StacktraceTest {
         JSONArray stacktraceJson = streamableToJsonArray(stacktrace);
 
         JSONObject firstFrame = (JSONObject) stacktraceJson.get(0);
-        assertEquals(34, firstFrame.get("lineNumber"));
+        assertEquals(36, firstFrame.get("lineNumber"));
         assertEquals("com.bugsnag.android.StacktraceTest.setUp", firstFrame.get("method"));
         assertEquals("StacktraceTest.java", firstFrame.get("file"));
         assertFalse(firstFrame.has("inProject"));
@@ -70,6 +72,19 @@ public class StacktraceTest {
 
         StackTraceElement[] ary = new StackTraceElement[elements.size()];
         Stacktrace stacktrace = new Stacktrace(elements.toArray(ary), config.getProjectPackages());
+        JSONArray jsonArray = streamableToJsonArray(stacktrace);
+        assertEquals(200, jsonArray.length());
+    }
+
+    @Test
+    public void testStacktraceTrimmingListCtor() throws Throwable {
+        List<Map<String, Object>> elements = new ArrayList<>();
+
+        for (int k = 0; k < 1000; k++) {
+            elements.add(Collections.singletonMap("Foo", new Object()));
+        }
+
+        Stacktrace stacktrace = new Stacktrace(elements);
         JSONArray jsonArray = streamableToJsonArray(stacktrace);
         assertEquals(200, jsonArray.length());
     }
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StacktraceTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StacktraceTest.java
@@ -74,6 +74,8 @@ public class StacktraceTest {
         Stacktrace stacktrace = new Stacktrace(elements.toArray(ary), config.getProjectPackages());
         JSONArray jsonArray = streamableToJsonArray(stacktrace);
         assertEquals(200, jsonArray.length());
+        assertEquals(0, jsonArray.getJSONObject(0).getInt("lineNumber"));
+        assertEquals(199, jsonArray.getJSONObject(199).getInt("lineNumber"));
     }
 
     @Test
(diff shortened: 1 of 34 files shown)


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StacktraceTest.java
```diff
@@ -74,6 +76,19 @@ public void testStacktraceTrimming() throws Throwable {
...
+        List<Map<String, Object>> elements = new ArrayList<>();
+
+        for (int k = 0; k < 1000; k++) {
+            elements.add(Collections.singletonMap("Foo", new Object()));
+        }
+
+        Stacktrace stacktrace = new Stacktrace(elements);
+        JSONArray jsonArray = streamableToJsonArray(stacktrace);
+        assertEquals(200, jsonArray.length());
+    }
```
kattrali: This test should also assert that the correct values are being truncated. The test object should have a way to differentiate each ordered frame and then later assert that the "top" 200 frames are what have been included.

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagException.java
```diff
@@ -39,11 +48,23 @@ public BugsnagException(@NonNull String name,
             this.message = ((BugsnagException) exc).getMessage();
             this.name = ((BugsnagException) exc).getName();
             this.type = ((BugsnagException) exc).getType();
+        } else if (exc instanceof JsonStream.Streamable) {
```
kattrali: The first conditional case might by lossy. If `exc instanceof BugsnagException` and it wraps a streamable, the streamable content might be lost. However, every `BugsnagException`	is an instance of `Streamable` which could be used directly.

bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorReader.java
```diff
@@ -238,40 +239,41 @@ private static BugsnagException readException(JsonReader reader) throws IOExcept
...
+            Object val = null;
+
+            switch (reader.peek()) {
+                case STRING:
+                    val = reader.nextString();
                     break;
-                case "lineNumber":
-                    lineNumber = reader.nextInt();
+                case NUMBER:
+                    val = reader.nextInt();
```
kattrali: The exception possibly thrown here should be caught - there are no known cases of non-integer values being used here but `JsonToken.NUMBER` is not guaranteed to be an integer. The minimum handling should be to still send the report, though the maximal handling could be to create an internal report for it.

bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorReader.java
```diff
@@ -238,40 +239,41 @@ private static BugsnagException readException(JsonReader reader) throws IOExcept
...
-                    method = reader.nextString();
-                    break;
-                case "file":
-                    file = reader.nextString();
+            String key = reader.nextName();
+            Object val = null;
+
+            switch (reader.peek()) {
+                case STRING:
+                    val = reader.nextString();
```
kattrali: Its a bit more concise to put the value in the map immediately rather than checking afterwards that it is non-null. If the token type is `STRING` or `NUMBER` then we know the value isn't null.
```suggestion
                    map.put(key, reader.nextString());
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Exceptions.java
```diff
@@ -20,23 +23,8 @@
...
-                String localizedMessage = currentEx.getLocalizedMessage();
-                StackTraceElement[] stackTrace = currentEx.getStackTrace();
-                exceptionToStream(writer, exceptionName, localizedMessage, stackTrace);
-            }
-            currentEx = currentEx.getCause();
-        }
-
-        writer.endArray();
+        exception.setProjectPackages(projectPackages);
+        exception.toStream(writer);
```
kattrali: I need to spend a bit more time understanding the interactions between `Exceptions` and `BugsnagException` but so far its a little confusing how `BugsnagException` now represents an array rather than just the individual exception value, and in particular might introduce a situation where the "cause" exception of a `Streamable` might be skipped from serialization.

The deleted logic in `Exceptions` may have been more correct and simplify the execution path between `BugsnagException.toStream`, `BugsnagException.serializeException`, and `BugsnagException.writeThrowable`.

fractalwrench: Looking back at the implementation this makes sense. I've updated the logic so that `Exceptions` is responsible for iterating through the throwable causes, and made `BugsnagException` responsible for serializing its own information only.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.java
```diff
@@ -15,41 +19,68 @@
...
+            String filename = el.getFileName() == null ? "Unknown" : el.getFileName();
+            map.put("file", filename);
+            map.put("lineNumber", el.getLineNumber());
+
+            if (inProject(el.getClassName(), projectPackages)) {
+                map.put("inProject", true);
+            }
+            return map;
+        } catch (Exception lineEx) {
+            return null;
```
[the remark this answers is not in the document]
fractalwrench: I'm not particularly sure what this is catching - possibly a `null` value returned from `methodName` or other methods? I think the safe thing to do would be to add the log back in and keep everything else as-is.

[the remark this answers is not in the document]
kattrali: Alright, thanks for looking. Class name and method name can't be null at least, though I wondered if you saw some case that I missed here that we should potentially be testing for.

PR #578, opened 2019-09-02
Mazerunner unification

fractalwrench:
## Goal

Unifies all remaining mazerunner scenarios which don't rely on linking against the NDK library. This means we no longer have to update each scenario in two places, and also gain the benefits of style checks/compilation errors on CI at a much earlier stage.

The few remaining mazerunner scenarios will be migrated in a separate PR (#579) as additional work will be required for the build system to pick up the SO file.

The changeset has been commented inline to explain the diff where appropriate


features/detect_anr.feature
```diff
@@ -9,7 +9,7 @@ Scenario: Sleeping the main thread with pending touch events when detectAnrs = t
     Then I should receive a request
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "ANR"
-    And the exception "message" equals "Application did not respond to UI input"
+    And the exception "message" starts with " Input dispatching timed out"
```
fractalwrench: This fixes the local ANR scenario message information which changed in #553

features/fixtures/mazerunner/src/main/AndroidManifest.xml
```diff
@@ -11,7 +11,12 @@
 
             </intent-filter>
         </activity>
-        <activity android:name=".SecondActivity"/>
+        <activity android:name=".SecondActivity">
```
fractalwrench: This now uses an implicit intent rather than an explicit intent. This means scenarios which previously required `SecondActivity` on its classpath can specify `"com.bugsnag.android.mazerunner.UPDATE_CONTEXT"` as the intent instead, and they can be moved to the `mazerunner-scenarios` module instead.

mazerunner-scenarios/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 3.4.1)
+
+add_library(entrypoint SHARED
+             src/main/cpp/entrypoint.cpp
+             src/main/cpp/bugsnags.cpp)
+
+#FIXME determine how best to link against bugsnag-ndk library in mazerunner module
```
fractalwrench: This will be addressed in a future PR. Any mazerunner scenarios which require direct use of bugsnag's C API have not been ported yet. e.g.

```
bugsnag_leave_breadcrumb() // etc
```

mazerunner-scenarios/src/main/cpp/bugsnags.cpp
```diff
@@ -0,0 +1,205 @@
...
+JNIEXPORT int JNICALL
+Java_com_bugsnag_android_mazerunner_scenarios_CXXExtraordinaryLongStringScenario_crash(JNIEnv *env,
+                                                                         jobject instance,
+                                                                         jint value) {
+  if (value > 0)
+    __builtin_trap();
+  return 12062 / value;
+}
+
+//JNIEXPORT void JNICALL
```
fractalwrench: Calls which use bugsnag's C API have been commented out for now, and are compiled from the individual `features` and `tests/features` folders instead.

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/TestCaseFactory.kt
```diff
@@ -4,7 +4,7 @@ import android.content.Context
 import com.bugsnag.android.Configuration
 import com.bugsnag.android.mazerunner.scenarios.Scenario
 
-internal class TestCaseFactory {
+class TestCaseFactory {
```
fractalwrench: This is required to make `TestCaseFactory` visible to other modules which depend on `mazerunner-scenarios`

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/AppNotRespondingDisabledScenario.kt
```diff
@@ -18,7 +19,7 @@ internal class AppNotRespondingDisabledScenario(config: Configuration,
     override fun run() {
         super.run()
         val main = Handler(Looper.getMainLooper())
-        main.postDelayed({
+        main.postDelayed(Runnable {
```
fractalwrench: Resolves the `ImplicitSAM` lint check

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/AutoContextScenario.kt
```diff
@@ -21,7 +19,7 @@ internal class AutoContextScenario(config: Configuration,
     override fun run() {
         super.run()
         registerActivityLifecycleCallbacks()
-        context.startActivity(Intent(context, SecondActivity::class.java))
+        context.startActivity(Intent("com.bugsnag.android.mazerunner.UPDATE_CONTEXT"))
```
fractalwrench: Uses an implicit intent as previously mentioned

PR #581, opened 2019-09-03
Report internal errors when serialization fails

fractalwrench:
## Goal

#570 added internal error reports for deserialization, which are delivered to Bugsnag when an unanticipated error occurred. This adds similar functionality for the serialization case, which will allow us to diagnose errors when writing files.

## Changeset

- Moved `delegate` property to `FileStore` so that serialisation errors are captured for errors/session payloads
- Sent internal reports in two additional cases:
    - When copying an NDK report to the JVM directory fails
    - When serializing a JVM report to disk fails
- Captured the following additional diagnostic info in reports:

```
canRead // whether the file is readable
canWrite // whether the file is writable
exists // whether the file exists
usableSpace // the amount of usable space in bytes in the cache directory
```


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -122,7 +124,30 @@ public Client(@NonNull Context androidContext, @NonNull Configuration configurat
...
+                @SuppressLint("UsableSpace") // requires API 26
+                long usableSpace = appContext.getCacheDir().getUsableSpace();
+                metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "usableSpace", usableSpace);
+                metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "filename", errorFile.getName());
+                metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "fileLength", errorFile.length());
+                Client.this.reportInternalBugsnagError(err);
+            }
+        };
+
+        sessionStore = new SessionStore(config, appContext, delegate);
```
kattrali: Should this be done at all, or even be done differently for the session store? It has the potential to be much more verbose since every app launch sends a session rather than just error prone ones, and the context values should be different from "Crash Report ..." which is specified in `FileStore`.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -122,7 +124,30 @@ public Client(@NonNull Context androidContext, @NonNull Configuration configurat
...
+                Thread thread = Thread.currentThread();
+                Error err = new Error.Builder(config, exc, null, thread, true).build();
+                err.setContext(context);
+
+                MetaData metaData = err.getMetaData();
+                metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "canRead", errorFile.canRead());
+                metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "canWrite", errorFile.canWrite());
+                metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "exists", errorFile.exists());
+
+                @SuppressLint("UsableSpace") // requires API 26
```
kattrali: What is the linting issue here? [`getUsableSpace()`](https://developer.android.com/reference/java/io/File.html#getUsableSpace()) seems to available in API 9.

fractalwrench: The comment was a bit short - I've updated it to say that the alternative API requires API 26.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: Avoiding this for now due to the volume of the sessions vs errors sounds like a good call - I'll update the PR to pass in null as the delegate, which will prevent internal error reporting.

fractalwrench:
I originally tested this by assembling a local artefact, altering the relevant code to always throw an exception, then checking on a local server that the payload was correct. I've also tested against our internal instance and verified the exceptions are picked up.

This doesn't seem like the sort of change that would be feasible to test in an automated way - at least, I can't think of a good way to achieve that.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## TBD
+
+* Report internal errors when serialization fails
+  [#581](https://github.com/bugsnag/bugsnag-android/pull/581)
+
+* Buffer io when reading from cached error file
```
kattrali: This should say what this is potentially fixing. Possible crash from running out of memory? Unbounded IO? etc

fractalwrench: Updated to clarify that from the end-user's perspective this will improve the SDK's performance.

PR #582, opened 2019-09-04
fix(native): Add interface for clearing a tab

kattrali:
The current revision is missing an interface for clearing a tab.

## Tests

* Manually and using a new automated test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 4.15.x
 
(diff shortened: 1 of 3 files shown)


kattrali:
Thanks for the quick review! Added changelog entry.

kattrali:
I just re-read your comment - everything on the backports branch is already present in master (This change in particular was in https://github.com/bugsnag/bugsnag-android/pull/505 which was released in 4.16.0).

PR #583, opened 2019-09-04
Remove unused gradle plugins from build scripts

fractalwrench:
## Goal

The following plugins are no longer used in the project:

- Coveralls (reported code coverage via Travis CI builds)
- DexCount (counted number of Dex methods in AAR)
- Infer (detected thread safety violations, needed to be manually run)

We should therefore remove them from the build script to avoid adding an unnecessary overhead when building the project.

## Changeset

- Remove coveralls + DexCount plugins
- Removed infer script which must be manually run
- Removed package-private `ThreadSafe` annotation which is used by Infer to find common thread safety violations


fractalwrench:
DexCount is the most widely used one, in practice the dex method count is not as big a deal anymore due to multidex solutions, and we should probably focus on tracking general artefact size if we go down that route.

PR #584, opened 2019-09-05
[Android] Fix support for native crashes containing multiple text encodings

kattrali:
## Goal

Fix a possible crash in `NativeBridge.deliverReportAtPath()` arising from the report payload or release stage containing bytes which cannot be encoded as Modified UTF-8.

## Design

NewStringUTF() is not safe to use unless you control the input 100% of
the time. It uses the Modified UTF-8 charset, which may exclude possible 
characters, crashing with "illegal byte" errors. This issue can be mitigated 
by passing the raw byte array to the JVM layer, and then decoding the 
string as UTF-8. Using the `String(bytes[], Charset)` constructor bypasses
possible exceptions, instead replacing unknown bytes with the
unidentified character marker.

This fixes a possible crash where a string value is added to a report
using an encoding other than ASCII or Unicode variants (such as 
SHIFT-JIS, EUC-*, etc) which then cannot be decoded as UTF-8. For 
example, half-width kana in EUC-JP is represented by 0x8E followed by
a value between 0xA1 and 0xDF.

## Changeset

* Changed the parameters of `NativeInterface.deliverReport()` to accept byte arrays instead of Strings (removing usage of `NewStringUTF` from the report decoding process)
* Re-encoded strings passed to `NativeBridge` functions as UTF-8 before passing them to the native layer. This replaces any malformed or unmappable bytes with the unidentifier character marker.

### Reproduction case

1. Using a third-party keyboard in a non-ASCII language, enter text which is added to metadata or breadcrumbs (Alternate version: crash the app, replacing a random set of bytes in a report with `\0x8e\0xde`)
2. Cause a C/C++ crash
3. Relaunch the app

A crash can occur in `deliverReportAtPath()` when attempting to construct a string with byte sequences containing control characters.

* Tested this change manually on ARMv7 (device), ARM64 (device), and x86 64-bit (emulator) within a Unity app with text inputs to cause the crash.

This change will also need to be cherry-picked to `master`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 4.15.x
 
+* Fix possible crash when recording reports and breadcrumbs containing values
+  using different text encodings or UTF-8 control characters, followed by a 
+  C/C++ crash.
+  [#584](https://github.com/bugsnag/bugsnag-android/pull/584)
 * Fix crash when calling `NativeInterface.clearTab()` (from an integration
   library)
   [#582](https://github.com/bugsnag/bugsnag-android/pull/582)
--- a/sdk/src/main/java/com/bugsnag/android/NativeInterface.java
+++ b/sdk/src/main/java/com/bugsnag/android/NativeInterface.java
@@ -8,6 +8,7 @@ import android.support.annotation.Nullable;
 
 import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -125,6 +126,9 @@ public class NativeInterface {
          */
         UPDATE_USER_ID,
     }
+
+    // The default charset on Android is always UTF-8
+    private static Charset UTF8Charset = Charset.defaultCharset();
 
     /**
      * Wrapper for messages sent to native observers
@@ -273,11 +277,43 @@ public class NativeInterface {
     }
 
     /**
+     * Sets the user
+     *
+     * @param idBytes id
+     * @param emailBytes email
+     * @param nameBytes name
+     */
+    @SuppressWarnings("unused")
+    public static void setUser(@Nullable final byte[] idBytes,
+                               @Nullable final byte[] emailBytes,
+                               @Nullable final byte[] nameBytes) {
+        String id = idBytes == null ? null : new String(idBytes, UTF8Charset);
+        String email = emailBytes == null ? null : new String(emailBytes, UTF8Charset);
+        String name = nameBytes == null ? null : new String(nameBytes, UTF8Charset);
+        setUser(id, email, name);
+    }
+
+    /**
      * Leave a "breadcrumb" log message
      */
     public static void leaveBreadcrumb(@NonNull final String name,
                                        @NonNull final BreadcrumbType type) {
-        getClient().leaveBreadcrumb(name, type, new HashMap<String, String>());
+        if (name == null) {
+            return;
+        }
+        getClient().leaveBreadcrumb(name, type, Collections.<String, String>emptyMap());
+    }
+
+    /**
+     * Leave a "breadcrumb" log message
+     */
+    public static void leaveBreadcrumb(@NonNull final byte[] nameBytes,
+                                       @NonNull final BreadcrumbType type) {
+        if (nameBytes == null) {
+            return;
+        }
+        String name = new String(nameBytes, UTF8Charset);
+        getClient().leaveBreadcrumb(name, type, Collections.<String, String>emptyMap());
     }
 
     /**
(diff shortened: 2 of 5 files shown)


sdk/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -6,6 +6,7 @@
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
+import java.nio.charset.StandardCharsets;
```
fractalwrench: This was added in API 19, so we'll need to use `Charset.forName("UTF-8")` instead to support API 14+

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 4.15.x
 
+* Fix possible crash when recording reports and breadcrumbs containing values
```
fractalwrench: Would be good to include a link to the PR so anybody looking at the changelog/release notes can consult the changeset

sdk/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -419,7 +420,9 @@ public static void registerSession(long startedAt, @Nullable String sessionId,
      *                     stages
      */
     @SuppressWarnings("unused")
-    public static void deliverReport(@Nullable String releaseStage, @NonNull String payload) {
+    public static void deliverReport(@Nullable byte[] releaseStageBytes, @NonNull byte[] payloadBytes) {
```
fractalwrench: There are several other methods in `NativeInterface` which take a `String` as a parameter. Are they also affected by this issue?

sdk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -271,8 +274,16 @@ private boolean is32bit() {
     private void handleAddBreadcrumb(Object arg) {
         if (arg instanceof Breadcrumb) {
             Breadcrumb crumb = (Breadcrumb) arg;
+            Map<String, String> metadata = new HashMap<String, String>();
```
fractalwrench: This can be replaced with the diamond type since 1.6 for conciseness: `Map<String, String> metadata = new HashMap<>()`

sdk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -271,8 +274,16 @@ private boolean is32bit() {
...
+            if (crumb.getMetadata() != null) {
```
fractalwrench: `getMetaData()` is non-null so this shouldn't be necessary

sdk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -389,7 +401,7 @@ private void handleForegroundActivityChange(Object arg) {
             @SuppressWarnings("unchecked")
             List<Object> metadata = (List<Object>)arg;
             if (metadata.size() == 2) {
-                updateInForeground((boolean) metadata.get(0), (String) metadata.get(1));
+                updateInForeground((boolean) metadata.get(0), makeSafe((String) metadata.get(1)));
                 return;
             }
         }
```
fractalwrench: `handleUserId` is missing a call to `makeSafe`

sdk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
kattrali: Resolved through #585

sdk/src/main/java/com/bugsnag/android/NativeInterface.java
kattrali: Yeah, that's a good call. There are only a few which are called from the NDK interface, and those are:

* `deliverReport`
* `leaveBreadcrumb`
* `setUser`
* `notify`

I've added additional APIs to cover the others so they can be called with byte arrays. Going forward, we should have a linting rule to disallow calling string methods from the native interface without an annotation (something I can look into adding later).

fractalwrench: Sounds like a good approach.

fractalwrench:
There are also some minor checkstyle violations that will break CI, which can be determined by running `./gradlew checkstyle lint detekt`.

sdk/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -271,6 +272,24 @@ public static void setUser(@Nullable final String id,
...
+     * @param idBytes id
+     * @param emailBytes email
+     * @param nameBytes name
+     */
+    @SuppressWarnings("unused")
+    public static void setUser(@Nullable final byte[] idBytes,
+                               @Nullable final byte[] emailBytes,
+                               @Nullable final byte[] nameBytes) {
+        // The default charset on Android is always UTF-8
+        String id = new String(idBytes, Charset.defaultCharset());
```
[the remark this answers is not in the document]
kattrali: Somewhat bizarrely, there's no documentation for this case. 🤨 I'll put in a check for it.

PR #588, opened 2019-09-09
Record storagemanager cache behaviour in internal error reports

fractalwrench:
## Goal

Android API 26 introduced `setCacheBehaviorGroup`, which tells the OS to delete all the files in a given directory if space needs reclaiming. Coupled with `setCacheBehaviorTombstone` which truncates files to 0, this could explain the volume of minimal error reports reported by some customers.

We should record this information in our internal error diagnostics to debug the issue further.

## Tests

Tested against a local artefact on API 23 + 29 which had been altered to always call `recordStorageCacheBehavior` when notifying.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -12,6 +12,8 @@ import android.content.SharedPreferences;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
 import android.content.res.Resources;
+import android.os.Build;
+import android.os.storage.StorageManager;
 import android.text.TextUtils;
 import android.view.OrientationEventListener;
 import androidx.annotation.NonNull;
@@ -21,6 +23,7 @@ import kotlin.Unit;
 import kotlin.jvm.functions.Function1;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -78,6 +81,7 @@ public class Client extends Observable implements Observer {
 
     private final OrientationEventListener orientationListener;
     private final Connectivity connectivity;
+    final StorageManager storageManager;
 
     /**
      * Initialize a Bugsnag client
@@ -123,6 +127,7 @@ public class Client extends Observable implements Observer {
         appContext = androidContext.getApplicationContext();
         config = configuration;
         sessionStore = new SessionStore(config, appContext);
+        storageManager = (StorageManager) appContext.getSystemService(Context.STORAGE_SERVICE);
 
         connectivity = new ConnectivityCompat(appContext, new Function1<Boolean, Unit>() {
             @Override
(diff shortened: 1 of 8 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -259,6 +265,21 @@ public void onOrientationChanged(int orientation) {
         loadPlugins();
     }
 
+    void recordStorageCacheBehavior(MetaData metaData) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            File cacheDir = appContext.getCacheDir();
+
+            try {
+                boolean tombstone = storageManager.isCacheBehaviorTombstone(cacheDir);
```
kattrali: Can this in any way be deceptively false, e.g. if `storageManager.isCacheBehaviorTombstone(cacheDir) == false` but `storageManager.isCacheBehaviorTombstone(cacheDir + "/bugsnag-errors") == true`?

fractalwrench: That's a possibility - although it would require for a user to have specifically called `storageManager.setCacheBehaviorTombstone` on the `bugsnag-errors` directory.

kattrali: Is there anything preventing checking the bugsnag-errors directory itself to avoid misleading values?

fractalwrench: not that I'm aware of - I plan on updating the PR to do this.

fractalwrench:
I've updated the PR to get the tombstone behaviour directly from the `bugsnag-errors` directory rather than `context.cacheDir`.

fractalwrench:
I've updated the mazerunner scenario for an internal error report to check the storagemanager cache values are set, and to more generally validate other fields that we add to `BugsnagDiagnostics`.

tests/features/internal_report.feature
```diff
@@ -11,3 +13,11 @@ Scenario: Sending internal error reports
     And the event "breadcrumbs" is null
     And the event "app.type" equals "android"
     And the event "device.osName" equals "android"
+    And the event "metaData.BugsnagDiagnostics.cacheTombstone" is false
```
kattrali: There's no test to assert that this is true when cache behavior is set, or that its null when using a lower Android version.

tests/features/internal_report.feature
```diff
@@ -11,3 +13,11 @@ Scenario: Sending internal error reports
...
+    And the event "metaData.BugsnagDiagnostics.filename" is not null
+    And the event "metaData.BugsnagDiagnostics.notifierName" is not null
```
kattrali: The `notifierName` is known and can be checked for equality.

fractalwrench:
I have updated the internal_report feature to run 3 scenarios:

- One for verifying reports are sent on API <26
- One for verifying cache tombstone/group behaviour is false by default
- One for verifying cache tombstone/group behaviour is true when enabled

I have also updated the assertion steps as requested.

PR #590, opened 2019-09-10
Buildkite tests: Enable ANRs

Cawllec:
## Goal

Enables ANR testing against real devices on BrowserStack.

There's potentially optimisation here by reducing the amount of times the screen is tapped in order to produce the correct response.


Cawllec:
Reduced tap count from 20 to 3, still seems to work correctly locally.

PR #591, opened 2019-09-10
Delete empty files left in cache directory

fractalwrench:
## Goal

API 26 added a cacheBehaviorTombstone API, which leaves behind a zero-length file rather than
deleting the file entirely. We should delete any empty files when attempting to send cached
errors/sessions.

## Changeset

Check for the presence of an empty file and delete it within `findStoredFiles()`.

## Tests

Manually verified in a local artefact that if an empty file is added to the `bugsnag-errors` directory it is discarded.


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Delete tombstoned files left in cache directory
```
kattrali: This entry should explain what this change means for crash reporting and why it might happen/be necessary to do this.

bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -137,6 +145,26 @@ void discardOldestFileIfNeeded() {
...
+     * zero-length file is left behind rather than deleting the file from the cache dir entirely.
+     *
+     * See https://developer.android.com/reference/android/os/storage/StorageManager.html
+     * @param file the cached file
+     * @return true if the file is a tombstone
+     */
+    private boolean isTombstone(File file) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            try {
+                return file.length() == 0 && storageManager.isCacheBehaviorTombstone(cacheDir);
```
kattrali: Would this value be true if `cacheDir` + `/bugsnag-errors` is using tombstone behavior but `cacheDir` is not? Can `path` be checked directly?

fractalwrench: Updated to check the path directly.

fractalwrench:
I've now added a mazerunner scenario to verify that cached tombstone files are deleted.

tests/features/cache_tombstone.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Tombstoned cache files are not reported
+
+@skip_below_android_9
+Scenario: A zero-length file is not reported when cache tombstone behavior is set
```
tomlongridge: Now that we're not checking tombstone behaviour and just removing any zero-length file, this scenario and associated test classes should be renamed to remove references to tombstone behaviour.

fractalwrench: I've renamed the test scenarios appropriately and have removed code which set the cache behaviour using the `StorageManager`.

PR #592, opened 2019-09-10
Query available cache size before writing report to disk

fractalwrench:
## Goal

Currently `FileStore` writes an arbitrary 128 files to disk, which was a decision made in #97. When additional metadata is added to a report and the system is low on space, this may have the effect of making it more likely for the OS to reclaim space from our application. 

## Changeset

The changeset makes several alterations which will reduce the likelihood of our cached files being deleted by the OS:

- Reduced max file count from 128 to 64. A typical error reports weigh around ~12Kb, which should fit into <1Mb storage if no additional metadata etc is added to a report
- Cached `context.cacheDir` and the `bugsnag-errors` directory File as fields, reducing duplication
- Moved `discardOldFilesIfNeeded()` into lock, reducing the potential for a `Report` serializing a cached `Error` to suffer a `FileNotFoundException`
- Called [`getUsableSpace()`](https://developer.android.com/reference/java/io/File#getUsableSpace()) to determine whether < 10Mb is available for use in the cache directory, and if so, delete old files.

## Discussion

`getCacheQuotaBytes()` and `getCacheSizeBytes()` are available on API 26+ which would allow us to determine whether we are over the cache quota more precisely. However, the documentation warns that these calls can take several seconds to complete, and would not resolve any caching issues on older devices, leading to the approach in this PR.

See: https://developer.android.com/reference/android/os/storage/StorageManager

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
@@ -60,7 +60,7 @@ class ErrorStore extends FileStore<Error> {
     };
 
     ErrorStore(@NonNull Configuration config, @NonNull Context appContext, Delegate delegate) {
-        super(config, appContext, "/bugsnag-errors/", 128, ERROR_REPORT_COMPARATOR);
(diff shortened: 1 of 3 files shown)


fractalwrench:
Closing for now, will reopen when we're confident about introducing this change.

PR #594, opened 2019-09-12
Prevent internal error reporting of FileNotFoundException during Delivery

fractalwrench:
## Goal

`FileNotFoundException` can occur due to a race where the oldest file is deleted from disk, and the
`Error` object sends a cached file directly in the request. If this does occur, we should suppress
internal error reports as it is a known issue that we cannot handle.

## Design

It is worth noting that redesigning the `ErrorStore` so that the `File` is always deserialized into an `Error` object would prevent this scenario from occurring.

However, given the complexity of these changes and incidence of this situation, it was seen as more appropriate to avoid sending an internal error report instead.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import androidx.annotation.NonNull;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorStore.java
```diff
@@ -174,7 +175,7 @@ private void flushErrorReport(File errorFile) {
             Logger.warn("Could not send previously saved error(s)"
                 + " to Bugsnag, will try again later", exception);
         } catch (Exception exception) {
-            if (delegate != null) {
+            if (delegate != null && !(exception instanceof FileNotFoundException)) {
```
tomlongridge: Rather than doing `instanceof`, how about a separate `catch` block for `FileNotFoundException` that only warns? There doesn't seem any need to attempt a delete in this case either.

tomlongridge: Warning may also act as a useful reminder to our future selves on why we're allowing FNFE's to go unreported.

fractalwrench:
Thanks for taking a look @tomlongridge - I've logged a warning in a separate catch block as requested.

PR #598, opened 2019-09-23
Deserialize longs/doubles stored in StackFrame object

fractalwrench:
## Goal

Some fields in the `StackFrame` object can be non-integers in some cases, but the current implementation assumes that they are always integers. This change uses existing code to deserialize numbers instead, which handles doubles/longs appropriately.

## Tests

Manually verified that it an error is read from a JSON file if the stackframe objects contain doubles/longs/ints.

### DIFF
--- a/bugsnag-android-core/src/androidTest/resources/stackframe_numbers.json
+++ b/bugsnag-android-core/src/androidTest/resources/stackframe_numbers.json
@@ -132,19 +132,19 @@
   "groupingHash": "400-b",
   "threads": [
     {
-      "id": 1,
+      "id": 11236722452451234,
       "name": "main",
       "type": "android",
       "stacktrace": [
         {
           "method": "dalvik.system.VMStack.getThreadStackTrace",
           "file": "VMStack.java",
-          "lineNumber": -2
+          "lineNumber": 160923409125093
         },
         {
           "method": "java.lang.Thread.getStackTrace",
           "file": "Thread.java",
-          "lineNumber": 1566
+          "lineNumber": 1566.5
         },
         {
           "method": "com.bugsnag.android.ThreadState.<init>",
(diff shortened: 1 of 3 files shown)


tomlongridge:
Why was the testing manual only? We should add some automated tests in this area - it seems like an area that is testable.

fractalwrench:
I've added an additional unit test that verifies longs/doubles are serialized successfully in stackframes from the ErrorReader class.

bugsnag-android-core/src/androidTest/resources/stackframe_numbers.json
```diff
@@ -0,0 +1,442 @@
...
+  "threads": [
+    {
+      "id": 1,
+      "name": "main",
+      "type": "android",
+      "stacktrace": [
+        {
+          "method": "dalvik.system.VMStack.getThreadStackTrace",
+          "file": "VMStack.java",
+          "lineNumber": -2
```
tomlongridge: Even though this is deserialised using the same method, it would be worth putting some assertions in for thread stacktrace line numbers.

bugsnag-android-core/src/androidTest/resources/stackframe_numbers.json
```diff
@@ -0,0 +1,442 @@
...
+  },
+  "user": {
+    "id": "9659cc1f-fec8-47a3-8ad2-0e3c7e84c24d",
+    "name": "Rasputin",
+    "email": "ras@example.com"
+  },
+  "groupingHash": "400-b",
+  "threads": [
+    {
+      "id": 1,
```
tomlongridge: Should we apply the same deserialisation logic to thread ID's?

fractalwrench:
I have added some additional assertions to the unit test to check a double/long deserialize into the lineNumber field, and that a long thread ID is also deserialized.

PR #599, opened 2019-09-23
Prevent SIGSEGV when passing NULL to bugsnag_set_user_env

fractalwrench:
## Goal

If `NULL` is passed as a parameter to `bugsnag_set_user_env` the process will crash with a SIGSEGV, because the method does not check for null.

This changeset adds a check for null in `bsg_byte_ary_from_string` and `bsg_release_byte_ary`. 

## Tests

The `CxxUserInfoScenario` in `native_api.feature` is now passing, whereas previously it was not. Also verified by passing NULL to all params in an example app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+* Prevent SIGSEGV when passing NULL to bugsnag_set_user_env
+  [#599](https://github.com/bugsnag/bugsnag-android/pull/599)
+
 * Alter value collected for device.freeDisk to collect usable space in internal storage,
  rather than total space in internal/external storage
   [#589](https://github.com/bugsnag/bugsnag-android/pull/589)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+* Prevent SIGSEGV when passing NULL to bugsnag_set_user_env
```
kattrali: This one actually doesn't need a changelog entry since the bug hasn't been released

fractalwrench: The mazerunner scenario fails with a SIGSEGV on master too as there isn't a null check there either: https://github.com/bugsnag/bugsnag-android/blob/master/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c#L140

fractalwrench:
I have now removed the changelog entry.

PR #606, opened 2019-10-01
Add parameterised JSON serialisation tests for JVM classes

fractalwrench:
## Goal

This changeset adds parameterised JSON serialization tests for all classes involved in JSON serialization. This will make it easier to expand our test coverage of serialization as all that will be required to write a new test case is the expected JSON as a test fixture file, and a properly configured object.

## Changeset

Supporting test code to serialize a `Streamable` object as JSON and verify it matches the contents of a file in the `resources` directory can be found in 77ed6ff0fc50d3f642ec1336cd620d2396ae68ab. `GSON` is added as a `testImplementation` dependency so that the tests do not need to use the `JsonWriter` supplied by the Android framework, and can therefore be executed as pure JVM tests.

A serialization test has been added for the following classes in individual commits. Where a single serialisation test already existed in `src/androidTest`, that test was converted to run in `src/test`.

```
Breadcrumb
Breadcrumbs
BugsnagException
CachedThread
Error
Exceptions
HandledState
MetaData
Notifier
Report
Session
SessionTrackingPayload
Severity
Stacktrace
ThreadState
User
AppData
DeviceData
```

Additionally, `SessionTrackingPayload` now injects a map and `Notifier` instance rather than `AppData/DeviceData`, to improve its testability.

`DeviceBuildInfo` has been created and injected into the constructor parameter of `DeviceData`. It encapsulates the values of static fields of `android.os.Build.*`, which allows serialization of these fields to be properly unit tested with mock values.


bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -164,6 +164,9 @@ private boolean isRooted() {
      */
     private boolean isEmulator() {
         String fingerprint = Build.FINGERPRINT;
+        if (fingerprint == null) {
```
Cawllec: Is this a case that should be added here, or in a separate PR?

bugsnag-android-core/build.gradle
```diff
@@ -16,3 +16,9 @@ apply from: "../gradle/release.gradle"
 apply from: "../gradle/detekt.gradle"
 apply from: "../gradle/checkstyle.gradle"
 apply from: "../gradle/license-check.gradle"
+
+dependencies {
+    // required to can run JSON serialization tests on the JVM, which is much quicker than
+    // relying on the version supplied by the Android Framework
+    testImplementation "com.google.code.gson:gson:2.8.5"
```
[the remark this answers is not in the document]
fractalwrench: Yes, this is the case. This has actually been addressed in #607, which vendors the `JsonReader` class to ensure equivalent behaviour and avoids the need for adding this `testImplementation` dependency.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
fractalwrench: `android.os.Build` is an Android framework class, which means the return value of `Build.FINGERPRINT` will be null when run in a JVM unit test. Checking for null passed the test without making larger changes to the existing code.

Another option would be to collect all the existing fields in this class such as `Build` and to pass them in within a wrapper class that can be populated in a JVM test. For example, something like:

```
class BuildInfo {
    val fingerprint: String
    val manufacturer: String
}
```

I can make this change if it'd be preferable

fractalwrench:
Thanks for the comments so far. There is a bit of coupling between the JVM serialization/deserialization PRs as they use the same code for reading JSON test fixtures etc, but no coupling between the NDK serialization tests and this.

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTrackingPayload.java
```diff
@@ -17,11 +17,12 @@
 
     SessionTrackingPayload(Session session,
                            List<File> files,
-                           AppData appData,
-                           DeviceData deviceData) {
-        this.appDataSummary = appData.getAppDataSummary();
-        this.deviceDataSummary = deviceData.getDeviceDataSummary();
-        this.notifier = Notifier.getInstance();
+                           Map<String, Object> appDataSummary,
```
Cawllec: While I understand these are fairly minor changes (i.e. just changing where the Object -> Map conversion is), could someone else (@kattrali ?) review any changes to the core SDK?

Are we certain there are no places that call this function that could have been missed?

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbsSerializationTest.kt
```diff
@@ -34,22 +29,6 @@ class BreadcrumbsSerializationTest {
...
-     * Verifies that breadcrumb metadata is serialised
-     */
-    @Test
-    fun testPayloadType() {
-        val metadata = mapOf(Pair("direction", "left"))
-        breadcrumbs.add(Breadcrumb("Rotated Menu", BreadcrumbType.STATE, metadata))
-
-        val json = streamableToJsonArray(breadcrumbs)
-        val node = json.getJSONObject(0)
-        assertEquals("Rotated Menu", node.get("name"))
```
Cawllec: Is this breadcrumb metadata now tested elsewhere?

bugsnag-android-core/src/test/resources/error_serialization_0.json
```diff
@@ -0,0 +1,161 @@
...
+    "type": "unhandledException"
+  },
+  "unhandled": true,
+  "incomplete": false,
+  "exceptions": [
+    {
+      "errorClass": "java.lang.RuntimeException",
+      "message": "whoops",
+      "type": "android",
+      "stacktrace": [
```
Cawllec: Is this stacktrace guaranteed?  Is there a possibility this could change on different Java versions?

fractalwrench: For this particular test case the stacktrace should be guaranteed, as the object has constructed its own `StacktraceElement` rather than asking the JVM to generate one via `getStacktrace()`.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbsSerializationTest.kt
fractalwrench: Updated the test case so that metadata is now included in the expected JSON.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
fractalwrench: I have removed the null check for `fingerprint` and have instead created a new `DeviceBuildInfo` class which encapsulates all the static fields accessed from `android.os.Build`. This class is injected as a constructor parameter into `DeviceData` which allows for it to be swapped out with suitable values in unit tests.

The serialization tests have been updated as more fields are now serialized, now that the `android.os.Build.*` fields do not return null.

tomlongridge:
Can you mention and justify the new DeviceBuildInfo class in your PR description?

fractalwrench:
Context on DeviceBuildInfo was here: https://github.com/bugsnag/bugsnag-android/pull/606#discussion_r331996079

I will amend the PR description to make this more prominent

fractalwrench:
Closing in favour of #616

PR #607, opened 2019-10-02
Add parameterised JSON deserialisation tests for JVM classes

fractalwrench:
## Goal

This changeset adds parameterised JSON deserialization tests for all classes involved in JSON deserialization that were practical to test with minor alterations. This will make it easier to expand our test coverage of deserialization as all that will be required to write a new test case is the expected JSON as a test fixture file, and a properly configured object.

Also see #606

## Changeset

A deserialization test has been added for the following classes in individual commits:

```
Breadcrumb
Breadcrumbs
BugsnagException
CachedThread
Error
Exceptions
MetaData
Severity
Stacktrace
ThreadState
User
AppData
DeviceData
```

The `JsonReader` class has been vendored from GSON so that we can use a consistent class for deserialization, rather than relying on the Android framework. This has the added bonus of allowing us to run tests on the JVM rather than relying on an Android device. `JsonWriterTest` and `JsonReaderTest` have been added to the `src/test` directory accordingly: 52c55768407cc816b857ae0160f8652d4390d2c5

The `ErrorReader` deserialization implementation has been altered so that individual methods are package-private and annotated as `@VisibleForTesting`. Where an individual method did not exist for deserialization, a method has been extracted from existing code where practical.

Package-private accessors have been added to `CachedThread` and `ThreadState` to facilitate testing.


bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorReader.java
```diff
@@ -496,13 +533,14 @@ private static CachedThread readThread(JsonReader reader) throws IOException {
     }
 
     @SuppressWarnings("unchecked")
-    private static <T> T deserializeNumber(JsonReader reader) throws IOException {
+    @VisibleForTesting
+    static <T> T deserializeNumber(JsonReader reader) throws IOException {
         try {
             return (T)(Integer) reader.nextInt();
-        } catch (NumberFormatException ex) {
+        } catch (NumberFormatException | IllegalStateException ex) {
```
fractalwrench: The `JsonReader` GSON implementation can throw an `IllegalStateException` also

kattrali: > Throws IllegalStateException if the next token is not a literal value.

If its not a literal value, none of these options will be valid. `IllegalStateException` should not be in this `catch` clause.

fractalwrench: This change is required to pass existing tests, specifically `RxErrorTest` which fails to serialize `metaData.device.screenDensity`: https://github.com/bugsnag/bugsnag-android/blob/23f2b4d25a997b5250dd6b371792cef9e8eeb19e/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/RxErrorTest.kt#L49, and `ErrorReaderNumberTest` which fails to read `lineNumber` for stackframe. This version of GSON appears to throw an `IllegalStateException` if a double is supplied when `nextLong()` is called.

bugsnag-android-core/src/test/java/com/bugsnag/android/JsonUtils.kt
```diff
@@ -21,7 +21,8 @@ internal fun verifyJsonMatches(map: Map<String, Any>, resourceName: String) {
 
 private fun validateJson(resourceName: String, json: String) {
     val whitespace = "\\s".toRegex()
-    val expectedJson = JsonParser().read(resourceName).replace(whitespace, "")
+    val rawJson = JsonParser().read(resourceName)
```
Cawllec: What's the reason for this?  Does it allow more access somewhere down the line?

bugsnag-android-core/src/main/java/com/bugsnag/android/CachedThread.java
```diff
@@ -49,4 +49,24 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
         }
         writer.endObject();
     }
+
+    long getId() {
```
Cawllec: Are the CachedThread objects accessible in non-Bugsnag code?

bugsnag-android-core/src/test/java/com/bugsnag/android/CachedThreadDeserializationTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    @Parameter
+    lateinit var testCase: Pair<CachedThread, String>
+
+    @Test
+    fun testJsonDeserialisation() {
+        val reader = JsonParser().parse(testCase.second)
+        val cachedThread = ErrorReader.readThread(reader)
+
+        val expected = testCase.first
+        assertEquals(expected.id, cachedThread.id)
```
Cawllec: We don't seem to be using the accessor methods here, are they required for these tests?

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.java
```diff
@@ -71,4 +71,8 @@ public void toStream(@NonNull JsonStream writer) throws IOException {
         }
         writer.endArray();
     }
+
+    CachedThread[] getCachedThreads() {
```
Cawllec: This is added in this commit but doesn't seem to be used here.

gradle/checkstyle.gradle
```diff
@@ -8,6 +8,6 @@ task("checkstyle", type: Checkstyle) {
     configFile rootProject.file("config/checkstyle/checkstyle.xml")
     source  "src/"
     include "**/*.java"
-    exclude "**/external/**/*.java"
+    exclude "**/external/**/*.java", "**/JsonReaderTest.java"
```
Cawllec: Is this excluding the test added from the GSON codebase?

fractalwrench: Yep. We've excluded that test from checkstyle as GSON uses a different codestyle to us, so the checkstyle would fail the build otherwise.

bugsnag-android-core/src/test/java/com/bugsnag/android/JsonUtils.kt
fractalwrench: This was only done for readability to avoid chaining the `replace()` method call - if it's not clear it can be changed back.

bugsnag-android-core/src/main/java/com/bugsnag/android/CachedThread.java
fractalwrench: No - the class is package-private so should not be accessible to end-users. The only case where the object could be accessed is if someone writes a class in their application in the `com.bugsnag.android` package, or if they use Java's reflection API.

bugsnag-android-core/src/test/java/com/bugsnag/android/CachedThreadDeserializationTest.kt
fractalwrench: This is a feature of Kotlin's interoperability - the Java getter is represented as a Kotlin property, which prevents us from having to write out the full `getId()` https://kotlinlang.org/docs/reference/java-interop.html#getters-and-setters

fractalwrench:
thanks for taking a look 👍

PR #608, opened 2019-10-03
Add parameterised JSON serialisation tests for NDK serializer

fractalwrench:
## Goal

This changeset adds parameterised JSON serialization tests for the NDK. This will make it easier to expand our test coverage of serialization as all that will be required to write a new test case is the expected JSON as a test fixture file, and a properly configured struct.

Also see #606 and #607

## Changeset

A parameterised serialization test has been added for the following objects in individual commits:

```
App
App metadata
Device
Device metadata
MetaData
Context
HandledState
Session
Breadcrumbs
Stackframe
Exception
```

Each test consists of:

1. A parameterised JUnit test which reads a JSON test fixture as a string and passes it via the JNI to a C test
2. A C test which loads a struct test case. The parameter index of the test case is passed to this method so it can conditionally load different test cases

The `bsg_serialize_report_to_file` method has been extracted into separate methods which are exposed in the header, to facilitate testing.


fractalwrench:
@Cawllec each method has had a test added in an isolated commit so that should be a nicer way to review than looking at one diff - there's also a separate one for reorganising the native serializer code. Is there any other approach that would help with reviewing this?

PR #609, opened 2019-10-04
Add E2E/instrumentation tests for JSON serialization/deserialization on Android

fractalwrench:
## Goal

Improves E2E and instrumentation test coverage for internal report behaviour.

## Changeset

This changeset deletes a file is an exception is thrown when writing it to disk. This is intended to prevent empty/partial files being unintentionally left on disk for delivery next launch.

Added/moved existing mazerunner scenarios so that tests cover:

- If an empty file is in the cache directory then zero requests should be made
- If a partial file is in the cache directory then an internal report should be sent
- If a file is deleted before it can be sent, 0 requests should be sent

Added instrumentations tests that cover:

- If an exception is thrown when writing a Streamable, an internal error report is sent (measured by invoking a delegate) and the partial file is deleted
- If an exception is thrown when copying a NDK report, an internal error report is sent (measured by invoking a delegate) and the partial file is deleted


bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -83,8 +83,18 @@ void enqueueContentForDelivery(String content) {
...
-                delegate.onErrorIOFailure(exc, new File(filename), "NDK Crash report copy");
+                delegate.onErrorIOFailure(exc, errorFile, "NDK Crash report copy");
+            }
+
+            try {
+                if (!errorFile.delete()) {
+                    errorFile.deleteOnExit();
+                }
+            } catch (Exception ex) {
+                Logger.warn("Failed to delete partially written file", ex);
```
tomlongridge: This functional change isn't mentioned in the PR description nor a changelog entry.

fractalwrench: This was mentioned at the start of the changeset in the PR description - I'll add a changelog entry and rethink how the change has been worded.

tomlongridge: FWIW, can you rename the PR and add the behaviour change to the Goal as it's easy to miss?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
```
tomlongridge: Just to check - this will be a bug fix, so patch release rather than minor release?

bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -83,8 +83,18 @@ void enqueueContentForDelivery(String content) {
...
             }
```
tomlongridge: Is it worth considering adding this to `IOUtils`?

CHANGELOG.md
fractalwrench: Yep

PR #610, opened 2019-10-04
Allow overriding the versionCode via Configuration

fractalwrench:
#  Goal

Allows overriding the `versionCode` either through a manifest meta-data element or programmatically on the `Configuration` object. This facilitates grouping for custom versioning schemes. By default, the `versionCode` still uses the `android:versionCode` manifest value once a `Client` has been instantiated.

## Changeset

Added integer `versionCode` field to `Configuration` which can also be read via the AndroidManifest `meta-data` element `com.bugsnag.android.VERSION_CODE`. By default if this value is not configured, the `Client` will set the `versionCode` to the value supplied in `android:versionCode` of the AndroidManifest.

## Tests

Ran the example app and confirmed that for a handled, unhandled, and NDK error, the versionCode reported in the dashboard differs for:

- the default `android:versionCode` value set in the manifest
- the `com.bugsnag.android.VERSION_CODE` value set in the manifest
- the value set on `config.versionCode`

Updated existing integration tests to verify that the value is loaded from the manifest and can be overridden.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/AppData.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/AppData.java
@@ -66,7 +66,7 @@ class AppData {
         map.put("type", calculateNotifierType());
         map.put("releaseStage", guessReleaseStage());
         map.put("version", calculateVersionName());
-        map.put("versionCode", calculateVersionCode());
+        map.put("versionCode", config.getVersionCode().intValue());
         map.put("codeBundleId", config.getCodeBundleId());
         return map;
     }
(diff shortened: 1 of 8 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/AppData.java
```diff
@@ -131,20 +134,6 @@ private String calculateNotifierType() {
...
-     */
-    @Nullable
-    @SuppressWarnings("deprecation")
-    private Integer calculateVersionCode() {
-        if (packageInfo != null) {
-            return packageInfo.versionCode;
-        } else {
-            return null;
-        }
-    }
```
tomlongridge: As I understand it, versionCode should behave exactly like appVersion: it can be included in the manifest under Bugsnag metadata, supplied programatically on `Configuration`, or otherwise it's read from the manifest under a common Android location. If so, the two fields should have the same logic in `Client` and `AppData`.

fractalwrench: Agreed to alter the implementation to follow the behaviour of appVersion for now. A ticket has been created to alter the implementation more generally as part of the immutable config work, to make Configuration the source of truth.

PR #612, opened 2019-10-10
Collect additional data in internal error reports

fractalwrench:
## Changeset

Collects the following values under the app/device tabs, using existing collection methods from `AppData` and `DeviceData`:

```
app.duration
app.durationInForeground
app.inForeground
device.freeDisk
```

## Tests

Added to existing mazerunner scenarios to verify that internal reports contain the new fields

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -215,9 +215,6 @@ public class Client extends Observable implements Observer {
                 metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "canWrite", errorFile.canWrite());
                 metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "exists", errorFile.exists());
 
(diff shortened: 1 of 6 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -215,9 +215,6 @@ public void onErrorIOFailure(Exception exc, File errorFile, String context) {
                 metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "canWrite", errorFile.canWrite());
                 metaData.addToTab(INTERNAL_DIAGNOSTICS_TAB, "exists", errorFile.exists());
 
-                @SuppressLint("UsableSpace") // storagemanager alternative API requires API 26
```
tomlongridge: Do we want to continue to store the cache's usable space in addition on our diagnostics tab?

fractalwrench: You're right, we do. I missed the `getCacheDir()` aspect of this and thought it was equivalent to `DeviceData#calculateFreeDisk()` - I've reinstated this.

PR #615, opened 2019-10-15
Remove unsupported armeabi from abiFilters

friederbluemle:
Trying to build or open the project in Android Studio is failing with:

```
ABIs [armeabi] are not supported for platform. Supported ABIs are [arm64-v8a, armeabi-v7a, x86, x86_64].
```

The reason is `armeabi` was removed from the NDK. See [NDK changelog](https://github.com/android/ndk/wiki/Changelog-r17)

> Support for ARMv5 (armeabi), MIPS, and MIPS64 has been removed. Attempting to build any of these ABIs will result in an error.

Also see:

- https://github.com/android/ndk/issues/961
- https://github.com/android/ndk/issues/911
- https://stackoverflow.com/questions/52233971/abis-armeabi-mips-are-not-supported-for-platform-android-ndk

There have been no new devices that only support `armeabi` but not `armeabi-v7a` **for 10 years or so**.

This PR removes the references to `armeabi` from the default list of `abiFilters`, allowing the build to pass again.

@kattrali - You already removed it from example app as part of #493, but in a0519a4cae3624a1b1c4ec05cd9659ca1f0a4a7d you're mentioning "Needed for lib release" - Can you elaborate on it?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,6 +98,15 @@ option:
 ./gradlew assembleRelease -PABI_FILTERS=x86,arm64-v8a
 ```
 
+To build for ARMv5 (`armeabi`):
+
+```shell
+./gradlew assembleRelease -PABI_FILTERS=armeabi
+```
+
+Note that `armeabi` has been removed in NDK r17 and requires the installation
+of an older NDK version.
+
 ## Testing
 
 Full details of how to build and run tests can be found in [the testing guide](`TESTING.md`)
--- a/bugsnag-plugin-android-anr/build.gradle
+++ b/bugsnag-plugin-android-anr/build.gradle
@@ -12,7 +12,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         externalNativeBuild.cmake.arguments "-DANDROID_CPP_FEATURES=exceptions", "-DANDROID_STL=c++_static"
         ndk.abiFilters = project.hasProperty("ABI_FILTERS") ? project.ABI_FILTERS.split(",") :
-            ["arm64-v8a", "armeabi-v7a", "armeabi", "x86", "x86_64"]
+            ["arm64-v8a", "armeabi-v7a", "x86", "x86_64"]
     }
     externalNativeBuild.cmake.path = "CMakeLists.txt"
 }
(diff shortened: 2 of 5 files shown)


kattrali:
> You already removed it from example app as part of #493, but in a0519a4 you're mentioning "Needed for lib release" - Can you elaborate on it?

Sure, @friederbluemle. Not everyone using bugsnag-android is using the latest and greatest, and some teams are still using NDK r10 and targeting armeabi devices - mostly TV set top boxes or similar. However, this is something we have to regularly re-evaluate to make sure the efforts we put into supporting old versions is well-spent. In the meantime, there's probably an easy workaround since you don't need those ABIs - we'll take a look to figure out what that should be.

friederbluemle:
Thanks for explaining @kattrali 👌 What you said makes sense.
I put the test cases back in this PR, and slightly amended the wording. To be clear, this PR does _not_ remove the ability to build for the `armeabi` ABI. It only updates the _default_ array of ABIs used for the standard `./gradlew build` task.

By using the default set of ABIs, we:

- Allow out-of-the-box successful builds of the project for most people
- Can avoid two `sed` in-place replacement hacks from the Dockerfiles which _already_ removed the API (and also left the working directory dirty)

In addition, if some teams _still_ need to build for removed/unsupported ABIs, they can easily do it using

```sh
./gradlew assembleRelease -PABI_FILTERS=armeabi
```

I added documentation next to the existing one in `CONTRIBUTING.md`, I could also add a note to `README.md`.

What do you think?

mattdyoung:
Hi @friederbluemle 

We've discussed this amongst the team and agree it would make sense so we'll look at reviewing / merging this when we have the bandwidth.

Thanks!

friederbluemle:
Thanks @mattdyoung - Please let me know if you want me to update anything else. 👌

friederbluemle:
Rebased onto latest default branch.

friederbluemle:
Rebased onto latest default branch.

friederbluemle:
Rebased onto latest default branch.

friederbluemle:
As of Feb 2020, the build on the `master` branch is still failing (without making manual changes first).

@mattdyoung - Just checking, any rough idea when this can be merged? Thanks!

mattdyoung:
@friederbluemle - we're just working on bringing together a few higher priority issues for a major release of the library and hope to pick this up soon after.

tomlongridge:
Work complete in #825

friederbluemle:
Thanks @tomlongridge!

PR #616, opened 2019-10-16
Update bugsnag-android to conform to notifier specification

fractalwrench:
# Summary

This changeset updates bugsnag-android's public JVM and NDK API to conform with the notifier specification.

Several of these changes break the previous interface and require a major version bump. Several of these changes are not backwards compatible, which was also agreed during development.

## Reviewing

By necessity this is a very large PR, so the changeset is best reviewed by checking out the branch and comparing it against the Notifier Spec, along with manual testing/upgrade testing/code review of critical sections.

# Code changes

This is not an exhaustive list of changes.

## Notable Interface changes

- `Report` -> `Event`
- `BeforeSend/BeforeNotify` -> `OnError`
- `Event` interface reworked to allow greater mutation of payload data
- `Metadata` now supports adding/getting/clearing, is no longer a public class
- Added `OnSession` and exposed `Session/SessionPayload`
- Removed various classes/methods which do not conform to the notifier interface, e.g. `notify()` variants which take `Severity` as a parameter
- `Bugsnag` now provides a facade of all `Client` methods
- Allowed loading configuration from a manifest

## Notable Internal changes

- Cached JSON files are no longer deserialized. This was error prone and is no longer necessary, because an `OnError` callback can now be invoked at the time of the crash in both the JVM + NDK.
- Classes have been converted to Kotlin where non-trivial changes were required to update the interface/behaviour
- The responsibilities of the `Client` class have been reduced and it now invokes methods on delegates which perform the work of notifying, updating metadata, etc
- `Configuration` is deep-copied to `ImmutableConfiguration` within the `Client` constructor
- `CallbackState`, `UserState`, `ContextState`,  and `MetadataState` encapsulate state that can be mutated via the `Client` and `Configuration`, and handle notifying any observers
- Added `UserAware` and `CallbackAware` internal interfaces, which keep the method signature the same between `Client/Configuration`

## NDK Changes

- Removed/made non-public unused code where possible
- Fixed setting the `handledState.severity` on an event directly, which previously gave the wrong severityReason
- Emitted Observable events from a `BaseObservable` that provides a more succinct `notifyObservers(type, args)` method
- `MessageType` and `NativeInterface.Message` have been replaced with a [sealed class hierarchy](https://kotlinlang.org/docs/reference/sealed-classes.html), which guarantees compile-time safety of observable messages and has greatly simplified the validation in the message receiver.
- Removed/renamed observable messages where appropriate
- Moved `appVersion`, `buildUuid`, and `releaseStage` into the initial install message as they are now immutable
- Implemented an `OnError` function that can be invoked in the crashing context, and that has a `bugsnag_event` param and returns `bool`
- Altered `bugsnag_report` to `bugsnag_event` and renamed fields where appropriate. Note: this does not 100% match the notifier spec, but future changes to remedy this can be additive and non-breaking.
- Added migration for deserialization of persisted values of the old `bugsnag_report` struct

## Tests

- Mazerunner scenarios have been updated to use the new interfaces where possible. New scenarios have not been added
- Instrumentation tests were converted to JVM tests when there were non-trivial interface changes
- The majority of tests are now JVM tests, which reduces the feedback loop as they execute in <5s and don't require an emulator/device
- JVM test coverage has increased from 22% to 68%
- The JVM serialization tests from #606 have been ported into this branch to verify that the new interfaces serialize JSON correctly

# Supporting material

## Example app

The example app is now a fixture that is an independent gradle project. It should therefore be opened separately in Android Studio, and gradlew commands should be changed accordingly.

## Docs

Docs will be written in a separate PR once the changeset is finalised.

## Upgrade guide

The upgrade guide will be written in a separate PR once the changeset is finalised.

## Proguard

We should update our Proguard rules in the docs + plugin when the changeset is finalised.


fractalwrench:
Closing as stale in favour of #643

PR #617, opened 2019-10-16
Rename pre-existing fields on Configuration

fractalwrench:
## Goal

Renames pre-existing fields on Configuration so that they match the notifier spec.

## Changeset

Updated the following fields:

notifierType -> appType
filters -> redactKeys
notifyReleaseStages -> enabledReleaseStages
autoNotify -> autoDetectErrors
detectNdkCrashes -> autoDetectNdkCrashes
detectAnrs -> autoDetectAnrs
autoCaptureSessions -> autoTrackSessions


bugsnag-android-core/src/main/java/com/bugsnag/android/AppData.java
```diff
@@ -120,17 +120,6 @@ String getActiveScreenClass() {
...
     }
 
-    @NonNull
-    private String calculateNotifierType() {
-        String notifierType = config.getNotifierType();
-
-        if (notifierType != null) {
-            return notifierType;
-        } else {
-            return "android";
```
fractalwrench: This is the default value on the Configuration field now, so is unnecessary here

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -5,21 +5,21 @@ import java.util.HashMap
...
-    val notifyReleaseStages: Collection<String>,
+    val enabledReleaseStages: Collection<String>,
     val projectPackages: Collection<String>,
     val releaseStage: String?,
     val buildUuid: String?,
     val appVersion: String?,
     val versionCode: Int,
     val codeBundleId: String?,
-    val notifierType: String?,
+    val appType: String,
```
fractalwrench: This is now guaranteed to be non-null

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -418,8 +413,8 @@ public boolean getAutoCaptureSessions() {
      *
      * @param autoCapture whether sessions should be captured automatically
      */
-    public void setAutoCaptureSessions(boolean autoCapture) {
-        this.autoCaptureSessions = autoCapture;
+    public void setAutoTrackSessions(boolean autoCapture) {
+        this.autoTrackSessions = autoCapture;
```
bengourley:
```suggestion
        this.autoTrackSessions = autoTrack;
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -418,8 +413,8 @@ public boolean getAutoCaptureSessions() {
      *
      * @param autoCapture whether sessions should be captured automatically
      */
-    public void setAutoCaptureSessions(boolean autoCapture) {
-        this.autoCaptureSessions = autoCapture;
+    public void setAutoTrackSessions(boolean autoCapture) {
```
bengourley:
```suggestion
    public void setAutoTrackSessions(boolean autoTrack) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
bengourley: See my other comments – I don't think this should have a default value. Am fine with it being an empty string if that helps in avoiding nullness.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -64,7 +59,7 @@
     private final Collection<BeforeSendSession> sessionCallbacks = new ConcurrentLinkedQueue<>();
 
     private String codeBundleId;
-    private String notifierType;
+    private String appType = "android";
```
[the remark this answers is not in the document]
fractalwrench: This is pre-existing behaviour that has been refactored as part of this PR - my understanding was that this could also affect how the pipeline processes events, although I'm not sure exactly how.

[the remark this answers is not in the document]
bengourley: Oh ok, I thought the previous behaviour was that it was empty. I'm not sure the pipeline does anything with this field but since this PR is just about renaming options I'm happy for the behaviour to be figured out another time.

PR #618, opened 2019-10-16
Update Breadcrumb functionality to match notifier spec

fractalwrench:
## Goal

Updates the breadcrumb functionality to better match the notifier spec.

## Changeset

- Renamed `breadcrumb.name` to `breadcrumb.message`
- Captured "Bugsnag loaded" breadcrumb when a `Client` is constructed
- Enforce 0-100 limit on valid values for `config.maxBreadcrumbs`
- Change default for `config.maxBreadcrumbs` from 32 to 25
- Removed 140 char limit from breadcrumb message length
- Altered `breadcrumb.metadata` to accept `Object` values rather than just `String`. (Note: this does not currently prevent nested values, which will be addressed in a separate PR that looks more fully at data redaction/serialization)

### New APIs

- Renamed `BeforeRecordBreadcrumb` to `OnBreadcrumb`
- Added `addOnBreadcrumb` and `removeOnBreadcrumb` to `Bugsnag` and `Client`
- Added `enabledBreadcrumbTypes to `Configuration`, which discards any breadcrumbs not of the type within this set

## Tests
- Added unit test to verify breadcrumbs with a type not contained in `enabledBreadcrumbTypes` are discarded
- Added unit test to verify that `enabledBreadcrumbTypes` can be configured and has a sensible default value
- Added unit test to verify that `OnBreadcrumb` callbacks can be removed

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
@@ -345,7 +345,7 @@ public final class Bugsnag {
      * Leave a "breadcrumb" log message representing an action or event which
      * occurred in your app, to aid with debugging
      *
-     * @param message     A short label (max 32 chars)
+     * @param message     A short label
      * @param type     A category for the breadcrumb
      * @param metadata Additional diagnostic information about the app environment
      */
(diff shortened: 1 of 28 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -342,14 +345,14 @@ public static void leaveBreadcrumb(@NonNull String message) {
      * Leave a "breadcrumb" log message representing an action or event which
      * occurred in your app, to aid with debugging
      *
-     * @param name     A short label (max 32 chars)
+     * @param message     A short label (max 32 chars)
```
bengourley: The limit has been removed

```suggestion
     * @param message     A short label
```

PR #622, opened 2019-10-17
Alter type visibility/remove unneeded classes to meet Notifier spec

fractalwrench:
## Goal

Alters the access modifiers of several classes which no longer need to be public due to changes in the notifier spec. Also removes unused annotations which have no effect.

## Changeset

Removed the `@ThreadSafe` and `@InternalApi` annotations. `@ThreadSafe` was used for the Infer static analysis tool which we no longer used and was kept as it was public and removing it would break semver. `@InternalApi` is no longer necessary as internal communication takes place via the `NativeInterface`/other methods but could not be removed for similar reasons.

The visibility of the following classes has been reduced as they no longer form part of the public API:

- BugsnagException
- EventReceiver
- HandledState
- IOUtils
- Notifier

Additionally, the `Notifier` class has been converted to Kotlin to simplify its implementation.

## Tests

Relied on existing test coverage, and converted `Notifier` coverage to Kotlin.


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ReportTest.java
```diff
@@ -54,7 +54,7 @@ public void testModifyingGroupingHash() throws JSONException, IOException {
     public void testModifyReportDetails() throws Exception {
         report.setApiKey("custom-api-key");
         report.getNotifier().setName("React Native");
-        report.getNotifier().setURL("https://bugsnag.com/reactnative");
+        report.getNotifier().setUrl("https://bugsnag.com/reactnative");
```
bengourley: related?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -6,6 +6,7 @@
 import androidx.annotation.Nullable;
 
 import java.lang.reflect.Field;
+import java.util.Arrays;
```
bengourley: related?

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ReportTest.java
fractalwrench: Our code style for JVM languages prefers that acronyms have only the first letter capitalised - we couldn't change this for the `Notifier` class without making a breaking change as these methods were public.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: An import was missing on the `immutable-config` branch (I think due to a merge conflict) which would otherwise fail the build.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ReportTest.java
bengourley: I just didn't see the update to this method's definition in this branch… or is this method synthesized on the `Notifier` class because it's now written in kotlin?

fractalwrench: Yep! `Notifier` has a `url` property, which when compiled down to Java automatically generates a getter + setter:
https://github.com/bugsnag/bugsnag-android/pull/622/files#diff-2a2f9b5da9e45dce6d45a9b80d5d2253R12

PR #636, opened 2019-11-12
Update example app to use latest notifier spec syntax

fractalwrench:
## Goal

Updates the example app so that it uses the latest notifier spec syntax. Additionally, the example app is now an independent fixture and not part of the same gradle project as the rest of the repository, as this separation made it easier to test breaking API changes.

## Changeset

- Updated example app to use new `addMetadata` syntax, and to alter severity in an `OnError` callback
- Updated manifest config option names to match new syntax
- Removed example app module from the main gradle project and extracted it to a separate fixture
- Added gradlew and bumped gradlew to 5.1 (minimum supported for this project setup)
- Removed lint warnings and instrumentation test runner as we don't use these for the example project
- Copied necessary parts from buildscript from the main gradle project into the example fixture (such as setting up the Android gradle plugin)

## Tests

Ran the example app. This can be done by checking out https://github.com/bugsnag/bugsnag-android/pull/616/, running `./gradlew assembleRelease publishToMavenLocal` with a `gradle.properties` version of 9.9.9, then opening the example app separately within Android Studio and running on an emulator.


examples/sdk-app-example/src/main/java/com.bugsnag.android/example/ExampleActivity.kt
```diff
@@ -39,17 +40,11 @@ class ExampleActivity : AppCompatActivity() {
     }
 
     private fun performAdditionalBugsnagSetup() {
-        // Execute some code before every bugsnag notification
-        Bugsnag.beforeNotify { error ->
```
bengourley: can this be replaced with `.addOnError(fn)` or is that not implemented yet?

examples/sdk-app-example/src/main/AndroidManifest.xml
```diff
@@ -35,11 +35,11 @@
       android:value="YOUR-API-KEY-HERE" />
 
     <meta-data
-      android:name="com.bugsnag.android.DETECT_ANRS"
+      android:name="com.bugsnag.android.AUTO_DETECT_ANRS"
       android:value="true" />
 
     <meta-data
-      android:name="com.bugsnag.android.DETECT_NDK_CRASHES"
+      android:name="com.bugsnag.android.AUTO_DETECT_NDK_CRASHES"
```
bengourley: Assuming the default value for this configuration option is true, it is potentially confusing for it to be in the manifest like this. Maybe having it commented out like this would be more clear so that people don't think they have to include it:

```xml
<!-- NDK crash detection is enabled by default. To disable it, set the following option: -->
<!--
  <meta-data
    android:name="com.bugsnag.android.AUTO_DETECT_NDK_CRASHES"
    android:value="false" />
-->
```

examples/sdk-app-example/build.gradle
```diff
@@ -38,14 +66,9 @@ android {
...
 
-project.afterEvaluate {
-    project.android.lintOptions.abortOnError = false
-}
-
 dependencies {
-    implementation "com.bugsnag:bugsnag-android:+"
-
-    implementation "androidx.appcompat:appcompat:1.0.2"
+    implementation "com.bugsnag:bugsnag-android:9.9.9"
```
bengourley: Will this get set to a real version when the spec refactor has properly landed?

examples/sdk-app-example/src/main/AndroidManifest.xml
fractalwrench: NDK crash detection is currently disabled by default so we need to include this in the manifest for the example app.

examples/sdk-app-example/src/main/java/com.bugsnag.android/example/ExampleActivity.kt
fractalwrench: I believe that hasn't been implemented yet.

examples/sdk-app-example/build.gradle
fractalwrench: Yep - the intention is to update this to a real version when we release the spec refactor.

PR #637, opened 2019-11-12
Rename bugsnag_report to bugsnag_event

fractalwrench:
## Goal

Renames `Report` fields to be consistent with the types defined in the notifier spec, and makes the `Report` publicly accessible to users in preparation for addition of the native `OnError` hook.

The majority of this diff is formed of renaming 'report' to 'event'.

## Changeset

- Rename `bugsnag_report` to `bugsnag_event`, and related variable/method names throughout the codebase to prefer 'event' terminology (this results in the majority of the diff)
- Rename `bugsnag_library` to `bugsnag_notifier`
- Rename `bugsnag_exception` to `bugsnag_error`
- Renames `report` source files to `event` source files
- Moves contents of the `bugsnag_event` struct into a publicly accessible header file
- Creates a copy of the pre-existing `bugsnag_report` as `bugsnag_report_v2` in `migrate.h`. A migration has also been added when deserializing the file which copies `bugsnag_report_v1` to `bugsnag_report_v2`, then to `bugsnag_event`.

## Tests

Added a unit test for verifying that `bugsnag_report_v2` can be migrated to `bugsnag_event`, and that the notifier/error fields are correctly populated.


bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
```diff
@@ -1,12 +1,38 @@
...
+     * The number of frames used in the stacktrace. Must be less than
+     * BUGSNAG_FRAMES_MAX.
+     */
+    ssize_t frame_count;
+    /**
+     * An ordered list of stack frames from the oldest to the most recent
+     */
+    bsg_stackframe stacktrace[BUGSNAG_FRAMES_MAX];
+} bsg_exception;
+
```
tomlongridge: Should we include some comments in this file (or rename the type) to make it clear that `bsg_exception` and `bsg_library` are for older version?

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -0,0 +1,291 @@
+#ifndef BUGSNAG_ANDROID_NDK_REPORT_API_H
+#define BUGSNAG_ANDROID_NDK_REPORT_API_H
```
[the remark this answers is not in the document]
fractalwrench: this should be renamed.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -0,0 +1,291 @@
...
+    int version;
+    /**
+     * 0 if big endian
+     */
+    int big_endian;
+    /**
+     * The value of device.runtimeVersions.osBuild
+     */
+    char os_build[64];
+} bsg_report_header;
```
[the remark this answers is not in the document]
fractalwrench: Renamed to `bsg_event_header`

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -368,23 +434,23 @@ char *bsg_serialize_report_to_json_string(bugsnag_report *report) {
   JSON_Object *exception = json_value_get_object(ex_val);
   JSON_Value *stack_val = json_value_init_array();
   JSON_Array *stacktrace = json_value_get_array(stack_val);
-  json_object_set_value(event, "exceptions", exceptions_val);
```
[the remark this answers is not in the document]
fractalwrench: The string should remain as `"exceptions"` to be a valid payload. I therefore think it could make sense to name the 3rd param as `exceptions_val`, unless some other variable needed renaming?

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -0,0 +1,291 @@
...
+ */
+#define BUGSNAG_DEFAULT_EX_TYPE "c"
+#endif
+/**
+ * Version of the bugsnag_report struct. Serialized to report header.
+ */
+#define BUGSNAG_EVENT_VERSION 3
+
+#define BUGSNAG_USER_INFO_LEN 64
+
```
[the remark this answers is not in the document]
fractalwrench: Not including a guard for `__cplusplus` is consistent with what was there before. Was that a pre-existing bug? https://github.com/bugsnag/bugsnag-android/pull/637/files#diff-8f292ff3995248bd79217aba9d6c6f2e

I have moved the comment header to `jni/event.h`.

bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
fractalwrench: Added some general documentation towards the top which should be helpful for future readers.

bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder_libcorkscrew.h
```diff
@@ -1,7 +1,7 @@
 #ifndef BUGSNAG_UTILS_STACK_UNWINDER_LIBCORKSCREW_H
 #define BUGSNAG_UTILS_STACK_UNWINDER_LIBCORKSCREW_H
 
-#include "../report.h"
+#include "event.h"
```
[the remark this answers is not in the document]
fractalwrench: I have updated these to use `../event.h` for consistency with what was here before.

fractalwrench:
Thanks for the comments - I've addressed most of the feedback and left comments for discussion where I wasn't 100% sure on the best course of action.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -0,0 +1,287 @@
...
+#endif
+#ifndef BUGSNAG_FRAMES_MAX
+/**
+ *  Number of frames in a stacktrace. Configures a default if not defined.
+ */
+#define BUGSNAG_FRAMES_MAX 192
+#endif
+#ifndef BUGSNAG_CRUMBS_MAX
+/**
+ *  Max number of breadcrumbs in a report. Configures a default if not defined.
```
tomlongridge:
```suggestion
 *  Max number of breadcrumbs in an event. Configures a default if not defined.
```

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
```diff
@@ -28,15 +28,15 @@ typedef struct {
      * Preferred unwinding style
      */
     bsg_unwinder unwind_style;
-    bsg_report_header report_header;
+    bsg_event_header report_header;
```
tomlongridge: As we're still calling the file a "report", this struct should probably still be bsg_report_header. Or else we should rename the variable and several of the methods.

Also, although this is a previous omission, can we give this a comment to document it?

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
```diff
@@ -28,15 +28,15 @@ typedef struct {
...
     /**
      * File path on disk where the next crash report will be written if needed.
      */
-    char next_report_path[384];
+    char next_event_path[384];
     /**
      * Cache of static metadata and report info. Exception/time information is populated at crash time.
```
tomlongridge:
```suggestion
     * Cache of static metadata and event info. Exception/time information is populated at crash time.
```

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
fractalwrench: Renamed the struct to `bsg_report_header` and added a comment to this field.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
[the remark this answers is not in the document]
kattrali: The header guards are added to prevent C++ function name mangling breaking interop between C and the JNI. Since there's no functions in this file, no header guards are needed.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -0,0 +1,287 @@
...
+   *  Change in application state
+   */
+  BSG_CRUMB_STATE,
+  /**
+   *  User event, such as authentication or control events
+   */
+  BSG_CRUMB_USER,
+} bsg_breadcrumb_t;
+
+typedef struct {
```
kattrali: The header in `assets/include/event.h` is the public header shown to users who use the Native API. It should not include private properties like the internal representations of stackframe and the `#define` values for stacktrace length, etc, as these aren't public APIs that we want to commit to forever, and we should be able to migrate those values internally without breaking people's code. 

The only public components of event are the severity enum and breadcrumb types, until we design and deploy an API which requires more, (e.g. for callbacks) which will need some design anyway to ensure that the API is clear.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -13,100 +13,166 @@
...
-          size_t breadcrumb_size = sizeof(bugsnag_breadcrumb) * BUGSNAG_CRUMBS_MAX;
-          memcpy(&report->breadcrumbs, report_v1->breadcrumbs, breadcrumb_size);
-
-          strcpy(report->context, report_v1->context);
-          report->severity = report_v1->severity;
-          strcpy(report->session_id, report_v1->session_id);
-          strcpy(report->session_start, report_v1->session_start);
-          report->handled_events = report_v1->handled_events;
-          report->unhandled_events = 1;
+bugsnag_event *map_v2_to_report(bugsnag_report_v2 *report_v2) {
```
kattrali: All of these functions should use the `bsg_` prefix, to avoid collisions with any other libraries running in the host app.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -13,100 +13,166 @@
...
+  }
+  free(report_v2);
+  return event;
+}
 
-          free(report_v1);
-      }
-  } else {
-      report = bsg_report_v2_read(fd);
+bugsnag_event *map_v1_to_report(bugsnag_report_v1 *report_v1) {
```
kattrali: Same as above.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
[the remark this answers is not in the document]
fractalwrench: Resolving as my understanding is that no further action is required here.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: Prefixed these methods.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
kattrali: _summary of in-person discussion_

There should be a more structured/user-friendly interface for the event prior to implementing `on_error`. In particular, the following points will be addressed as a part of a subsequent document/changeset:

* Having a way to safely set and remove metadata/app/device properties
* Reviewing the interface to ensure it is hardened for arbitrary input as currently the functions were designed for internal use
* Protecting report components which could be incorrectly altered and break reporting (for example, metadata storage and the breadcrumb circular buffer could be broken entirely by changing the index properties incorrectly)

fractalwrench: I've updated `assets/include/event.h` to only include the severity/breadcrumb type structs, and moved the other structs back to an internal header.

The remaining work on clariying the event API for on_error callbacks will take place in separate PRs.

PR #639, opened 2019-11-12
Remove BeforeSend callback and associated deserialization code

fractalwrench:
## Goal

The `BeforeSend` callback has been superseded by the concept of an `OnError` callback, which runs in the crashing context for the JVM/NDK. Therefore the `BeforeSend` callback should be removed from our public API to prevent user confusion over which callback to use.

Additionally, the only reason we deserialize cached JSON files is so they can be exposed in a `BeforeSend` callback. This has traditionally been error prone and a source of bugs - as this is no longer required, the deserialization code can be removed from bugsnag-android.

## Changeset

Removed `BeforeSend` interface, unit tests, and mazerunner scenarios (new tests will be added for `OnError` in a separate diff). Additionally the `ErrorReader` and associated test code have been removed, as it is no longer required to deserialize JSON from disk.


tests/features/cached_error_reports.feature
```diff
@@ -11,79 +11,22 @@ Scenario: Sending internal error reports on API <26
...
-    And the event "device.osName" equals "android"
-    And the event "metaData.BugsnagDiagnostics.filename" is not null
-    And the event "metaData.BugsnagDiagnostics.notifierName" equals "Android Bugsnag Notifier"
-    And the event "metaData.BugsnagDiagnostics.apiKey" equals "ABCDEFGHIJKLMNOPQRSTUVWXYZ012345"
-    And the event "metaData.BugsnagDiagnostics.packageName" equals "com.bugsnag.android.mazerunner"
-    And the event "metaData.BugsnagDiagnostics.notifierVersion" is not null
-    And the event "metaData.BugsnagDiagnostics.fileLength" equals 4
-    And the event "metaData.BugsnagDiagnostics.cacheGroup" is false
-    And the event "metaData.BugsnagDiagnostics.cacheTombstone" is false
+    Then I should receive no requests
```
snmaynard: I dont get these scenario changes. Reading these tests its not at all clear. What are they testing?

fractalwrench: These scenarios used to verify that when sending an invalid JSON file, the notifier sent an incomplete report to our own internal instance. Because we no longer need to deserialize cached JSON files the payload will be dropped by the payload instead - so the scenario has been updated to verify that 0 valid requests are received by mazerunner.

PR #640, opened 2019-11-12
Rename SessionTrackingPayload to SessionPayload

fractalwrench:
## Goal

Renames `SessionTrackingPayload` to `SessionPayload` to achieve consistency with the notifier spec, and ensures that the `app` and `device` fields are mutable so that values can be altered in an `OnSession` callback.

## Changeset

- Renamed `SessionTrackingPayload` to `SessionPayload` and associated methods/variables
- Updated `JsonStream` to avoid an ambiguous method signature by removing `value(Streamable)`, as an `Object` and a `Streamable` can be the same type, causing a compile-time error
- Updated `SessionPayload` to have a `MutableMap` for app/device which allows redaction/addition of information to the session

## Tests

Relied on existing test coverage, and converted serialization test to a parameterised unit test that runs on the JVM.


fractalwrench:
`getAppDataSummary()` is an existing method on `AppData` that was previously called internally within `SessionTrackingPayload`. One of the interface requirements is that the app/device should be mutable public properties - the simplest way of doing this was to replace the `AppData/DeviceData` constructor parameters with a `MutableMap`.  This also avoids the painful process of having to mock `AppData/DeviceData` in the JVM tests.

PR #641, opened 2019-11-15
Size comparison

Cawllec:
## Goal

Add an extra testing stage that outputs an Android sizing table to a pull request using the Danger tool.


tests/features/fixtures/minimalapp/.gitignore
```diff
@@ -0,0 +1,13 @@
+*.iml
+.gradle
+/local.properties
+/.idea/caches
```
fractalwrench: `.idea` should be sufficient to ignore any additional files that intellij might generate in future

tests/features/fixtures/minimalapp/app/bugsnag.template
```diff
@@ -0,0 +1,47 @@
+apply plugin: 'com.android.application'
+apply plugin: 'com.bugsnag.android.gradle'
```
fractalwrench: It might be possible to depend on bugsnag conditionally by passing a command line argument when making the build. this would avoid the scenario where the dependency versions of something like `com.android.support:appcompat` gets out of sync between the two versions.

  https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_task_using_project_properties

tests/features/fixtures/minimalapp/app/base.template
```diff
@@ -0,0 +1,33 @@
...
+        applicationId "com.bugsnag.android.minimalapp"
+        minSdkVersion 14
+        targetSdkVersion 28
+        versionCode 1
+        versionName "1.0"
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+    }
+    buildTypes {
+        release {
+            minifyEnabled false
```
fractalwrench: If it's not too much extra effort, it would also be quite useful information to know what the size difference is when an APK is minified.

tests/features/fixtures/minimalapp/app/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation 'com.android.support:appcompat-v7:28.0.0'
+    testImplementation 'junit:junit:4.12'
+    androidTestImplementation 'com.android.support.test:runner:1.0.2'
+    androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.2'
+}
+
+if (project.hasProperty("withBugsnag")) {
+    println("Building with Bugsnag")
```
fractalwrench: `project.logger.info` might be more appropriate if a message needs to be logged here as it will show up in any build logs on CI

tests/features/fixtures/minimalapp/Dangerfile
```diff
@@ -0,0 +1,95 @@
...
+}
+
+def format_kbs(size)
+  (size/1000.0).round(2)
+end
+
+markdown(%Q{
+  ### Android notifier sizes
+
+  | Format      | Size (kB)                                     | Minified Size (kB)                                |
```
fractalwrench: Could we rename the Size/Minified Size titles? This made me think that it was the size of the total APK, rather than the impact of bugsnag-android. Maybe "size impact of bugsnag-android"?

Cawllec:
Addressed the above comments, unfortunately the `You used 'puts'` notice is due to the use of `outputs` in the directory strings, not because anything is being logged.

PR #644, opened 2019-11-18
Rename Error to Event

fractalwrench:
## Goal

Renames `Error` to `Event`, to gain consistency with the notifier spec. Note that this does not alter the actual interface which is still non-compliant - this will be addressed in separate PRs.

## Changeset

Renamed:
- `Error` to `Event`
- `ErrorStore` to `EventStore`
- `EventReceiver` to `SystemBroadcastReceiver`
- Local variable/parameter/accessors of "error" to "event"

Additionally the IntelliJ refactor tool removed some unused imports that are not required.

Originally raised in #624


fractalwrench:
Thanks for a thorough review - I've updated the test code local variables to `event`. I've also updated the JavaDoc, although later commits will remove that method and the general intention is that docs will be rewritten after everything has been merged down to the integration branch and the spec fully agreed upon.

I've also pushed a small fix that was failing some of the mazerunner scenarios on CI that was missing from this changeset. Essentially the value for `config.delivery` was null when disabling session/report delivery in isolation, so it is now seeded with a default value (this was in a separate commit on the other integration branch).

PR #648, opened 2019-11-18
Remove obsolete notify() calls and supply 4 variants instead 

fractalwrench:
## Goal

Removes obsolete `notify()` method signatures and supplies 4 variants instead: 2 of which take a `Throwable` and optional callback, the other 2 of which take an error message/class, stacktrace, and optional callback. This matches the notifier spec - use cases such as altering the severity/adding metadata should now be achieved by supplying an `OnError` callback.

Originally raised in #626

## Changeset
- Removed unnecessary `notify()` calls from `Bugsnag` and `Client` classes
- Removed obsolete instrumentation tests `BeforeSendTest` and `ClientNotifyTest` which covered removed `notify()` calls
- Removed obsolete mazerunner scenarios which covered removed `notify()` calls
- Removed unused options from `DeliveryStyle`

Note: this commit also fixes test code for #645 which migrated the device time/locale fields.

## Tests
Relied on running existing tests.


bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryStyle.java
```diff
@@ -1,8 +1,6 @@
 package com.bugsnag.android;
 
 enum DeliveryStyle {
-    SAME_THREAD,
```
bengourley: What purpose did `SAME_THREAD` delivery style / `notifyBlocking()` method serve and why is it ok that it hasn't been replaced?

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryStyle.java
```diff
@@ -1,8 +1,6 @@
...
     ASYNC,
-    ASYNC_WITH_CACHE,
-    NO_CACHE
```
bengourley: Was this simply unused?

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -228,24 +229,23 @@ void bsg_serialize_device(const bsg_device_info device, JSON_Object *event) {
   }
 
   json_object_dotset_number(event, "device.totalMemory", device.total_memory);
+
+  char report_time[sizeof "2018-10-08T12:07:09Z"];
```
bengourley: I don't think this should be part of this PR – it doesn't seem related to removing obsolete `notify()` calls?

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/NewSessionScenario.kt
```diff
@@ -50,7 +50,7 @@ internal class NewSessionScenario(config: Configuration,
             flushAllSessions()
             Log.d("Bugsnag - New", "Second session flushed")
             Thread.sleep(SLEEP_MS)
-            client.notifyBlocking(generateException())
+            client.notify(generateException())
```
bengourley: Is this now a non-blocking call? Seems like a pretty fundamental change so I just want to understand the impact of of it better.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryStyle.java
fractalwrench: `notifyBlocking()` generated and then delivered a `Report` on the same thread. In contrast, we now always generate an `Event`, then deliver it on a background thread.

Blocking delivery is not what the vast majority of users would want as this is likely to cause ANRs in a production app, by blocking the main thread. `notifyBlocking()` also exposes additional complexity to the user and forces them to make a conscious choice between `notify()` and `notifyBlocking()` - when the first option is the right one. Additionally this is no longer in the notifier spec so removing this method is more consistent.

Removing `notifyBlocking()` means that the enum only has two options. The `DeliveryStyle` really wants to check whether a report is unhandled and cache it to disk before attempting delivery if that is the case. Therefore the `Client` has been updated to do this and `DeliveryStyle` has been removed as it is obsolete.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: It appears that this also contains the smaller changeset for #645, so this should disappear from the diff once that is merged.

PR #651, opened 2019-11-20
Rename MetaData to Metadata

fractalwrench:
## Goal

Renames `MetaData` to `Metadata` to gain consistency with the notifier spec. This changeset also changes any local variable usage of `metaData`.


features/support/env.rb
```diff
@@ -6,15 +6,15 @@
 # Install latest versions of bugsnag-android
   run_required_commands([
     [
-      "./gradlew", "assembleRelease", "publishToMavenLocal", "-x", "lintVitalRelease"
```
fractalwrench: This task is no longer in the project after the example app restructure, so needs to be removed to make mazerunner run.

bengourley: Should it have been removed in the the example app restructure PR then?

fractalwrench: Ideally yes - this is only needed to run the local version of mazerunner, which is why it was originally missed.

fractalwrench:
Thanks - I've updated the additional cases to use the new naming convention for metadata.

fractalwrench:
Spotted a few more cases myself after doing a global search/replace. I believe that's the last of them.

PR #652, opened 2019-11-20
Remove unused methods from Severity and add test coverage

fractalwrench:
## Goal

Removes unused public methods from the `Severity` class to meet the notifier spec, and adds test coverage for the serialization method.

## Changeset

- Removed `fromString()` and `fromChar()` as these are no longer used (they were once required for JSON deserialization)
- Convert `Severity` to Kotlin

## Tests

Added parameterised JVM test to verify serializer.


fractalwrench:
The methods are public so removing them will be a breaking change - which is why it'd be good to do them as part of this project. I agree that functionally there is no change as far as Severity is concerned.

PR #654, opened 2019-11-20
Add Error interface

fractalwrench:
## Goal

Adds the `Error` interface to conform to the Notifier spec. Note that this PR only adds the interface, and this is not used by the `Event` yet.

## Tests

Added a JVM serialization test to verify that JSON is constructed correctly.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Error.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Error.kt
@@ -13,8 +13,8 @@ class Error @JvmOverloads internal constructor(
 
             var currentEx: Throwable? = exc
             while (currentEx != null) {
-                val trace = Stacktrace(exc.stackTrace, projectPackages)
-                errors.add(Error(exc.javaClass.name, exc.localizedMessage, trace.trace))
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Error.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    companion object {
+        fun createError(exc: Throwable, projectPackages: Collection<String>): List<Error> {
+            val errors = mutableListOf<Error>()
+
+            var currentEx: Throwable? = exc
+            while (currentEx != null) {
+                val trace = Stacktrace(exc.stackTrace, projectPackages)
+                errors.add(Error(exc.javaClass.name, exc.localizedMessage, trace.trace))
+                currentEx = currentEx.cause
```
bengourley: Do you plan to add a test for adding multiple exceptions in another PR? If not could you ensure the test cover this scenario here?

fractalwrench: I have added a test for adding multiple exceptions - this uncovered a bug caused by using `exc` rather than `currentExc`, so I've also fixed that.

PR #656, opened 2019-11-20
Add Stackframe serialization test

fractalwrench:
## Goal

Adds a JVM serialization test to verify the constructed JSON for the `Stackframe` object.

Additionally, the `customFields` field has been updated to allow nullable values in its map, and the serializer has been updated to allow nullable objects (which are not omitted from serialization by default)

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
@@ -7,7 +7,7 @@ class Stackframe internal constructor(
     var file: String?,
     var lineNumber: Number?,
     var inProject: Boolean?,
-    private var customFields: Map<String, Any>
+    private var customFields: Map<String, Any?>
 ): JsonStream.Streamable {
 
     @Throws(IOException::class)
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -7,7 +7,7 @@ class Stackframe internal constructor(
     var file: String?,
     var lineNumber: Number?,
     var inProject: Boolean?,
-    private var customFields: Map<String, Any>
+    private var customFields: Map<String, Any?>
```
bengourley: is the added `?` because the `Any` type does not include `null`?

fractalwrench:
>Interface-wise, is columnNumber never a thing on Android?
This is currently the case. `Stackframe` includes a `customFields` property that will allow for the addition of columnNumber for JS exceptions, and loadAddress etc for NDK errors.

We could alter this during the JS refactor if it's desirable to have a structured field, rather than a Map that accepts arbitrary values.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
fractalwrench: Yep.  `Any` and `Any?` are separate types in Kotlin: the `?` means that it allows null values. There's a good article explaining Kotlin's type hierarchy in a bit more detail here: http://www.natpryce.com/articles/000818.html

PR #658, opened 2019-11-21
Update Breadcrumb interface to match notifier spec

fractalwrench:
## Goal

Updates the `Breadcrumb` interface to match the notifier spec, and alters the supporting classes `BreadcrumbState` and `BreadcrumbType` to support this change.

## Changeset

- Altered `Breadcrumb` metadata field to a `MutableMap<String, Any?> to allow mutation as per the spec
- Moved payload size/pruning calculations to `BreadcrumbState` to reduce the responsibilities of `Breadcrumb`
- Replaced breadcrumb instrumentation test with a JVM serialization test to verify JSON is generated in the correct way for both `Breadcrumb` and `BreadcrumbState` in isolation
- Renamed `Breadcrumbs` to `BreadcrumbState` to better reflect that it holds the state of breadcrumbs collected by the `Client`, and emits observable events to notify the NDK
- Converted Breadcrumb classes to Kotlin
- Renamed variable usage from `breadcrumbs` to `breadcrumbState` where applicable
- Updated ProGuard rule to retain `BreadcrumbState` rather than `Breadcrumbs`


bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbState.kt
```diff
@@ -0,0 +1,78 @@
...
+            pruneBreadcrumbs()
+            setChanged()
+            notifyObservers(
+                NativeInterface.Message(
+                    NativeInterface.MessageType.ADD_BREADCRUMB, breadcrumb
+                )
+            )
+        } catch (ex: IOException) {
+            Logger.warn("Dropping breadcrumb because it could not be serialized", ex)
+        }
```
tomlongridge: Suggest moving this `catch` either closer to the `payloadSize` call or modifying the method to make it a validation function?

bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbState.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private fun pruneBreadcrumbs() {
+        // Remove oldest breadcrumbState until new max size reached
+        while (store.size > maxBreadcrumbs) {
+            store.poll()
+        }
+    }
+
+    companion object {
+        private const val MAX_PAYLOAD_SIZE = 4096
```
tomlongridge: Seems odd to define this here, if you're only referring to it in the main class.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -234,7 +234,7 @@ public static String getNativeReportPath() {
      */
     @NonNull
     public static List<Breadcrumb> getBreadcrumbs() {
-        Queue<Breadcrumb> store = getClient().breadcrumbs.store;
+        Queue<Breadcrumb> store = getClient().breadcrumbState.getStore();
```
tomlongridge: Possibly my lack of Kotlin familiarity, but why call `getStore` and use the `store` field?

bugsnag-android-core/src/test/java/com/bugsnag/android/BreadcrumbMutabilityTest.kt
```diff
@@ -1,24 +1,12 @@
...
 class BreadcrumbMutabilityTest {
-
-    @Test
-    fun breadcrumbCopiesMap() {
-        val data = mutableMapOf<String, Any>()
-        val breadcrumb = Breadcrumb("foo", BreadcrumbType.MANUAL, data)
-        data["a"] = "bar"
-        assertTrue(breadcrumb.metadata.isEmpty())
-        assertNotSame(data, breadcrumb.metadata)
-    }
```
tomlongridge: Has not making a copy been specified as the behaviour?

bugsnag-android-core/src/test/java/com/bugsnag/android/BreadcrumbSerializationTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@RunWith(Parameterized::class)
+internal class BreadcrumbSerializationTest {
+
+    companion object {
+        @JvmStatic
+        @Parameters
+        fun testCases() = generateSerializationTestCases(
+            "breadcrumb",
+            Breadcrumb("hello world", BreadcrumbType.MANUAL, mutableMapOf(), Date(0))
```
tomlongridge: Should this have non-null metadata test as well?

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -256,11 +261,29 @@ private boolean is32bit() {
         return is32bit;
     }
 
+    // TODO temporary - will be removed in later commit
```
tomlongridge: Is this change made necessary by the refactor? If so, what later commit will remove it and can you add some discussion on the PR about it?

If it's not related to the refactor, I suggest we remove this from the refactor entirely and address as a separate change, targeting master.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: Kotlin properties automatically synthesise getters + setters. So when called from Java the `store` property actually looks like this behind the scenes: 

```
private final Map<String, Object> store;

public Map<String, Any> getStore() {
    return store;
}
```

Therefore anywhere that previously used a field has been updated to use the getter, as the field is not accessible.

bugsnag-android-core/src/test/java/com/bugsnag/android/BreadcrumbMutabilityTest.kt
fractalwrench: Yes, the map is now accessible and mutable, so we no longer need to assert that these objects are not the same. Mutating the map is already covered by other tests.

bugsnag-android-core/src/test/java/com/bugsnag/android/BreadcrumbSerializationTest.kt
fractalwrench: This test has been ported from #606 which aimed to only add one test case each to start off with. I can add more test cases if that's desirable.

bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbState.kt
```diff
@@ -0,0 +1,78 @@
...
+
+        @Throws(IOException::class)
+        internal fun payloadSize(breadcrumb: Breadcrumb): Int {
```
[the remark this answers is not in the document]
fractalwrench: Renamed method

bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbState.kt
fractalwrench: I agree - the catch should be more specific than it was historically written as, I'll update this.

bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbState.kt
fractalwrench: Kotlin replaces the concept of `static` with a [Companion object](https://kotlinlang.org/docs/tutorials/kotlin-for-py/objects-and-companion-objects.html) - essentially this is a singleton that can be defined for each class. This would be the idiomatic way of defining a constant in Kotlin and is equivalent to `private static final int MAX_PAYLOAD_SIZE = 4096` in Java.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
fractalwrench: Yes - the spec states that `breadcrumb.timestamp` should be a `Date`, whereas before it was a `String`. `NativeBridge` does not have access to `DateUtils` because it is defined in a separate package, meaning it is necessary to perform the conversion elsewhere.

This will be resolved by the following commit, which alters the `ADD_BREADCRUMB` message to use individual primitive fields rather than the `Breadcrumb` object - meaning that the date serialization is done up front when sending the message: https://github.com/bugsnag/bugsnag-android/commit/a581c825eeb4f510ebd2295914246f7ba096eaa2#diff-a28efd499ea99991ba5a7fb937719768R87

Completing this change without a temporary duplication either requires waiting for #659 to be merged, creating a temporary fix that will be overwritten, or just accepting that this will break CI.

bugsnag-android-core/src/test/java/com/bugsnag/android/BreadcrumbSerializationTest.kt
fractalwrench: Added a test case to sanity-check that the map is serialized.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
tomlongridge: OK - understood.

bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbState.kt
tomlongridge: I meant more the fact that the "constant" is in the companion object, but used in the class only.

PR #660, opened 2019-11-21
Add serialization test for stacktrace

fractalwrench:
## Goal

Adds a serialization test to verify that `Stacktrace` generates the expected JSON. This changeset also allows adding nullable values to the `customFields` field, which accepts arbitrary values from RN + NDK - so we should allow null values.

## Changeset

- Make `JsonStream` accept nullable values when serializing (these are omitted from the JSON so there is no functional change here, but this is required in order to change the `customFields` field to accept null values)
- Converted existing Stacktrace instrumentation test to run on the JVM


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StacktraceTest.java
```diff
@@ -1,118 +0,0 @@
...
-        StackTraceElement stackTraceElement = new StackTraceElement("",
-            "someMethod", "someFile", 12);
-        Stacktrace stacktrace1 = new Stacktrace(
-            new StackTraceElement[]{stackTraceElement}, Collections.<String>emptyList());
-        stacktraceJson = streamableToJsonArray(stacktrace1);
-
-        frame = (JSONObject) stacktraceJson.get(0);
-        assertEquals("someMethod", frame.get("method"));
-    }
-}
```
tomlongridge: This test appears to be missing in the refactor.

fractalwrench: This particular test case is now implicit, as there is now a JSON fixture which holds the expected output. If the class name/method name/file name is wrong, the generated JSON will be wrong and the test will fail as the two are different.

PR #661, opened 2019-11-21
Refactor AppData data collection and add JVM serialization test

fractalwrench:
## Changeset

- Rename `getAppMetadata()` for consistency with `Metadata` capitalization
- Use `config.versionCode` directly within `AppData` to simplify implementation
- Move serialization tests from instrumentation sourceset to a JVM test, consolidating in one class
- Enabled mockmaker-inline, which is required for Mockito to mock final classes


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/AppDataTest.java
```diff
@@ -63,20 +63,4 @@ public void testAccessors() {
...
-    public void testJsonSerialisation() throws JSONException {
-        appData.put("buildUUID", "fa54de");
-        JSONObject appDataJson = mapToJson(appData);
-
-        assertEquals(1, appDataJson.getInt("versionCode"));
-        assertEquals("1.0", appDataJson.get("version"));
-        assertEquals("development", appDataJson.get("releaseStage"));
-        assertEquals("android", appDataJson.get("type"));
-        assertEquals("com.bugsnag.android.core.test", appDataJson.get("id"));
-        assertNotNull(appDataJson.get("buildUUID"));
```
simonbowring: Do we need to check the serialization of `buildUUIID`, `duration`, `durationInForeground` and `inForeground` in the new tests like this test was?

tomlongridge: It feels like we have a lot of different test classes - AppDataSummaryTest, AppDataTest, AppDataOverrideTest and AppDataSerialisationTest - but they don't seem to cover all the fields between them.

Can these be rationalised and make sure that we include each of the fields of App?

fractalwrench: I have removed the various different test classes in favour of `AppDataSerializationTest`, which now covers the serialization of each value defined within AppData.

fractalwrench: This is also an issue with tests for `DeviceData` - I'll address that in a separate PR.

bugsnag-android-core/src/main/java/com/bugsnag/android/AppData.java
```diff
@@ -191,11 +186,8 @@ private long getMemoryUsage() {
      * Check if the device is currently running low on memory.
      */
     @Nullable
-    private Boolean isLowMemory() {
+    private Boolean isLowMemory(ActivityManager activityManager) {
```
tomlongridge: This doesn't need to be passed in as a parameter - it can use the member variable.

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataSerializationTest.kt
```diff
@@ -0,0 +1,71 @@
...
+            // construct AppData object
+            val appData = AppData(context, pm, convert(config), sessionTracker, am)
+            appData.setBinaryArch("x86")
+
+            // serializes the 3 different maps that AppData can generate:
+            // 1. summary (used in session payloads)
+            // 2. regular (used in event payloads)
+            // 3. metadata (used in event payloads)
+            val metadata = appData.appDataMetadata
+            metadata.remove("memoryUsage")
```
tomlongridge: Can this not be tested using a mock?

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/AppDataTest.java
tomlongridge: Should we not also have basic unit tests for the accessors (separate from serialisation)? They would be a unit test, not an instrumentation test, but would check the getters return an appropriate value.

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataSerializationTest.kt
fractalwrench: I don't believe this is possible with the current implementation as this calculation uses `Runtime`, which is not mockable as it should only be instantiated once per application.

tomlongridge: This makes me more confident that we should have a more basic unit test for the getters in AppData - then we can at least assert that it's non-null?

https://github.com/bugsnag/bugsnag-android/pull/661#discussion_r351794414

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/AppDataTest.java
fractalwrench: `AppData#getAppDataSummary()` returns a `Map`, so I'm not sure that testing the accessors would be necessary as we're not defining any of our own. Unless I've got the wrong end of the stick here?

bugsnag-android-core/src/main/java/com/bugsnag/android/AppData.java
fractalwrench: Updated to use the field.

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataSerializationTest.kt
tomlongridge: I'll raise this separately to address unit test structure for this and Device.

PR #662, opened 2019-11-21
Update DeviceData and add serialization test case

fractalwrench:
## Goal

Ports changes to `DeviceData` made in this PR to the notifier spec update branch, and adds JVM serialization tests: https://github.com/bugsnag/bugsnag-android/pull/606#discussion_r331988213

## Changeset

- Renames `getDeviceMetadata()` for consistency with `Metadata` capitalization
- Convert `DeviceData` instrumentation tests to single JVM serialization test that verifies JSON structure
- Adds `DeviceBuildInfo`, which holds static `Build` fields directly so that their values can be faked under test
- Updated `DeviceData` to take `DeviceBuildInfo` as a field

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
@@ -5,24 +5,21 @@ import android.content.ContentResolver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.os.BatteryManager;
 import android.os.Build;
 import android.os.Environment;
-import android.os.StatFs;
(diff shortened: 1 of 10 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -115,7 +114,7 @@
         return map;
     }
 
-    Map<String, Object> getDeviceMetaData() {
+    Map<String, Object> getDeviceMetadata() {
```
Pezzah: Could the Build.BRAND be extracted in buildInfo as well?

fractalwrench: Good catch - I've updated `DeviceBuildInfo` to contain this field as well.

PR #664, opened 2019-11-22
Update JSON serializer to allow redacting specific values

fractalwrench:
## Goal

The `ObjectJsonStreamer` implementation is used to serialize JSON, but currently redacts the value of _any_ key that matches `redactKeys`. This is inconsistent with the specification, which states that only `metadata` and `breadcrumbs[n].metadata` should be redacted.

This changeset updates redaction so that only matching values in `Breadcrumb` and `Metadata` are replaced with `[REDACTED]`. It also adds additional test coverage and converts existing instrumentation tests to run on the JVM.

## Changeset

- Adds a `redactKeys` flag to `objectToStream` to indicate whether the object should have its values redacted. This is false by default
- Update `Metadata` and `Breadcrumb` serializers to redact values
- Converted `ObjectJsonStreamer` to Kotlin and refactored value redaction
- Converted `JsonStreamTest` to a JVM test, and split individual test cases into `JsonStreamFileTest` and `JsonStreamTest`
- Add test coverage for redaction in `JsonStreamTest`
- Move `JsonWriterTest` to the JVM test directory and remove android instrumentation test annotation
- Add test cases for verifying data is redacted from `Metadata` when serialized, replacing existing coverage


bugsnag-android-core/src/test/java/com/bugsnag/android/MetadataRedactionTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MetadataRedactionTest {
+
```
tomlongridge: Should we include a test for "password" being the default redacted key?

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/JsonStreamTest.java
```diff
@@ -1,104 +0,0 @@
...
-        stream.name("float").value(123.45f);
-        stream.endObject();
-
-        JSONObject json = new JSONObject(writer.toString());
-        assertTrue(json.isNull("nullLong"));
-        assertTrue(json.isNull("nullBoolean"));
-        assertTrue(json.isNull("nullString"));
-        assertTrue(json.isNull("nullInteger"));
-        assertTrue(json.isNull("nullFloat"));
-        assertTrue(json.isNull("nullDouble"));
```
tomlongridge: There doesn't seem to be an equivalent test for null values.

bugsnag-android-core/src/test/java/com/bugsnag/android/MetadataRedactionTest.java
fractalwrench: This is covered in a [separate test](https://github.com/bugsnag/bugsnag-android/blob/json-redaction/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ManifestConfigLoaderTest.kt#L49) which verifies 'password' is the default value of `redactKeys`

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/JsonStreamTest.java
fractalwrench: Added an equivalent test case for checking null values.

bugsnag-android-core/src/test/java/com/bugsnag/android/MetadataRedactionTest.java
tomlongridge: This would be the default even if Manifest files were not used. This feels like a lower-level requirement that should have a test here unless you are going to handle the default value in Configuration and not have any default at this level.

fractalwrench: Added an additional test case to verify the default value on `Metadata` specifically.

PR #667, opened 2019-11-22
Add test coverage for session JSON serialization

fractalwrench:
## Goal

Adds missing test coverage to verify that `Session` constructs JSON correctly. This functionality is ported from #606


bugsnag-android-core/src/test/java/com/bugsnag/android/SessionSerializationTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@RunWith(Parameterized::class)
+internal class SessionSerializationTest {
+
+    companion object {
+        @JvmStatic
+        @Parameters
+        fun testCases() = generateSerializationTestCases(
+            "session",
+            Session("123", Date(0), User(null, null, null), 1, 0)
```
tomlongridge: I'd rather `User` had some non-null data to make sure the object is used in the output.

fractalwrench: Added a placeholder value for User to ensure it is serialized in the session JSON.

PR #668, opened 2019-11-22
Refactor ThreadState implementation

fractalwrench:
## Goal

Refactors the `ThreadState` implementation to reduce the knowledge required to create an object.

This is not a public interface so does not result in any publicly facing change, but is designed to make the construction of an `Event` easier (and forms the groundwork of a future PR to alter the `Event` interface).

## Changeset

- Removed obsolete constructor which takes a list of Threads, as this was only required when deserializing JSON which we no longer do
- Made `threads` mutable to allow users to alter the value on `Event`
- Alter the constructor so that only a configuration and `Array<StacktraceElement>` are required parameters, and a sensible default is chosen for the current thread/thread trace
- Remove `testSerialization()` test case as this is covered by `ThreadSerializationTest`

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
@@ -419,13 +419,7 @@ public class Event implements JsonStream.Streamable, MetadataAware {
                 boolean unhandled,
                 Metadata globalMetadata) {
             Throwable exc = unhandled ? exception : null;
-            StackTraceElement[] stackTrace;
-            if (exc != null) {
-                stackTrace = exc.getStackTrace();
-            } else {
-                stackTrace = thread.getStackTrace();
-            }
-            this.threadState = new ThreadState(config, stackTrace, thread);
+            this.threadState = new ThreadState(config, exc, thread);
             this.config = config;
             this.exception = exception;
             this.severityReasonType = HandledState.REASON_USER_SPECIFIED; // default
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
@@ -419,7 +419,13 @@ public class Event implements JsonStream.Streamable, MetadataAware {
                 boolean unhandled,
                 Metadata globalMetadata) {
             Throwable exc = unhandled ? exception : null;
-            this.threadState = new ThreadState(config, thread, Thread.getAllStackTraces(), exc);
+            StackTraceElement[] stackTrace;
+            if (exc != null) {
+                stackTrace = exc.getStackTrace();
+            } else {
+                stackTrace = thread.getStackTrace();
+            }
+            this.threadState = new ThreadState(config, stackTrace, thread);
             this.config = config;
             this.exception = exception;
             this.severityReasonType = HandledState.REASON_USER_SPECIFIED; // default
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
@@ -5,37 +5,29 @@ import java.io.IOException
 /**
  * Capture and serialize the state of all threads at the time of an exception.
  */
-internal class ThreadState : JsonStream.Streamable {
+internal class ThreadState @JvmOverloads constructor
+    (
+    config: ImmutableConfig,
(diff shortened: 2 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -5,37 +5,29 @@ import java.io.IOException
...
+        // unhandled errors use the exception trace
         // API 24/25 don't record the currentThread, add it in manually
         // https://issuetracker.google.com/issues/64122757
-        if (!stackTraces.containsKey(currentThread)) {
-            stackTraces[currentThread] = currentThread.stackTrace
-        }
-        if (exc != null) { // unhandled errors use the exception trace
-            stackTraces[currentThread] = exc.stackTrace
-        }
+        stackTraces[currentThread] = excStacktrace
```
tomlongridge: Can you guarantee that for non-API-24/25 stracktraces, it has no impact to reassign the existing one? I would rather see the original guard still in place to check the key.

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -5,37 +5,29 @@ import java.io.IOException
...
-        currentThread: java.lang.Thread,
-        stackTraces: MutableMap<java.lang.Thread, Array<StackTraceElement>>,
-        exc: Throwable?
-    ) {
+internal class ThreadState @JvmOverloads constructor
+    (
+    config: ImmutableConfig,
+    excStacktrace: Array<StackTraceElement>,
+    currentThread: java.lang.Thread = java.lang.Thread.currentThread(),
+    stackTraces: MutableMap<java.lang.Thread, Array<StackTraceElement>> = java.lang.Thread.getAllStackTraces()
```
tomlongridge: Does this need to be a parameter? Can't it just be a local in init?

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -5,37 +5,29 @@ import java.io.IOException
...
-        config: ImmutableConfig,
-        currentThread: java.lang.Thread,
-        stackTraces: MutableMap<java.lang.Thread, Array<StackTraceElement>>,
-        exc: Throwable?
-    ) {
+internal class ThreadState @JvmOverloads constructor
+    (
+    config: ImmutableConfig,
+    excStacktrace: Array<StackTraceElement>,
+    currentThread: java.lang.Thread = java.lang.Thread.currentThread(),
```
tomlongridge: Does this need a default? When wouldn't we specify one?

fractalwrench: Having this as a parameter is useful primarily for testing - for example, we use it within `ThreadStateTest` to verify that a different thread can be set as the 'errorReportingThread': https://github.com/bugsnag/bugsnag-android/pull/668/files#diff-5957b89ae75ed52bfc6a969a2a5ed6adR43

Otherwise, the report generation always takes place on the thread an exception was delivered on, so the value of the `currentThread` parameter should always be equivalent to `Thread.currentThread()`

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
fractalwrench: I believe the intention behind this would be to make testing easier. However, as we're not currently using this any test code I can move this out of the constructor.

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
fractalwrench: I have reverted this change to use the original guard.

bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
```diff
@@ -419,7 +419,8 @@ void setProjectPackages(Collection<String> projectPackages) {
                 boolean unhandled,
                 Metadata globalMetadata) {
             Throwable exc = unhandled ? exception : null;
-            this.threadState = new ThreadState(config, thread, Thread.getAllStackTraces(), exc);
+            // unhandled errors use the exception trace for thread traces
```
tomlongridge: Comment no longer required here as you've moved the if-else clause back to the `ThreadState`.

PR #670, opened 2019-11-22
Add CallbackAware interface

fractalwrench:
## Goal

Adds a non-public interface which contains the methods for adding/removing callbacks.

According to the notifier spec these methods share the same signatures on both `Client` and `Configuration`, therefore we should ensure they retain consistency by implementing the same interface.


fractalwrench:
I have removed `addOnError` from the `BugsnagConfiguration` interface.

It's worth noting that `BugsnagConfiguration` will be eventually removed - for now we have to hold onto it for the integration branch to compile.

PR #672, opened 2019-11-25
Add Logger interface as property on Configuration

fractalwrench:
## Goal

Adds a `Logger` interface that a user can configure to alter the default logging in Bugsnag. By default, log messages are written to [Logcat](https://developer.android.com/studio/command-line/logcat) in development, and are disabled entirely in production apps.

This changeset removes the previous static logger and injects the logger as a dependency from `Configuration`, where a user can set their own implementation.

## Changeset

- Added a public `Logger` interface which matches the notifier spec, which allows users to supply their own Logger implementation
- Disabled logcat logging if the `releaseStage` matches `"production"`
- Added accessors to `logger` on `Configuration`, and removed accessors for `loggingEnabled`
- Added a `DebugLogger` implementation which logs to Logcat
- Added a `NoopLogger` implementation which logs to nothing
- Removed logging where verbose/not appropriate
- Updated test code to pass in a `NoopLogger` where necessary


bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -67,9 +67,9 @@ public static Client init(@NonNull Context androidContext, @NonNull Configuratio
     }
 
     private static void logClientInitWarning() {
-        Logger.warn("It appears that Bugsnag.init() was called more than once. Subsequent "
-            + "calls have no effect, but may indicate that Bugsnag is not integrated in an"
-            + " Application subclass, which can lead to undesired behaviour.");
+        DebugLogger.INSTANCE.w("It appears that Bugsnag.init() was called more than once. "
```
bengourley: is `DebugLogger.INSTANCE.w` a sneaky way of calling the logger even when logging is disabled?

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -31,7 +31,6 @@
     private static final int METADATA_SECTION = 0;
     private static final int METADATA_KEY = 1;
     private static final int METADATA_VALUE = 2;
-    private static final String LOG_TAG = "BugsnagNDK:NativeBridge";
```
bengourley: Worth noting for the changelog that the log tag has changed for NDK logs? Was this intentional?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -91,7 +91,33 @@ public Configuration(@NonNull String apiKey) {
             autoDetectNdkCrashes = false;
         }
 
-        loggingEnabled = !AppData.RELEASE_STAGE_PRODUCTION.equals(releaseStage);
+        boolean loggingEnabled = !AppData.RELEASE_STAGE_PRODUCTION.equals(releaseStage);
+
+        if (loggingEnabled) {
```
bengourley: I think doing this logic in the constructor may be error-prone… what happens if a `Configuration` object is instantiated, the the developer calls `config.setReleaseStage(str)`. Does this get re-evalutated?

bugsnag-android-core/src/main/java/com/bugsnag/android/Exceptions.java
```diff
@@ -72,7 +72,7 @@ private void exceptionToStream(@NonNull JsonStream writer,
         writer.name("message").value(message);
         writer.name("type").value(exceptionType);
 
-        Stacktrace stacktrace = new Stacktrace(frames, projectPackages);
+        Stacktrace stacktrace = new Stacktrace(frames, projectPackages, NoopLogger.INSTANCE);
```
bengourley: why does a stacktrace which is created here get a noop logger, but a stacktrace for a thread gets an actual logger?

bugsnag-android-core/src/main/java/com/bugsnag/android/Logger.kt
```diff
@@ -0,0 +1,12 @@
+package com.bugsnag.android
+
+interface Logger {
+    fun e(msg: String) = Unit
```
bengourley: I noticed that the logger that got removed had fully spelled method names (`debug`, `info` etc.) rather than these single-char abbreviations. Is this definitely the prevailing convention? Would this cause friction for people or is there a way to allow both?

bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
fractalwrench: Yep - this will always be logged. It feels like we would always want to inform the user of this as it would indicate a potential bug in how they have integrated bugsnag.

bugsnag-android-core/src/main/java/com/bugsnag/android/Exceptions.java
fractalwrench: The `Exceptions` class will be deleted in #673 - so I've just passed in a `NoopLogger` here as its the path of least resistance.

bugsnag-android-core/src/main/java/com/bugsnag/android/Logger.kt
fractalwrench: This is more consistent with the [default Android logger](https://developer.android.com/reference/android/util/Log), although I think it'd be conventional to name things as either `e()` or `error()`.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
fractalwrench: I don't think we have ever documented the log tag before. Having the same log tag should make it somewhat easier to search for all the 'bugsnag' logs - or at least that was the reasoning behind this change.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: If the developer calls `config.setReleaseStage()` the `logger` implementation does not currently change.

I think it would be best to avoid these sorts of side-effects in setters, but this may be something we want to discuss more broadly if it's not already mentioned in the spec.

bengourley: Yeah I think doing it as a side effect in setters would be a complex and error-prone way to implement it.

I think it's important to fix this though – I'd consider it a bug, but at best it's difficult to explain: "in production we will switch off logging, except if you set releaseStage=production in a particular way"

I would use the following steps to achieve consistent logic, and I'd place in the client constructor:

```
if config.logger is set by the developer
  this.logger = config.logger
else
  if config.releaseStage is 'production'
    this.logger = NoopLogger
  else
    this.logger = DebugLogger

File: bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java (hunk 2 above)
Comment (bengourley) [reply bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java]: Yeah even though we haven't documented it before, it's possible people may have used it.
I agree that the change is a positive one, I'm just saying we should ensure it goes in the changelog/upgrade guide.

File: bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java (hunk 1 above)
Comment (bengourley) [reply bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java]: I get that, but what if a user has passed in a custom logger and for some reason they don't see things that go via the Android logger? Seems strange to accept a logger as a configuration option and then bypass it.

Comment (fractalwrench) [reply bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java]: I have updated this to use the logger on configuration.

File: bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java (hunk 2 above)
Comment (fractalwrench) [reply bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java]: Added a note in the upgrading section

File: bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java (hunk 3 above)
Comment (fractalwrench) [reply bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java]: I've updated this so that if the `releaseStage` has been set to 'production' after `Configuration` has been constructed, the `Client` will use a `DebugLogger` if `config.logger` hasn't been overridden by the user.

**Comment by fractalwrench** [comment]
Ok, think I understand now - I've updated the setting of the default logger to take place in the `Client` constructor and rebased.

File: bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java [hunk 6]
```diff
@@ -109,9 +114,24 @@ public Client(@NonNull Context androidContext, @NonNull String apiKey) {
      * @param configuration  a configuration for the Client
      */
     public Client(@NonNull Context androidContext, @NonNull final Configuration configuration) {
+        // if the user has set the releaseStage to production manually, disable logging
+        if (configuration.getLogger() == null) {
+            String releaseStage = configuration.getReleaseStage();
+            boolean loggingEnabled = !AppData.RELEASE_STAGE_PRODUCTION.equals(releaseStage);
+
+            if (loggingEnabled) {
+                configuration.setLogger(DebugLogger.INSTANCE);
```
bengourley: Why do you mutate configuration here? I guess it's not a deal breaker for this PR but it just doesn't make sense in my head to make changes after it has been supplied.

fractalwrench: The object is converted to an object of type ImmutableConfiguration, and setting a logger on the original object was the simplest way to pass existing sanitisation code.

PR #674, opened 2019-11-25
Update Configuration interface to match notifier spec

fractalwrench:
## Goal

Updates `Configuration` so that its interface matches the notifier spec, and alters code which relies on `Configuration` to use the new APIs.

## Changeset

- Converted existing class to Kotlin
- Removed `Observable` inheritance hierarchy, as configuration is now considered immutable after initialisation therefore we do not want to observe any changes
- Updated fields which used `Collection` to use the more specific `Set` instead
- Removed `autoCaptureBreadcrumbs` in favour of `enabledBreadcrumbTypes`
- Removed `BugsnagConfiguration`, as this interface is no longer in the notifier spec and not required
- Made `OnSession` publicly visible
- Removed redundant checks for whether automatic breadcrumbs are enabled when leaving an internal breadcrumb
- Updated mazerunner scenarios to use property access syntax rather than Java setters (`setAutoTrackSessions(false)` -> `autoTrackSessions = false`)

### Tests
- Removed `ConfigCollectionMutabilityTest` as the types used now guarantee immutability
- Removed references to removed fields and updated existing tests to use new API


bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -96,7 +95,7 @@ internal fun convertToImmutableConfig(config: Configuration): ImmutableConfig {
         versionCode = config.versionCode!!,
         codeBundleId = config.codeBundleId,
         appType = config.appType,
-        delivery = config.delivery,
+        delivery = config.delivery!!,
         endpoints = config.endpoints,
```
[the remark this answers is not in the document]
fractalwrench: `DefaultDelivery` is the implementation used by default, and it relies on an Android `Context` object, which is only supplied after the `Client` has been initialised. We therefore allow a nullable `Delivery` in `Configuration` and sanitise the object in the `Client` constructor: https://github.com/bugsnag/bugsnag-android/blob/configuration-interface/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java#L260

`Endpoints` does not require a `Context` so we enforce that is should be non-null.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/NullMetadataTest.java
```diff
@@ -45,13 +45,6 @@ public void tearDown() {
         client.close();
     }
 
-    @Test
-    public void testConfigSetNullMetadata() throws Exception {
```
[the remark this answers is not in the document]
fractalwrench: Yep - we no longer allow setting a `Metadata` object directly on `Configuration`, in favour of `addMetadata/getMetadata/clearMetadata` methods that mutate a non-public field.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
```diff
@@ -0,0 +1,330 @@
...
+
+        loggingEnabled = AppData.RELEASE_STAGE_PRODUCTION != releaseStage
+    }
+
+    /**
+     * Set which exception classes should be ignored (not sent) by Bugsnag.
+     *
+     * For example:
+     *
+     * client.setIgnoreClasses("java.lang.RuntimeException");
```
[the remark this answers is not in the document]
fractalwrench: Probably not! I'll update this to some other class.

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationTest.java
```diff
@@ -76,48 +77,84 @@ public void testLaunchThreshold() throws Exception {
...
+    public void testAddOnBreadcrumb() {
+        OnBreadcrumb cb = new OnBreadcrumb() {
+            @Override
+            public boolean run(@NonNull Breadcrumb breadcrumb) {
+                return false;
+            }
+        };
+        config.addOnBreadcrumb(cb);
+        config.removeOnBreadcrumb(cb);
+    }
```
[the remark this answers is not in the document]
fractalwrench: I've removed these tests as they are not doing anything as-is.

simonbowring:
@fractalwrench, are the test failures going to be addressed in this PR?

fractalwrench:
@simonbowring the `filtering_metadata` scenario has been fixed in #684 as it's a pre-existing issue.

I have fixed the `breadcrumb` scenario in this PR, which was failing due to `config.enabledBreadcrumbTypes` erroneously being set as an empty set - preventing the expected breadcrumbs from being sent.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
```diff
@@ -0,0 +1,330 @@
...
+     */
+    var autoDetectErrors = true
+
+    /**
+     * Intended for internal use only - sets the code bundle id for React Native
+     */
+    var codeBundleId: String? = null
+
+    /**
+     * Intended for internal use only - sets the type of the notifier (e.g. Android, React Native)
```
bengourley: `appType` can be used externally too

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
```diff
@@ -0,0 +1,330 @@
...
+     * Use this when you want to ensure sensitive information, such as passwords
+     * or credit card information is stripped from metadata you send to Bugsnag.
+     * Any keys in metadata which contain these strings will be marked as
+     * REDACTED when send to Bugsnag.
+     *
+     * For example:
+     *
+     * client.setRedactKeys("password", "credit_card");
+     */
+    var redactKeys: Set<String>
```
bengourley: this should be `redactedKeys`

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
```diff
@@ -0,0 +1,330 @@
...
+        get() = Collections.unmodifiableSet(metadata.redactKeys)
+        set(redactKeys) = metadata.setRedactKeys(redactKeys)
+
+
+    var loggingEnabled: Boolean
```
bengourley: I suspect this should this go away after the logger changes have been combined?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
```diff
@@ -0,0 +1,330 @@
...
+    /**
+     * Removes an on session callback
+     *
+     * @param onSession the on session callback
+     */
+    override fun removeOnSession(onSession: OnSession) {
+        sessionCallbacks.remove(onSession)
+    }
+
+    override fun addMetadata(section: String, value: Any?) =
```
bengourley: I think the type of `value` can be more specific here – it needs to be a `Map<String, any>`

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
fractalwrench: This is correct

fractalwrench:
Thanks for the review!

>I didn't see any logic in the diff depending on the value of enabledBreadcrumbTypes (i.e only storing a breacrumb if its type is in the list of enabled types). Did I miss something or is this in another PR?

This is implemented by adding an `OnBreadcrumb` callback within the `Client`: https://github.com/bugsnag/bugsnag-android/blob/31ccd4e7e8c6545719f67270c5bcc5b057a015db/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java#L229

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
fractalwrench: good catch - I've updated the implementation to use `Map<String, Any?>`

PR #680, opened 2019-11-26
Update NativeBridge implementation to use StateEvent

fractalwrench:
## Goal

Updates the `NativeBridge` class to respond to `StateEvent` messages instead of `NativeInterface.Message` `StateEvent` was added in #659 and is intended to replace the previous implementation. This changeset begins that replacement, starting with the `Client` class.

## Changeset

- Created a `ClientObservable` class, which extracts the responsibility for sending state change events from the `Client`, such as when the orientation changes
- Updated `ClientObservable` to use `StateEvent` rather than `NativeInterface.Message`
- Updated the `NativeBridge` implementation to use `StateEvent`. (It's worth noting that not everywhere in the Client has been updated to emit StateEvent messages, so this will fail some mazerunner scenarios)


bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -0,0 +1,164 @@
...
+    private val is32bit: Boolean
+        get() {
+            val abis = NativeInterface.getCpuAbi()
+            return !abis.toList().any { it.contains("64") }
+        }
+
+    external fun install(
+        reportingDirectory: String,
+        autoDetectNdkCrashes: Boolean,
+        apiLevel: Int, is32bit: Boolean,
```
bengourley: style nit: should `is32bit` be on a new line?

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -0,0 +1,164 @@
...
+            warn("The native reporting directory cannot be created.")
+        }
+    }
+
+    override fun update(observable: Observable, arg: Any?) {
+        if (isInvalidMessage(arg)) return
+
+        when (val msg = arg as StateEvent) {
+            is Install -> handleInstallMessage(msg)
+            DeliverPending -> deliverPendingReports()
```
bengourley: what is the difference between the entries in this pattern match that have `is` and those that don't?

bengourley:
Tagging @simonbowring for additional implementation review on this one 🙇

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
fractalwrench: `is Install` checks whether `msg` is of type `Install`, whereas `DeliverPending` is a singleton so we only need to perform an equality check.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
fractalwrench: Yep, it should. I'll update this after Si's review.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -0,0 +1,164 @@
...
+            lock.unlock()
+        }
+    }
+
+    private fun handleInstallMessage(arg: Install) {
+        lock.lock()
+        try {
+            if (installed.get()) {
+                warn("Received duplicate setup message with arg: $arg")
+            }
```
tomlongridge: Should we proceed in the `else` case? Previously we just returned.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -0,0 +1,164 @@
...
+                is String -> addMetadataString(arg.section, arg.key!!, newValue)
+                is Boolean -> addMetadataBoolean(arg.section, arg.key!!, newValue)
+                is Number -> addMetadataDouble(arg.section, arg.key!!, newValue.toDouble())
+                else -> Unit
+            }
+        }
+    }
+
+    private fun warn(msg: String) {
+        Log.d("Bugsnag", msg)
```
tomlongridge: Shouldn't this be `Log.w`? Also, it should use the configured logger rather than Android's directly.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -0,0 +1,164 @@
...
+    private fun isInvalidMessage(msg: Any?): Boolean {
+        if (msg !is StateEvent) {
+            return true
+        }
+        if (!installed.get() && msg !is Install) {
+            warn("Received message before INSTALL: $msg")
+            return true
+        }
+
+        warn(String.format("Received NDK message %s", msg))
```
tomlongridge: Not a warning.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
```diff
@@ -1,460 +0,0 @@
...
-    public static native void clearMetadataTab(@NonNull String tab);
-
-    public static native void removeMetadata(@NonNull String tab,@NonNull  String key);
-
-    public static native void startedSession(@NonNull String sessionID, @NonNull String key,
-                                             int handledCount, int unhandledCount);
-
-    public static native void pausedSession();
-
-    public static native void updateAppVersion(@NonNull String appVersion);
```
tomlongridge: Is this now completely unused? (Along with others that aren't in the refactor?) Should we remove the `MessageType` in the native interface?

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -0,0 +1,164 @@
...
+            NotifyUnhandled -> addUnhandledEvent()
+            PauseSession -> pausedSession()
+            is StartSession -> startedSession(msg.id, msg.startedAt, msg.handledCount, msg.unhandledCount)
+            is UpdateContext -> updateContext(msg.context ?: "")
+            is UpdateInForeground -> updateInForeground(msg.inForeground, msg.contextActivity ?: "")
+            is UpdateOrientation -> updateOrientation(msg.orientation)
+            is UpdateUserEmail -> updateUserEmail(msg.email ?: "")
+            is UpdateUserName -> updateUserName(msg.name ?: "")
+            is UpdateUserId -> updateUserId(msg.id ?: "")
+        }
```
tomlongridge: Do we need any sort of warning if we reach the default case? We didn't before, but it seemed like a simpler mapping between the `MessageType` enum values and the switch statement.

fractalwrench: A default case is not required in Kotlin when using a sealed class in a `when` expression, as the compiler will verify that all cases are covered: https://kotlinlang.org/docs/reference/sealed-classes.html

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
fractalwrench: It's possible to use the configured logger now that the logger PR has landed, so I'll update the changeset to use that.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
fractalwrench: changed to Info level

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.java
fractalwrench: Yep this is unused - I've updated `MessageType` to remove messages which aren't currently used, and the rest will be removed in separate PRs which replaces them with `StateEvent`.

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
fractalwrench: I believe that the current implementation makes sense in that it will perform no operation if the message is not `StateEvent` and the plugin has not been installed, unless I'm missing an edge case?

fractalwrench:
Note: CI will fail until #680 is merged

fractalwrench:
@tomlongridge after running mazerunner locally I believe the `native_api` and `native_event_tracking` features will fail as they verify that the NDK has received messages of state changes. I plan to address this individually in separate PRs against the integration branch as otherwise the diff may become fairly large to review.

PR #683, opened 2019-11-27
Update device data serialization test to cover more fields

fractalwrench:
## Goal

Consolidates the tests for `DeviceData` so that they exist only in `DeviceDataSerialisationTest`. This was suggested for `AppData` in this PR: https://github.com/bugsnag/bugsnag-android/pull/661#discussion_r350203740

## Changeset

- Removed existing `DeviceData` tests
- Updated `DeviceDataSerialization` to ensure all fields are serialized (where appropriate/possible without drastically altering existing code)


bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -61,9 +61,10 @@
 
     @NonNull
     final String[] cpuAbi;
+    private final File dataDirectory;
 
     DeviceData(Connectivity connectivity, Context appContext, Resources resources,
-               String installId, DeviceBuildInfo buildInfo) {
+               String installId, DeviceBuildInfo buildInfo, File dataDirectory) {
```
tomlongridge: I'm not sure I understand the rationale for this change - it makes it different to the other calculations for memory/orientation, so doesn't seem ideal.

bugsnag-android-core/src/test/java/com/bugsnag/android/DeviceDataSerializationTest.kt
```diff
@@ -17,19 +21,49 @@ internal class DeviceDataSerializationTest {
...
+
+            // construct devicedata object
+            val deviceData = DeviceData(connectivity, context, res, "123", buildInfo, File(""))
+
+            // serializes the 3 different maps that DeviceData can generate:
+            // 1. summary (used in session payloads)
+            // 2. regular (used in event payloads)
+            // 3. metadata (used in event payloads)
+            val regularMap = deviceData.deviceData
+            arrayOf("freeDisk", "freeMemory", "totalMemory", "locale", "time").forEach { regularMap.remove(it) }
```
tomlongridge: Why are these removed from the test? Can't we mock the data providers?

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/DeviceDataSummaryTest.java
```diff
@@ -1,40 +0,0 @@
...
-        this.deviceData = deviceData.getDeviceDataSummary();
-    }
-
-    @Test
-    public void testAccessors() {
-        assertNotNull(deviceData.get("manufacturer"));
-        assertNotNull(deviceData.get("model"));
-        assertNotNull(deviceData.get("osName"));
-        assertNotNull(deviceData.get("osVersion"));
-    }
```
tomlongridge: I was expecting to see these (and `DeviceDataTest`) tests moved to a unit test (as opposed to instrumentation) with assertions on each of the getters for each of the fields in DeviceData.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
fractalwrench: This allows for the static method `Environment.getDataDirectory()` to be mocked in tests, which lets us test the value of `device.freeDisk`. If we left it as-is it would also throw a NPE when run in a JVM test as this is an Android-specific test.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/DeviceDataSummaryTest.java
fractalwrench: `deviceData` in this test is a `Map`, so adding similar assertions would just be testing `Map#get()`. I have therefore avoiding doing this as we already check the values of each field by verifying the JSON matches the test fixture.

bugsnag-android-core/src/test/java/com/bugsnag/android/DeviceDataSerializationTest.kt
fractalwrench: freeMemory/totalMemory cannot be mocked in their current form as they rely on the `Runtime` class directly. We could add a wrapper class that decorates all calls to `Runtime` which would be mocked if this is desirable.

`freeDisk` requires a mock `File`, which could be added.

`locale` requires a mock/fake `Locale`, which would require adding another constructor parameter.

`time` requires passing in a `Date` as a constructor parameter.

If you're happy with making these alterations to the class then I can update the test accordingly.

tomlongridge: OK - let's address the existing testability constraints separately at the end of the refactor.

However, in this case I don't see how the changes to pass in dataDirectory have gained us anything as this isn't being mocked in the test anyway. Unless I'm missing something, I believe this would be the same regardless of whether the directory was set internally or externally.

Suggest we remove this for now, or if we are keeping the parameter to the constructor then I don't see a need to have the parameter to `calculateFreeDisk` as it can use the member variable (otherwise it's effectively a rather pointless static function).

fractalwrench: `Environment.getDataDirectory()` returns null when run in a JVM test, which causes a NPE that fails the whole test case. I'd therefore suggest one additional change to this PR of mocking the parameter and then address whether we need to use PowerMock/other mocking frameworks at the end of the refactor, if you're happy with that approach.

tomlongridge: Yes, as long as `calculateFreeDisk` is also corrected to use member variable, I'm happy with that.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceData.java
```diff
@@ -224,10 +226,9 @@ static long calculateTotalMemory() {
      * Get the usable disk space on internal storage's data directory
      */
     @SuppressLint("UsableSpace")
-    long calculateFreeDisk() {
+    long calculateFreeDisk(File dataDirectory) {
```
tomlongridge: As discussed - this should use the member variable not a parameter.

fractalwrench: Updated to use the member variable.

PR #685, opened 2019-12-03
feat: (internal) Support initializing custom plugins before init

kattrali:
Fixes an issue where a plugin cannot be configured until _after_ Bugsnag
is started, potentially sending reports prior to a plugin being able to
modify them.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagPluginInterface.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagPluginInterface.kt
@@ -15,7 +15,8 @@ object BugsnagPluginInterface {
         registeredPluginClasses.add(clz)
     }
 
-    fun loadRegisteredPlugins(client: Client) {
+    @JvmName("loadRegisteredPlugins")
+    internal fun loadRegisteredPlugins(client: Client) {
         registeredPluginClasses.forEach { loadPlugin(client, it) }
     }
 
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagPluginInterface.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagPluginInterface.kt
@@ -9,6 +9,15 @@ interface BugsnagPlugin {
 object BugsnagPluginInterface {
 
     private var plugins = mutableMapOf<Class<*>, BugsnagPlugin>()
+    private var registeredPluginClasses = mutableSetOf<Class<*>>()
+
+    fun registerPlugin(clz: Class<*>) {
+        registeredPluginClasses.add(clz)
+    }
+
+    fun loadRegisteredPlugins(client: Client) {
+        registeredPluginClasses.forEach { loadPlugin(client, it) }
+    }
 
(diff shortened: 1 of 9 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagPluginInterface.kt
```diff
@@ -9,6 +9,15 @@ interface BugsnagPlugin {
 object BugsnagPluginInterface {
 
     private var plugins = mutableMapOf<Class<*>, BugsnagPlugin>()
+    private var registeredPluginClasses = mutableSetOf<Class<*>>()
+
+    fun registerPlugin(clz: Class<*>) {
```
fractalwrench: These methods should either have `protected` access visibility or `internal` with the `@JvmName` annotation, as this will avoid exposing them as part of our public API.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -318,6 +318,7 @@ private void loadPlugins() {
         NativeInterface.setClient(this);
         enableOrDisableNdkReporting();
         enableOrDisableAnrReporting();
+        BugsnagPluginInterface.INSTANCE.loadRegisteredPlugins(this);
```
fractalwrench: This is missing an E2E test to verify that the functionality works.

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagPluginInterface.kt
kattrali: Is this class public? Are things public by default?

fractalwrench: Yes, in Kotlin everything is public by default. Other than that, the modifiers are reasonably similar to Java but have some subtle differences: https://kotlinlang.org/docs/reference/visibility-modifiers.html

kattrali: I see. So when this function was removed in #603, it broke a public interface. 😞 I did not realize that at the time.

kattrali: I've made an update to protect `loadRegisteredPlugins` as it should only be called from within the same module. However, I've left `registerPlugin` as-is because calling it from a class which was not compiled as a part of the same source set means its not in the same module:

> The internal visibility modifier means that the member is visible within the same module. More specifically, a module is a set of Kotlin files compiled together:
> * an IntelliJ IDEA module;
> * a Maven project;
> * a Gradle source set (with the exception that the test source set can access the internal declarations of main);
> * a set of files compiled with one invocation of the <kotlinc> Ant task.

There doesn't seem to be a package access modifier or equivalent.

fractalwrench: Yep - this is the right approach as we need to expose `registerPlugin` to other modules. The only alternative I'm aware of would be reflection

PR #688, opened 2019-12-03
Add CallbackState class to encapsulate callbacks

fractalwrench:
## Goal

Extracts the logic for adding and running a callback to a single class named `CallbackState`, rather than spreading this code across `Configuration` and `Client`.

## Changeset

- Updated `Configuration` and `Client` to forward method invocations to a `CallbackState` field
- Added `CallbackState` which contains a collection of all callbacks, checks whether a duplicate callback has been added, and provides methods for running callbacks on an object
- Updated `Client` and `SessionTracker` to run callbacks via `CallbackState`

## Tests

- Updated `ConcurrentCallbackTest` to run on JVM and added cases for breadcrumb/session
- Added new `CallbackStateTest`

Note: CI will fail until #680 and #686 are merged


fractalwrench:
The spec as currently written allows adding/removing of callbacks on both `Configuration/Client` therefore the methods have been extracted to a common interface named `CallbackAware`.  `CallbackState` is an internal implementation detail that we could change at a later date if needed - however, it felt simplest to effectively forward the method invocation on from either Client/Configuration.

PR #689, opened 2019-12-03
Create ContextState class to encapsulate context

fractalwrench:
## Goal

Creates a class named `ContextState` which holds the current context and notifies observers. This passes a failing mazerunner scenario that checks the NDK is informed of any update to context.

## Changeset

- Created `ContextState` class
- Updated change in context to emit a `StateEvent` object as an observable, which passes the 
 failing `CXXUpdateContextCrashScenario`
- Deep copy context from `Configuration` to `Client`
- Removed obsolete `UPDATE_CONTEXT` from `NativeInterface`

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientConfigTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientConfigTest.java
@@ -1,6 +1,7 @@
 package com.bugsnag.android;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 import android.content.Context;
 
@@ -38,7 +39,7 @@ public class ClientConfigTest {
     public void testSetContext() throws Exception {
         client.setContext("JunitTest");
         assertEquals("JunitTest", client.getContext());
-        assertEquals("JunitTest", config.getContext());
+        assertNull(config.getContext());
     }
 
     @Test
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
@@ -18,9 +18,6 @@ class Configuration(
 
     var metadata = Metadata()
 
-    @JvmField
-    internal val contextState: ContextState
-
     protected val onErrorTasks = ConcurrentLinkedQueue<OnError>()
     protected val breadcrumbCallbacks = ConcurrentLinkedQueue<OnBreadcrumb>()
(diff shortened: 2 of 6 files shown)


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientConfigTest.java
```diff
@@ -38,7 +39,7 @@ public void tearDown() {
     public void testSetContext() throws Exception {
         client.setContext("JunitTest");
         assertEquals("JunitTest", client.getContext());
-        assertEquals("JunitTest", config.getContext());
+        assertNull(config.getContext());
```
tomlongridge: I don't understand why this test how this test even passed before! Isn't the purpose of this test script to ensure that values passed to config are passed to client? Therefore shouldn't we be setting it on config and asserting on client (like the delivery test below)?

tomlongridge:
Also - can you state in the PR description which tests you expect to fail and why.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientConfigTest.java
fractalwrench: The test passed previously because the value of `context` on `Client was backed by a field on `Configuration`. That's not the case anymore as we take a deep-copy of `Configuration`, so the field can be asserted as null here.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.kt
```diff
@@ -164,7 +168,11 @@ class Configuration(
      * name of the top-most activity at the time of a report, and use this
      * as the context, but sometime this is not possible.
      */
-    var context: String? = null
+    var context: String?
+        get() = contextState.context
+        set(context) {
+            contextState.context = context
+        }
```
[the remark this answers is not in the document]
fractalwrench: I've updated this to just use a `String` on `Configuration`.

fractalwrench:
I've made an additional change to use `contextState` rather than `clientState` which was failing an `error_context` mazerunner scenario. The remaining scenarios are failing due to the integration branch not fully merging all the PRs which update the NDK messaging to use `StateEvent`.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientConfigTest.java
tomlongridge: OK, but would a more useful test not be to verify the other way round? However this is not just for context, but for the whole client:config relationship, so I think we could do with an overall review of unit test structure at the end of this refactor.

PR #690, opened 2019-12-03
Make AddBreadcrumb use StateEvent observable message

fractalwrench:
## Goal

Updates the `BreadcrumbState` class to use `StateEvent` observable messages rather than `NativeInterface`. This is required to pass the failing breadcrumb scenarios in the `native_api` mazerunner feature.


fractalwrench:
The JVM test failure appears to be a flake after retrying the job.

Some mazerunner scenarios (native_api, native_event_tracking, native_session_tracking) are expected to fail as the integration branch has not been updated to use the new `StateEvent` message which is required to keep the NDK layer updated with JVM state. These will be fixed in separate PRs which are waiting to be reviewed + merged.

PR #691, opened 2019-12-03
Make SessionTracker use StateEvent messages

fractalwrench:
## Goal

Updates the `SessionTracker` class to use `StateEvent` observable messages rather than `NativeInterface`. This is required to pass failing scenarios in the `native_session_tracking` mazerunner feature.

## Testing

The remaining scenarios are failing due to the integration branch not fully merging all the PRs which update the NDK messaging to use StateEvent.

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
@@ -113,23 +113,20 @@ public class ObserverInterfaceTest {
     }
 
     @Test
-    public void testStartSessionSendsMessage() throws InterruptedException {
+    public void testStartSessionSendsMessage() {
         client.startSession();
-        List<Object> sessionInfo = (List<Object>)findMessageInQueue(
-                NativeInterface.MessageType.START_SESSION, List.class);
-        assertEquals(4, sessionInfo.size());
-        assertTrue(sessionInfo.get(0) instanceof String);
-        assertTrue(sessionInfo.get(1) instanceof String);
-        assertTrue(sessionInfo.get(2) instanceof Integer);
-        assertTrue(sessionInfo.get(3) instanceof Integer);
+        StateEvent.StartSession sessionInfo = findMessageInQueue(StateEvent.StartSession.class);
+        assertNotNull(sessionInfo.getId());
+        assertNotNull(sessionInfo.getStartedAt());
+        assertEquals(0, sessionInfo.getHandledCount());
+        assertEquals(0, sessionInfo.getUnhandledCount());
     }
 
     @Test
-    public void testPauseSessionSendsmessage() {
+    public void testPauseSessionSendsMessage() {
         client.startSession();
         client.pauseSession();
-        Object msg = findMessageInQueue(NativeInterface.MessageType.PAUSE_SESSION, null);
-        assertNull(msg);
+        assertNotNull(findMessageInQueue(StateEvent.PauseSession.class));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -148,9 +143,7 @@ private void notifySessionStartObserver(Session session) {
             session = new Session(sessionId, date, user, unhandledCount, handledCount);
             notifySessionStartObserver(session);
         } else {
-            setChanged();
-            notifyObservers(new NativeInterface.Message(
-                NativeInterface.MessageType.PAUSE_SESSION, null));
+            notifyObservers(StateEvent.PauseSession.INSTANCE);
```
tomlongridge: Is register session still used?

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
```diff
@@ -114,23 +115,20 @@ public void testNotifySendsMessage() {
...
 
     @Test
-    public void testPauseSessionSendsmessage() {
+    public void testPauseSessionSendsMessage() {
         client.startSession();
         client.pauseSession();
-        Object msg = findMessageInQueue(NativeInterface.MessageType.PAUSE_SESSION, null);
-        assertNull(msg);
+        assertNotNull(findMessageInQueue(StateEvent.PauseSession.class));
     }
```
tomlongridge: We should add a test for `SessionTracker.registerSession` sending a message.

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -24,7 +24,7 @@
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
-class SessionTracker extends Observable implements Application.ActivityLifecycleCallbacks {
+class SessionTracker extends BaseObservable implements Application.ActivityLifecycleCallbacks {
```
tomlongridge: There are some unused imports.

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -121,12 +119,9 @@ boolean resumeSession() {
     }
 
     private void notifySessionStartObserver(Session session) {
-        setChanged();
         String startedAt = DateUtils.toIso8601(session.getStartedAt());
```
tomlongridge: Can this date conversion be handled in the NativeBridge so that each caller doesn't have to do it?

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
fractalwrench: It's currently used by React Native and Unity - that may change after the latest alterations.

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
fractalwrench: This would mean we would need to duplicate `DateUtils` or make it public, as the class is not accessible from `NativeBridge`. As there are currently only two calls which serialize a `Date` I felt this was not worth the duplication.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
fractalwrench: added a test for this case

PR #693, opened 2019-12-04
Update Metadata to use StateEvent observable message

fractalwrench:
## Goal

Ensures that `Metadata` changes are synced to the NDK layer by using the `StateEvent` message rather than the obsolete `NativeInterface.Message` object.

## Changeset

- Added `MetadataState` which now sends `Observable` messages when the `Metadata` changes
- Updated `addMetadata()` so that the `key` is not null when it is an explicit parameter
- Changed return type of one `getMetadata()` to a more specific `Map` type
- Held `metadata` in a `MetadataState` field within `Client` that is copied from `Configuration`

## Tests
- Updated existing test code to use new `addMetadata()` and `getMetadata()` signatures
- Updated existing test code to check for presence of `StateEvent`
- Added unit test for `MetadataState`
- Updated `MetadataSerializationTest` test fixture to fix incorrect assertion

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
@@ -50,59 +50,42 @@ public class ObserverInterfaceTest {
     @Test
     public void testAddMetadataToClientSendsMessage() {
         client.addMetadata("foo", "bar", "baz");
-        List<Object> metadataItem = (List<Object>)findMessageInQueue(
-                NativeInterface.MessageType.ADD_METADATA, List.class);
-        assertEquals(3, metadataItem.size());
-        assertEquals("foo", metadataItem.get(0));
-        assertEquals("bar", metadataItem.get(1));
-        assertEquals("baz", metadataItem.get(2));
+        StateEvent.AddMetadata msg = findMessageInQueue(StateEvent.AddMetadata.class);
+        assertEquals("foo", msg.getSection());
+        assertEquals("bar", msg.getKey());
+        assertEquals("baz", msg.getValue());
     }
 
     @Test
     public void testAddNullMetadataToClientSendsMessage() {
         client.addMetadata("foo", "bar", null);
-        List<Object> metadataItem = (List<Object>)findMessageInQueue(
-                NativeInterface.MessageType.REMOVE_METADATA, List.class);
-        assertEquals(2, metadataItem.size());
-        assertEquals("foo", metadataItem.get(0));
-        assertEquals("bar", metadataItem.get(1));
+        StateEvent.RemoveMetadata msg = findMessageInQueue(StateEvent.RemoveMetadata.class);
+        assertEquals("foo", msg.getSection());
+        assertEquals("bar", msg.getKey());
     }
 
     @Test
     public void testAddMetadataToMetadataSendsMessage() {
         client.addMetadata("foo", "bar", "baz");
-        List<Object> metadataItem = (List<Object>)findMessageInQueue(
-                NativeInterface.MessageType.ADD_METADATA, List.class);
-        assertEquals(3, metadataItem.size());
-        assertEquals("foo", metadataItem.get(0));
-        assertEquals("bar", metadataItem.get(1));
-        assertEquals("baz", metadataItem.get(2));
+        StateEvent.AddMetadata metadataItem = findMessageInQueue(StateEvent.AddMetadata.class);
+        assertEquals("foo", metadataItem.getSection());
+        assertEquals("bar", metadataItem.getKey());
+        assertEquals("baz", metadataItem.getValue());
     }
 
     @Test
     public void testClearTabFromClientSendsMessage() {
-        client.clearMetadata("axis", null);
-        Object value = findMessageInQueue(
-                NativeInterface.MessageType.CLEAR_METADATA_TAB, String.class);
-        assertEquals("axis", value);
-    }
-
-    @Test
-    public void testClearTabFromMetadataSendsMessage() {
-        client.clearMetadata("axis", null);
-        Object value =  findMessageInQueue(
-                NativeInterface.MessageType.CLEAR_METADATA_TAB, String.class);
-        assertEquals("axis", value);
+        client.clearMetadata("axis");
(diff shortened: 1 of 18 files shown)


fractalwrench:
This PR fixes the remaining mazerunner scenarios, as it converts the remainder of the SDK to using `StateEvent` to communicate changes to the NDK rather than the obsolete `NativeInterface.Message`.

bugsnag-android-core/src/main/java/com/bugsnag/android/MetadataState.kt
```diff
@@ -0,0 +1,35 @@
...
+        when (value) {
+            null -> notifyClear(section, key)
+            else -> notifyObservers(StateEvent.AddMetadata(section, key, metadata.getMetadata(section, key)))
+        }
+    }
+
+    override fun clearMetadata(section: String) {
+        metadata.clearMetadata(section)
+        notifyClear(section, null)
+        notifyObservers(StateEvent.ClearMetadataTab(section))
```
tomlongridge: This appears to be a duplicate call to `notifyObservers` - it's already called through `notifyClear`.

fractalwrench: Good catch - I've removed the duplicate call.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
```diff
@@ -50,59 +50,42 @@ public void tearDown() {
...
-        assertEquals("bar", metadataItem.get(1));
-        assertEquals("baz", metadataItem.get(2));
+        StateEvent.AddMetadata msg = findMessageInQueue(StateEvent.AddMetadata.class);
+        assertEquals("foo", msg.getSection());
+        assertEquals("bar", msg.getKey());
+        assertEquals("baz", msg.getValue());
     }
 
     @Test
     public void testAddNullMetadataToClientSendsMessage() {
```
tomlongridge: This seems to be a duplicate of `testAddNullMetadataToMetadataSendsMessage`. We're now testing everything through `Client`, so it seems we can remove the `ToClient`, `FromClient` and `ToMetadata` part of the test names and remove this test.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
```diff
@@ -50,59 +50,42 @@ public void tearDown() {
...
-        assertEquals("bar", metadataItem.get(1));
-        assertEquals("baz", metadataItem.get(2));
+        StateEvent.AddMetadata metadataItem = findMessageInQueue(StateEvent.AddMetadata.class);
+        assertEquals("foo", metadataItem.getSection());
+        assertEquals("bar", metadataItem.getKey());
+        assertEquals("baz", metadataItem.getValue());
     }
 
     @Test
     public void testClearTabFromClientSendsMessage() {
```
tomlongridge: We should also have a test for clearing metadata with key.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -214,7 +214,11 @@ public static void leaveBreadcrumb(@NonNull String message,
      * Remove metadata from subsequent exception reports
      */
     public static void clearMetadata(@NonNull String section, @Nullable String key) {
-        getClient().clearMetadata(section, key);
+        if (key == null) {
+            getClient().clearMetadata(section);
+        } else {
+            getClient().clearMetadata(section, key);
+        }
```
tomlongridge: We seem to have lost something in the expressiveness of the API if we have to do this rather than rely on the method overrides, but I guess that's because we don't want an extra method in the native interface?

bugsnag-android-core/src/main/java/com/bugsnag/android/MetadataState.kt
```diff
@@ -0,0 +1,35 @@
...
+internal data class MetadataState(val metadata: Metadata = Metadata()) : BaseObservable(),
+    MetadataAware {
+
+    override fun addMetadata(section: String, value: Map<String, Any?>) = metadata.addMetadata(section, value)
+    override fun addMetadata(section: String, key: String, value: Any?) {
+        metadata.addMetadata(section, key, value)
+
+        when (value) {
+            null -> notifyClear(section, key)
+            else -> notifyObservers(StateEvent.AddMetadata(section, key, metadata.getMetadata(section, key)))
```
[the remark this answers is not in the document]
fractalwrench: Calling `metadata.getMetadata(section, key)` guarantees that the message will contain the actual value in the metadata store - for example, if the new value is merged with an existing map.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: We can add another method to the native interface if needed but I believe this would require updating downstream code in the NDK + Unity - supporting this with the existing method signature seemed the simplest option for now, and we could always alter the implementation at a later date.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
fractalwrench: I have removed this test, renamed the other test cases, and removed all duplicates.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ObserverInterfaceTest.java
fractalwrench: I have added a case for `testAddNullMetadataSendsMessage` which I believe is what this is asked for here

tomlongridge: We also need one for `clearMetadata` with a key to generate a `RemoveMetadata` event.

fractalwrench: Got it - I've added a separate test for this case.

PR #695, opened 2019-12-09
Remove unused code from public API

fractalwrench:
## Goal

Removes unused code and alters the public type hierarchy in miscellaneous files.

## Changeset

- Made `NetworkChangeCallback` non-public
- Removed `DateUtils` deserialization function as we no longer deserialize JSON
- Made `Session` final to prevent end-users extending the class
- Removed unused constants + methods from `SessionTracker`

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
@@ -1,7 +1,5 @@
 package com.bugsnag.android;
 
-import static com.bugsnag.android.MapUtils.getStringFromMap;
-
 import android.app.Activity;
 import android.app.Application;
 import android.os.Bundle;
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -392,7 +373,6 @@ void updateForegroundTracker(String activityName, boolean activityStarting, long
                 lastExitedForegroundMs.set(nowMs);
             }
         }
-        setChanged();
```
tomlongridge: What's the justification for this change?

fractalwrench: The call is redundant as `setChanged()` is called as a result of `notifyNdkInForeground()` being invoked - therefore this can be removed.

fractalwrench:
The `strict_mode` feature failure is addressed in #704 - all other scenarios are expected to pass.

PR #697, opened 2019-12-09
Make native interface use immutable configuration

fractalwrench:
## Goal

Updates `NativeInterface` to use the `ImmutableConfiguration` object rather than `Configuration` to prevent the user from mutating fields by holding onto a reference. This also renames methods for consistency with the Client's terminology.

## Changeset

- Removed non-public getter for `Configuration` on `Client`
- Added non-public getters/setters for `Client` fields as needed
- Removed `Message/MessageType` from `NativeInterface` as these are no longer used
- Renamed methods on `NativeInterface` to match terminology in rest of codebase
- Updated `NativeInterface` to use getters on `Client` rather than direct field access
- Updated JNI calls to use new method names
- Added a unit test to verify that `NativeInterface` calls the appropriate method on `Client`

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
@@ -18,59 +18,6 @@ import java.util.Queue;
  */
 public class NativeInterface {
 
-    public enum MessageType {
-        /**
-         * Add a new metadata value. The Message object should be an array
-         * containing [tab, key, value]
-         */
-        ADD_METADATA,
-        /**
-         * Clear all metadata on a tab. The Message object should be the tab
-         * name
-         */
-        CLEAR_METADATA_TAB,
-        /**
-         * Remove a metadata value. The Message object should be a string array
-         * containing [tab, key]
-         */
-        REMOVE_METADATA,
-        /**
-         * A new session was started. The Message object should be a string
-         * array
-         * containing [id, startDateIsoString]
-         */
-        START_SESSION,
-
-        /**
-         * A session was paused.
-         */
-        PAUSE_SESSION,
-
-        /**
-         * Set a new value for `app.inForeground`. The message object should be a
-         * List containing the values [inForeground (Boolean),
-         * foregroundActivityName (String)]
-         */
-        UPDATE_IN_FOREGROUND,
-    }
-
-    /**
-     * Wrapper for messages sent to native observers
-     */
-    public static class Message {
-
-        @NonNull
-        public final MessageType type;
-
-        @Nullable
-        public final Object value;
-
-        public Message(@NonNull MessageType type, @Nullable Object value) {
-            this.type = type;
-            this.value = value;
-        }
-    }
-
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -154,23 +95,23 @@ public static String getNativeReportPath() {
...
+        return getClient().getDeviceData().getCpuAbi();
     }
 
     /**
      * Retrieves global metadata from the static Client instance as a Map
      */
     @NonNull
     public static Map<String, Object> getMetadata() {
-        return new HashMap<>(getClient().metadataState.getMetadata().toMap());
+        return new HashMap<>(getClient().getMetadataState().getMetadata().toMap());
```
tomlongridge: Can this call `Client.getMetadata` rather than require a separate method to retrieve `MetadataState`?

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -154,23 +95,23 @@ public static String getNativeReportPath() {
...
     }
 
     /**
-     * Retrieves breadcrumbs from the static Client instance as a Map
+     * Retrieves breadcrumbState from the static Client instance as a Map
      */
     @NonNull
     public static List<Breadcrumb> getBreadcrumbs() {
-        Queue<Breadcrumb> store = getClient().breadcrumbState.getStore();
+        Queue<Breadcrumb> store = getClient().getBreadcrumbState().getStore();
```
tomlongridge: Can this call `Client.getBreadcrumbs` rather than require a separate method to retrieve `BreadcrumbState`?

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: Yep, I've updated this to use the existing method.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: Yep, I've updated this to use the existing method instead.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -154,24 +95,23 @@ public static String getNativeReportPath() {
...
      */
     @NonNull
     public static Map<String, Object> getMetadata() {
-        return new HashMap<>(getClient().metadataState.getMetadata().toMap());
+        return getClient().getMetadata();
     }
 
     /**
-     * Retrieves breadcrumbs from the static Client instance as a Map
+     * Retrieves breadcrumbState from the static Client instance as a Map
```
tomlongridge:
```suggestion
     * Retrieves a list of stored breadcrumbs from the static Client instance
```

PR #698, opened 2019-12-09
Extract InternalReportDelegate class from Client

fractalwrench:
## Goal

Extracts the logic for sending an internal bugsnag report from the `Client` to its own class.

There should be no functional change here, but this enables the addition of a unit test to verify the behaviour of internal reporting on top of existing E2E tests. It also reduces the responsibilities of the large `Client` class.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
@@ -32,7 +32,8 @@ internal data class ImmutableConfig(
 
     companion object {
         private const val HEADER_API_PAYLOAD_VERSION = "Bugsnag-Payload-Version"
-        private const val HEADER_API_KEY = "Bugsnag-Api-Key"
+        internal const val HEADER_API_KEY = "Bugsnag-Api-Key"
+        internal const val HEADER_INTERNAL_ERROR = "Bugsnag-Internal-Error"
         private const val HEADER_BUGSNAG_SENT_AT = "Bugsnag-Sent-At"
     }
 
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -32,7 +32,7 @@ internal data class ImmutableConfig(
 
     companion object {
         private const val HEADER_API_PAYLOAD_VERSION = "Bugsnag-Payload-Version"
-        private const val HEADER_API_KEY = "Bugsnag-Api-Key"
+        internal const val HEADER_API_KEY = "Bugsnag-Api-Key"
```
tomlongridge: I'm not sure why this change is required.

fractalwrench: `HEADER_API_KEY` is now accessed from the `InternalReportDelegate` rather than using a string constant. Because this is in another class it needs to be `internal` rather than `private` to be accessed.

fractalwrench: The usage of this constant within `InternalReportDelegate` seemed to have escaped the changeset - I've updated this to replace the string constant with `HEADER_API_KEY`

bugsnag-android-core/src/main/java/com/bugsnag/android/InternalReportDelegate.java
```diff
@@ -0,0 +1,134 @@
...
+                @Override
+                public void run() {
+                    try {
+                        Delivery delivery = immutableConfig.getDelivery();
+                        DeliveryParams params = immutableConfig.errorApiDeliveryParams();
+
+                        // can only modify headers if DefaultDelivery is in use
+                        if (delivery instanceof DefaultDelivery) {
+                            Map<String, String> headers = params.getHeaders();
+                            headers.put("Bugsnag-Internal-Error", "true");
```
tomlongridge: Can you extract this to a header in this class for consistency?

fractalwrench: Extracted to a header in `ImmutableConfig`.

PR #700, opened 2019-12-09
Extract Report Delivery logic from Client class

fractalwrench:
## Goal

Extracts the logic used to deliver reports from the `Client` to the `ReportDeliveryDelegate` class. This reduces the responsibilities of the `Client` and allows for the addition of a unit test - the code remains functionally equivalent.

Note that the `DeliveryStyle` enum has been deleted as part of this change. Instead, the `ReportDeliveryDelegate` will check the `unhandled` state of an `Event`, and write it to disk before attempting delivery - as this indicates the app is about to terminate.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
@@ -1,0 +1,109 @@
+package com.bugsnag.android;
+
+import androidx.annotation.NonNull;
+
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.RejectedExecutionException;
+
+class ReportDeliveryDelegate extends BaseObservable {
+
+    final Logger logger;
+    final EventStore eventStore;
+    final ImmutableConfig immutableConfig;
+    final BreadcrumbState breadcrumbState;
+
+    ReportDeliveryDelegate(Logger logger, EventStore eventStore,
+                           ImmutableConfig immutableConfig, BreadcrumbState breadcrumbState) {
+        this.logger = logger;
+        this.eventStore = eventStore;
+        this.immutableConfig = immutableConfig;
+        this.breadcrumbState = breadcrumbState;
+    }
+
+    void deliverEvent(@NonNull Event event) {
+        // Build the report
+        Report report = new Report(immutableConfig.getApiKey(), event);
+        Session session = event.getSession();
+
+        if (session != null) {
+            setChanged();
+
+            if (event.isUnhandled()) {
+                event.setSession(session.incrementUnhandledAndCopy());
+                notifyObservers(StateEvent.NotifyUnhandled.INSTANCE);
+            } else {
+                event.setSession(session.incrementHandledAndCopy());
+                notifyObservers(StateEvent.NotifyHandled.INSTANCE);
+            }
+        }
+
+        if (event.isUnhandled()) {
+            eventStore.write(event);
+            eventStore.flushAsync();
+        } else {
+            deliverReportAsync(event, report);
+        }
+    }
+
+    private void deliverReportAsync(@NonNull Event event, Report report) {
+        final Report finalReport = report;
+        final Event finalEvent = event;
+
+        // Attempt to send the report in the background
+        try {
+            Async.run(new Runnable() {
+                @Override
+                public void run() {
+                    deliver(finalReport, finalEvent);
+                }
+            });
+        } catch (RejectedExecutionException exception) {
+            eventStore.write(event);
+            logger.w("Exceeded max queue count, saving to disk to send later");
+        }
+    }
+
+    DeliveryStatus deliver(@NonNull Report report, @NonNull Event event) {
+        DeliveryParams deliveryParams = immutableConfig.errorApiDeliveryParams();
+        Delivery delivery = immutableConfig.getDelivery();
+        DeliveryStatus deliveryStatus = delivery.deliver(report, deliveryParams);
+
+        switch (deliveryStatus) {
+            case DELIVERED:
+                logger.i("Sent 1 new event to Bugsnag");
+                leaveErrorBreadcrumb(event);
+                break;
+            case UNDELIVERED:
+                logger.w("Could not send event(s) to Bugsnag,"
+                        + " saving to disk to send later");
+                eventStore.write(event);
+                leaveErrorBreadcrumb(event);
+                break;
+            case FAILURE:
+                logger.w("Problem sending event to Bugsnag");
+                break;
+            default:
+                break;
+        }
+        return deliveryStatus;
+    }
+
+    private void leaveErrorBreadcrumb(@NonNull Event event) {
+        // Add a breadcrumb for this event occurring
+        List<Error> errors = event.getErrors();
+
+        String name = "";
+        String msg = "";
+
+        if (errors.size() > 0) {
+            name = errors.get(0).getErrorClass();
+            msg = errors.get(0).getErrorMessage();
+        }
+
+        Map<String, Object> message = Collections.<String, Object>singletonMap("message", msg);
+        breadcrumbState.add(new Breadcrumb(name, BreadcrumbType.ERROR, message, new Date()));
+    }
+}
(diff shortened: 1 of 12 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+        this.logger = logger;
+        this.eventStore = eventStore;
+        this.immutableConfig = immutableConfig;
+        this.breadcrumbState = breadcrumbState;
+    }
+
+    void deliverEvent(@NonNull Event event) {
+        // Build the report
+        Report report = new Report(immutableConfig.getApiKey(), event);
+        Session session = event.getSession();
```
tomlongridge: Shouldn't this be set from `SessionTracker`'s current session?

We should also have assertions in the tests for this session to be incremented.

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+                }
+            });
+        } catch (RejectedExecutionException exception) {
+            eventStore.write(event);
+            logger.w("Exceeded max queue count, saving to disk to send later");
+        }
+    }
+
+    DeliveryStatus deliver(@NonNull Report report, @NonNull Event event) {
+        DeliveryParams deliveryParams = immutableConfig.errorApiDeliveryParams();
```
tomlongridge: Not a change from this PR directly, but should these functions be named `getErrorApiDeliveryParams` and `getSesssionApiDeliveryParams`?

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+
+import androidx.annotation.NonNull;
+
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.RejectedExecutionException;
+
+class ReportDeliveryDelegate extends BaseObservable {
```
tomlongridge: We are moving away from the concept of a `Report` in favour of `Event`. Should this be named just `DeliveryDelegate`?

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+
+    ReportDeliveryDelegate(Logger logger, EventStore eventStore,
+                           ImmutableConfig immutableConfig, BreadcrumbState breadcrumbState) {
+        this.logger = logger;
+        this.eventStore = eventStore;
+        this.immutableConfig = immutableConfig;
+        this.breadcrumbState = breadcrumbState;
+    }
+
+    void deliverEvent(@NonNull Event event) {
```
tomlongridge: Should this be called just `deliver` and rename the non-public method `deliverReportInternal` (or similar)?

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+                    deliver(finalReport, finalEvent);
+                }
+            });
+        } catch (RejectedExecutionException exception) {
+            eventStore.write(event);
+            logger.w("Exceeded max queue count, saving to disk to send later");
+        }
+    }
+
+    DeliveryStatus deliver(@NonNull Report report, @NonNull Event event) {
```
tomlongridge: Can this be private?

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+        // Add a breadcrumb for this event occurring
+        List<Error> errors = event.getErrors();
+
+        String name = "";
+        String msg = "";
+
+        if (errors.size() > 0) {
+            name = errors.get(0).getErrorClass();
+            msg = errors.get(0).getErrorMessage();
+        }
```
tomlongridge: This was discussed and refactored here: https://github.com/bugsnag/bugsnag-android/pull/673/files#diff-d01d476975cb3e9ebc74fb3cec71c943R759-R764

Is this a deliberate change or a regression?

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
```diff
@@ -0,0 +1,109 @@
...
+                notifyObservers(StateEvent.NotifyUnhandled.INSTANCE);
+            } else {
+                event.setSession(session.incrementHandledAndCopy());
+                notifyObservers(StateEvent.NotifyHandled.INSTANCE);
+            }
+        }
+
+        if (event.isUnhandled()) {
+            eventStore.write(event);
+            eventStore.flushAsync();
```
tomlongridge: This feels like the wrong level of abstraction and gives this method too many things to know.

For example, consider extracting to a `cacheEvent` function which takes a boolean for whether to flush straightaway. This can be called by all the methods that use `eventStore` in the class.

bugsnag-android-core/src/test/java/com/bugsnag/android/ReportDeliveryDelegateTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        // verify message sent
+        assertNotNull(msg)
+
+        // check session count incremented
+        assertEquals(0, event.session!!.unhandledCount)
+        assertEquals(1, event.session!!.handledCount)
+    }
+
+    @Test
+    fun deliverReport() {
```
tomlongridge: Consider tests for events with no/empty errors - impact on breadcrumbs

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
fractalwrench: Renamed class

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
fractalwrench: Renamed methods

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
fractalwrench: This needs to be package-private for testing - I've added an annotation to document this

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
fractalwrench: This was a regression due to copying over from the old diff, I've updated to use the refactored version

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
fractalwrench: My understanding is that the previous `currentSession` variable is equivalent to `event.getSession()` - and that therefore we can just remove that variable as its redundant.

fractalwrench: We have test coverage for session counts being incremented in `generateUnhandledReport` and  `generateHandledReport` - if more is required it'd be good to understand what needs adding.

bugsnag-android-core/src/main/java/com/bugsnag/android/ReportDeliveryDelegate.java
fractalwrench: Extracted to a separate function.

bugsnag-android-core/src/test/java/com/bugsnag/android/ReportDeliveryDelegateTest.kt
fractalwrench: I've added a test to ensure that if an Event has no errors the payload is dropped before attempting delivery.

PR #704, opened 2019-12-09
Merge strictmode metadata with global Metadata

fractalwrench:
## Goal

When a [StrictMode](https://developer.android.com/reference/android/os/StrictMode) exception is captured by our `UncaughtExceptionHandler` we attempt to extract additional information about the reason for the strictmode violation from the throwable object. This information is attached to the `Event` via a `Metadata` object passed to a notify call.

The implementation of `Client` did not merge this metadata object with the global metadata store, so the strictmode information was not included in events. This change fixes the `strict_mode.feature` scenarios which are only triggered on a full test run.


fractalwrench:
I can't think of a good + easy way to unit test this behaviour, as it's contained within `Client` which is a large class whose dependencies we would need to mock.

Perhaps we could supply a non-public constructor that injects mock dependencies, then verify that the `Event` contains the appropriate metadata when the `ReportDeliveryDelegate` is invoked. This would rely on #700

PR #705, opened 2019-12-10
Remove unused imports

fractalwrench:
## Goal

Removes unused imports revealed by the IntelliJ inspection tool. These are unnecessary and can therefore be removed from the codebase.


fractalwrench:
I've not been able to find an Android Lint or other static analysis check for this - presumably because IntelliJ offers tools to do the same.

PR #711, opened 2019-12-11
Reduce event storage size to 32

fractalwrench:
The notifier spec states that 32 events should be stored by default - this updates the implementation from the previous 128.


bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -53,7 +53,7 @@ public int compare(File lhs, File rhs) {
 
     EventStore(@NonNull ImmutableConfig config,
                @NonNull Context appContext, @NonNull Logger logger, Delegate delegate) {
-        super(appContext, "/bugsnag-errors/", 128, EVENT_COMPARATOR, logger, delegate);
+        super(appContext, "/bugsnag-errors/", 32, EVENT_COMPARATOR, logger, delegate);
```
tomlongridge: Can you extract a constant whilst we're at it?

PR #718, opened 2019-12-11
Alter bugsnag loaded breadcrumb type

fractalwrench:
The "Bugsnag Loaded" breadcrumb should be of type Navigation according to the spec. This changeset updates the type to match the spec, and test assertions to pass the change.


fractalwrench:
The `breadcrumb.feature` scenario seems to be failing due to the change in the breadcrumbtype - I believe the scenario filters out `Navigation` breadcrumbs by `config.enabledBreadcrumbTypes`

fractalwrench:
I have fixed the test and updated the type to be `State` rather than `Navigation`.

PR #721, opened 2019-12-12
Validate API key to be 16 hexadecimal characters long

fractalwrench:
## Goal

Validates the API key to be 16 hexadecimal characters long, as per the notifier spec.

## Changeset

- Made `Bugsnag.init(apiKey: String)` load `Configuration` with an `apiKey` override, rather than just use a `Configuration` object
- Validated that an API key supplied to `Configuration` must be 32 chars in length and only contains hexadecimal alphanumerics

## Tests

Added additional unit test coverage for `Configuration` API key validation, and updated existing test cases to pass in a valid API key.


fractalwrench:
I believe I've fixed the mazerunner scenarios by altering the API key in the browserstack mazerunner scenario suite. The local suite does pass on my machine so this was an issue where the two codebases have diverged and do different things.

PR #722, opened 2019-12-12
Prevent complex types from being serialized in breadcrumb metadata

fractalwrench:
## Goal

Prevents complex types from being serialized in breadcrumb metadata. This brings the implementation in line with the spec, which states that only strings, booleans, and numbers should be serialized as JSON.

## Changeset

Filtered the `metadata` added to the `Breadcrumb` during serialization to remove any complex objects, and added a unit test to verify that only primitive objects are serialized.


tomlongridge:
Review pending bugsnag/platforms-notifier-spec#156

PR #723, opened 2019-12-12
Clone metadata for configuration and event

fractalwrench:
## Goal

Clones metadata when setting the initial Client value from Configuration, and setting metadata on an Event. This matches the notifier spec and avoids global state mutating event payloads in-flight.

## Changeset

Made `Metadata` a data class which allows for its values to be cloned via a `copy()` method when setting on the `Event`.

## Tests

Added unit tests to verify that metadata is deep-cloned when necessary.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -145,7 +145,11 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
                 logger);
         systemBroadcastReceiver = new SystemBroadcastReceiver(this, logger);
 
-        metadataState = configuration.metadataState.copy(configuration.metadataState.getMetadata());
+        // performs deep copy of metadata to preserve immutability of Configuration interface
+        Metadata orig = configuration.metadataState.getMetadata();
+        Metadata copy = orig.copy();
+        metadataState = configuration.metadataState.copy(copy);
+
         // Set up and collect constant app and device diagnostics
         sharedPrefs = appContext.getSharedPreferences(SHARED_PREF_KEY, Context.MODE_PRIVATE);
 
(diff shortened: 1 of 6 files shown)


fractalwrench:
I've added some additional assertions and altered the implementation so that this performs a true deep clone of the Metadata sections.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -145,7 +145,10 @@ public Unit invoke(Boolean connected) {
                 logger);
         systemBroadcastReceiver = new SystemBroadcastReceiver(this, logger);
 
-        metadataState = configuration.metadataState.copy(configuration.metadataState.getMetadata());
+        Metadata orig = configuration.metadataState.getMetadata();
+        Metadata copy = orig.copy(orig.toMap(), orig.getJsonStreamer(), orig.getRedactedKeys());
```
tomlongridge: Should this be the parameterless override of `copy` from `Metadata`?

```suggestion
        Metadata copy = orig.copy();
```

Also, can you either comment on the purpose of this or extract to private function to document.

bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
```diff
@@ -80,7 +79,17 @@ internal class Metadata @JvmOverloads constructor(map: Map<String, Any> = Concur
...
     }
 
-    fun toMap(): Map<String, Any> = HashMap(store)
+    fun toMap(): Map<String, Any> {
+        val hashMap = HashMap(store)
+
+        // deep copy each section
+        store.entries.forEach {
+            if (it.value is ConcurrentHashMap<*, *>) {
+                hashMap[it.key] = ConcurrentHashMap(it.value as ConcurrentHashMap<*, *>)
```
tomlongridge: Do we not have to clone nested complex objects too?

fractalwrench: No - these would have been added by the user as a metadata value and it would be confusing to clone this IMO.

We would also not know how to clone nested objects unless they were a Map/other known type.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: I've updated this to use the parameterless override, and added an inline comment for documentation.

PR #727, opened 2019-12-12
Remove notify method variant

fractalwrench:
## Goal

Removes a `notify()` method variant that accepts a name, message, and stacktrace. It is possible to achieve this by using a `Throwable` with a custom stacktrace set on it and altering the name/message in a callback. This variant should therefore be removed.


tomlongridge:
Review on hold pending https://github.com/bugsnag/platforms-notifier-spec/issues/129

fractalwrench:
I believe this is ready for review now so I've taken it off hold.

PR #728, opened 2019-12-12
Rename Endpoints -> EndpointConfiguration

fractalwrench:
Renames `Endpoints` to `EndpointConfiguration` to match the notifier spec. Note that the field name should rename the same and only the type should change.


fractalwrench:
The test failure was a flake and the merge conflict has now been resolved.

PR #729, opened 2019-12-12
Use consistent case for enums

fractalwrench:
## Goal

The casing we use for enums is inconsistent in the codebase. While [both are valid for Kotlin](https://kotlinlang.org/docs/reference/coding-conventions.html#property-names), we should try and pick one to ensure our API remains consistent.

## Changeset

Changed `Error.Type` and `Thread.Type` to use upper-case. Existing test code was also updated to use these new names.


tomlongridge:
I don't see any enums in our codebase with camel case? Given that Java has a strong convention for all-caps and Kotlin opts for either, I don't see the need to change and cause potential extra work for upgrading users unless there's a stronger reason.

fractalwrench:
`Error.Type` and `Thread.Type` spring to mind. We could go for consistency in the other direction by making those caps if required.

tomlongridge:
(Apologies - I hadn't pulled the latest integration branch.)

From what I can see, the conventions are strongest for all caps so I would suggest we match that with the new enums.

fractalwrench:
I've updated the changeset to use upper case for everything.

PR #730, opened 2019-12-13
build(core): Prevent obfuscation of classes for native access

kattrali:
Fixes app compilation warnings and potential crash when calling the `Bugsnag` class via JNI.

Compiling against latest bugsnag-android showed warnings for keeping
`NativeInterface` while `Client` was obfuscated and a part of its public 
API, and similarly for `JsonStream` and `Breadcrumb`.

Additionally added the `Bugsnag` class to make it available via the JNI 
when obfuscation is in effect (as classes are loaded via the JNI by
name).


fractalwrench:
Closing as this was tested during UAT by @Cawllec and no unexpected behaviour was observed in an obfuscated app.

PR #732, opened 2019-12-13
Add App/AppWithState interface to Event

fractalwrench:
## Goal

Currently `event.app` is an unstructured `Map`. This change creates an `App/AppWithState` interface that defines each individual field in the payload. This updates the implementation to match the notifier spec.

## Changeset

- Added `App` interface which is used in the `SessionPayload` and contains stateless information only
- Added `AppWithState` interface which extends `App`. This is used in the `Event` payload and contains stateful information, such as the `durationInForeground`
- Ensured that it is possible to redact `App/AppWithState` by using nullable types for every field
- Removed `packageName ` from the `app` metadata tab, as the field is duplicated by `app.id`
- Removed `versionName ` from the `app` metadata tab, as the field is duplicated by `app.version` and this has been known to cause confusion amongst users in the past
- Updated internal reports to send the full `AppWithState` information
- Renamed `AppData` to `AppDataCollector` and converted to Kotlin
- Cached stateless results in fields within `AppDataCollector` rather than calculating each time

## Tests

Updated existing test code to use new App interfaces. Assertions for the expected JSON have also been updated where fields have changed. The mazerunner assertion steps for native payloads have also been updated.


bugsnag-android-core/src/test/java/com/bugsnag/android/InternalReportDelegateTest.kt
```diff
@@ -59,9 +56,7 @@ internal class InternalReportDelegateTest {
         delegate.reportInternalBugsnagError(event)
 
         // app
-        assertEquals(2, event.app["foo"])
-        assertEquals(500L, event.app["durationInForeground"])
-        assertEquals(true, event.app["inForeground"])
```
tomlongridge: I dislike removing assertions! Can't this data be easily set in the dummy app that was generated?

tests/features/steps/android_steps.rb
```diff
@@ -127,8 +127,6 @@
     And the payload field "events.0.device.totalMemory" is not null
     And the payload field "events.0.device.runtimeVersions.osBuild" is not null
     And the payload field "events.0.metaData.app.name" equals "MazeRunner"
-    And the payload field "events.0.metaData.app.packageName" equals "com.bugsnag.android.mazerunner"
-    And the payload field "events.0.metaData.app.versionName" is not null
```
tomlongridge: Should `versionName` assertion be dropped?

fractalwrench: I think that both `packageName` and `versionName` should no longer be automatically collected.

In the case of `versionName` this is because we already collect it with the key `app.version`. This has also caused a bit of confusion in the past about why we collect 3 separate keys: version/versionCode/versionName.

bugsnag-android-core/src/test/java/com/bugsnag/android/InternalReportDelegateTest.kt
fractalwrench: I've added this assertion back in.

PR #734, opened 2019-12-16
Add Device/DeviceWithState interface rather than unstructured Map

fractalwrench:
## Goal

Currently `event.device` is an unstructured `Map`. This change creates an `Device/DeviceWithState` interface that defines each individual field in the payload. This updates the implementation to match the notifier spec.

## Changeset

- Added `Device` interface which is used in the `SessionPayload` and contains stateless information only
- Added `DeviceWithState` interface which extends `Device`. This is used in the `Event` payload and contains stateful information, such as `freeDisk`
- Ensured that it is possible to redact `Device/DeviceWithState` by using nullable types for every field
- Updated internal reports to send the full `DeviceWithState` information rather than a limited subset
- Renamed `DeviceData` to `DeviceDataCollector` and converted to Kotlin
- Cached stateless results in fields within `DeviceDataCollector` rather than calculating each time

## Tests

Updated existing test code to use new Device interfaces. Assertions for the expected JSON have also been updated where fields have changed. The detekt baselines have also been updated to suppress static analysis warnings that we don't need to address.


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -146,7 +146,10 @@ public Unit invoke(Boolean connected) {
                 sessionStore, logger);
         systemBroadcastReceiver = new SystemBroadcastReceiver(this, logger);
 
-        metadataState = configuration.metadataState.copy(configuration.metadataState.getMetadata());
+        Metadata orig = configuration.metadataState.getMetadata();
+        Metadata copy = orig.copy(orig.toMap(), orig.getJsonStreamer(), orig.getRedactedKeys());
+        metadataState = configuration.metadataState.copy(copy);
```
bengourley: This logic is confusing – it seems like the thing that needs to happen here is a copy of the metadata, but `copy` is mentioned 4x as well as `getMetaData()` and `toMap()`. Maybe it just needs to be this gnarly but if so can you add comments as to why and what it's doing?

For instance, why can you call `orig.copy()` but then it looks like you have to pass in the data that you want to copy… `orig.toMap()`?

I also don't see what this change has to do with the description of the PR?

fractalwrench: I've rebased and updated this to avoid copying like this.

PR #743, opened 2020-01-10
Allow disabling previous signal handler invocation for Unity ANRs

fractalwrench:
Currently the previously installed signal handler is always invoked for SIGQUIT. In Unity apps invoking the previous handler triggers a SIGSEGV in the Unity signal handler, which is an apparent bug in the implementation. As the handler only appears to print a dump of the threads to the console, we should therefore avoid calling it as doing so will allow us to detect ANRs on Unity.

This is controlled by adding a non-public `unityApp` flag to `Configuration`, which is set by Unity during initialisation, then passed onto the `AnrPlugin` and ultimately the native layer.

Methods which enable/disable native error capture such as `enableAnrReporting/disableAnrReporting` have also been tweaked so that they do not flip the configuration option when invoked. This is because when the releaseStage was changed from the native layer, it would automatically enable/disable ANRs via `ClientConfigObserver` without respecting the current `detectAnrs` flag value.

Tested by verifying that an Android + Unity example app can both send ANRs without crashing.

### DIFF
--- a/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
+++ b/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
@@ -1,5 +1,6 @@
 package com.bugsnag.android
 
+import android.os.Handler
 import android.os.Looper
 
 internal class AnrPlugin : BugsnagPlugin {
@@ -14,13 +15,17 @@ internal class AnrPlugin : BugsnagPlugin {
     private lateinit var client: Client
     private val collector = AnrDetailsCollector()
 
-    private external fun enableAnrReporting()
+    private external fun enableAnrReporting(callPreviousSigquitHandler: Boolean)
     private external fun disableAnrReporting()
 
     override fun loadPlugin(client: Client) {
-        enableAnrReporting()
-        Logger.info("Initialised ANR Plugin")
-        this.client = client
+        // this must be run from the main thread as the SIGQUIT is sent to the main thread,
+        // and if the handler is installed on a background thread instead we receive no signal
+        Handler(Looper.getMainLooper()).post(Runnable {
+            this.client = client
+            enableAnrReporting(client.config.callPreviousSigquitHandler)
+            Logger.warn("Initialised ANR Plugin")
+        })
     }
 
     override fun unloadPlugin() = disableAnrReporting()
--- a/bugsnag-plugin-android-anr/src/main/jni/anr_handler.h
+++ b/bugsnag-plugin-android-anr/src/main/jni/anr_handler.h
@@ -26,8 +26,9 @@ extern "C" {
  * Monitor for ANRs, writing to a byte buffer when detected
  * @param env the JNIEnv used to notify when an ANR occurs
  * @param plugin the AnrPlugin object that shouild be used to notify bugsnag
+ * @param callPreviousSigquitHandler whether the previous SIGQUIT handler should be invoked
  */
-bool bsg_handler_install_anr(JNIEnv *env, jobject plugin);
+bool bsg_handler_install_anr(JNIEnv *env, jobject plugin, jboolean callPreviousSigquitHandler);
 
 void bsg_handler_uninstall_anr(void);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## TBD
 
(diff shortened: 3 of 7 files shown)


snmaynard:
I think having a configuration option that is called something like `callPreviousSignalHandler` is preferable because it is obvious what it does. Having an option that it just true if its in unity and then passing that into the ANR handler seems somewhat shortsighted.

fractalwrench:
I don't see a need for this to be publicly configurable by the end-user as this workaround is only meant to avoid an apparent bug in Unity's signal handler implementation. I'm fine with renaming the option itself.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -316,30 +316,17 @@ void recordStorageCacheBehavior(MetaData metaData) {
 
     private void loadPlugins() {
         NativeInterface.setClient(this);
-        enableOrDisableNdkReporting();
-        enableOrDisableAnrReporting();
-        BugsnagPluginInterface.INSTANCE.loadRegisteredPlugins(this);
```
kattrali: Removing this line breaks the plugin interface.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -14,13 +15,15 @@ internal class AnrPlugin : BugsnagPlugin {
...
 
-    private external fun enableAnrReporting()
+    private external fun enableAnrReporting(callPreviousSigquitHandler: Boolean)
     private external fun disableAnrReporting()
 
     override fun loadPlugin(client: Client) {
-        enableAnrReporting()
-        Logger.info("Initialised ANR Plugin")
-        this.client = client
+        Handler(Looper.getMainLooper()).post(Runnable {
```
kattrali: This could use a comment to indicate why it must be called from the main thread.

fractalwrench:
The key reason changes were needed to `Client/NativeInterface` is because when the `notifyReleaseStages` state changed, it automatically enabled ANR/NDK detection and disregarded any previous value set by the user: https://github.com/bugsnag/bugsnag-android/pull/743/files#diff-878031eeabeac11e0f7a2bd58b6aebf8L25

In our example app, the C# layer set the `notifyReleaseStages` and `releaseStage` in `Configuration.cs` after disabling ANR/NDK detection. This called through to the `ClientConfigObserver` which enabled ANR/NDK detection again. Therefore I have separated the code which flips the `detectAnrs/detectNdkCrashes` and the code which enables/disables signal handlers, as it avoids this scenario.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: I've updated this as I hadn't considered plugins such as cocos2dx which aren't part of this repository.

kattrali:
> The key reason changes were needed to `Client/NativeInterface` is because when the `notifyReleaseStages` state changed, it automatically enabled ANR/NDK detection and disregarded any previous value set by the user: https://github.com/bugsnag/bugsnag-android/pull/743/files#diff-878031eeabeac11e0f7a2bd58b6aebf8L25

And I agree, the change to the `ClientConfigObserver` makes sense. What I'm asking though is why the `Client` and `NativeInterface` changes are still necessary after changing `ClientConfigObserver`? For the specific Unity example app case, updating `config.setDetect...()` in `client.{dis,en}ableAnrReporting()` is then a no-op.

fractalwrench:
I've updated the implementation so that changes are no longer required to `Client/NativeInterface` and we only handle Unity's specific case by altering `ClientConfigObserver`.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.h
```diff
@@ -26,8 +26,9 @@ extern "C" {
  * Monitor for ANRs, writing to a byte buffer when detected
  * @param env the JNIEnv used to notify when an ANR occurs
  * @param plugin the AnrPlugin object that shouild be used to notify bugsnag
+ * @param callPreviousSigquitHandler whether the execution environment is a Unity app
```
kattrali: Stale comment

fractalwrench: Updated stale comment

PR #744, opened 2020-01-13
Implement on_error in NDK layer

fractalwrench:
## Goal

Adds an `on_error` implementation which consists of a callback that is invoked when a fatal signal/exception occurs. This will allow users to discard native errors at the time of capture and to append additional metadata if required.

It's worth noting that the intention is to expose more methods that will allow for the manipulation of the `bugsnag_event` struct in future PRs. This changeset focuses mainly on the ability to add/remove `on_error` callbacks and only adds mutators as required for testing.

## Changeset

- Added a public `on_error` typedef that consists of a single-param method with an `intptr_t` to a non-public `bugsnag_event` struct
- Added `add_on_error/remove_on_error`, which sets the value of a single global `on_error` variable which is invoked when a fatal error occurs. This is a slight deviation from the notifier specification but I believe it makes more sense in C than allowing a user to specify multiple callbacks
- Added `bugsnag_event_set_context` which allows users to alter the context of the `bugsnag_event` supplied in an `on_error` callback
- Invoked the `on_error` callback within the signal/C++ exception handler and conditionally serialized the event to disk depending on its return value

## Spec deviation

This changeset deviates from the notifier specification by only allowing for one `on_error` callback to be added.

The rationale behind this is to limit the complexity of callback implementations. This is because code in a callback needs to be async-safe and never crash, because if it is not the program may terminate and bugsnag-android will be unable to capture any useful information. This is in contrast to the JVM where we can just catch the thrown exception and carry on.

There is no technical reason preventing us from allowing multiple callbacks, but my feeling is we should only allow users to supply one to reduce the complexity in their implementation, which reduces the risk of a crash.

## Tests

Added E2E tests to verify that the event context can be mutated in `on_error` and that the return value dictates whether an event is serialized or not, for both signals and C++ exceptions.

Added an additional unit test to verify that `bugsnag_event_set_context/bugsnag_event_get_context` allow for the `context` field to be mutated. Going forward the intention is that all fields in the struct will be tested at the unit level this way.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
@@ -8,6 +8,9 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+typedef bool (*bsg_on_error)(void *);
+
 /**
  * Configure the Bugsnag interface, optionally including the JNI environment.
  * @param env  The JNI environment to use when using convenience methods
@@ -35,6 +38,12 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
  */
 void bugsnag_leave_breadcrumb(char *message, bsg_breadcrumb_t type);
 void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message, bsg_breadcrumb_t type);
+
+void bugsnag_add_on_error(bsg_on_error on_error);
+void bugsnag_remove_on_error(bsg_on_error on_error);
+
+char *bugsnag_event_get_context(void *event_ptr);
+void bugsnag_event_set_context(void *event_ptr, char *value);
 
 #ifdef __cplusplus
 }
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
@@ -39,10 +39,30 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
 void bugsnag_leave_breadcrumb(char *message, bsg_breadcrumb_t type);
 void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message, bsg_breadcrumb_t type);
 
+/**
+ * Adds a callback which is invoked whenever a fatal error occurs. The callback will be passed a
+ * pointer to the event payload as a parameter, allowing for data to be added/removed.
+ * @param on_error the callback
+ */
 void bugsnag_add_on_error(bsg_on_error on_error);
-void bugsnag_remove_on_error(bsg_on_error on_error);
 
+/**
+ * Removes any callback previously added in bugsnag_add_on_error
+ */
+void bugsnag_remove_on_error();
+
+/**
+ * Retrieves the event context
+ * @param event_ptr a pointer to the event supplied in an on_error callback
+ * @return the event context, or NULL if this has not been set
+ */
 char *bugsnag_event_get_context(void *event_ptr);
+
+/**
+ * Sets the event context
+ * @param event_ptr a pointer to the event supplied in an on_error callback
+ * @param value the new event context value, which can be NULL
+ */
 void bugsnag_event_set_context(void *event_ptr, char *value);
 
 #ifdef __cplusplus
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -48,7 +48,7 @@ void bugsnag_add_on_error(bsg_on_error on_error) {
   }
 }
 
-void bugsnag_remove_on_error(bsg_on_error on_error) {
+void bugsnag_remove_on_error() {
   if (bsg_global_env != NULL) {
     bsg_global_env->on_error = NULL;
   }
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -42,6 +42,26 @@ bsg_unwinder bsg_configured_unwind_style() {
   return BSG_CUSTOM_UNWIND;
 }
 
+void bugsnag_add_on_error(bsg_on_error on_error) {
+  if (bsg_global_env != NULL) {
+    bsg_global_env->on_error = on_error;
+  }
+}
+
+void bugsnag_remove_on_error(bsg_on_error on_error) {
+  if (bsg_global_env != NULL) {
+    bsg_global_env->on_error = NULL;
+  }
+}
+
+bool bsg_run_on_error() {
+  bsg_on_error on_error = bsg_global_env->on_error;
+  if (on_error != NULL) {
+      return on_error(&bsg_global_env->next_event);
+  }
+  return true;
+}
+
 JNIEXPORT void JNICALL Java_com_bugsnag_android_NdkPlugin_enableCrashReporting(
         JNIEnv *env, jobject _this) {
   if (bsg_global_env == NULL) {
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
@@ -64,6 +64,13 @@ typedef struct {
 
 bsg_unwinder bsg_configured_unwind_style();
 
+/**
+ * Invokes the user-supplied on_error callback, if it has been set. This allows users to mutate
+ * the bugsnag_event payload before it is persisted to disk, and to discard the report
+ * by returning false..
+ *
+ * @return true if the report should be delivered, false if it should be discarded
+ */
 bool bsg_run_on_error();
 
 #ifdef __cplusplus
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
@@ -9,6 +9,7 @@
 
 #include "event.h"
 #include "utils/stack_unwinder.h"
+#include "../assets/include/bugsnag.h"
 
 #ifndef BUGSNAG_LOG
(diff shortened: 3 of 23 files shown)


fractalwrench:
@tomlongridge asking for preliminary review before a more in-depth one by Delisa. I did run a high-level sketch of this design by Delisa yesterday which mitigates some of the risk.

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -8,6 +8,9 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+typedef bool (*on_error)(intptr_t);
```
tomlongridge: Do all our types have `bsg_` prefix - maybe `bsb_on_error_cb_t`?

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,11 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
 void bugsnag_leave_breadcrumb(char *message, bsg_breadcrumb_t type);
 void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message, bsg_breadcrumb_t type);
 
+void bugsnag_add_on_error(on_error on_error);
+void bugsnag_remove_on_error(on_error on_error);
```
tomlongridge: Is there a reason why this can't/shouldn't follow the pattern of declaring an `_env` version of each function that does the work and a version with this parameter that uses the global pointer (and validates that Bugsnag has been initialised)?

Currently the `env` parameter in the implementation of `bugsnag_add/remove_on_error_env` isn't used.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -42,6 +42,26 @@ bsg_unwinder bsg_configured_unwind_style() {
...
+  }
+}
+
+void bugsnag_remove_on_error_env(JNIEnv *env, on_error on_error) {
+  if (bsg_global_env != NULL) {
+    bsg_global_env->on_error = NULL;
+  }
+}
+
+bool bsg_run_on_error_cbs(bsg_environment *env) {
```
tomlongridge: I'm not sure about using `cb`/`cbs` abbreviation - it's not immediately clear to me what it stands for. It's also not consistently applied - only used for this run method. Consider using `_callback` and/or applying to other declarations outside the public interface.

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
tomlongridge: Is it confusing to have these be add and remove operations if there's just one allowed? Should we deliberately deviate from the spec here to differentiate the behaviour?

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
fractalwrench: Renamed the type as `bsg_on_error` - this is consistent with how we name other structs/types.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
fractalwrench: Renamed to `bsg_run_on_error`

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,11 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
...
+
+void bugsnag_event_set_context(intptr_t event_ptr, char *value);
```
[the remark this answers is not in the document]
fractalwrench: I've updated the parameter type to `void *` instead.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -42,6 +42,26 @@ bsg_unwinder bsg_configured_unwind_style() {
...
+  on_error cb = env->on_error;
+  if (cb!= NULL) {
+      return cb((intptr_t) &env->next_event);
```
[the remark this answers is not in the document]
fractalwrench: I'm not sure what this warning is referring to, so believe it might have been addressed by changing `intptr_t` to `void *`. Lmk if that's not the case

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
fractalwrench: I have added method variants which also have an `_env` parameter

fractalwrench: Added some additional context on the rationale behind this in the description.

fractalwrench:
Added an additional unit test to verify the behaviour of `bugsnag_event_set_context` - the plan is to add similar unit tests for each mutator we add that interacts with `bugsnag_event`.

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,14 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
 void bugsnag_leave_breadcrumb(char *message, bsg_breadcrumb_t type);
 void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message, bsg_breadcrumb_t type);
 
+void bugsnag_add_on_error(bsg_on_error on_error);
+void bugsnag_add_on_error_env(JNIEnv *env, bsg_on_error on_error);
+void bugsnag_remove_on_error(bsg_on_error on_error);
+void bugsnag_remove_on_error_env(JNIEnv *env, bsg_on_error on_error);
```
tomlongridge: As discussed IRL, there doesn't seem any point in adding the `_env` variant unless the method actually requires access to the JNI.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -42,6 +42,26 @@ bsg_unwinder bsg_configured_unwind_style() {
...
+}
+
+void bugsnag_remove_on_error_env(JNIEnv *env, bsg_on_error on_error) {
+  if (bsg_global_env != NULL) {
+    bsg_global_env->on_error = NULL;
+  }
+}
+
+bool bsg_run_on_error(bsg_environment *env) {
+  bsg_on_error on_error = env->on_error;
```
tomlongridge: Other methods in this file use `bsg_global_env` directly and all callers pass this in as the value, so I would suggest we keep the established pattern and remove the parameter.

fractalwrench:
Rebased to latest integration branch.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.h
```diff
@@ -57,10 +58,14 @@ typedef struct {
      * true if a handler has completed crash handling
      */
     bool crash_handled;
+
+    bsg_on_error on_error;
 } bsg_environment;
 
 bsg_unwinder bsg_configured_unwind_style();
 
+bool bsg_run_on_error();
```
kattrali: Could use some docs here to say what this does, and what the return value means. There might also be a more descriptive name, since its always used as a condition to persist reports? "should save report", "update and filter report" maybe?

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,12 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
 void bugsnag_leave_breadcrumb(char *message, bsg_breadcrumb_t type);
 void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message, bsg_breadcrumb_t type);
 
+void bugsnag_add_on_error(bsg_on_error on_error);
```
kattrali: The new functions in this file need some documentation in particular, as this is the public header.

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,12 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
...
+void bugsnag_remove_on_error(bsg_on_error on_error);
```
kattrali: If the argument is not going to be used, there's no need to require it and it seems a bit confusing to include it.

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,12 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
...
+
+char *bugsnag_event_get_context(void *event_ptr);
```
kattrali: In the docs in this function I'd clarify if its possible to get a null value.

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -36,6 +39,12 @@ void bugsnag_set_user_env(JNIEnv *env, char* id, char* email, char* name);
...
+void bugsnag_event_set_context(void *event_ptr, char *value);
```
kattrali: Is null a valid value for context?

fractalwrench:
Updated `bsg_remove_on_error` to remove unneeded parameter and added docs to `bsg_run_on_error`.

As discussed IRL, docs for Java + C will be added once the main docs site has been updated to avoid rework.

tomlongridge:
> As discussed IRL, docs for Java + C will be added once the main docs site has been updated to avoid rework.

For this level of detail, I'd be more comfortable doing the in-code commenting now so that this doesn't get forgotten. If just the key information is there, it can be finessed if required and made consistent with the docs later but I'm not sure of my ability to pick this up later and get all the details.

fractalwrench:
Added some provisional doc comments to the PR.

PR #745, opened 2020-01-16
Add ndk accessor methods for event.app

fractalwrench:
## Goal

Adds accessor methods for manipulating the fields on `event.app`. This is necessary now that the `bugsnag_event` pointer is exposed in an `on_error` callback, and allows users to mutate the generated report.

## Changeset

Added get/set methods for each field defined on `event.app` that is present in the notifier spec.

## Tests

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
@@ -51,20 +51,6 @@ void bugsnag_add_on_error(bsg_on_error on_error);
  */
 void bugsnag_remove_on_error();
 
-/**
- * Retrieves the event context
- * @param event_ptr a pointer to the event supplied in an on_error callback
- * @return the event context, or NULL if this has not been set
- */
-char *bugsnag_event_get_context(void *event_ptr);
-
-/**
- * Sets the event context
- * @param event_ptr a pointer to the event supplied in an on_error callback
- * @param value the new event context value, which can be NULL
- */
-void bugsnag_event_set_context(void *event_ptr, char *value);
-
 #ifdef __cplusplus
 }
 #endif
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
@@ -1,6 +1,9 @@
 #ifndef BUGSNAG_ANDROID_NDK_EVENT_API_H
 #define BUGSNAG_ANDROID_NDK_EVENT_API_H
 
+#include <stdbool.h>
+#include <sys/types.h>
+
 typedef enum {
   /** An unhandled exception */
   BSG_SEVERITY_ERR,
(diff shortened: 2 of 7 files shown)


fractalwrench:
Discussed IRL and agreed that where a method already exists, we should update it to use our public API to reduce the complexity of the implementation.

As a note, I couldn't find `bugsnag_app_set_binary_arch(bugsnag_event *event, char *value);` given as an example. I'm assuming that the comment only refers to the following methods:

```

void bugsnag_event_add_metadata_double(bugsnag_event *event, char *section,
                                       char *name, double value);
void bugsnag_event_add_metadata_string(bugsnag_event *event, char *section,
                                       char *name, char *value);
void bugsnag_event_add_metadata_bool(bugsnag_event *event, char *section,
                                     char *name, bool value);
void bugsnag_event_add_breadcrumb(bugsnag_event *event,
                                  bugsnag_breadcrumb *crumb);
void bugsnag_event_clear_breadcrumbs(bugsnag_event *event);
void bugsnag_event_remove_metadata(bugsnag_event *event, char *section,
                                   char *name);
void bugsnag_event_remove_metadata_tab(bugsnag_event *event, char *section);
void bugsnag_event_set_orientation(bugsnag_event *event, int value);
void bugsnag_event_set_app_version(bugsnag_event *event, char *value);
void bugsnag_event_set_build_uuid(bugsnag_event *event, char *value);
void bugsnag_event_set_release_stage(bugsnag_event *event, char *value);
void bugsnag_event_set_user_email(bugsnag_event *event, char *value);
void bugsnag_event_set_user_id(bugsnag_event *event, char *value);
void bugsnag_event_set_user_name(bugsnag_event *event, char *value);
void bugsnag_event_start_session(bugsnag_event *event, char *session_id,
                                 char *started_at, int handled_count, int unhandled_count);
bool bugsnag_event_has_session(bugsnag_event *event);
```

bugsnag-plugin-android-ndk/src/main/assets/include/bugsnag.h
```diff
@@ -45,6 +46,40 @@ void bugsnag_remove_on_error(bsg_on_error on_error);
...
+
+time_t bugsnag_app_get_duration(void *event_ptr);
+void bugsnag_app_set_duration(void *event_ptr, time_t value);
+
+time_t bugsnag_app_get_duration_in_foreground(void *event_ptr);
+void bugsnag_app_set_duration_in_foreground(void *event_ptr, time_t value);
+
+bool bugsnag_app_get_in_foreground(void *event_ptr);
+void bugsnag_app_set_in_foreground(void *event_ptr, bool value);
+
```
tomlongridge: Should these be in declared `event.h`?

fractalwrench: Moved these declarations to `assets/include/event.h`

tomlongridge: Should we therefore take them out of `jni/event.h` and use the public header internally where we use these methods? Otherwise we have to maintain both definitions.

fractalwrench:
Rebased to latest integration branch.

PR #749, opened 2020-01-21
Buildkite agent 4.3.5

twometresteve:
## Goal

Prepare the Buildkite pipeline for use with the 4.3.5 Buildkite agent.  In particular, allow docker-compose build and push operations to be performed in the same step.

## Changeset

Version of docker-compose-buildkite-plugin upgraded to v3.1.0, which states that docker push can now be combined with build or run. 

## Tests

Implicitly tested by virtue of a successful build.


tomlongridge:
@twometresteve - suggest you merge/rebase `next` into your branch to deal with at least some of these failures.

PR #752, opened 2020-01-23
Convert sendThreads from boolean to enum

fractalwrench:
## Goal

Converts the `sendThreads` configuration option from a boolean to an enum type. This allows for multiple states to be specified in future, and also allows for a distinction between always reporting thread traces, never reporting, and reporting only for unhandled errors.

## Changeset

- Updated `Event` logic that gates whether a thread trace should be collected or not
- Removed `com.bugsnag.android.SEND_THREADS` as a key that can be specified in the manifest as this is no longer a primitive type
- Updated existing test coverage to assert the new default values/overrides for the `sendThreads` option


fractalwrench:
`BreadcrumbType` is arguably also in the same boat. I don't think there's a strong benefit in making the change but am happy to do so in a separate PR if others feel differently

PR #756, opened 2020-01-24
Add enabledErrorTypes configuration option

fractalwrench:
## Goal

Reworks the configuration for controlling error capture to match the notifier spec, by adding an `ErrorType` class

## Changeset

- Replaces the previous boolean flags with `ErrorType` class on `Configuration`
- Removes the ability to set ANR/NDK/Exception detection from the manifest as these are no longer represented on `Configuration` as primitive types
- Updated existing code to use new `ErrorType` field

## Tests

Updated existing test code to use new `ErrorType` field.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorTypes.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorTypes.kt
@@ -1,0 +1,29 @@
+package com.bugsnag.android
+
+class ErrorTypes(
+
+    /**
+     * Sets whether [ANRs](https://developer.android.com/topic/performance/vitals/anr)
+     * should be reported to Bugsnag. When enabled, Bugsnag will record an ANR whenever the main
+     * thread has been blocked for 5000 milliseconds or longer.
+     *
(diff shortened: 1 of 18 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorTypes.kt
```diff
@@ -0,0 +1,29 @@
+package com.bugsnag.android
+
+class ErrorTypes(
+
+    /**
+     * Sets whether [ANRs](https://developer.android.com/topic/performance/vitals/anr)
+     * should be reported to Bugsnag. When enabled, Bugsnag will record an ANR whenever the main
+     * thread has been blocked for 5000 milliseconds or longer.
```
tomlongridge: This comment should probably be amended now that we are handling the signal rather than checking the main thread. Possibly fix in master and merge up?

fractalwrench: I plan on updating all the Javadoc once changes are complete on the integration branch and docs have been written for the docs site. As discussed IRL, leaving this as-is for now

PR #758, opened 2020-01-27
Add autoDetectErrors flag to Configuration

fractalwrench:
## Goal

Adds the `autoDetectErrors` flag to `Configuration` which disables all error capture, matching the notifier spec.

## Changeset

- Added `autoDetectErrors` boolean field to `Configuration`
- Added `com.bugsnag.android.AUTO_DETECT_ERRORS` as boolean field that can be read from manifest
- Disabled all error detection if `autoDetectErrors` is false when constructing an `ImmutableConfig` object

## Tests

Added to existing unit/instrumentation tests.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
@@ -8,6 +8,7 @@ import java.util.HashMap
 internal data class ImmutableConfig(
     val apiKey: String,
     val enabledErrorTypes: ErrorTypes,
+    val autoDetectErrors: Boolean,
     val autoTrackSessions: Boolean,
     val sendThreads: Thread.ThreadSendPolicy,
     val ignoreClasses: Collection<String>,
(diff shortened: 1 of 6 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -83,9 +84,15 @@ internal data class ImmutableConfig(
 }
 
 internal fun convertToImmutableConfig(config: Configuration): ImmutableConfig {
+    val errorTypes = when {
+        config.autoDetectErrors -> config.enabledErrorTypes.copy()
+        else -> ErrorTypes(anrs = false, ndkCrashes = false, unhandledExceptions = false)
```
tomlongridge: With this implementation we'll need to remember to add any new errors types to this function (which is not located nearby) to allow the flag to work correctly.

An explicit call to `autoDetectErrors` in the places that it's used seems preferable to me as it makes the intent clearer. Alternatively, could `ErrorTypes` take the `autoDetectErrors` Boolean parameter that could be logically and'd with the defaults to turn them all off?

fractalwrench: Updated `ErrorTypes` to contain a non-public secondary constructor which switches off detection for all flags.

PR #759, opened 2020-01-27
Add ndk accessors for event.device

fractalwrench:
## Goal

Adds accessor methods for manipulating the fields on `event.device`. This is necessary now that the `bugsnag_event` pointer is exposed in an `on_error` callback, and allows users to mutate the generated report.

## Changeset

- Added get/set methods for each field defined on `event.device` that is present in the notifier spec.
- Updated `bugsnag_event_set_orientation` to use new public accessor method
- Added a migration for `bsg_device_info_v1` as the struct value has changed

**Note:** this PR does _not_ include accessors for `cpuAbi` and `runtimeVersions`. I'm not 100% sure how we'd want to approach exposing array values from the `bugsnag_event` struct, and am also not convinced that users would want to manipulate these fields in the context of a crashing process.

## Tests

Added unit tests for each new accessor method, and migration tests for any changed structs.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.h
@@ -107,6 +107,7 @@ typedef struct {
     char network_access[64];
     char os_build[64];
     char os_version[64];
+    char os_name[64];
     float screen_density;
     char screen_resolution[32];
     long total_memory;
@@ -258,7 +259,6 @@ void bugsnag_event_clear_breadcrumbs(bugsnag_event *event);
 void bugsnag_event_remove_metadata(bugsnag_event *event, char *section,
                                    char *name);
 void bugsnag_event_remove_metadata_tab(bugsnag_event *event, char *section);
-void bugsnag_event_set_orientation(bugsnag_event *event, int value);
 void bugsnag_event_set_user_email(bugsnag_event *event, char *value);
 void bugsnag_event_set_user_id(bugsnag_event *event, char *value);
 void bugsnag_event_set_user_name(bugsnag_event *event, char *value);
--- a/bugsnag-plugin-android-ndk/src/main/jni/metadata.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/metadata.c
@@ -266,6 +266,10 @@ void bsg_populate_app_data(JNIEnv *env, bsg_jni_cache *jni_cache,
   (*env)->DeleteLocalRef(env, data);
 }
 
+char *bsg_os_name() {
+  return "android";
+}
+
 void bsg_populate_device_data(JNIEnv *env, bsg_jni_cache *jni_cache,
                               bugsnag_event *event) {
   jobject data = (*env)->CallStaticObjectMethod(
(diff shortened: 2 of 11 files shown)


bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
```diff
@@ -18,6 +31,106 @@ TEST test_event_context(void) {
...
+TEST test_event_os_name(void) {
+    bugsnag_event *event = init_event();
+    ASSERT_STR_EQ("android", event->device.os_name);
+    bugsnag_app_set_os_name(event, "samsung");
+    ASSERT_STR_EQ("samsung", bugsnag_app_get_os_name(event));
+    free(event);
+    PASS();
+}
+
 SUITE(event_mutators) {
```
tomlongridge: Missing cpu_abi test to make this check all `device` fields.

bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -109,6 +109,7 @@ typedef struct {
     char network_access[64];
     char os_build[64];
     char os_version[64];
+    char os_name[64];
```
tomlongridge: Also missing free disk and free memory

bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
```diff
@@ -2,10 +2,23 @@
...
 #include "../../main/assets/include/bugsnag.h"
+#include <event.h>
 
 bugsnag_event *init_event() {
     bugsnag_event *event = calloc(1, sizeof(bugsnag_event));
     bsg_strncpy_safe(event->context, "Foo", sizeof(event->context));
+
+    event->device.jailbroken = true;
+    event->device.total_memory = 1095092340;
+    bsg_strncpy_safe(event->device.id, "my-id-123", sizeof(event->device.id));
```
tomlongridge: Missing jailbroken, cpu_abi and total memory (free disk and free memory too, if we add them)

bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
fractalwrench: `cpu_abi` was left out of this PR as it will require a bit more thought since it's a struct, and the changeset was already quite large. I plan on addressing this in a separate PR.

bugsnag-plugin-android-ndk/src/main/jni/event.h
fractalwrench: It's not possible to capture these values in a C signal handler so we omit them from the `device` struct.

bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
fractalwrench: `jailbroken` and `totalMemory` appear to be populated in `init_event()` and there are tests for both of them, is something else missing?

`cpuAbi` will be addressed in a separate PR as its a struct and will require a bit of thought (mentioned elsewhere) and we don't collect free disk/memory as it's not possible from a signal handler.

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -297,6 +297,7 @@ void bsg_populate_device_data(JNIEnv *env, bsg_jni_cache *jni_cache,
   bsg_copy_map_value_string(env, jni_cache, data, "screenResolution",
                             event->device.screen_resolution,
                             sizeof(event->device.screen_resolution));
+  bsg_strcpy(event->device.os_name, "android");
```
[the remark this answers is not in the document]
fractalwrench: Extracted this to a function.

bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
tomlongridge: Apologies - it's out of horizontal alignment as it's not a string, so I missed that!

PR #761, opened 2020-01-27
Add v5 upgrade guide

fractalwrench:
Note: This is reopened PR #757 due to a requested rename of the base branch.

Adds an upgrade guide from v4 -> v5 of the notifier.

I've tried to focus on mapping legacy method signatures onto the replacement APIs, and linked off to the docs when an in-depth explanation is required (as the upgrade guide will inevitably get out of sync with the docs as time goes on).

I've borrowed a few bits and pieces from the JS upgrade guide so feedback might apply there also: https://github.com/bugsnag/bugsnag-js/blob/v7/UPGRADING.md#6x-to-7x

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -4,7 +4,291 @@ Upgrading Guide
 Upgrade from 4.X to 5.X
 -----------------------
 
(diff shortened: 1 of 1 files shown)


fractalwrench:
Merging this PR into v5, will review #772 shortly

PR #763, opened 2020-01-28
Add accessors for unhandled/severity

fractalwrench:
## Goal

Adds accessor methods for manipulating the `unhandled` and `severity` fields on `event`. This is necessary now that the `bugsnag_event` pointer is exposed in an `on_error` callback, and allows users to mutate the generated report.

## Changeset

Added get/set methods for `severity`, and a getter only for `unhandled`.

## Tests


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -138,6 +138,9 @@ bugsnag_event *bsg_map_v2_to_report(bugsnag_report_v2 *report_v2) {
     event->error.frame_count = report_v2->exception.frame_count;
     size_t error_size = sizeof(bsg_stackframe) * BUGSNAG_FRAMES_MAX;
     memcpy(&event->error.stacktrace, report_v2->exception.stacktrace, error_size);
+
+    // unhandled
+    event->unhandled = true;
```
tomlongridge: Can you improve the comment? I assume it's set to true because this was the previous default (L250).

PR #764, opened 2020-01-28
Add setUser accessor to event

fractalwrench:
## Goal

Adds accessor methods for manipulating `event.user`. This is necessary now that the `bugsnag_event` pointer is exposed in an `on_error` callback, and allows users to mutate the generated report.

## Changeset

Added a setter method for `event.user`.

Note that a getter method has not been added. This would also require adding a getter method to the main `Bugsnag` interface, and exposing some form of a `bsg_user` struct - as this would be an additive change this has been left for #768 

## Tests

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.c
@@ -113,6 +113,13 @@ void bugsnag_event_set_user_id(bugsnag_event *event, char *value) {
   bsg_strncpy_safe(event->user.id, value, sizeof(event->user.id));
 }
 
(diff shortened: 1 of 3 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -169,3 +169,10 @@ void bugsnag_event_clear_breadcrumbs(bugsnag_event *event) {
 bool bugsnag_event_has_session(bugsnag_event *event) {
     return strlen(event->session_id) > 0;
 }
+
+void bugsnag_event_set_user(void *event_ptr, char* id, char* email, char* name) {
```
tomlongridge: Can you reposition near set_user_xxx?

PR #766, opened 2020-01-28
Remove versionName/packageName from NDK payload

fractalwrench:
## Goal

Removes the `packageName` and `versionName` fields from the serialized payload in `event.app`. These are replaced by `id` and `version` respectively, so are redundant.

The fields have already been removed from the JVM payload - this removes them from the NDK payload.

## Tests

Verified in an example app that the payload is not present for a fatal C error.


fractalwrench:
Renamed `binaryArch` to `binary_arch`.

PR #767, opened 2020-01-28
Add grouping_hash to ndk event interface

fractalwrench:
## Goal

Adds `grouping_hash` to `bugsnag_event` so that users can alter the grouping on the `bugsnag_event` exposed in an `on_error` callback.

## Tests

Added unit tests for each new accessor method, and to verify that the new field is serialized correctly.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
@@ -282,6 +282,12 @@ void bsg_serialize_context(const bugsnag_event *event, JSON_Object *event_obj) {
   }
 }
 
+void bsg_serialize_grouping_hash(const bugsnag_event *event, JSON_Object *event_obj) {
+  if (strlen(event->grouping_hash) > 0) {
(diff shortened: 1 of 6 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -242,6 +242,12 @@ void bsg_serialize_context(const bugsnag_event *event, JSON_Object *event_obj) {
   }
 }
 
+void bsg_serialize_grouping_hash(const bugsnag_event *event, JSON_Object *event_obj) {
+  if (strlen(event->grouping_hash) > 0) {
```
tomlongridge: Most of the other fields are set regardless of whether there is a value in `event` - is there a reason this is guarded?

fractalwrench: A groupingHash should only be serialized if a user has set the value on the struct. If the value is `NULL` then I believe the current JSON implementation won't serialize it anyway but it felt good to be explicit on this point.

PR #772, opened 2020-01-30
Proposed corrections to V5 upgrading.md

tomlongridge:
- Slimmed-down client table
- Reworded some headings and combined some smaller sections
- Tweaked the callback example for clarity
- Combined final section to a table

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -8,12 +8,12 @@ __This version contains many breaking changes__. It is part of an effort to unif
 
 # Interfaces
 
-## Bugsnag
+## Bugsnag interface
 
 `Bugsnag.init()` has been renamed to `Bugsnag.start()`.
 
-Many of the previous methods on `Bugsnag` should now be called on `Configuration` instead.
-For example, `maxBreadcrumbs` can now only be set on `Configuration` before `Bugsnag` is started:
+Many of the previous methods on `Bugsnag` must now be called on `Configuration` and provided before Bugsnag starts instead.
+For example, `maxBreadcrumbs` can now only be set on `Configuration` and supplied as part of the `start` method:
 
 ```kotlin
 val config = Configuration("my-api-key")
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -210,85 +160,74 @@ Several existing methods have been renamed, a full list of which is shown below:
 
 See the [full documentation](https://docs.bugsnag.com/platforms/android) for more information.
 
-## BeforeNotify/BeforeSend/Callback
+## Event callbacks
```
fractalwrench: Users do not have prior knowledge about what an 'event callback' is. They will be familiar with `BeforeNotify/BeforeSend/Callback` and will search for those terms in the upgrade guide. I think we should therefore use the exact class name in at least part of each of the section titles to reduce confusion.

UPGRADING.md
```diff
@@ -68,70 +68,22 @@ The full list of altered methods and their intended replacements can be found be
...
+-Client(this, "my-api-key")
++Bugsnag.start(this, "my-api-key")
 ```
 
-The full list of altered methods and their intended replacements can be found below:
-
 | v4.x API  | v5.x API |
 | ------------- | ------------- |
 | `Client#Client(Context, String, boolean)`  | `Bugsnag#start(Context, Configuration)` |
-| `Client#addToTab`  | `Bugsnag#addMetadata(String, String, Object)` and `Configuration#addMetadata` |
```
fractalwrench: These methods should be kept in the table. When reading this document our users will have no prior knowledge that `Bugsnag` is now favoured over using `Client` and it could seem like no upgrade path has been documented for the `Client` class.

UPGRADING.md
```diff
@@ -185,7 +135,7 @@ Several methods on `Configuration` have been renamed for greater API consistency
 
 See the [full documentation](https://docs.bugsnag.com/platforms/android) for more information.
 
-## Error
+## Events & Errors
```
fractalwrench: Users don't know what an `Event` is - we should use the old v4 name.

UPGRADING.md
```diff
@@ -210,85 +160,74 @@ Several existing methods have been renamed, a full list of which is shown below:
 
 See the [full documentation](https://docs.bugsnag.com/platforms/android) for more information.
 
-## BeforeNotify/BeforeSend/Callback
+## Event callbacks
 
-These three callbacks have been superseded and replaced by `OnError`, a single callback which runs immediately after an error has occurred. This can run globally on all errors, or on an individual error.
+`BeforeNotify` and `BeforeSend` callbacks have been superseded and replaced by `OnError`, a single callback which runs immediately after an error has occurred. This can run globally on all errors, or on an individual handled error:
```
fractalwrench: IMO 'individual error' is clearer, particularly as we're about to allow end-users to alter the handled state of errors.

fractalwrench:
Closing as obsolete.

PR #774, opened 2020-02-03
Allow editing metadata in ndk payload

fractalwrench:
## Goal

Allows accessing and manipulating data in the NDK `bugsnag_event` struct.

## Changeset

As C doesn't allow overloading methods this implementation deviates slightly from the notifier spec. The changes include:

- Made `bsg_metadata_t` a public type
- Added method for getting metadata for char/bool/double: `bsg_get_metadata_<type>`
- Added method for adding metadata for char/bool/double: `bsg_add_metadata_<type>`
- Added `bsg_clear_metadata/bsg_clear_metadata_section`
- Added `bsg_has_metadata` - it is expected that users call this to discern the type of metadata for the given key/section, if any

**Note**: this implementation focuses on adding these methods to access fields on the `bugsnag_event` payload only. Once this PR has been approved a similar approach will be followed on the `bugsnag` interface and the `breadcrumb` interface.

## Tests

Added unit tests to verify new methods.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
@@ -56,6 +56,13 @@ typedef struct {
     char id[64];
 } bsg_user_t;
 
+typedef enum {
+    BSG_NONE_VALUE,
+    BSG_BOOL_VALUE,
+    BSG_CHAR_VALUE,
+    BSG_NUMBER_VALUE,
+} bsg_metadata_t;
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -162,6 +169,20 @@ bool bugsnag_event_is_unhandled(void *event_ptr);
 
 char *bugsnag_event_get_grouping_hash(void *event_ptr);
 void bugsnag_event_set_grouping_hash(void *event_ptr, char *value);
+
+/* Metadata */
+
+void bugsnag_event_add_metadata_double(void *event_ptr, char *section, char *name, double value);
+void bugsnag_event_add_metadata_string(void *event_ptr, char *section, char *name, char *value);
+void bugsnag_event_add_metadata_bool(void *event_ptr, char *section, char *name, bool value);
+
+void bugsnag_event_clear_metadata_section(void *event_ptr, char *section);
+void bugsnag_event_clear_metadata(void *event_ptr, char *section, char *name);
+
+bsg_metadata_t bugsnag_event_has_metadata(void *event_ptr, char *section, char *name);
+double bugsnag_event_get_metadata_double(void *event_ptr, char *section, char *name);
+char *bugsnag_event_get_metadata_string(void *event_ptr, char *section, char *name);
+bool bugsnag_event_get_metadata_bool(void *event_ptr, char *section, char *name);
 
 #ifdef __cplusplus
 }
--- a/bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
+++ b/bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
@@ -297,6 +297,49 @@ TEST test_event_unhandled(void) {
     free(event);
     PASS();
 }
+
+TEST test_event_metadata(void) {
+    bugsnag_event *event = init_event();
+    bugsnag_event_add_metadata_string(event, "str", "foo", "hello");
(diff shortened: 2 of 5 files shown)


bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -163,6 +170,20 @@ bool bugsnag_event_is_unhandled(void *event_ptr);
...
+void bugsnag_event_add_metadata_double(void *event_ptr, char *section, char *name, double value);
+void bugsnag_event_add_metadata_string(void *event_ptr, char *section, char *name, char *value);
+void bugsnag_event_add_metadata_bool(void *event_ptr, char *section, char *name, bool value);
+
+void bugsnag_event_clear_metadata_section(void *event_ptr, char *section);
+void bugsnag_event_clear_metadata(void *event_ptr, char *section, char *name);
+
+bsg_metadata_t bugsnag_event_has_metadata(void *event_ptr, char *section, char *name);
+double bugsnag_event_get_metadata_double(void *event_ptr, char *section, char *name);
+void bugsnag_event_get_metadata_string(void *event_ptr, char *section, char *name, char dst[64]);
```
tomlongridge: All the other `get` methods for strings return a `char *` to the actual string in the event structure. Is there a reason we can't take this approach for metadata strings rather than passing in a write parameter?

bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
```diff
@@ -297,6 +297,44 @@ TEST test_event_unhandled(void) {
...
+
+    // test return value for present key
+    bugsnag_event_get_metadata_string(event, "str", "foo", dst);
+    ASSERT_STR_EQ("hello", dst);
+    ASSERT_EQ(5.9, bugsnag_event_get_metadata_double(event, "double", "foo"));
+    ASSERT_EQ(true, bugsnag_event_get_metadata_bool(event, "bool", "foo"));
+
+    // test clearing values
+    bugsnag_event_clear_metadata_section(event, "str");
+    ASSERT_EQ(BSG_NONE_VALUE, bugsnag_event_has_metadata(event, "str", "foo"));
```
tomlongridge: For completeness, maybe test clearing a value with `bugsnag_event_clear_metadata` and also assert that other keys still exist in the section?

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -163,6 +170,20 @@ bool bugsnag_event_is_unhandled(void *event_ptr);
...
+/* Metadata */
+
+void bugsnag_event_add_metadata_double(void *event_ptr, char *section, char *name, double value);
+void bugsnag_event_add_metadata_string(void *event_ptr, char *section, char *name, char *value);
+void bugsnag_event_add_metadata_bool(void *event_ptr, char *section, char *name, bool value);
+
+void bugsnag_event_clear_metadata_section(void *event_ptr, char *section);
+void bugsnag_event_clear_metadata(void *event_ptr, char *section, char *name);
+
+bsg_metadata_t bugsnag_event_has_metadata(void *event_ptr, char *section, char *name);
```
tomlongridge: The use of this method means that we need to search metadata twice (once to check existence and once to return the value). Do you foresee that this will be a particularly useful public function? If not, I would suggest that we remove this and just use `bugsnag_get_metadata_value` (or rename to the simpler `bugsnag_get_metadata`).

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
fractalwrench: C doesn't allow returning arrays from functions

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
fractalwrench: I think this has utility in that it allows users to check whether a key exists or not. `false` might be a valid return value for `bugsnag_event_get_metadata_bool`, or it might signify that the key doesn't exist. Without this method an end-user has no way of knowing.

bugsnag-plugin-android-ndk/src/test/cpp/test_bsg_event.c
fractalwrench: Added an extra assertion here.

PR #775, opened 2020-02-03
Add ndk accessor for device.cpuAbi

fractalwrench:
## Goal

Adds a getter/setter for the `event.device.cpuAbi` field, which finishes the work of adding getters/setters to all of the fields which will remain on the `bsg_device` struct.

## Changeset

Added a getter/setter which copy the value into/from a char array.

## Tests

Added a unit test for the new accessor methods.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
@@ -113,6 +113,8 @@ void bugsnag_app_set_in_foreground(void *event_ptr, bool value);
 
 /* Accessors for event.device */
 
+void bugsnag_device_get_cpu_abi(void *event_ptr, char value[8][32]);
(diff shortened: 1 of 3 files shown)


bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -113,6 +113,8 @@ void bugsnag_app_set_in_foreground(void *event_ptr, bool value);
 
 /* Accessors for event.device */
 
+void bugsnag_device_get_cpu_abi(void *event_ptr, char value[8][32]);
+void bugsnag_device_set_cpu_abi(void *event_ptr, char value[8][32]);
```
tomlongridge: This feels a little clunky to use with a multi-dimensional array parameter. How about managing the top-level array internally and having get-by-index, size, add and clear methods?

PR #777, opened 2020-02-05
Allow editing stacktrace in ndk payload

fractalwrench:
## Goal

Allows editing the stacktrace in the NDK event payload, as per the notifier spec.

## Changeset

Users can request the size of the stacktrace, and then get a stackframe at a particular index and inspect/mutate it. This was seen as preferable to exposing the entire array.

Additionally, `bsg_stackframe_t` has been added as a public type.

## Tests

Added a unit test for getting/mutating a stacktrace

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.c
@@ -463,3 +463,18 @@ void bugsnag_event_set_grouping_hash(void *event_ptr, char *value) {
   bugsnag_event *event = (bugsnag_event *) event_ptr;
   bsg_strncpy_safe(event->grouping_hash, value, sizeof(event->grouping_hash));
 }
+
+int bugsnag_event_get_stacktrace_size(void *event_ptr) {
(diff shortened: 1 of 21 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -463,3 +463,14 @@ void bugsnag_event_set_grouping_hash(void *event_ptr, char *value) {
...
 }
+
+int bugsnag_event_get_stacktrace_size(void *event_ptr) {
+  bugsnag_event *event = (bugsnag_event *) event_ptr;
+  return event->error.frame_count;
+}
+
+bsg_stackframe_t *bugsnag_event_get_stackframe(void *event_ptr, int index) {
+  bugsnag_event *event = (bugsnag_event *) event_ptr;
+  return &event->error.stacktrace[index];
```
tomlongridge: We should probably guard for out-of-bounds and return null if the index isn't >0 and <size

fractalwrench: Adding a bounds check here and returned `NULL`.

PR #778, opened 2020-02-05
Rename ignoreClasses to discardClasses

fractalwrench:
## Goal

Renames `ignoreClasses` to `discardClasses` as per the notifier spec, as this better fits terminology in the dashboard.

## Changeset

Renamed `ignoreClasses` to `discardClasses`

## Tests

Relied on existing test coverage


bugsnag-android-core/src/main/java/com/bugsnag/android/ManifestConfigLoader.kt
```diff
@@ -107,7 +107,7 @@ internal class ManifestConfigLoader {
             if (data.containsKey(ENABLED_RELEASE_STAGES)) {
                 enabledReleaseStages = getStrArray(data, ENABLED_RELEASE_STAGES, emptySet())
             }
-            ignoreClasses = getStrArray(data, IGNORE_CLASSES, ignoreClasses)
+            discardClasses = getStrArray(data, IGNORE_CLASSES, discardClasses)
```
tomlongridge: Need to rename the manifest key name - IGNORE_CLASSES.

fractalwrench: Renamed the manifest key.

PR #780, opened 2020-02-05
Add methods for manipulating breadcrumbs in event payload

fractalwrench:
## Goal

Adds methods for manipulating breadcrumbs in the event payload, as per the notifier spec.

## Changeset

- Renamed breadcrumb type to `bsg_breadcrumb_type_t`
- Made `bsg_breadcrumb_t` public
- Added method for getting/setting breadcrumbs, which map the internal breadcrumb struct onto the externally visible struct

**Note:** the metadata is not currently accessible on the breadcrumb struct, and the `onBreadcrumb` callback has not yet been implemented. This will be addressed in separate PRs.

## Tests

Added a unit test to verify clearing, getting, and setting breadcrumbs.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/assets/include/event.h
@@ -48,6 +48,12 @@ typedef enum {
    *  User event, such as authentication or control events
    */
   BSG_CRUMB_USER,
+} bsg_breadcrumb_type_t;
+
+typedef struct {
+    char message[128];
+    char timestamp[37];
+    bsg_breadcrumb_type_t type;
 } bsg_breadcrumb_t;
 
 typedef struct {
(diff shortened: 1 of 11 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -195,7 +195,7 @@ typedef struct {
 typedef struct {
     char name[33];
```
tomlongridge: Should we rename this to `message` for consistency?

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -48,6 +48,12 @@ typedef enum {
    *  User event, such as authentication or control events
    */
   BSG_CRUMB_USER,
+} bsg_breadcrumb_type_t;
+
+typedef struct {
+    char message[33];
```
tomlongridge: This character limit has now been lifted. Assuming that we still need a fixed size array on the NDK layer, can we increase this to a less restrictive size? 128 perhaps?

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
```diff
@@ -184,6 +190,13 @@ double bugsnag_event_get_metadata_double(void *event_ptr, char *section, char *n
 char *bugsnag_event_get_metadata_string(void *event_ptr, char *section, char *name);
 bool bugsnag_event_get_metadata_bool(void *event_ptr, char *section, char *name);
 
+/* Breadcrumbs */
+
+void bugsnag_event_clear_breadcrumbs(void *event_ptr);
+int bugsnag_event_get_breadcrumbs_size(void *event_ptr);
+bsg_breadcrumb_t bugsnag_event_get_breadcrumb(void *event_ptr, int index);
+void bugsnag_event_set_breadcrumb(void *event_ptr, bsg_breadcrumb_t breadcrumb, int index);
```
tomlongridge: These should return/accept `bsg_breadcrumb_t *`: returning a reference to the actual breadcrumb when getting and performing a copy of the struct when setting.

bugsnag-plugin-android-ndk/src/main/assets/include/event.h
fractalwrench: Added a migration to achieve this.

PR #784, opened 2020-02-06
Migrate fields in NDK payload to metadata

fractalwrench:
## Goal

Migrates legacy fields from the `bsg_app` and `bsg_device` structs to metadata instead. This allows users to redact or mutate the information if they wish, as we don't expose accessors for these fields.

**Note**: `app.activeScreen` and `device.osBuild` have been left in place as they affect the context/runtimeversions respectively so should be kept in the struct.

## Changeset

Moved the following fields on app to metadata:

- `name`
- `lowMemory` (removed entirely, as this value is stateful and cannot be set to an accurate value)
- `memoryUsage` (removed entirely, as never used)

And the following fields on device:

- `batteryLevel`
- `brand`
- `dpi`
- `emulator`
- `locationStatus`
- `networkAccess`
- `screenDensity`
- `screenResolution`

Serialized reports are migrated to metadata when they are read.

## Tests

- Updated unit tests to remove metadata values from serialization tests (as this is covered elsewhere in metadata serialization test).
- Updated unit tests to serialize `bugsnag_report_v1` to disk when testing migrations, as the existing code was erroneously trying to fit a `bugsnag_report_v1` into `bugsnag_event` which led to test failures with these changes
- Ensured that the native mazerunner scenarios validate that these fields are still in the payload
- Compared an NDK report against a JVM error report in an example app, verifying the app/device tabs


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -340,8 +339,6 @@ void bsg_serialize_app(const bsg_app_info app, JSON_Object *event_obj) {
 
 void bsg_serialize_app_metadata(const bsg_app_info app, JSON_Object *event_obj) {
   json_object_dotset_string(event_obj, "metaData.app.activeScreen", app.active_screen);
```
tomlongridge: From your comments on the PR:

> Note: app.activeScreen and device.osBuild have been left in place as they affect the context/runtimeversions respectively so should be kept in the struct.

Is this consistent if the value is written to `metaData.app. activeScreen`? Can we not move it to metadata everywhere?

tomlongridge: Also, should we lower-case `metaData`?

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -372,14 +369,7 @@ void bsg_serialize_device(const bsg_device_info device, JSON_Object *event_obj)
 }
 
 void bsg_serialize_device_metadata(const bsg_device_info device, JSON_Object *event_obj) {
-  json_object_dotset_string(event_obj, "metaData.device.brand", device.brand);
-  json_object_dotset_boolean(event_obj, "metaData.device.emulator", device.emulator);
   json_object_dotset_boolean(event_obj, "metaData.device.jailbroken", device.jailbroken);
```
tomlongridge: I believe this should be in `device.jailbroken`, not metadata.

fractalwrench:
In person discussion explained that these fields should be migrated because currently they are serialized to the metadata, but there is no way for a user to redact them, because they exist on the `bsg_app_info` and `bsg_device_info` structs, and we haven't exposed accessors for them.

Migrating to metadata allows for the values to be redacted if required by end-users, and makes our implementation align more closely with the error payload.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: This would require changes to the Error Reporting API so the key can be altered after that happens.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: I've updated the serializer to put this in `device.jailbroken` instead.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -351,7 +351,7 @@ Java_com_bugsnag_android_ndk_NativeBridge_updateLowMemory(JNIEnv *env,
   if (bsg_global_env == NULL)
     return;
   bsg_request_env_write_lock();
-  bsg_global_env->next_event.app.low_memory = (bool)new_value;
+  bugsnag_event_add_metadata_bool(&bsg_global_env->next_event, "app", "lowMemory", (bool)new_value);
```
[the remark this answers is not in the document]
fractalwrench: This method doesn't appear to be invoked by the JNI anywhere, so is dead code, and nothing will be serialized.

This seems to be the case on master too - and I'm not convinced that it makes sense to put this value in fatal native reports, as the information will always be out-of-date by the time a crash occurs, and we won't be able to query for a fresh value via the JNI at this point.

Leaving the JNI method in place seemed like the option with the least risk but we could remove this if that's preferable.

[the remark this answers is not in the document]
tomlongridge: Can you raise a ticket to cover this? It seems like we either need to instate the JNI update call and consider what happens if the low memory warnings comes and goes throughout the app's lifecycle; or else remove it from NDK and JVM layers.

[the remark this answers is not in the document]
fractalwrench: Created a ticket.

PR #789, opened 2020-02-07
Alter data collected in navigation breadcrumbs

fractalwrench:
## Goal

Alters the data collected in navigation breadcrumbs to bring the metadata structure in line with the notifier spec.

## Changeset

- Extracted activity lifecycle breadcrumb logging into `ActivityBreadcrumbCollector`, separating this responsibility from the `SessionTracker`
- Captured `from` and `to` in the metadata of lifecycle breadcrumbs
- Captured `hasBundle` to detect whether the `onCreate()` method was called with a bundle. This indicates whether the activity has been launched for the first time or is being recreated, and is useful information for developers
- Extracted lifecycle callbacks into `SessionLifecycleCallback` for the `SessionTracker`, reducing the class responsibilities
- Updated client to register new callbacks

## Tests

Tested in an example app to verify the navigation breadcrumbs in the dashboard.

Added unit tests and updated existing ones to increase coverage.

Suppressed detekt static analysis check about empty functions in `SessionLifecycleCallback`.

Screenshot 2020-02-07 at 17 43 06


snmaynard:
I agree with @bengourley. State is a better type for this. Also how can an activity go from onDestroy to onCreate?

fractalwrench:
An activity can go from `onDestroy()` to `onCreate()` when the Activity is being recreated due to a [Configuration Change](https://developer.android.com/guide/topics/resources/runtime-changes), which is typically caused by a change in screen orientation. The Android Framework tears down the existing activity and invokes `onCreate()` with a bundle of savedState. The expectation is that the developer can then use that saved state to draw whatever is needed on screen.

I'll update the breadcrumb type to State, and reword the metadata keys.

fractalwrench:
After discussing with Ben I've updated the breadcrumb so that data is structured in the following way: Screenshot 2020-02-11 at 10 45 35

PR #791, opened 2020-02-10
Collect breadcrumb for device Orientation Change

fractalwrench:
## Goal

Collects a breadcrumb when the device's orientation changes. This gives Android parity with other platforms and the notifier spec, which states that this information should be automatically captured.

Additionally, this changeset fixes an inconsistency in how the device orientation was calculated between the JVM/NDK layers by removing `OrientationEventListener`. This class used the _physical_ orientation of the device to calculate orientation, which may be inconsistent with the _software_ orientation. For example, a user can lock apps to only display in portrait/landscape. Removing this class also avoids the need for bugsnag-android to use an accelerometer which can be a privacy/battery concern.

## Changeset

- Added `ConfigChangeReceiver`, which listens for the `CONFIGURATION_CHANGED` system broadcast. Apps can capture this event when the system undergoes a [configuration change](https://developer.android.com/guide/topics/resources/runtime-changes) - this is commonly a change in orientation, but might also be a change in Locale/other events
- When `CONFIGURATION_CHANGED` is received, the device orientation is queried, and if it has changed, a breadcrumb is logged and the NDK layer's orientation value is updated via the `ClientObservable`
- Removed `OrientationEventListener` as this class is no longer required
- Updated the NDK layer to accept a `String` value for orientation, rather than an `Int`. This means we no longer need to calculate the string value for orientation from the degrees passed to the native layer

## Tests

- Verified in an example app that a breadcrumb is left for each orientation change
- Verified that the orientation is correct for a device in both a JVM + NDK crash
- Added unit test coverage for `ConfigChangeReceiver`


fractalwrench:
This is intentional, as these could be two separate events emitted by the system - for example, if the user updates their device's locale.

PR #792, opened 2020-02-12
Populate releaseStage in config during initial sanitisation

fractalwrench:
## Goal

The `releaseStage` was not set on the `ImmutableConfig` object when a `Client` was initialised, meaning it remained null if not set explicitly. This updates the sanitiser function to set a default value on the field, rather than just the serialized payload.

## Changeset

Moved the `guessReleaseStage` method to `ImmutableConfig` which is populated on `Client` initialisation.

## Tests

Added a unit test to verify that the field is set to a sensible default.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
@@ -1,6 +1,7 @@
 package com.bugsnag.android
 
 import android.content.Context
+import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
 import java.util.Date
 import java.util.HashMap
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -160,5 +161,38 @@ internal fun sanitiseConfiguration(
             configuration.buildUuid = buildUuid
         }
     }
+
+    // populate releaseStage
+    if (configuration.releaseStage == null) {
```
tomlongridge: We're only entering this block if `releaseStage` is null in config, so the check in `guessReleaseStage` is redundant.

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -160,5 +161,38 @@ internal fun sanitiseConfiguration(
...
+        try {
+            val packageManager = appContext.packageManager
+            val ai = packageManager.getApplicationInfo(
+                packageName, PackageManager.GET_META_DATA
+            )
```
tomlongridge: How much overhead is there in doing this? Can the ApplicationInfo be shared with the logic to get the buildUuid above?

fractalwrench: There's minimal overhead in retrieving this information IIRC.

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
fractalwrench: The check is not redundant in `guessReleaseStage` as `config.releaseStage` is a mutable property whose value could have been mutated on another thread - the Kotlin compiler will not allow us to remove the null-check from that method as-is. We could refactor the method to take a `String` only and copy `config.releaseStage` to a local variable, but I don't think there's much benefit in doing so.

fractalwrench:
Inlined `guessReleaseStage()` as suggested, and simplified the implementation by putting `PackageInfo` and `ApplicationInfo` into local variables up front.

PR #794, opened 2020-02-18
V5 docs

fractalwrench:
Adds remaining Javadoc to the codebase.

This mainly adapts what was written on the docs site PR. There are a few cases (e.g. individual methods on `Logger`) where there was no equivalent doc, in which case I've written docs from scratch.


bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.kt
```diff
@@ -3,10 +3,31 @@ package com.bugsnag.android
 import java.io.IOException
 import java.util.Date
 
+/**
+ * In order to understand what happened in your application before each crash, it can be helpful
+ * to leave short log statements that we call breadcrumbs. The last several breadcrumbs are
```
tomlongridge:
```suggestion
 * to leave short log statements that we call breadcrumbs. Breadcrumbs are
```

bugsnag-android-core/src/main/java/com/bugsnag/android/OnBreadcrumbCallback.java
```diff
@@ -11,8 +11,8 @@
  * <p>
  * For example:
  * <p>
- * Bugsnag.beforeRecordBreadcrumb(new BeforeRecordBreadcrumb() {
- * public boolean shouldRecord(Breadcrumb breadcrumb) {
+ * Bugsnag.onBreadcrumb(new OnbreadcrumbCallback() {
```
tomlongridge:
```suggestion
 * Bugsnag.onBreadcrumb(new OnBreadcrumbCallback() {
```

PR #796, opened 2020-02-18
Make notifier publicly accessible on session/event

fractalwrench:
Makes the `notifier` property publicly accessible on `Session/EventPayload`. This allows users to access it if they have implemented a custom `Delivery`.


tomlongridge:
`DefaultDelivery` doesn't reference `Notifier` - why would another implementation need it?

fractalwrench:
A custom implementation of `Delivery` would likely use a completely different method for serialising JSON. We can get away without adding a getter/setter in our own JSON serializer implementation as the field can just be added directly.

PR #801, opened 2020-02-19
Clarify points in upgrade guide

fractalwrench:


twometresteve:
Changes as they are look fine, but I see a point was raised about the lack of Java example meaning it wasn't obvious that the type passed to the callback needs to changes.  It was marked as fixed in this PR, but I don't see an obvious corresponding change.

Cawllec:
Yes, that was my mistake, I still think we should either link to the Java example in the docs or add on here.

fractalwrench:
I've referenced the type of the callback here: https://github.com/bugsnag/bugsnag-android/pull/801/files#diff-8f9c1338eaf62f816d813b24392771cbR220

PR #803, opened 2020-02-20
Log a warning when null values are supplied to Configuration, rather than throwing an exception

fractalwrench:
## Goal

Logs a warning when null values are supplied to Configuration, rather than throwing an exception.

## Changeset

- Renamed `Configuration.kt` to `ConfigImpl.kt` and made it non-public
- Added `Configuration.java` which uses the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) on a `ConfigImpl` field and acts as the public interface
- Moved code documentation to `Configuration.java`

## Tests

Added unit tests to verify that the correct method on `ConfigImpl` is invoked when a valid value is set, and that a warning message is logged when null is passed.


bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+    public static Configuration load(@NonNull Context context) {
+        return ConfigImpl.load(context);
+    }
+
+    @NonNull
+    static Configuration load(@NonNull Context context, @NonNull String apiKey) {
+        return ConfigImpl.load(context, apiKey);
+    }
+
+    private void warn(String property) {
```
twometresteve: Following the guidance to the letter, this would log an error.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+    /**
+     * Sets a unique identifier for the app build to be included in all events sent to Bugsnag.
+     *
+     * This is used to identify proguard
+     * mapping files in the case that you publish multiple different apps with
+     * the same appId and versionCode. The default value is read from the
+     * com.bugsnag.android.BUILD_UUID meta-data field in your app manifest.
+     */
+    public void setBuildUuid(@Nullable String buildUuid) {
+        impl.setBuildUuid(buildUuid);
```
twometresteve: Could there be some merit in adding a short comment on each nullable property to say, e.g. "buildUuid is nullable".  Would just make future reviews easier and more obvious that a guard hasn't been missed off.

twometresteve: Oh hang on, maybe the annotation on the parameter is sufficient - ignore that.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+    }
+
+    /**
+     * Sets the maximum number of breadcrumbs which will be stored. Once the threshold is reached,
+     * the oldest breadcrumbs will be deleted.
+     *
+     * By default, 25 breadcrumbs are stored: this can be amended up to a maximum of 100.
+     */
+    public void setMaxBreadcrumbs(int maxBreadcrumbs) {
+        impl.setMaxBreadcrumbs(maxBreadcrumbs);
```
twometresteve: Is there logic downstream to limit this to within a range of 0 to 100?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+        return impl.getDiscardClasses();
+    }
+
+    /**
+     * Allows you to specify the fully-qualified name of error classes that will be discarded
+     * before being sent to Bugsnag if they are detected. The notifier performs an exact
+     * match against the canonical class name.
+     */
+    public void setDiscardClasses(@Nullable Set<String> discardClasses) {
+        if (discardClasses != null) {
```
twometresteve: Do we also need to check each String within the Set for null?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+    public Set<String> getEnabledReleaseStages() {
+        return impl.getEnabledReleaseStages();
+    }
+
+    /**
+     * By default, Bugsnag will be notified of events that happen in any releaseStage.
+     * If you would like to change which release stages notify Bugsnag you can set this property.
+     */
+    public void setEnabledReleaseStages(@Nullable Set<String> enabledReleaseStages) {
+        impl.setEnabledReleaseStages(enabledReleaseStages);
```
twometresteve: Set is nullable, but not the Strings within it.... do we need to check them (personally I think so, based only what I readon the Kotlin null safety page, even if it it is a nause)?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Yep, we have a custom setter here: https://github.com/bugsnag/bugsnag-android/blob/null-config/bugsnag-android-core/src/main/java/com/bugsnag/android/ConfigImpl.kt#L46

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Yep - we need to do this for all our collections. Should the behaviour be to not set the field in this case, or to carry on without that null value?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Resolving as this will be covered by the other comment above.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Updated to error log level.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+        return impl.getEndpoints();
+    }
+
+    /**
+     * Set the endpoints to send data to. By default we'll send error reports to
+     * https://notify.bugsnag.com, and sessions to https://sessions.bugsnag.com, but you can
+     * override this if you are using Bugsnag Enterprise to point to your own Bugsnag endpoints.
+     */
+    public void setEndpoints(@Nullable EndpointConfiguration endpoints) {
+        if (endpoints != null) {
```
twometresteve: This one is nullable, so the guard shouldn't be there.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+
+    /**
+     * Sets which values should be removed from any Metadata objects before
+     * sending them to Bugsnag. Use this if you want to ensure you don't send
+     * sensitive data such as passwords, and credit card numbers to our
+     * servers. Any keys which contain these strings will be filtered.
+     *
+     * By default, redactedKeys is set to "password"
+     */
+    public void setRedactedKeys(@Nullable Set<String> redactedKeys) {
```
twometresteve: Possibly overtaken by my other comment, but this and setDiscardClasses take a different approach than other nullable sets (setEnabledReleaseStages and setEnabledBreadcrumbTypes).

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationFacadeTest.java
```diff
@@ -0,0 +1,426 @@
...
+import org.junit.Test;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+@SuppressWarnings("ConstantConditions") // suppress warning about making redundant null checks
+public class ConfigurationFacadeTest {
```
twometresteve: Tests missing for removeOnError, removeOnBreadcrumb and removeOnSession.
More tediously, there's no coverage for the getters - but I can see why you've put assertions on the values from impl and I think we should keep them.  Covering the getters may feel like nugatory effort, but the benefit for us is that 100% coverage means we never need to look closer to check what is/isn't covered (and potentially we can also put a build check on it).  Having said that I suspect Alex already has some tests that would cover them.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+     * Returns a map of data in the specified section.
+     */
+    @Nullable
+    @Override
+    public Map<String, Object> getMetadata(@NonNull String section) {
+        if (section != null) {
+            return impl.getMetadata(section);
+        } else {
+            error("getMetadata");
+            return null;
```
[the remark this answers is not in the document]
fractalwrench: The spec states that if a section/value doesn't exist the notifier should return null - so this is an error state that users need to handle anyway.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,833 @@
...
+    @NonNull
+    public String getApiKey() {
+        return impl.getApiKey();
+    }
+
+    /**
+     * Changes the API key used for events sent to Bugsnag.
+     */
+    public void setApiKey(@NonNull String apiKey) {
+        if (apiKey != null) {
```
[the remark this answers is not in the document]
fractalwrench: Good catch - I've altered the implementation to just pass in the param instead, and `ConfigImpl` will validate everything.

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationFacadeTest.java
fractalwrench: Added tests for the missing methods. Assuming no change is required on how the getters are tested since we have some test coverage on them.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: This is because the underlying implementation in `Metadata` relies on there being non-null sets - I've updated this general area so that Sets now behave consistently.

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationFacadeTest.java
```diff
@@ -0,0 +1,514 @@
...
+
+    @Test
+    public void getMetadataValid() {
+        config.addMetadata("foo", "test", true);
+        assertTrue((Boolean) config.getMetadata("foo", "test"));
+    }
+
+    @Test
+    public void getMetadataInvalid() {
+        config.getMetadata(null);
```
twometresteve: assertNull on the return?

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,821 @@
...
+     *
+     * You can use the return type from the deliver functions to control the strategy for
+     * retrying the transmission at a later date.
+     *
+     * If DeliveryStatus.UNDELIVERED is returned, the notifier will automatically cache
+     * the payload and trigger delivery later on. Otherwise, if either DeliveryStatus.DELIVERED
+     * or DeliveryStatus.FAILURE is returned the notifier will removed any cached payload
+     * and no further delivery will be attempted.
+     */
+    public void setDelivery(@NonNull Delivery delivery) {
```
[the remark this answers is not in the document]
fractalwrench: It wouldn't make sense for users to set `Delivery` to null as no payloads would be delivered to our API. It is nullable in Kotlin as an implementation detail - we can't instantiate `DefaultDelivery` in `Configuration` because it requires a `Context`. This context object is supplied when a user initialises the notifier, so we then instantiate `Delivery` if nothing else has been set.

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationFacadeTest.java
```diff
@@ -0,0 +1,514 @@
...
+        config.setRedactedKeys(redactedKeys);
+        assertEquals(redactedKeys, config.impl.getRedactedKeys());
+        config.setRedactedKeys(null);
+        assertNull(config.impl.getRedactedKeys());
+    }
+
+    @Test
+    public void discardClassesValid() {
+        Set<String> discardClasses = new HashSet<>();
+        config.setDiscardClasses(discardClasses);
```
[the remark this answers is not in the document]
fractalwrench: Added something to the set (also for projectPackages)

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationFacadeTest.java
```diff
@@ -0,0 +1,514 @@
...
+    @Test
+    public void removeOnErrorValid() {
+        OnErrorCallback cb = new OnErrorCallback() {
+            @Override
+            public boolean onError(@NonNull Event event) {
+                return false;
+            }
+        };
+        config.addOnError(cb);
+        config.removeOnError(cb);
```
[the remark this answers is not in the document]
fractalwrench: Combined the two tests as suggested (also for session/breadcrumb callbacks)

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationFacadeTest.java
```diff
@@ -0,0 +1,514 @@
...
+
+    @Test
+    public void projectPackagesValid() {
+        Set<String> projectPackages = new HashSet<>();
+        config.setProjectPackages(projectPackages);
+        assertEquals(projectPackages, config.impl.getProjectPackages());
+    }
+
+    @Test
+    public void projectPackagesInvalid() {
```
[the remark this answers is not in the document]
fractalwrench: This property is non-null in the spec - I've updated the Kotlin property to reflect this.

bugsnag-android-core/src/test/java/com/bugsnag/android/InterceptingLogger.kt
```diff
@@ -0,0 +1,21 @@
+package com.bugsnag.android
+
+class InterceptingLogger: Logger {
```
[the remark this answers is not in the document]
fractalwrench: Agreed to stick with using a fake `Logger` after in-person discussion as there was no major advantage either way.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,821 @@
...
+
+    /**
+     * Sets which values should be removed from any Metadata objects before
+     * sending them to Bugsnag. Use this if you want to ensure you don't send
+     * sensitive data such as passwords, and credit card numbers to our
+     * servers. Any keys which contain these strings will be filtered.
+     *
+     * By default, redactedKeys is set to "password"
+     */
+    public void setRedactedKeys(@Nullable Set<String> redactedKeys) {
```
tomlongridge: I don't think this is quite as expected. `Nullable` indicates the user that `null` is a valid input and we accept it but do nothing.

We should mark this as `NonNull` and handle as such - making the underlying `redactedKeys` property also non-null.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,821 @@
...
+    public Set<String> getDiscardClasses() {
+        return impl.getDiscardClasses();
+    }
+
+    /**
+     * Allows you to specify the fully-qualified name of error classes that will be discarded
+     * before being sent to Bugsnag if they are detected. The notifier performs an exact
+     * match against the canonical class name.
+     */
+    public void setDiscardClasses(@Nullable Set<String> discardClasses) {
```
tomlongridge: This should be `NonNull` as per redacted keys.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,821 @@
...
+    /**
+     * Returns a map of data in the specified section.
+     */
+    @Nullable
+    @Override
+    public Map<String, Object> getMetadata(@NonNull String section) {
+        if (section != null) {
+            return impl.getMetadata(section);
+        } else {
+            error("getMetadata");
```
tomlongridge: I don't think logging makes sense for an accessor method.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,821 @@
...
+    /**
+     * Returns the value of the specified key in the specified section.
+     */
+    @Nullable
+    @Override
+    public Object getMetadata(@NonNull String section, @NonNull String key) {
+        if (section != null && key != null) {
+            return impl.getMetadata(section, key);
+        } else {
+            error("getMetadata");
```
tomlongridge: I don't think logging makes sense for an accessor method.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -135,7 +135,7 @@ public Client(@NonNull Context androidContext, @NonNull final Configuration conf
         storageManager = (StorageManager) appContext.getSystemService(Context.STORAGE_SERVICE);
 
         // Set up breadcrumbs
-        callbackState = configuration.callbackState.copy();
```
tomlongridge: I was expecting `ConfigImpl` to be used everywhere internally (rather than `Configuration`) - this is the advantage of retaining the Kotlin class: there should be no changes internally.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
tomlongridge: I'd like to see all validation applied in this class so it can be handled consistently and there's only one place to maintain it. `ConfigImpl` then need not worry about any sort of input validation.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: We need to indicate that something went wrong. How should we do this?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: This is not my understanding of the decorator pattern or the design that was sketched out, perhaps we can discuss what is required.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: This depends on what we want to use the decorator pattern for - I assumed that the sole responsibility of this class should be to check for null references.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
twometresteve: I don't have a problem with logging here - checks are based on how the parameters are annotated rather than what the method is for.  We state that the parameter should not be null, so to me it consistent to log an error and noop if this is broken.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
twometresteve: As above, we're validating values passed to all of our public API methods, whether they are setters, getters, or otherwise.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
tomlongridge: I think the main reason for logging was to inform the user that a mutation didn't occur as they expected. I'm not sure this applies to a getter, but I'm not too concerned as it applies the rule more consistently :)

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
tomlongridge: As discussed IRL, we'll rename `ConfigImpl` to `ConfigInternal` and use `Configuration` to do _all_ validation but use `ConfigInternal` everywhere internally - this should avoid any change wider in library and maintain a more pure Kotlin class.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: From in-person discussion we decided to alter this so that if _any_ value in a collection is null, the whole parameter is treated as null

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Updated the two properties to be non-null.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Moved breadcrumb validation to this class.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
tomlongridge: Client should use the Internal class everywhere internally so that we get the benefits of nullability checks.

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -90,13 +90,14 @@ internal fun convertToImmutableConfig(config: Configuration): ImmutableConfig {
         else -> ErrorTypes(false)
     }
 
+    @Suppress("UNNECESSARY_SAFE_CALL", "USELESS_ELVIS")
```
tomlongridge: I hope all changes in this file aren't required if we're using ConfigInternal everywhere internally.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,836 @@
...
+    public EndpointConfiguration getEndpoints() {
+        return impl.getEndpoints();
+    }
+
+    /**
+     * Set the endpoints to send data to. By default we'll send error reports to
+     * https://notify.bugsnag.com, and sessions to https://sessions.bugsnag.com, but you can
+     * override this if you are using Bugsnag Enterprise to point to your own Bugsnag endpoints.
+     */
+    public void setEndpoints(@Nullable EndpointConfiguration endpoints) {
```
tomlongridge: I think `endpoints` should be non-nullable.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -0,0 +1,836 @@
...
+     * Sets the maximum number of breadcrumbs which will be stored. Once the threshold is reached,
+     * the oldest breadcrumbs will be deleted.
+     *
+     * By default, 25 breadcrumbs are stored: this can be amended up to a maximum of 100.
+     */
+    public void setMaxBreadcrumbs(int maxBreadcrumbs) {
+        if (maxBreadcrumbs >= MIN_BREADCRUMBS && maxBreadcrumbs <= MAX_BREADCRUMBS) {
+            impl.setMaxBreadcrumbs(maxBreadcrumbs);
+        } else {
+            getLogger().e("Invalid null value supplied to config.maxBreadcrumbs, expected "
```
tomlongridge: There's an error message format in the spec - see Configuration page.

bugsnag-android-core/src/main/java/com/bugsnag/android/ConfigInternal.kt
```diff
@@ -0,0 +1,113 @@
...
+
+import android.content.Context
+import java.util.Collections
+
+internal class ConfigInternal(apiKey: String) : CallbackAware, MetadataAware, UserAware {
+
+    var apiKey: String = apiKey
+        set(value) {
+            require(value.matches(API_KEY_REGEX.toRegex())) { "You must provide a valid Bugsnag API key" }
+            field = value
```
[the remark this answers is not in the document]
fractalwrench: Moved to `Configuration.java`

bugsnag-android-core/src/main/java/com/bugsnag/android/ConfigInternal.kt
```diff
@@ -0,0 +1,113 @@
...
+    var sendThreads: Thread.ThreadSendPolicy = Thread.ThreadSendPolicy.ALWAYS
+    var persistUser: Boolean = false
+
+    var launchCrashThresholdMs: Long = DEFAULT_LAUNCH_CRASH_THRESHOLD_MS
+        set(launchCrashThresholdMs) {
+            field = when {
+                launchCrashThresholdMs <= MIN_LAUNCH_CRASH_THRESHOLD_MS -> MIN_LAUNCH_CRASH_THRESHOLD_MS
+                else -> launchCrashThresholdMs
+            }
+        }
```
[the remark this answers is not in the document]
fractalwrench: Moved to `Configuration.java`

bugsnag-android-core/src/main/java/com/bugsnag/android/ConfigInternal.kt
```diff
@@ -0,0 +1,113 @@
...
+    var redactedKeys: Set<String>? = metadataState.metadata.redactedKeys
+        set(value) {
+            if (value != null) {
+                metadataState.metadata.setRedactedKeys(value)
+            }
+            field = value
+        }
+
+    init {
+        require(apiKey.matches(API_KEY_REGEX.toRegex())) { "You must provide a valid Bugsnag API key" }
```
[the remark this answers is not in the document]
fractalwrench: Moved to `Configuration.java`

bugsnag-android-core/src/main/java/com/bugsnag/android/ConfigInternal.kt
```diff
@@ -0,0 +1,113 @@
...
+    var autoDetectErrors: Boolean = true
+    var codeBundleId: String? = null
+    var appType: String? = "android"
+    var logger: Logger? = null
+    var delivery: Delivery? = null
+    var endpoints: EndpointConfiguration = EndpointConfiguration()
+    var maxBreadcrumbs: Int = DEFAULT_MAX_SIZE
+    var context: String? = null
+
+    var redactedKeys: Set<String>? = metadataState.metadata.redactedKeys
```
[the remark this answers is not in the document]
fractalwrench: Updated to be non-nullable.

bugsnag-android-core/src/main/java/com/bugsnag/android/ConfigInternal.kt
```diff
@@ -0,0 +1,113 @@
...
+            // or bugsnag-android-ndk
+            val clz = Class.forName("com.bugsnag.android.BuildConfig")
+            val field = clz.getDeclaredField("AUTO_DETECT_NDK_CRASHES")
+            field.getBoolean(null)
+        } catch (exc: Throwable) {
+            false
+        }
+    }
+
+    var discardClasses: Set<String>? = emptySet()
```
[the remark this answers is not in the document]
fractalwrench: Updated to be non-nullable.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: Updated setter validation and test to assume this property is non-null.

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
fractalwrench: updated to follow message format

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
fractalwrench: These changes have been reverted as they are no longer necessary.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: I don't think this is necessary - we convert `Configuration` to an `ImmutableConfig` object whose properties cannot be mutated and are non-null where possible. We only need to access `Configuration` directly at the very start of the `Client` constructor and none of the properties used would gain any type-safety benefit from using `ConfigInternal` instead.

fractalwrench:
Addressed latest review feedback

PR #806, opened 2020-02-20
Add android 8.1 and 10

twometresteve:
## Goal

Add Android 8.1 and 10.0 Maze Runner tests, as well as Instrumentation tests on Android 10 to the CI build.

## Design

Following existing pattern for builds, except using the "soft fail" functionality of BuildKite for the Maze Runner tests.  We know there are issues causing some tests to fail and these need more investigation in due course, but for now we would rather receive the feedback from running them and failing than having nothing at all.

## Changeset

Pipeline build updated to include new versions.
Logic for determining skipping of tests on specific versions of Android updated to account for new device names in Maze Runner.

## Tests

Verified by inspecting the build output.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -325,3 +325,15 @@ steps:
       NDK_VERSION: "r19"
     concurrency: 5
     concurrency_group: 'browserstack-app'
+
+  - label: ':android: NDK 21 SDK 10.0 Instrumentation tests'
+    timeout_in_minutes: 60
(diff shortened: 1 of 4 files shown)


twometresteve:
> LGTM. After a week or so of builds we should evaluate how often the new jobs are failing and move them over to a hard fail or fix any flakiness. Do we have an internal ticket to track this action?

Yes, PLAT-4018 and PLAT-4020 should cover 8.1 and 10 respectively.

PR #807, opened 2020-02-21
Update test assertions against v5 changes

Cawllec:
## Goal

Adds test coverage to previously untested areas relating to ongoing v5 changes.
These include:
- Clearing metadata on the client
- Behaviour of breadcrumb callbacks
- Adding and clearing configuration metadata
- Event object error and unhandled states
- PersistUser configuration option

## Design

Where possible additions were made at the lowest test level practical, the JVM layer.  Tests that required more complex interaction between elements of the notifier and the OS were made at the Instrumentation layer, and tests that required multiple runs of a test application or object were added to maze-runner.


tests/features/session_tracking.feature
```diff
@@ -27,3 +27,37 @@ Scenario: Set Auto Capture Sessions sends
     When I run "SessionSetAutoCaptureScenario"
     And I wait to receive a request
     Then the request is valid for the session reporting API version "1.0" for the "Android Bugsnag Notifier" notifier
+
+Scenario: User is persisted between sessions
```
fractalwrench: Do these scenarios also need to be added to the local feature which uses mazerunner v1?

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationTest.java
```diff
@@ -180,4 +181,45 @@ public void testVersionCode() {
...
+        assertNull(configuration.getMetadata("keyVal", "foo"));
+        assertEquals("val1", configuration.getMetadata("mapVal", "key1"));
+        assertNull(configuration.getMetadata("mapVal", "key2"));
+    }
+
+    @Test
+    public void testSetUser() {
+        Configuration configuration = generateConfiguration();
+        configuration.setUser("24601", "m@rp.fr", "MM");
+        assertEquals(User.class, configuration.getUser().getClass());
```
fractalwrench: This assertion could be removed as it seems to test the Java type system rather than our code

tests/features/session_tracking.feature
Cawllec: This should be added in now

bugsnag-android-core/src/test/java/com/bugsnag/android/ConfigurationTest.java
```diff
@@ -180,4 +181,44 @@ public void testVersionCode() {
         configuration.setVersionCode(577);
         assertEquals(577, (int) configuration.getVersionCode());
     }
+
+    @Test
+    public void testAddMetadata() {
```
twometresteve: I guess just a head's up that some rationalisation will be needed whenever the changes meet the null check changes that Jamie currently has in review.

Cawllec: I don't believe the additional null checking should invalidate any current behaviour, but we can hold this PR until the change has been merged and re-review if necessary.

twometresteve: Sorry, didn't mean any behaviour would be invalidated, it's just that he's created a new ConfigurationFacadeTest.java file that we may want to merge with.
https://github.com/bugsnag/bugsnag-android/pull/803/files#diff-9bae5430ee1c302bbe1fcbe38404051d

bugsnag-android-core/src/test/java/com/bugsnag/android/BreadcrumbStateTest.kt
```diff
@@ -121,8 +121,50 @@ class BreadcrumbStateTest {
...
+        assertEquals(breadcrumb, breadcrumbState.store.peek())
+    }
+
+    /**
+     * Verifies that returning false in one callback will halt subsequent callbacks and not apply breadcrumb
+     */
+     @Test
+     fun testOnBreadcrumbCallbackFalse() {
+        val breadcrumb = Breadcrumb("Whoops")
+        breadcrumbState.callbackState.addOnBreadcrumb(OnBreadcrumbCallback { givenBreadcrumb ->
```
twometresteve: Assuming I understand the requirements correctly, a slightly better test would be to have an extra callback before this that returns true, verifying that a single entry makes it into the store.

Cawllec:
Made a couple of changes, as well as merging the loading config maze-runner tests into this branch.  At the moment the `BuildUUID` assertion is commented out in the manifest test, but should be fully runnable once https://github.com/bugsnag/bugsnag-android-gradle-plugin/commit/7ff9dc92d84b176c222ee6c5bffc178857975c5d is released.

tests/features/fixtures/mazerunner/build.gradle
```diff
@@ -26,7 +26,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.4.2'
-        classpath 'com.bugsnag:bugsnag-android-gradle-plugin:4.5.2'
+        classpath 'com.bugsnag:bugsnag-android-gradle-plugin:4+'
```
fractalwrench: This change would need to be `4.+` to pickup the dependency. Is there a benefit to bumping it to 4.7.3 instead though? It feels like it'd be good to keep a specific version of the plugin to avoid introducing an extra moving part to CI.

tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/LoadConfigurationNullsScenario.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.bugsnag.android.OnErrorCallback
+import java.lang.RuntimeException
+
+
+internal class LoadConfigurationNullsScenario(
+    config: Configuration,
+    context: Context
+) : Scenario(config, context) {
+
+    init {
```
fractalwrench: nit: this empty init block can be removed

tests/features/load_configuration.feature
```diff
@@ -0,0 +1,43 @@
...
+    And I run "LoadConfigurationFromManifestScenario"
+    Then I wait to receive a request
+    And the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier with the apiKey "abc12312312312312312312312312312"
+    And the exception "message" equals "LoadConfigurationFromManifestScenario"
+    And the event "app.releaseStage" equals "testing"
+    And the payload field "events.0.breadcrumbs" is an array with 1 elements
+    And the event "metaData.test.foo" equals "bar"
+    And the event "metaData.test.filter_me" equals "[REDACTED]"
+    And the event "app.versionCode" equals 753
+    #And the event "app.buildUUID" equals "test-7.5.3"
```
fractalwrench: We should create an internal ticket to enable this again once the gradle plugin fix has been released

tests/features/fixtures/mazerunner/build.gradle
Cawllec: Sounds good.  If there's an internal ticket to reactivate assertions around buildUUID this can be updated at the same time.

tests/features/load_configuration.feature
```diff
@@ -0,0 +1,43 @@
...
+    And the payload field "events.0.threads" is a non-empty array
+
+Scenario: Load configuration initialised with nulls
+    When I configure the app to run in the "skipBugsnag" state
+    And I run "LoadConfigurationNullsScenario"
+    Then I wait to receive a request
+    And the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier with the apiKey "12312312312312312312312312312312"
+    And the exception "message" equals "LoadConfigurationNullsScenario"
+    And the event "metaData.test.foo" equals "bar"
+    And the event "metaData.test.filter_me" equals "foobar"
```
twometresteve: Need more expectations that fields are not present/null/set to defaults to reflect the setting of nulls in the config.

Cawllec: I agree, however it's likely the behaviour will change with the `null` -> `no-op` update.
In that case, I can either set expectations at this point based on how I think it'll look, or wait until I know how it should work.

twometresteve: Ok, that makes sense.  Could be a case for adding a "# TODO blah" comment in the meantime, just to guard a little against forgetting?

Cawllec:
I've made several changes since the last reviews, it would be good to halt the work on this branch now and merge if able.  @twometresteve @fractalwrench would you be able to take a quick look?

tests/features/load_configuration.feature
fractalwrench: Can this comment be resolved now?

tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXConfigurationMetadataNativeCrashScenario.java
```diff
@@ -0,0 +1,37 @@
+package com.bugsnag.android.mazerunner.scenarios;
+
+import com.bugsnag.android.Bugsnag;
+import com.bugsnag.android.Configuration;
+
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+public class CXXConfigurationMetadataNativeCrashScenario extends Scenario {
```
fractalwrench: There doesn't seem to be corresponding scenarios for any of the `features/fixtures` test suite. If we want to make a decision to stop writing tests for that suite and focus on the browser tests instead I'd be ok with that, as long as it's documented somewhere in the repo/confluence.

tests/features/support/env.rb
```diff
@@ -28,6 +28,10 @@
   skip_this_scenario("Skipping scenario") unless %w[ANDROID_8_0 ANDROID_8_1 ANDROID_9_0 ANDROID_10_0].include? bs_device
 end
 
+Before('@skip_android_8_1') do |scenario|
+  skip_this_scenario("Skipping scenario") unless %w[ANDORID_8_1].include? bs_device
```
twometresteve: I think you what an "if", not "unless".  But I'd also be interested to know why we want to skip on 8.1 specifically (it is soft fail anyway, for example).

Cawllec: Due to the timings of the ANRs triggered on the Samsung Galaxy device we're using for 8.1 they will never pass our current tests.  While we could leave it to soft fail, as this is a known issue I'm comfortable with skipping this particular combination and allowing the test to run the through the rest of the scenarios.

twometresteve: Makes sense, could we just write a short comment to that effect?  Especially as it's due to the specific device, as opposed to 8.1 in general.

PR #808, opened 2020-02-21
Log a warning when null values are supplied to Breadcrumb, rather than throwing an exception

fractalwrench:
## Goal

Logs a warning when null values are supplied to Breadcrumb, rather than throwing an exception.

## Changeset

- Renamed `Breadcrumb.kt` to `BreadcrumbImpl.kt` and made it non-public
- Added `Breadcrumb.java` which uses the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) on a `BreadcrumbImpl` field and acts as the public interface
- Moved code documentation to `Breadcrumb.java`
- Passed in `logger` as required to `Breadcrumb` constructor
- Updated breadcrumb properties to be mutable on `Breadcrumb.kt`

## Tests

Added unit tests to verify that the correct method on `BreadcrumbImpl` is invoked when a valid value is set, and that a warning message is logged when null is passed.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
@@ -1,0 +1,78 @@
+package com.bugsnag.android;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.Date;
+import java.util.Map;
+
+@SuppressWarnings("ConstantConditions")
+public class Breadcrumb implements JsonStream.Streamable {
+
+    private final BreadcrumbImpl impl;
+    private final Logger logger;
+
+    Breadcrumb(@NonNull String message, @NonNull Logger logger) {
+        this.impl = new BreadcrumbImpl(message);
+        this.logger = logger;
+    }
+
+    Breadcrumb(@NonNull String message,
+               @NonNull BreadcrumbType type,
+               @Nullable Map<String, Object> metadata,
+               @NonNull Date timestamp,
+               @NonNull Logger logger) {
+        this.impl = new BreadcrumbImpl(message, type, metadata, timestamp);
+        this.logger = logger;
+    }
+
+    private void error(String property) {
+        logger.e("Invalid null value supplied to breadcrumb." + property + ", ignoring");
+    }
+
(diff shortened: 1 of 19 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
```diff
@@ -0,0 +1,101 @@
...
+    Breadcrumb(@NonNull String message,
+               @NonNull BreadcrumbType type,
+               @Nullable Map<String, Object> metadata,
+               @NonNull Date timestamp,
+               @NonNull Logger logger) {
+        this.impl = new BreadcrumbImpl(message, type, metadata, timestamp);
+        this.logger = logger;
+    }
+
+    private void error(String property) {
```
tomlongridge:
```suggestion
    private void logNull(String property) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
```diff
@@ -0,0 +1,101 @@
...
+    @NonNull
+    public BreadcrumbType getType() {
+        return impl.getType();
+    }
+
+    /**
+     * Sets diagnostic data relating to the breadcrumb
+     */
+    public void setMetadata(@NonNull Map<String, Object> metadata) {
+        impl.setMetadata(metadata);
```
tomlongridge: This is Nullable I believe?

bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.kt
```diff
@@ -8,26 +8,10 @@ import java.util.Date
...
-
-    /**
-     * Diagnostic data relating to the breadcrumb
-     */
-    val metadata: MutableMap<String, Any?>,
-
-    /**
-     * The timestamp that the breadcrumb was left
-     */
+internal class BreadcrumbImpl internal constructor(
```
tomlongridge:
```suggestion
internal class BreadcrumbInternal internal constructor(
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Breadcrumb.java
fractalwrench: Yes - I've now updated the implementation to account for this.

PR #809, opened 2020-02-21
Log a warning when null values are supplied to Thread, rather than throwing an exception

fractalwrench:
## Goal

Logs a warning when null values are supplied to Thread, rather than throwing an exception.

## Changeset

- Renamed `Thread.kt` to `ThreadImpl.kt` and made it non-public
- Added `Thread.java` which uses the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) on a `BreadcrumbImpl` field and acts as the public interface
- Moved code documentation to `Thread.java`
- Passed in `logger` as required to `Thread` constructor
- Updated properties to be mutable on `Thread.kt`
- Moved `Thread.Type` and `Thread.ThreadSendPolicy` to live in separate Kotlin files, renamed usages accordingly

## Tests

Added unit tests to verify that the correct method on `BreadcrumbImpl` is invoked when a valid value is set, and that a warning message is logged when null is passed.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Thread.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Thread.java
@@ -1,0 +1,74 @@
+package com.bugsnag.android;
+
+import androidx.annotation.NonNull;
+
+import java.io.IOException;
+
+@SuppressWarnings("ConstantConditions")
+public class Thread implements JsonStream.Streamable {
+
+    private final ThreadImpl impl;
+    private final Logger logger;
+
+    Thread(
+            long id,
+            @NonNull String name,
+            @NonNull ThreadType type,
+            boolean errorReportingThread,
+            @NonNull Stacktrace stacktrace,
+            @NonNull Logger logger) {
+        this.impl = new ThreadImpl(id, name, type, errorReportingThread, stacktrace);
+        this.logger = logger;
+    }
+
+    private void error(String property) {
+        logger.e("Invalid null value supplied to thread." + property + ", ignoring");
+    }
+
+    public void setId(long id) {
+        impl.setId(id);
+    }
+
+    public long getId() {
+        return impl.getId();
+    }
+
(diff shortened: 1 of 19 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Thread.java
```diff
@@ -0,0 +1,121 @@
...
+            @NonNull String name,
+            @NonNull ThreadType type,
+            boolean errorReportingThread,
+            @NonNull Stacktrace stacktrace,
+            @NonNull Logger logger) {
+        this.impl = new ThreadImpl(id, name, type, errorReportingThread, stacktrace);
+        this.logger = logger;
+    }
+
+    private void error(String property) {
```
tomlongridge:
```suggestion
    private void logNull(String property) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Thread.java
```diff
@@ -0,0 +1,121 @@
...
+     * Gets whether the thread was the thread that caused the event
+     */
+    public boolean getErrorReportingThread() {
+        return impl.isErrorReportingThread();
+    }
+
+    /**
+     * Sets a representation of the thread's stacktrace
+     */
+    public void setStacktrace(@NonNull List<Stackframe> stacktrace) {
```
tomlongridge: This should be sanitized for null elements

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadImpl.kt
```diff
@@ -0,0 +1,32 @@
+package com.bugsnag.android
+
+import java.io.IOException
+
+class ThreadImpl internal constructor(
```
tomlongridge:
```suggestion
class ThreadInternal internal constructor(
```

fractalwrench:
Addressed review feedback

PR #810, opened 2020-02-21
Log a warning when null values are supplied to Error, rather than throwing an exception

fractalwrench:
## Goal

Logs a warning when null values are supplied to Error, rather than throwing an exception.

## Changeset

- Renamed `Error.kt` to `ErrorImpl.kt` and made it non-public
- Added `Error.java` which uses the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) on a `ErrorImpl` field and acts as the public interface
- Moved code documentation to `Error.java`
- Passed in `logger` as required to `Error` constructor
- Moved `Error.Type` to its own file

## Tests

Added unit tests to verify that the correct method on `ErrorImpl` is invoked when a valid value is set, and that a warning message is logged when null is passed.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
@@ -1,0 +1,76 @@
+package com.bugsnag.android;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+
+@SuppressWarnings("ConstantConditions")
+public class Error implements JsonStream.Streamable {
+
+    private final ErrorImpl impl;
+    private final Logger logger;
+
+    Error(@NonNull ErrorImpl impl,
+          @NonNull Logger logger) {
+        this.impl = impl;
+        this.logger = logger;
+    }
+
+    private void error(String property) {
+        logger.e("Invalid null value supplied to error." + property + ", ignoring");
+    }
+
+    public void setErrorClass(@NonNull String errorClass) {
+        if (errorClass != null) {
+            impl.setErrorClass(errorClass);
+        } else {
+            error("errorClass");
+        }
+    }
+
+    @NonNull
+    public String getErrorClass() {
+        return impl.getErrorClass();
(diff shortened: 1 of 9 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -0,0 +1,101 @@
...
+    private final ErrorImpl impl;
+    private final Logger logger;
+
+    Error(@NonNull ErrorImpl impl,
+          @NonNull Logger logger) {
+        this.impl = impl;
+        this.logger = logger;
+    }
+
+    private void error(String property) {
```
tomlongridge:
```suggestion
    private void logNull(String property) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorImpl.kt
```diff
@@ -0,0 +1,32 @@
+package com.bugsnag.android
+
+internal class ErrorImpl @JvmOverloads internal constructor(
```
tomlongridge:
```suggestion
internal class ErrorInternal @JvmOverloads internal constructor(
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Error.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Gets a representation of the stacktrace
+     */
+    @NonNull
+    public List<Stackframe> getStacktrace() {
+        return impl.getStacktrace();
+    }
+
+    @Override
+    public void toStream(@NonNull JsonStream stream) throws IOException {
```
twometresteve: IntelliJ reckons that throws IOException is unnecessary?

fractalwrench: I believe IntelliJ is wrong in this case as this method calls into Kotlin code which can ultimately throw an IOException.

twometresteve: Not overly surprised and happy to ignore it.

PR #811, opened 2020-02-21
Log a warning when null values are supplied to Session

fractalwrench:
## Goal

Logs a warning when null values are supplied to Session, rather than allowing an NPE if the user invokes a getter.

## Changeset

- If a null value is passed to a non-null option, a warning is now logged
- Passed in `logger` as required to `Session` constructors

## Tests

Added unit tests to verify that the correct method on `Session` is invoked when a valid value is set, and that a warning message is logged when null is passed.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
@@ -12,12 +12,14 @@ import java.util.concurrent.atomic.AtomicInteger;
 /**
  * Represents a contiguous session in an application.
  */
+@SuppressWarnings("ConstantConditions")
 public final class Session implements JsonStream.Streamable, UserAware {
 
     private final File file;
     private String id;
     private Date startedAt;
     private User user;
+    private final Logger logger;
     private App app;
     private Device device;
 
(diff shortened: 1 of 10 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
```diff
@@ -28,30 +30,37 @@
...
         this.tracked.set(true);
     }
 
-    Session(File file) {
+    Session(File file, Logger logger) {
         this.file = file;
+        this.logger = logger;
+    }
+
+    private void error(String property) {
```
tomlongridge:
```suggestion
    private void logNull(String property) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
```diff
@@ -12,12 +12,14 @@
 /**
  * Represents a contiguous session in an application.
  */
+@SuppressWarnings("ConstantConditions")
 public final class Session implements JsonStream.Streamable, UserAware {
```
tomlongridge: Shouldn't this have a `SessionInternal` backing class as per the others?

fractalwrench: The `Session` class has not been implemented in Kotlin so there's no need to do this.

PR #812, opened 2020-02-24
Log a warning when null values are supplied to Event, rather than throwing an exception

fractalwrench:
## Goal

Logs a warning when null values are supplied to Event, rather than throwing an exception.

## Changeset

- Renamed `Event.kt` to `EventInternal.kt` and made it non-public
- Added `Event.java` which uses the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) on a `EventInternal` field and acts as the public interface
- Moved code documentation to `Event.java`
- Altered errors/breadcrumbs/threads on `Event` to be a readonly mutable list, as this better reflects the spec as it is currently written

## Tests

Added unit tests to verify that the correct method on `EventInternal` is invoked when a valid value is set, and that a warning message is logged when null is passed.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
@@ -1,0 +1,220 @@
+package com.bugsnag.android;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings("ConstantConditions")
+public class Event implements JsonStream.Streamable, MetadataAware, UserAware {
+
+    final EventImpl impl;
+    private final Logger logger;
+
+    Event(@Nullable Throwable originalError,
+          @NonNull ImmutableConfig config,
+          @NonNull HandledState handledState,
+          @NonNull Logger logger) {
+        this(originalError, config, handledState, new Metadata(), logger);
+    }
+
+    Event(@Nullable Throwable originalError,
+          @NonNull ImmutableConfig config,
+          @NonNull HandledState handledState,
+          @NonNull Metadata metadata,
+          @NonNull Logger logger) {
+        this(new EventImpl(originalError, config, handledState, metadata), logger);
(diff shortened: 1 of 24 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
```diff
@@ -0,0 +1,325 @@
...
+    protected boolean shouldDiscardClass() {
+        return impl.shouldDiscardClass();
+    }
+
+    protected void updateSeverityInternal(@NonNull Severity severity) {
+        impl.updateSeverityInternal(severity);
+    }
+
+    void setApp(@NonNull AppWithState app) {
+        impl.app = app;
```
tomlongridge: Is there reason behind how you access this wrt to property or accessor function? Perhaps we should change all these uses to getters/setters?

bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
```diff
@@ -0,0 +1,325 @@
...
+
+/**
+ * An Event object represents a Throwable captured by Bugsnag and is available as a parameter on
+ * an {@link OnErrorCallback}, where individual properties can be mutated before an error report is
+ * sent to Bugsnag's API.
+ */
+@SuppressWarnings("ConstantConditions")
+public class Event implements JsonStream.Streamable, MetadataAware, UserAware {
+
+    final EventInternal impl;
```
tomlongridge: Shall we call this `_event` or something similar to indicate internal-ness?

bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
fractalwrench: I've altered this to use the setter for app/device.

bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
fractalwrench: Agreed to leave this as-is since `_event` would go against our checkstyle rules.

PR #816, opened 2020-02-26
Amend internal use of Configuration to ConfigInternal

tomlongridge:
I think we should use ConfigInternal everywhere inside the library to benefit from it being a Kotlin class. Therefore Client should convert it and santitise it to give defaults early on and nothing else should access it.

TODO - the tests probably need looking at to ensure they are testing the right thing. I wanted to get opinion on this before spending too much time on this.


tomlongridge:
Agreed to close for now and readdress structure in future.

PR #822, opened 2020-02-27
feat: transfer remaining bugsnag-js implementation to repo

fractalwrench:
## Goal

Transfers the remaining Android code from bugsnag-js to this repository, following on from #821. 

## Changeset

- Automatically initialises the `bugsnag-plugin-react-native` plugin if the class is on the classpath
- Moved `leaveBreadcrumb/configure/getPayloadInfo` implementations to `BugsnagReactNativePlugin`
- Reduce visibility modifier of serializers and other internal classes
- Resolve lint warnings (as `InternalHooks` is no longer publicly visible)
- Added method to the `NativeInterface` for retrieving an instantiated plugin (used by `BugsnagReactNative` in the JS repo)

## Tests

Migrated existing unit tests and added a new unit test for the `BugsnagReactNativePlugin`. Additionally checked that the example app sends reports.


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -0,0 +1,104 @@
+package com.bugsnag.android
+
+import java.util.Locale
+
+@Suppress("unused", "UNUSED_PARAMETER")
```
tomlongridge: Move to `configure` to avoid missing other valid warnings in the class.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -0,0 +1,104 @@
...
+class BugsnagReactNativePlugin : Plugin {
```
tomlongridge: Some Javadoc here would be good to set the scene.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    fun configure(): Map<String, Any?> {
+        try {
+            // see if bugsnag-android is already initialised
+            client = Bugsnag.getClient()
+            internalHooks = InternalHooks(client)
+            logger = client.logger
+        } catch (exc: IllegalStateException) {
+            throw exc
+        }
```
tomlongridge: This doesn't appear to do anything.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    lateinit var logger: Logger
+
+    override fun load(client: Client) {
+        client.logger.i("Initialized React Native Plugin")
+    }
+
+    override fun unload() {}
+
+    fun configure(): Map<String, Any?> {
```
tomlongridge: Can these configure steps to be done in the `load` function? Then expose a `getConfig` method for the map?

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -384,4 +385,19 @@ public static Event createEvent(@NonNull Throwable exc,
     public static Logger getLogger() {
         return getClient().getConfig().getLogger();
     }
+
+    /**
+     * Retrieves an instantiated plugin of the given type, or null if none has been created
+     */
+    @SuppressWarnings("unchecked")
+    @Nullable
+    public static <T extends Plugin> T getPlugin(@NonNull Class<T> clz) {
```
tomlongridge: I think we should consider making a getPlugin method in Client and simply using the interface class to wrap it. This avoids anything else need to know about the PluginClient and having to open up the the `pluginClient` and `plugins` member variables for access here.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
fractalwrench: This is part of the previous implementation in the bugsnag-js repo. I agree that it's redundant but I left it unchanged for now to make the changeset easier to review.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
fractalwrench: I don't think we should do this as it complicates the implementation and effectively adds two entrypoints that could interact in unexpected ways. `configure()` will always be the first method called by the bugsnag-js layer so in my mind it makes sense to gather information at the time we need it. This is also part of the design that has been reviewed + approved so I think any architectural change here would need re-review from everyone to make sure it works on all platforms and scenarios.

bugsnag-android-core/src/main/java/com/bugsnag/android/PluginClient.kt
```diff
@@ -28,7 +29,7 @@ internal class PluginClient(
             val pluginClz = Class.forName(clz)
             pluginClz.newInstance() as Plugin
         } catch (exc: Throwable) {
-            logger.w("Plugin '$clz' is not on the classpath - functionality will not be enabled.")
+            logger.d("Plugin '$clz' is not on the classpath - functionality will not be enabled.")
```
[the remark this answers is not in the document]
fractalwrench: This sounds reasonable as it's an expected condition that certain plugins (such as the BugsnagReactNativePlugin) won't be on the classpath by default. If the class is present but fails to load then logging an error message is sensible.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: I'm happy to go with the approach of adding a `getPlugin()` method to `Client`. Although I'm a bit wary in a general sense that there's a danger of us adding _every_ method like this onto `Client`, which is already quite a large class.

tomlongridge: Agreed on the general point, although I feel like in this case it's reasonable.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -84,6 +84,7 @@
     final DeliveryDelegate deliveryDelegate;
 
     final ClientObservable clientObservable = new ClientObservable();
+    PluginClient pluginClient;
```
tomlongridge:
```suggestion
    private PluginClient pluginClient;
```

PR #825, opened 2020-02-28
Remove armeabi from default build

tomlongridge:
## Goal

As per #615 - removes the now unsupported "armeabi" ABI from the default build.

## Changeset

* bugsnag-plugin-android-anr/build.gradle - removed "armeabi" from default list of targets
* bugsnag-plugin-android-ndk/build.gradle - removed "armeabi" from default list of targets
* examples/sdk-app-example/build.gradle - removed "armeabi" from default list of targets and added ABI_FILTERS property for consistency with main build
* Makefile - added full list of ABIs for inclusion when making a release
* gradle/release.gradle - added check during before assembleRelease to throw exception if the NDK library version or ABI list is unexpected (supports both Android and Unity ABI lists). 

## Tests

Run locally to verify validation check, CI will show no difference without "armeabi".

### DIFF
--- a/gradle/release.gradle
+++ b/gradle/release.gradle
@@ -11,7 +11,7 @@ def expectedNdkVersion = "16" // Controls which NDK library major version should
 
 def expectedAbiLists = // Controls which combinations of ABIs are expected for release
 [
-    ["arm64-v8a", "armeabi-v7a", "x86", "x86_64"] as Set,
+    ["arm64-v8a", "armeabi", "armeabi-v7a", "x86", "x86_64"] as Set,
     ["armeabi-v7a", "x86"] as Set
 ]
 
@@ -68,7 +68,7 @@ task validatePublishOptions() {
 }
 
 tasks.whenTaskAdded { task ->
-    if (task.name == 'assembleRelease') {
+    if ((task.name == 'assembleRelease') || (task.name == 'publish') || (task.name == 'bintrayUpload')) {
         task.dependsOn validatePublishOptions
     }
 }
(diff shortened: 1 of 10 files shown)


gradle/release.gradle
```diff
@@ -30,6 +30,24 @@ task sourceJar(type: Jar) {
...
+allprojects { proj ->
+    afterEvaluate {
+        validatePublishedAbis(proj)
+    }
+}
+
+// Throws an exception if the build is not targeting the required ABIs for publishing
+private void validatePublishedAbis(proj) {
+    if (proj.android.defaultConfig.ndk.abiFilters != null) {
+        for (String abi : ["arm64-v8a", "armeabi", "armeabi-v7a", "x86", "x86_64"]) {
```
fractalwrench: I've just realised that checking the ABIs like this will break the Unity build: https://github.com/bugsnag/bugsnag-unity/blob/master/Rakefile#L83

Perhaps there's something in the Gradle DSL that allows us to check the version of the NDK used instead, and verify that it's r16b? https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.LibraryExtension.html

gradle/release.gradle
```diff
@@ -30,7 +38,43 @@ task sourceJar(type: Jar) {
...
+                    return
+                }
+            }
+            throw new GradleException("Unexpected list of ABIs for release: ${project.android.defaultConfig.ndk.abiFilters}")
+        }
+    }
+}
+
+tasks.whenTaskAdded { task ->
+    if (task.name == 'assembleRelease') {
```
fractalwrench: It might be worth depending on the 'publish' and 'bintrayUpload' tasks as these can be invoked without having run `assembleRelease`.

PR #826, opened 2020-02-28
Add deserializers for error, thread, and stackframe

fractalwrench:
## Goal

Adds deserializers which convert a `Map` into `Error`, `Thread`, and `Stackframe`. This is required by the React Native plugin when a JS error is sent over the React bridge for the native layer to deliver.

## Changeset

- Added deserializers for `Error/Thread/Stackframe`.
- Added `MapUtils` with convenience methods for retrieving values from a map with type safety, either returning null or throwing an exception
- Added additional constructor to `Stacktrace`
- Minor tweak of `UserDeserializer` implementation

## Tests

Added unit tests to cover each deserializer.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
@@ -32,6 +32,14 @@ internal class Stacktrace : JsonStream.Streamable {
         this.logger = logger
     }
 
+    constructor(logger: Logger, frames: List<Stackframe>) {
+        trace = when {
+            frames.size >= STACKTRACE_TRIM_LENGTH -> frames.subList(0, STACKTRACE_TRIM_LENGTH)
+            else -> frames
+        }
+        this.logger = logger
+    }
+
     private fun mapToStackframe(it: Map<String, Any?>) =
         Stackframe(
             it["method"] as String?,
--- a/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ThreadDeserializer.java
+++ b/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ThreadDeserializer.java
@@ -1,0 +1,37 @@
+package com.bugsnag.android;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+class ThreadDeserializer implements MapDeserializer<Thread> {
+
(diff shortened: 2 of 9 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
```diff
@@ -32,6 +32,14 @@ internal class Stacktrace : JsonStream.Streamable {
         this.logger = logger
     }
 
+    constructor(logger: Logger, frames: List<Stackframe>) {
```
tomlongridge: Reverse parameters for consistency with existing constructors.

```suggestion
    constructor(frames: List<Stackframe>, logger: Logger) {
```

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ThreadDeserializer.java
```diff
@@ -0,0 +1,37 @@
...
+    public Thread deserialize(Map<String, Object> map) {
+        String type = MapUtils.getOrThrow(map, "type");
+        List<Map<String, Object>> stacktrace = MapUtils.getOrThrow(map, "stacktrace");
+        List<Stackframe> frames = new ArrayList<>();
+
+        for (Map<String, Object> frame : stacktrace) {
+            frames.add(stackframeDeserializer.deserialize(frame));
+        }
+
+        return new Thread(
```
tomlongridge: Are we using public or internal version? This is different in `ErrorDeserializer`, so probably needs to be consistent.

fractalwrench: We are consistently returning the public interface from all the deserializers.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
fractalwrench: Discussed in-person that this isn't possible due to type-erasure causing a constructor signature clash. As these constructors are internal and not part of our public interface, no action is required here.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ThreadDeserializer.java
tomlongridge: Ah yes - OK. I misread Error.

PR #827, opened 2020-02-28
Capture state events relevant to react native

fractalwrench:
## Goal

Captures state events relevant to React Native and converts them to a `MessageEvent` class, which will be sent across the React Native bridge. The JS layer needs to be informed of 3 changes on the native layer: context, user, and metadata. This changeset subscribes to these events and converts them to an agreed interface the JS layer can understand.

**Note**: the actual emission of the event data across the React Native bridge has not been included and will be implemented in a separate PR.

## Changeset

- Added `BugsnagReactNativeBridge` which converts state events to `MessageEvent` then invokes a callback
- Subscribed the plugin to events for context/user/metadata
- Converted separate user StateEvent types into one, `UpdateUser`
- Renamed old `StateEvent` types to `ClearMetadataSection` and `ClearMetadataValue` to keep in line with current terminology

## Tests

Added unit tests to verify that each state event message is converted correctly. Updated existing test coverage where necessary.

### DIFF
--- a/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/MessageEvent.kt
+++ b/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/MessageEvent.kt
@@ -1,6 +1,19 @@
 package com.bugsnag.android
 
+/**
+ * Holds information about a state change event which is serialized by the BugsnagReactNative
+ * class and emitted to the JS layer.
+ */
 class MessageEvent(
+
+    /**
+     * The type of the event. E.g. `UpdateContext`
+     */
     val type: String,
+
+    /**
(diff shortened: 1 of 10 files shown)


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/MessageEvent.kt
```diff
@@ -0,0 +1,6 @@
+package com.bugsnag.android
+
+class MessageEvent(
```
tomlongridge: Some javadoc here would be good to give context.

fractalwrench: Added JavaDoc for this interface

PR #829, opened 2020-03-02
V5 app device deserialize

fractalwrench:
## Goal

Adds deserializers which convert a `Map` into `AppWithState` and `DeviceWithState`. This is required by the React Native plugin when a JS error is sent over the React bridge for the native layer to deliver.

## Changeset

- Added deserializers for `AppWithState/DeviceWithState`.
- Altered `App/AppWithState` constructors to allow passing in buildUuid and other fields without a `ImmutableConfig` object
- Added `MapUtils` with convenience methods for retrieving values from a map with type safety, either returning null or throwing an exception
- Minor tweak of `UserDeserializer` implementation

## Tests

Added unit tests to cover each deserializer.

### DIFF
--- a/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
+++ b/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
@@ -1,5 +1,6 @@
 package com.bugsnag.android;
 
+import java.util.Date;
 import java.util.Map;
 
 class DeviceDeserializer implements MapDeserializer<DeviceWithState> {
@@ -19,9 +20,13 @@ class DeviceDeserializer implements MapDeserializer<DeviceWithState> {
         );
 
         String time = MapUtils.getOrNull(map, "time");
+        Date date = null;
+
+        if (time != null) {
+            date = DateUtils.fromIso8601(time);
+        }
         return new DeviceWithState(
                 buildInfo,
-                buildInfo.getCpuAbis(),
                 MapUtils.<Boolean>getOrNull(map, "jailbroken"),
                 MapUtils.<String>getOrNull(map, "id"),
                 MapUtils.<String>getOrNull(map, "locale"),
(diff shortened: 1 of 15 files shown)


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
```diff
@@ -0,0 +1,35 @@
...
+                MapUtils.<String>getOrThrow(map, "fingerprint"),
+                MapUtils.<String>getOrThrow(map, "tags"),
+                MapUtils.<String>getOrThrow(map, "brand"),
+                MapUtils.<String[]>getOrThrow(map, "cpuAbis")
+        );
+
+        String time = MapUtils.getOrNull(map, "time");
+        return new DeviceWithState(
+                buildInfo,
+                buildInfo.getCpuAbis(),
```
tomlongridge: Why does `Device` have cpuAbis? (I realise this is pre-existing, but seems odd to be in `DeviceBuildInfo` too.)

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
```diff
@@ -0,0 +1,35 @@
...
+                MapUtils.<String>getOrThrow(map, "osVersion"),
+                MapUtils.<Integer>getOrThrow(map, "apiLevel"),
+                MapUtils.<String>getOrThrow(map, "osBuild"),
+                MapUtils.<String>getOrThrow(map, "fingerprint"),
+                MapUtils.<String>getOrThrow(map, "tags"),
+                MapUtils.<String>getOrThrow(map, "brand"),
+                MapUtils.<String[]>getOrThrow(map, "cpuAbis")
+        );
+
+        String time = MapUtils.getOrNull(map, "time");
```
tomlongridge: Can be `null` but `DateUtils.fromIso8601` is marked non-null - it should probably end up setting it to `null` rather than error

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
fractalwrench: There doesn't appear to be a reason. I've updated the constructor to remove this parameter and populate it from `buildInfo` instead.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
fractalwrench: Added a null check to avoid this throwing an exception.

PR #832, opened 2020-03-02
Move thread serialization logic into ThreadState

fractalwrench:
## Goal

Moves the logic for determining whether to collect a thread trace from `Event` into `ThreadState`. This allows for the React Native implementation to return an empty list of threads if `config.sendThreads` is not enabled for the given error.

## Changeset

- Moved logic for collecting thread trace from `Event` to `ThreadState` constructor
- Updated `getPayloadInfo()` to accept a boolean value which holds the unhandled state

## Tests

Updated existing test coverage to handle new `ThreadState` constructor

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
@@ -5,33 +5,61 @@ import java.io.IOException
 /**
  * Capture and serialize the state of all threads at the time of an exception.
  */
-internal class ThreadState @JvmOverloads constructor
-    (
-    config: ImmutableConfig,
+internal class ThreadState @JvmOverloads constructor(
     exc: Throwable?,
+    isUnhandled: Boolean,
+    sendThreads: ThreadSendPolicy,
+    projectPackages: Collection<String>,
+    logger: Logger,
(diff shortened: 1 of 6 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -5,33 +5,61 @@ import java.io.IOException
...
-    logger: Logger = config.logger
+    stackTraces: MutableMap<java.lang.Thread, Array<StackTraceElement>> = java.lang.Thread.getAllStackTraces()
 ) : JsonStream.Streamable {
 
-    internal val threads: MutableList<Thread>
+    internal constructor(
+        exc: Throwable?,
+        isUnhandled: Boolean,
+        config: ImmutableConfig
+    ) : this(exc, isUnhandled, config.sendThreads, config.projectPackages, config.logger)
```
tomlongridge: Do we need a secondary constructor?

I find this class very difficult to read - can we consolidate the constructors and make some of the parameters member variables?

fractalwrench: Discussed in-person and agreed to merge as-is since the other constructor is required for testing.

PR #834, opened 2020-03-03
React Native fixes after integration testing in example app

fractalwrench:
## Goal

Fixes bugs in the RN module implementation which prevented the example app from receiving errors during integration testing.

## Changeset

- Altered error/thread types 'browser_js' to 'reactnativejs'
- Coerce int/long/double values into `Number` rather than casting directly to Int/Long, as this led to a crash since the JS layer could serialize these values as doubles.

## Tests

Relied on existing test coverage and manual testing within a RN example app.

### DIFF
--- a/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
+++ b/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
@@ -7,14 +7,11 @@ class DeviceDeserializer implements MapDeserializer<DeviceWithState> {
 
     @Override
     public DeviceWithState deserialize(Map<String, Object> map) {
-        Number num = MapUtils.getOrNull(map, "apiLevel");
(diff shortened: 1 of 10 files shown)


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/DeviceDeserializer.java
```diff
@@ -7,11 +7,14 @@
 
     @Override
     public DeviceWithState deserialize(Map<String, Object> map) {
+        Number num = MapUtils.getOrNull(map, "apiLevel");
+        Integer apiLevel = num != null ? num.intValue() : null;
```
tomlongridge: Worth pushing this (and getLong) into MapUtils?

fractalwrench: Extracted to MapUtils class.

PR #838, opened 2020-03-04
Amend notifier payload to allow including notifier dependencies

fractalwrench:
## Goal

Alters the notifier payload to include notifier dependencies for both events and sessions. This is used for React Native to signify the relationship between Android + the JS layer.

## Changeset

- Made `Notifier` a public class rather than singleton object
- Added `notifier` field on `Client` which can be altered by a `Plugin`
- Added `dependencies` collection to `Notifier` which contains other notifiers
- Updated existing code to inject `Notifier` as a parameter rather than relying on the singleton interface
- Updated serializer to include `dependencies` array in payload
- `bugsnag-plugin-react-native` updates the notifier version information accordingly when it is initialised

## Tests

Updated existing test coverage to pass in notifier where possible

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -124,7 +124,7 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
                 leaveBreadcrumb("Connectivity change", BreadcrumbType.STATE, data);
 
                 if (hasConnection) {
-                    eventStore.flushAsync(notifier);
+                    eventStore.flushAsync();
                 }
                 return null;
             }
@@ -205,10 +205,10 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
         InternalReportDelegate delegate = new InternalReportDelegate(appContext, logger,
                 immutableConfig, storageManager, appDataCollector, deviceDataCollector,
                 sessionTracker, notifier);
-        eventStore = new EventStore(immutableConfig, appContext, logger, delegate);
+        eventStore = new EventStore(immutableConfig, appContext, logger, notifier, delegate);
 
         deliveryDelegate = new DeliveryDelegate(logger, eventStore,
-                immutableConfig, breadcrumbState);
+                immutableConfig, breadcrumbState, notifier);
 
         // Install a default exception handler with this client
         if (immutableConfig.getEnabledErrorTypes().getUnhandledExceptions()) {
@@ -232,7 +232,7 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
         registerOrientationChangeListener();
 
         // Flush any on-disk errors
-        eventStore.flushOnLaunch(notifier);
+        eventStore.flushOnLaunch();
         Map<String, Object> data = Collections.emptyMap();
         leaveBreadcrumb("Bugsnag loaded", BreadcrumbType.STATE, data);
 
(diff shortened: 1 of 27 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -86,6 +85,8 @@
     final ClientObservable clientObservable = new ClientObservable();
     private PluginClient pluginClient;
 
+    final Notifier notifier = new Notifier();
```
tomlongridge: Can `bugsnag-android` notifier details be set in `Client` on init rather than baked into the `Notifier` object? I think it makes it slightly more like a normal data object.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
```diff
@@ -24,9 +24,9 @@
         this.breadcrumbState = breadcrumbState;
     }
 
-    void deliver(@NonNull Event event) {
```
tomlongridge: Making `Notifier` a member variable would avoid some of the method signature changes - I think this makes sense as it's more similar to the logger, config and breadcrumb state objects that you're setting in the constructor.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -59,7 +59,7 @@ public int compare(File lhs, File rhs) {
         this.delegate = delegate;
     }
 
-    void flushOnLaunch() {
```
tomlongridge: Making Notifier a member variable would avoid some of the method signature changes here too.

bugsnag-android-core/src/main/java/com/bugsnag/android/InternalReportDelegate.java
```diff
@@ -87,16 +90,15 @@ void recordStorageCacheBehavior(Event event) {
      * generated and sent asynchronously with no callbacks, retry attempts, or writing to disk.
      * This is intended for internal use only, and reports will not be visible to end-users.
      */
-    void reportInternalBugsnagError(@NonNull Event event) {
```
tomlongridge: You should be able to use the member variable here shouldn't you?

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -26,6 +26,12 @@ class BugsnagReactNativePlugin : Plugin {
         logger = client.logger
         internalHooks = InternalHooks(client)
         client.logger.i("Initialized React Native Plugin")
+
+        val notifier = client.notifier
+        notifier.name = "Bugsnag React Native"
+        notifier.url = "https://github.com/bugsnag/bugsnag-js"
+        // use the version supplied by bugsnag-android, as this is equivalent for this module
```
tomlongridge: This should be the JS version I think?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: Our release script which bumps the version relies on the values being set in the class - unless you feel strongly I'll keep this as-is to reduce the amount of change.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
fractalwrench: Addressed similar comments by changing these to member variables.

PR #841, opened 2020-03-06
V5 UAT - Additional NDK tests

Cawllec:
## Goal

Adds Maze-runner tests to cover added functionality in the V5 release.


Cawllec:
I realise the commits look a bit janky, but the actual file changes are correct.

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXRemoveDataScenario.java
```diff
@@ -0,0 +1,44 @@
...
+        activate();
+        Handler main = new Handler(Looper.getMainLooper());
+        main.postDelayed(new Runnable() {
+            @Override
+            public void run() {
+                Bugsnag.clearMetadata("remove");
+                Bugsnag.clearMetadata("persist", "remove");
+                activate();
+            }
+        }, 1500);
```
fractalwrench: Can this value be reduced to something smaller, e.g. 150?

features/fixtures/mazerunner/src/main/cpp/bugsnags.cpp
```diff
@@ -293,4 +293,83 @@ Java_com_bugsnag_android_mazerunner_scenarios_CXXConfigurationMetadataNativeCras
...
+JNIEXPORT void JNICALL
+Java_com_bugsnag_android_mazerunner_scenarios_CXXRemoveOnErrorScenario_activate(
+    JNIEnv *env,
+    jobject instance) {
+  bugsnag_add_on_error(&on_err_false);
+  // Confirm that the callback set correctly
+  bugsnag_notify((char *)"Foo", (char *)"Bar", BSG_SEVERITY_ERR);
+  // Confirms that only the correct callback is removed
+  bugsnag_add_on_error(&on_err_true);
+  bugsnag_remove_on_error(&on_err_false);
```
fractalwrench: This shouldn't compile as the method has zero parameters

tests/features/fixtures/mazerunner/src/main/cpp/bugsnags.cpp
```diff
@@ -291,4 +291,82 @@ Java_com_bugsnag_android_mazerunner_scenarios_CXXConfigurationMetadataNativeCras
...
+  bugsnag_leave_breadcrumb((char *)"Start scenario crumb", BSG_CRUMB_LOG);
+  bugsnag_notify((char *)"Start scenario",
+                     (char *)"Testing env", BSG_SEVERITY_INFO);
+}
+
+JNIEXPORT void JNICALL
+Java_com_bugsnag_android_mazerunner_scenarios_CXXRemoveOnErrorScenario_activate(
+    JNIEnv *env,
+    jobject instance) {
+  bugsnag_add_on_error(&on_err_false);
```
twometresteve: I'm not convinced this is going to prove behaviour as intended.... The first callback returning false would mean no further ones are run anyway, right?  So I don't think adding and then removing a true return would prove what we want to.

Assuming I've understood everything correctly, it might be better to create a second true-returning callback that set different fields to the first.  Something along the lines of:
1. Add the first callback
2. (Optionally) notify 
3. add the second
4. notify
5. remove callback
6. notify one more time.... 

Where we'd expect 3 events to be sent, with the 1st and 3rd being the same and the 2nd having addition fields set.

tests/features/native_on_error.feature
```diff
@@ -23,3 +23,10 @@ Feature: Native on error callbacks are invoked
     And I wait to receive a request
     And the request payload contains a completed unhandled native report
     And the event "context" equals "Some custom context"
+
+  Scenario: on_error returning false is removed
+    When I run "CXXRemoveOnErrorScenario" and relaunch the app
```
twometresteve: This scenario does include an onError returning false, but did you intend to bundle separate scenarios in together?

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXGetJavaDataScenario.java
```diff
@@ -0,0 +1,39 @@
...
+
+    static {
+        System.loadLibrary("bugsnag-ndk");
+        System.loadLibrary("monochrome");
+        System.loadLibrary("entrypoint");
+    }
+
+    public native void activate();
+
+    /**
```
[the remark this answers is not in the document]
Cawllec: Unfortunately these are necessary due to the linter I believe.

tests/features/fixtures/mazerunner/src/main/cpp/bugsnags.cpp
Cawllec: I've actually checked the implementation and I don't think this is necessary for proving the `on_error` callbacks can be removed.  It seems there can only be one callback present in the NDK layer at once, so to verify this we'd only need to:
1. Set some data
2. Add a callback that mutates said data
2. Notify
3. Verify the data was mutated
4. Remove the callback
5. Notify
6. Verify the data is not mutated

features/native_api.feature
```diff
@@ -204,3 +204,45 @@ Feature: Native API
...
+    Scenario: Get Java data in the Native layer
+        When I run "CXXGetJavaDataScenario"
+        And I configure the app to run in the "non-crashy" state
+        And I relaunch the app
+        And I should receive a request
+        Then the request payload contains a completed unhandled native report
+        And the event "unhandled" is true
+        And the event "metaData.data.context" equals "passContext"
+        And the event "metaData.data.appVersion" equals "passAppVersion"
+        And the event "metaData.data.user" equals "passUser"
```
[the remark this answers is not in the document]
Cawllec: I'd only intended to test a subset of behaviour (e.g. one from App, Device, User etc), but can add these easily enough.

tests/features/native_on_error.feature
Cawllec: As above, I think we can remove some ambiguity from this by simplifying it to just modifying data.

features/native_api.feature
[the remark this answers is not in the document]
twometresteve: Yes, you're quite right, that is the world we want to get to, lower level tests should verify the detail.  Please disregard.

tests/features/fixtures/mazerunner/src/main/cpp/bugsnags.cpp
twometresteve: Ok, I like that approach too.

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXGetJavaDataScenario.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     */
+    public CXXGetJavaDataScenario(@NonNull Configuration config, @NonNull Context context) {
+        super(config, context);
+        config.setAutoTrackSessions(false);
+        config.addMetadata("notData", "vals", "passMetaData");
+        config.setAppVersion("passAppVersion");
+        config.setContext("passContext");
+        config.setUser("passUser", "passUser", "passUser");
```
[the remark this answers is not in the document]
twometresteve: Comment retracted, unit/instrumentation tests should handle the breadth.  e2e are for the depth through the software.

mazerunner-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXGetJavaDataScenario.java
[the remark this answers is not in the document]
twometresteve: Because it whats any JavaDoc?  Fair enough, but in that case perhaps put a simple sentence in there for the actual doc - just so it doesn't look an omission to the untrained (i.e. my) eye.

PR #842, opened 2020-03-09
Change enabledBreadcrumbTypes to affect automatic breadcrumb hooks only

tomlongridge:
## Goal

The new `enabledBreadcrumbTypes` should only affect the automatically collected breadcrumbs and should disable the hooks to listen for the breadcrumbs to avoid unnecessary computations.

## Changeset

* Created `leaveAutoBreadcrumb` in `Client` for internally-left breadcrumbs to wrap the config check
* Wrapped the various hooks for listening to system events with a check of the config and didn't register the hook if it is disabled

## Tests

* Unit tests updated
* End-to-end tests for no breadcrumbs, manual and the loaded 'state' breadcrumb.

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbFilterTest.kt
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbFilterTest.kt
@@ -1,34 +1,46 @@
 package com.bugsnag.android
 
-import com.bugsnag.android.BugsnagTestUtils.generateClient
-import com.bugsnag.android.BugsnagTestUtils.generateConfiguration
-import org.junit.After
+import com.bugsnag.android.BreadcrumbType.MANUAL
 import org.junit.Assert.assertEquals
-import org.junit.Before
+
+import androidx.test.filters.SmallTest
+import com.bugsnag.android.BugsnagTestUtils.generateClient
+
+import org.junit.After
 import org.junit.Test
 
+@SmallTest
 class BreadcrumbFilterTest {
 
     private var client: Client? = null
 
-    @Before
-    fun setUp() {
-        val configuration = generateConfiguration()
-        configuration.enabledBreadcrumbTypes = setOf(BreadcrumbType.REQUEST)
-        client = generateClient(configuration)
-    }
-
     @After
     fun tearDown() {
-        client!!.close()
+        client?.close()
     }
 
     @Test
-    fun zeroBreadcrumbsReceived() {
-        client!!.leaveBreadcrumb("Hello")
-        assertEquals(0, client!!.breadcrumbState.store.size.toLong())
+    fun testManualBreadcrumbNotFiltered() {
+        val configuration = BugsnagTestUtils.generateConfiguration()
+        configuration.enabledBreadcrumbTypes = emptySet()
+        client = generateClient(configuration)
 
-        client!!.leaveBreadcrumb("Hello", BreadcrumbType.REQUEST, emptyMap())
-        assertEquals(1, client!!.breadcrumbState.store.size.toLong())
+        client?.leaveBreadcrumb("Hello World")
+
+        assertEquals(1, client!!.breadcrumbState.store.size)
     }
+
+    @Test
+    fun testManualBreadcrumbNotFilteredOnManual() {
+        val configuration = BugsnagTestUtils.generateConfiguration()
+        configuration.enabledBreadcrumbTypes = setOf(MANUAL)
+        client = generateClient(configuration)
+
+        client?.leaveBreadcrumb("Hello World")
+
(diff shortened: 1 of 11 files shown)


features/breadcrumb.feature
```diff
@@ -17,3 +17,16 @@ Scenario: Manually added breadcrumbs are sent in report
...
     And the event "breadcrumbs.0.metaData.message" equals "Hello Breadcrumb!"
+
+Scenario: Manually added breadcrumbs are sent in report when auto breadcrumbs are disabled
+    When I run "BreadcrumbDisabledScenario"
+    Then I should receive a request
+    And the request is a valid for the error reporting API
+    And the event has 1 breadcrumbs
+
+Scenario: Manually added breadcrumbs are sent in report when auto breadcrumbs are disabled
+    When I run "BreadcrumbAutoScenario"
```
fractalwrench: Both these classes are missing from the changeset, which causes the new scenarios to fail.

features/breadcrumb.feature
```diff
@@ -17,3 +17,16 @@ Scenario: Manually added breadcrumbs are sent in report
...
+    And I relauch the app
```
fractalwrench: This is an undefined step as `features` uses v1 of mazerunner, whereas `test/features` uses v2.

features/breadcrumb.feature
```diff
@@ -17,3 +17,16 @@ Scenario: Manually added breadcrumbs are sent in report
     And the event "breadcrumbs.0.type" equals "manual"
     And the event "breadcrumbs.0.metaData" is not null
     And the event "breadcrumbs.0.metaData.message" equals "Hello Breadcrumb!"
+
+Scenario: Manually added breadcrumbs are sent in report when auto breadcrumbs are disabled
```
fractalwrench: These scenarios need to be added to `test/features` in addition to run on CI.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbFilterTest.kt
```diff
@@ -1,34 +1,46 @@
...
-    fun setUp() {
-        val configuration = generateConfiguration()
-        configuration.enabledBreadcrumbTypes = setOf(BreadcrumbType.REQUEST)
-        client = generateClient(configuration)
-    }
-
     @After
     fun tearDown() {
-        client!!.close()
+        client?.close()
```
fractalwrench: This is an existing issue, but declaring the client as `lateinit var client: Client` would avoid having to use the safe operator at all.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -70,11 +70,12 @@
 
     private final SessionStore sessionStore;
 
-    final SystemBroadcastReceiver systemBroadcastReceiver;
     final SessionTracker sessionTracker;
 
-    final ActivityBreadcrumbCollector activityBreadcrumbCollector;
-    final SessionLifecycleCallback sessionLifecycleCallback;
```
fractalwrench: Nit: these could still be declared `final` and assigned null in the constructor if the particular breadcrumb type is not enabled. This would avoid the field being reassigned internally after the `Client` has been instantiated.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -216,16 +211,21 @@ public Unit invoke(String activity, Map<String, ?> metadata) {
...
-            Async.run(new Runnable() {
-                @Override
-                public void run() {
-                    IntentFilter intentFilter = SystemBroadcastReceiver.getIntentFilter();
-                    appContext.registerReceiver(systemBroadcastReceiver, intentFilter);
-                }
-            });
-        } catch (RejectedExecutionException ex) {
-            logger.w("Failed to register for automatic breadcrumb broadcasts", ex);
+        final SystemBroadcastReceiver systemBroadcastReceiver =
```
fractalwrench: I believe this can be assigned to the field, unless there's something I'm missing?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
tomlongridge: OK - I've done this. It adds quite a few else clauses, so not sure it's particularly nicer code but has the benefits of being final.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
tomlongridge: I wanted to avoid setting the field if there was nothing to listen to in case attempting to unregister it on `finalize` causes an error (this doesn't seem to be documented).

PR #847, opened 2020-03-10
Alter NDK event field sizes

fractalwrench:
Alters the size of fields in the `bugsnag_event` struct to that specified in the notifier spec, and updates existing test coverage to ensure that the migration works as expected.

This alters the breadcrumb array length from 30 to 25, and the breadcrumb name field to 64 from 33.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
@@ -153,10 +153,11 @@ bugsnag_event *bsg_map_v2_to_report(bugsnag_report_v2 *report_v2) {
 }
 
 void migrate_breadcrumb_v1(bugsnag_report_v2 *report_v2, bugsnag_event *event) {
-  for (int k = 0; k < report_v2->crumb_count; k++) {
-    int crumb_index = (report_v2->crumb_first_index + k) % BUGSNAG_CRUMBS_MAX;
+  for (int k = 0; k < event->crumb_count; k++) {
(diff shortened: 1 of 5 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -153,10 +153,11 @@ bugsnag_event *bsg_map_v2_to_report(bugsnag_report_v2 *report_v2) {
 }
 
 void migrate_breadcrumb_v1(bugsnag_report_v2 *report_v2, bugsnag_event *event) {
-  for (int k = 0; k < report_v2->crumb_count; k++) {
-    int crumb_index = (report_v2->crumb_first_index + k) % BUGSNAG_CRUMBS_MAX;
+  for (int k = 0; k < event->crumb_count; k++) {
+    int crumb_index = (report_v2->crumb_first_index + k) % V1_BUGSNAG_CRUMBS_MAX;
```
tomlongridge: Can we take account of whether the number of crumbs is > `BUGSNAG_CRUMBS_MAX` and if so add to the `crumb_first_index` (i.e. drop the earliest breadcrumbs and not the latest)?

It would be good to have a test to check this logic.

fractalwrench: I've reworked this method to drop the earliest breadcrumbs, and have extracted the index calculations to separate functions with added unit tests. The new implementation also makes use of `bugsnag_event_add_breadcrumb` which should result in less potential for things to go wrong with the index overlapping.

PR #850, opened 2020-03-11
Removed old settings from manifest in example

tomlongridge:
AUTO_DETECT_ANRS and AUTO_DETECT_NDK_CRASHES are now set in `enabledErrorTypes` not through manifest.


examples/sdk-app-example/src/main/AndroidManifest.xml
```diff
@@ -34,14 +34,6 @@
       android:name="com.bugsnag.android.API_KEY"
       android:value="your-api-key" />
```
fractalwrench: It'd be good to update this to a valid dummy API key, as currently this crashes the example app.

PR #852, opened 2020-03-12
feat(bugsnag-plugin-react-native): Update configure() param to be a map

bengourley:
## Goal
Updates the `configure()` method in `BugsnagReactNativePlugin` to take a `Map` as a parameter. When `configure()` is called by the JS layer it passes runtime information about the JS notifier version, React Native version, and JSEngine in this map. The native layer consumes this 

## Changeset

- Made `Device` take a `MutableMap` of runtimeVersion information in its initialiser
- Made `DeviceDataCollector` retain a `MutableMap` of version information
- Added `addRuntimeVersionInfo()` to `Client` which adds info to the `DeviceDataCollector` runtimeVersions map
- Read `reactNativeVersion` and `engine` from map supplied to `configure()` and added them to runtimeVersions for events + sessions

## Tests
- Relied on existing test coverage that was updated
- Updated unit test JSON fixtures as the underlying `Map` implementation of `runtimeVersions` changed. This appears to have altered the order in which keys are serialized as JSON.
- Added test to verify that `addRuntimeVersionInfo()` on `Client` calls the appropriate method on `DeviceDataCollector`
- Deleted file left in obsolete `sdk` directory


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -39,10 +46,12 @@ class BugsnagReactNativePlugin : Plugin {
     override fun unload() {}
 
     @Suppress("unused")
-    fun configure(jsVersion: String): Map<String, Any?> {
+    fun configure(env: Map<String, Any?>?): Map<String, Any?> {
```
tomlongridge: How can we best keep this interface maintained/documented? It feels a bit unstructured at the moment.

I'd like to minimise the number of top-level keys, so how about making "runtimeVersions" the key to a map of String->String runtime version elements? That way there are fewer magic strings and also the JS layer is in control of what it adds.

fractalwrench: Using a map value would add additional complexity to the deserializer which I'd prefer to avoid at this stage given there will only be 3 values present, and it's known that this solution works.

It's also worth noting that we can change the structure of values in the map without making a breaking change, so could revisit if we do start sending more values over the bridge in future.

PR #854, opened 2020-03-16
Alter api key validation as per change in notifier spec

fractalwrench:


bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -47,7 +47,11 @@ static Configuration load(@NonNull Context context, @NonNull String apiKey) {
 
     private void validateApiKey(String value) {
         if (value == null || !value.matches(API_KEY_REGEX)) {
-            throw new IllegalArgumentException("You must provide a valid Bugsnag API key");
+            if (value == null) {
+                value = "";
```
bengourley: `null` and empty API key should throw

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
```diff
@@ -46,8 +46,13 @@ static Configuration load(@NonNull Context context, @NonNull String apiKey) {
...
     private void validateApiKey(String value) {
-        if (value == null || !value.matches(API_KEY_REGEX)) {
-            throw new IllegalArgumentException("You must provide a valid Bugsnag API key");
+        if (Intrinsics.isEmpty(value)) {
+            throw new IllegalArgumentException("No Bugsnag API Key set");
+        }
+
+        if (!value.matches(API_KEY_REGEX)) {
+            DebugLogger.INSTANCE.w(String.format("Invalid configuration. apiKey should be a "
+                    + "32-character hexademical string, got %s", value));
```
tomlongridge:
```suggestion
                    + "32-character hexademical string, got \"%s\"", value));
```

fractalwrench:
Updated placeholder API key in example.

PR #857, opened 2020-03-19
Update example app method signatures

fractalwrench:
Updates the example app method signature for leaveBreadcrumb, whose parameter order has changed.

Verified that this works by running the example app up manually.


twometresteve:
Just another observation from the NDK errors change that you may like to account for here... I don't think we need the following line in ExampleApplication.kt any more:
```       
 config.enabledErrorTypes.ndkCrashes = true
```

fractalwrench:
Removed unnecessary line from the example application.

PR #859, opened 2020-03-24
Changes to upgrade guide following review

tomlongridge:
As per review comments - update the upgrade guide to emphasise use of manifest.

I've also condensed some of the content - combining the Client and Bugsnag changes into one table and putting all the Private API moves into a single table.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -6,29 +6,18 @@ Upgrade from 4.X to 5.X
 
 __This version contains many breaking changes__. It is part of an effort to unify our notifier libraries across platforms, making the user interface more consistent, and implementations better on multi-layered environments where multiple Bugsnag libraries need to work together (such as React Native).
 
-### Overview
+### Key points
+
+- `Bugsnag.init` has been renamed to `Bugsnag.start`.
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -6,304 +6,221 @@ Upgrade from 4.X to 5.X
...
 __This version contains many breaking changes__. It is part of an effort to unify our notifier libraries across platforms, making the user interface more consistent, and implementations better on multi-layered environments where multiple Bugsnag libraries need to work together (such as React Native).
 
-# Interfaces
+### Overview
 
-## Bugsnag
+`Bugsnag.init` has been renamed to `Bugsnag.start`.
 
-`Bugsnag.init()` has been renamed to `Bugsnag.start()`.
+The configuration options that can be set via your `AndroidManifest.xml` has been expanded:
```
duncanhewett: It would be good to make it clearer that this is an example, how about:
> Many more configuration options can be set via your `AndroidManifest.xml`, for example:

tomlongridge: The overview was getting a bit noisy, so I've made this "Key points" and move the configuration example back down to the Configuration section.

PR #861, opened 2020-03-26
Gracefully fail if native library could not be loaded

fractalwrench:
## Goal

The ANR + NDK plugins load native libraries to detect errors. On some versions of Android these libraries cannot be loaded (e.g. due to PackageManager bugs) which results in an `UnsatisfiedLinkError` being thrown.

This error is now caught and reported to the Bugsnag dashboard, and the affected plugin(s) is disabled to allow graceful fallback.

## Changeset

- Moved native library loading to `LibraryLoader` class which attempts to load the library once, reporting any `UnsatisfiedLinkError` to bugsnag
- Updated ANR/NDK plugins to use `LibraryLoader` class and only continue loading if successful

Note: the error message format will need to be updated to link to documentation before this is merged. A documentation PR will be raised separately and in the meanwhile the rest of this PR can be reviewed.

## Tests

Added a unit test to verify the behaviour of the new `LibraryLoader` class.

Tested the error message format in the dashboard by running an example app where the library name supplied for loading was incorrect.

### DIFF
--- a/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
+++ b/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
@@ -5,6 +5,11 @@ import android.os.Looper
 
 internal class AnrPlugin : Plugin {
 
+    private companion object {
+        private const val LOAD_ERR_MSG = "Native library could not be linked. Bugsnag will " +
+                "not report ANRs. See https://docs.bugsnag.com/platforms/android/anr-link-errors"
+    }
+
     private val loader = LibraryLoader()
     private lateinit var client: Client
     private val collector = AnrDetailsCollector()
@@ -16,7 +21,7 @@ internal class AnrPlugin : Plugin {
         val loaded = loader.loadLibrary("bugsnag-plugin-android-anr", client) {
             val error = it.errors[0]
             error.errorClass = "AnrLinkError"
-            error.errorMessage = "Native library could not be linked. Bugsnag will not report ANRs."
+            error.errorMessage = LOAD_ERR_MSG
             true
         }
 
@@ -29,8 +34,7 @@ internal class AnrPlugin : Plugin {
                 client.logger.i("Initialised ANR Plugin")
             })
         } else {
-            client.logger.e("Native library could not be linked. Bugsnag will not report " +
-                    "ANRs. See https://docs.bugsnag.com/platforms/android/anr-link-errors")
+            client.logger.e(LOAD_ERR_MSG)
         }
     }
 
--- a/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
+++ b/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
@@ -29,7 +29,8 @@ internal class AnrPlugin : Plugin {
                 client.logger.i("Initialised ANR Plugin")
             })
         } else {
-            client.logger.e("Native library could not be linked. Bugsnag will not report ANRs.")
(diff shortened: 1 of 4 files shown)


bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -5,26 +5,30 @@ import android.os.Looper
...
+
+        if (loaded) {
+            // this must be run from the main thread as the SIGQUIT is sent to the main thread,
+            // and if the handler is installed on a background thread instead we receive no signal
+            Handler(Looper.getMainLooper()).post(Runnable {
+                this.client = client
+                enableAnrReporting(true)
+                client.logger.w("Initialised ANR Plugin")
+            })
+        }
```
tomlongridge: Shall we log a warning to the `logger` if it fails to load as well? Alternatively perform the log message in LibraryLoader and log "ANR reporting enabled" above.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -5,26 +5,30 @@ import android.os.Looper
...
+            true
+        }
+
+        if (loaded) {
+            // this must be run from the main thread as the SIGQUIT is sent to the main thread,
+            // and if the handler is installed on a background thread instead we receive no signal
+            Handler(Looper.getMainLooper()).post(Runnable {
+                this.client = client
+                enableAnrReporting(true)
+                client.logger.w("Initialised ANR Plugin")
```
tomlongridge: Existing issue - but why is this a warning?

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/NdkPlugin.kt
```diff
@@ -4,26 +4,31 @@ import com.bugsnag.android.ndk.NativeBridge
...
+        if (loaded) {
+            if (nativeBridge == null) {
+                nativeBridge = NativeBridge()
+                client.registerObserver(nativeBridge)
+                client.sendNativeSetupNotification()
+                client.syncInitialState()
+            }
+            enableCrashReporting()
+            client.logger.i("Initialised NDK Plugin")
         }
```
tomlongridge: As above, let's log a warning if not loaded.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
fractalwrench: Updated to info level

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/NdkPlugin.kt
```diff
@@ -4,26 +4,34 @@ import com.bugsnag.android.ndk.NativeBridge
...
     override fun load(client: Client) {
-        if (nativeBridge == null) {
-            nativeBridge = NativeBridge()
-            client.registerObserver(nativeBridge)
-            client.sendNativeSetupNotification()
-            client.syncInitialState()
+        val loaded = loader.loadLibrary("bugsnag-ndk", client) {
+            val error = it.errors[0]
+            error.errorClass = "NdkLinkError"
+            error.errorMessage = "Native library could not be linked. Bugsnag will not report NDK errors."
```
tomlongridge: We need to update the Bugsnag message too.

fractalwrench: I've extracted this message to a constant that is now used in both places.

PR #864, opened 2020-03-30
Maze runner retry on failure and fail fast

snmaynard:
Companion to bugsnag/maze-runner#91

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,7 +56,7 @@ services:
       BROWSER_STACK_LOCAL_IDENTIFIER: ${BUILDKITE_JOB_ID:-maze-runner}
       DEVICE_TYPE:
       APP_LOCATION:
-    command: --tags ~@ignore --fail-fast --retry 2
+    command: --tags ~@ignore --fail-fast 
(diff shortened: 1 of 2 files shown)


fractalwrench:
Failing fast looks good.

I would prefer to avoid retries if at all possible, as it doesn't fix the fundamental issue that the mazerunner scenarios are flaky. Increasing resources for the Android CI build has already done a lot to improve stability, and IMO it would be worth spending a little time investigating the known flakes before resorting to automatic retries. The majority of failures seem to be coming from `session_stopping.feature` and other scenarios involving sessions.

snmaynard:
Whatever you think is best is good with me, I was just moving the CLI options that were applied globally to maze runner down to individual repos as I think that is the better place for them. Feel free to make the changes you want and merge it down as you see fit.

twometresteve:
Great, a consensus.  I've removed the retries.  Are you happy with this going into master, @fractalwrench , as opposed to v5?  I'll leave you to approve and merge it though; I'll be merging the change to remove the options from MazeRunner in due course.

PR #867, opened 2020-04-14
Remove buildUuid from Configuration

fractalwrench:
## Goal

The `buildUuid` field should not be settable in Configuration and should be read via the manifest only to prevent the value getting out of date with the value generated by the gradle plugin.

## Changeset

- Removed `buildUuid` accessors from `Configuration` and `ManifestConfigLoader`
- Updated `ImmutableConfig` to load `buildUuid` from manifest only
- Updated upgrading guide

## Tests
- Removed `buildUuid` accessor/loading tests as appropriate
- Updated `event.app` serialization tests to avoid checking `buildUuid`, as the property cannot currently be mocked or faked for test
- Tested changes in example app and verified that the `buildUuid` is still reported


tomlongridge:
Looks like some actual test failures...

fractalwrench:
Updated test assertion which was attempting to check a buildUuid set via Configuration, which is no longer possible. Waiting on CI to run the test as it's not possible to run the scenario locally.

PR #873, opened 2020-04-23
Run v2 MazeRunner tests locally

twometresteve:
## Goal

Make the v2 MazeRunner tests efficient to run locally, whilst removing the v1 features.

## Design

This change makes greater use of locally built artifacts rather than building a docker image every time.  With the docker image setup we currently have, if the source code changes then the layer that does the Gradle build is invalidate and a full build occurs, adding a lot of time to running the tests.  I tried to change the Docker build to make sure the Gradle build cache was maintained to avoid this, but ultimately it was far easier to fall back to using the local build environment.

## Changeset

Makefile updated
MR v1 artifacts removed
TESTING.md updated
Dockerfile further enhanced to avoid duplicate downloads after code changes (especially noticeable on low-bandwidth connections).

## Tests

Standard CI run.


twometresteve:
> I tested this out locally and this LGTM, particularly the speed 😎. I had a couple of queries:
> 
> I saw this warning message when running locally: `WARNING: The BRANCH_NAME variable is not set. Defaulting to a blank string.`

Yes, I noticed it too.  Seems like annoying behaviour to me as that variable is only used by other services in the compose file.  It isn't needed for the maze one (but we could set it artificially just to quash the warning).

> Is there any way to obtain requests sent if a mazerunner scenario failed? Previously these were stored to the `maze_output` directory and they simplified debugging.

I wasn't fully aware of the v1 behaviour, but I have noticed today that we could do more wrt providing diagnostic data.  I was going to experiment with build artifacts in Buildkite for diagnosing CI failures and this could tie in directly with running locally.  Would you rather we hold off merging this until tht's in place in MR v2?

fractalwrench:
>Would you rather we hold off merging this until tht's in place in MR v2?

From my POV this can be merged as-is, a ticket can be created internally to address future work and we can prioritise that as necessary

PR #874, opened 2020-04-23
Create observer before registering plugin

fractalwrench:
## Goal

Creates the observer before it is registered, otherwise this triggers an exception as the `lateinit var` will not have been initialised. This crashed the Bugsnag react native plugin

## Tests

Manually verified with a local artefact on the v7-react-native example app


fractalwrench:
>How can we prevent this kind of thing regressing?

I think packaging AARs in the node_modules should be helpful at preventing this kind of thing, as I assume this was tested with a local AAR artefact on the React Native example app but Gradle incorrectly used a stale version of bugsnag-android. This will be addressed in the following PR: https://github.com/bugsnag/bugsnag-js/pull/815

PR #885, opened 2020-06-16
Update dependencies to use AGP 4 and Gradle 6.1.1

fractalwrench:
## Goal

Updates dependencies to use AGP 4 and Gradle 6.1.1, which are the recommended versions to be used with Android Studio 4 that has now been released into the stable channel. This changeset also removes unused code referring to the removed infer dependency.


fractalwrench:
I believe this requires another change to the publishing plugin which is currently failing the build. It's probably worth migrating to the maven publish plugin supported by Android Studio 3.6+, for which there is an existing ticket: https://developer.android.com/studio/build/maven-publish-plugin

PR #893, opened 2020-06-24
Increase max breadcrumbs

### LINKED ISSUE #891: Increase max breadcrumbs
At Square we currently use a custom tab to log our own breadcrumbs. I was looking at migrating us to using Bugsnag breadcrumbs. In Bugsnag 4.x there was no max breadcrumb, while in 5.0 the max is 100. We currently have a max of 400.

I can't use the 4.x breadcrumbs, the implementation is clearly broken: it serializes the entire stream of breadcrumbs on every add() just to check the json isn't greater than 4Kb (I'm glad to see this was removed in 5.x).

pyricau:
This doesn't change the default behavior but allows users to change the max breadcrumbs to 1000.

Fixes #891

## Goal

Square uses 400 breadcrumbs in reports, the current max is 100.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
@@ -17,7 +17,7 @@ import java.util.Set;
 public class Configuration implements CallbackAware, MetadataAware, UserAware {
 
     private static final int MIN_BREADCRUMBS = 0;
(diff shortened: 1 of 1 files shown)


xander-jones:
Hey again @pyricau, thanks for the contribution. However, as discussed in https://github.com/bugsnag/bugsnag-android/issues/891, we're not looking to increase the maximum number of breadcrumbs to avoid issues delivering the payload.

PR #897, opened 2020-07-01
Update example app and test fixture dependencies

fractalwrench:
## Goal

Updates the example app and test fixtures to use the latest AGP version (4.0.0), and updates their dependencies to ensure they are not left abandoned on an obsolete version.

This is part of the groundwork of preparing the entire project to use AGP 4.0.0.

## Changeset

- Updated android sizer task to use 1.0.0 of bundletool
- Updated apps to use latest compileSdkVersion of 29, AGP 4.0.0, Kotlin 1.3.72
- Added one Java file to `minimalapp` fixture to prevent Android Bundle failing to generate APKs with AGP 4.0.0 due to no code being present
- Removed unneeded test dependencies from minimalapp
- Updated android sizer script to get APKs/AABs from new generated location
- Updated test fixture to allow clear text traffic required by bs-local.com, which was failing the scenarios on Android 9 after upgrading the compileSdkVersion

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,17 +111,6 @@ For release purposes, the Makefile's build command includes the "armeabi" ABI fo
 
 Full details of how to build and run tests can be found in [the testing guide](TESTING.md)
 
-## Building the Example App
-
-You can build and install the example app as follows:
-
-```shell
-./gradlew installDebug
-```
-
-This builds the latest version of the library and installs an app onto your
-device/emulator.
-
 ## Installing/testing against a local maven repository
 
 Change `VERSION_NAME` in `gradle.properties` to a version higher than the currently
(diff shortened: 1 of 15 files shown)


tomlongridge:
This now seems to preclude the use of Android Studio < 4.0:

> This version of Android Studio cannot open this project, please retry with Android Studio 4.0 or newer.

fractalwrench:
Android Studio 4 only officially supports AGP 4, so I believe we'll ultimately need to upgrade the plugin in the example app and choose which Android Studio versions we care about more. I'd be inclined to bump it as we're using pretty ancient versions as it stands.

tomlongridge:
> Android Studio 4 only officially supports AGP 4, so I believe we'll ultimately need to upgrade the plugin in the example app and choose which Android Studio versions we care about more. I'd be inclined to bump it as we're using pretty ancient versions as it stands.

Android Studio 4.0 is not old (it was released in May), so we need to find a way to be able to still run our example in 3.x.

fractalwrench:
Updated as per review feedback

fractalwrench:
Using 3.6.0 for the example app _should_ work but it's unsupported by Android Studio and is subject to breaking/showing a warning message in future versions of Android Studio. Android Studio 4.0 is reasonably old in Android terms as it's very common to use the beta channel or canary channel.

fractalwrench:
Downgraded AGP to 3.6.0 for the example app only.

PR #900, opened 2020-07-10
Ensure plugins are loaded before flushing errors

fractalwrench:
## Goal

Ensures that plugins are loaded before flushing errors. This avoids the potential for a race condition where a plugin sets a callback that filters out a particular error class.

## Tests

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -241,8 +241,8 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
         loadPlugins(configuration);
 
         // Flush any on-disk errors
-        eventStore.flushOnLaunch();
         connectivity.registerForNetworkChanges();
+        eventStore.flushOnLaunch();
 
         // leave auto breadcrumb
         Map<String, Object> data = Collections.emptyMap();
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -234,17 +234,19 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
         } else {
             this.systemBroadcastReceiver = null;
         }
-        connectivity.registerForNetworkChanges();
 
         registerOrientationChangeListener();
 
+        // initialise plugins before attempting to flush any errors
+        loadPlugins(configuration);
+
         // Flush any on-disk errors
         eventStore.flushOnLaunch();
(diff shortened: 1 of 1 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -234,17 +234,19 @@ public void run() {
...
-        connectivity.registerForNetworkChanges();
 
         registerOrientationChangeListener();
 
+        // initialise plugins before attempting to flush any errors
+        loadPlugins(configuration);
+
         // Flush any on-disk errors
         eventStore.flushOnLaunch();
+        connectivity.registerForNetworkChanges();
```
tomlongridge, line 245: We seem to be left with a seemingly illogical order - do the positions of the two event registering lines (connectivity and orientation) impact the issue you're seeing? Logically, I would expect them both to be done together.

fractalwrench: The orientation listener doesn't affect whether errors are flushed and has therefore been left as-is to minimise change. L244/245 are the only places in the `Client` constructor that trigger delivery of error reports so have therefore been moved after the plugin loading to avoid any potential race.

This is unrelated to the React Native flake issue but is worth addressing nonetheless.

tomlongridge: So why move `registerForNetworkChanges`? I can see this affects the flushing of reports as a connection may not be present. Why is it desirable not to have this updated until after the flush attempt?

fractalwrench: I'll switch L245 with L244 - the important thing is that both occur after plugins are loaded

tomlongridge: But I'm still not seeing any justification for moving `registerForNetworkChanges`. I'd like for it either to be moved back to its original location so there's no net change, or if it's key to the change, it should be mentioned in the PR or comments why it should be after  loadPlugins for future reference.

fractalwrench: `registerForNetworkChanges` can invoke the following callback which triggers error flushing, so it's therefore essential to call `loadPlugins` before this: https://github.com/bugsnag/bugsnag-android/blob/82b8236c42099b68920fff1f353cab34d85f1f86/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java#L127

tomlongridge: OK - ☝️ , that's the detail I was looking for.

PR #902, opened 2020-07-13
Avoid attempting to flush unhandled error reports in same session

fractalwrench:
## Goal

When an unhandled error is captured the process is usually about to terminate. Currently bugsnag attempts to send the request which can lead to a condition where a request is successfully sent but the process is terminated before the SDK can act on the response. This can lead to duplicate error reports on the relaunch of the app (and is theorised to cause flakes in the E2E tests).

Disabling this behaviour for unhandled errors will reduce the likelihood of these duplicate requests being sent. This excludes the following ANRs, because the process will not directly terminate and there's a good chance the request will be delivered in the same session.

## Tests

Added unit tests to cover ANR calculations.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
@@ -43,10 +43,22 @@ class DeliveryDelegate extends BaseObservable {
         }
 
         if (event.isUnhandled()) {
-            cacheEvent(event, true);
+            // should only send unhandled errors if they don't terminate the process (i.e. ANRs)
+            cacheEvent(event, isAnr(event));
         } else {
             deliverPayloadAsync(event, eventPayload);
         }
+    }
+
+    boolean isAnr(@NonNull Event event) {
+        List<Error> errors = event.getErrors();
+        String errorClass = null;
+
+        if (!errors.isEmpty()) {
+            Error error = errors.get(0);
+            errorClass = error.getErrorClass();
+        }
+        return "ANR".equals(errorClass);
     }
 
     private void deliverPayloadAsync(@NonNull Event event, EventPayload eventPayload) {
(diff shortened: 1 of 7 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
```diff
@@ -43,12 +43,32 @@ void deliver(@NonNull Event event) {
...
+            Error error = errors.get(0);
+            errorClass = error.getErrorClass();
+        }
+        return "ANR".equals(errorClass);
+    }
+
+    boolean isStackoverflow(@NonNull Event event) {
+        Throwable err = event.getOriginalError();
+        return (err != null) && StackOverflowError.class.equals(err.getClass());
+    }
```
tomlongridge, line 70: These feel like useful methods that could be on `Event` - can they be made protected methods in there?

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
```diff
@@ -43,12 +43,32 @@ void deliver(@NonNull Event event) {
         }
 
         if (event.isUnhandled()) {
-            cacheEvent(event, true);
+            // should only send unhandled errors if they don't terminate the process or
+            // if they can't be reported on next launch. ANRs don't terminate the process
+            // directly and stackoverflow can't be persisted on disk
```
tomlongridge, line 48: Can you investigate this limitation and either clarify the comment (as it still persists these errors) or allow stack overflows to be sent as normal?

fractalwrench: Clarified comment

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
fractalwrench: Moved method to `Event`

PR #911, opened 2020-07-23
Prevent duplicate notifier dependencies being added

fractalwrench:
## Goal

Prevents duplicate notifier dependencies being added when RN's JS layer is reloaded, by resetting the dependency list each time.

## Tests

Updated existing unit test coverage to ensure the notifier version information serializes correctly.

### DIFF
--- a/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
+++ b/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
@@ -51,7 +51,7 @@ class BugsnagReactNativePlugin : Plugin {
         notifier.name = "Bugsnag React Native"
         notifier.url = "https://github.com/bugsnag/bugsnag-js"
         notifier.version = jsVersion
-        notifier.dependencies.add(Notifier()) // depend on bugsnag-android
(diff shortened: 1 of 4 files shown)


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/BugsnagReactNativePlugin.kt
```diff
@@ -51,7 +51,7 @@ class BugsnagReactNativePlugin : Plugin {
         notifier.name = "Bugsnag React Native"
         notifier.url = "https://github.com/bugsnag/bugsnag-js"
         notifier.version = jsVersion
-        notifier.dependencies.add(Notifier()) // depend on bugsnag-android
+        client.notifier.dependencies = listOf(Notifier()) // depend on bugsnag-android
```
tomlongridge, line 54: Small nit: notifier == client.notifier, so I don't think this is necessary and it makes the code harder to understand why .dependencies is set differently.

```suggestion
        notifier.dependencies = listOf(Notifier()) // depend on bugsnag-android
```

fractalwrench: Updated as per suggestion.

PR #916, opened 2020-08-02
Test improvement: removed conditional operator test smell

eas5:
This is a test refactoring

**Problem:**
Conditions within the test method will alter the behavior of the test and its expected output and would lead to situations where the test fails to detect defects in the production method since test statements were not executed as a condition was not met.

**Solution:**
Instead of using a conditional to control the test execution, It is important to use the proper JUnit API with assumeNotNull(). In this case, using assumeNotNull(), the test method will not pass or fail without effectively testing the subject under test, but be ignored if it can not be performed.

**Result:**
_Before:_
`if(exception != null){ ... }`
_After:_
`assumeNotNull(exception);`

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StrictModeTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/StrictModeTest.java
@@ -1,10 +1,12 @@
 package com.bugsnag.android;
 
+import static org.junit.Assert.*;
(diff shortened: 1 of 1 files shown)


johnkiely1:
Hi @eas5.

Thanks for this, looks like it makes sense to add, we will review in due course.

xander-jones:
#925 has been closed out. This has been released in https://github.com/bugsnag/bugsnag-android/releases/tag/v5.1.0

PR #920, opened 2020-08-07
fix(metadata): Ensure metadata updates set with a map notify observers

bengourley:
Metadata updates with a map did not trigger any updates for observers. This fix ensures the observers are notified for each tab that changed.


bugsnag-android-core/src/main/java/com/bugsnag/android/MetadataState.kt
```diff
@@ -55,4 +57,10 @@ internal data class MetadataState(val metadata: Metadata = Metadata()) : BaseObs
             else -> notifyObservers(AddMetadata(section, key, metadata.getMetadata(section, key)))
         }
     }
+
+    private fun notifyMetadataAdded(section: String, value: Map<String,Any?>) {
```
twometresteve, line 61: I think the unit test should be east enough to update to cover this addition.

fractalwrench:
LGTM

PR #933, opened 2020-09-15
[PLAT-4818] Add manual testing mode to Maze-runner test fixture

Cawllec:
## Goal

Enables easier development and testing of the notifier and associated E2E tests by enabling them to be run locally.

## Tests
Run several scenarios locally.

### DIFF
--- a/tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -47,6 +47,8 @@ class MainActivity : Activity() {
 
         clearUserData.setOnClickListener {
             clearStoredApiKey()
+            val apiKeyField = findViewById<EditText>(R.id.manualApiKey)
+            apiKeyField.text.clear()
         }
 
         if (apiKeyStored()) {
--- a/tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -121,6 +121,7 @@ class MainActivity : Activity() {
         val config: Configuration
         if (apiKeyField.text.isNotEmpty()) {
             val manualApiKey = apiKeyField.text.toString()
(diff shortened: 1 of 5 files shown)


Cawllec:
There appears to be an error in building due to the removal of a `null` from the maze-runner scenarios project. This will be address ASAP before merging.

twometresteve (commented):
I haven't had a chance to exercise this dynamically, but on visual inspection it certainly looks like the functionality that I was expecting to see.  I'd be happy to see this merged once the build passes and approved by another.

tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -14,6 +15,8 @@ import com.bugsnag.android.mazerunner.scenarios.Scenario
 
 class MainActivity : Activity() {
 
+    private val apiKeyKey = "BUGSNAG_API_KEY"
```
fractalwrench, line 18:
```suggestion
    private val apiKey = "BUGSNAG_API_KEY"
```

fractalwrench (approved):

tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
twometresteve: Although it looks like a typo I think this is actually correct - "BUGSNAG_API_KEY" is the key for getting the apiKey.

twometresteve (approved):
LGTM.  I built and tested the fixture locally and it behaved as I expected, with errors appearing in the dashboard and apiKey saved/cleared accordingly.

tests/features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
fractalwrench: Ah - I'd assumed that this was a variable that holds the API key. Perhaps something like the following would make it more clear that it's a constant:

`private const val EXTRA_API_KEY = "BUGSNAG_API_KEY"`

twometresteve:
Although I've approved this, should it actually be going into `next`?

PR #937, opened 2020-09-18
Improve stack traces and grouping for React Native promise rejections

fractalwrench:
## Goal

Improves the stack traces reported for React Native promise rejections by including the native stack trace in the reported error.

## Changeset

- Added `NativeErrorDeserializer` class which grabs the `nativeStack` property if set by the JS layer and creates an `Error` object, which is appended as a separate stacktrace
- Passed the JS error to the `NativeErrorDeserializer` and used its error message/class as the default values for the native error
- Moved `inProject` to Stacktrace's [companion object](https://kotlinlang.org/docs/tutorials/kotlin-for-py/objects-and-companion-objects.html) for accessibility from the `NativeErrorDeserializer`

## Testing

In addition to unit testing of deserialization logic, relied on E2E scenarios from https://github.com/bugsnag/bugsnag-js/pull/1047 and manually verified that the stacktrace appeared as expected in the dashboard:

Screenshot 2020-09-18 at 14 38 46

https://github.com/bugsnag/bugsnag-js/compare/bengourley/rn-native-stack...PLAT-5001/native-stack

The RN changes can be trialled in an example app on the following branch: https://github.com/bugsnag/bugsnag-js/tree/PLAT-5001/native-stack

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
@@ -11,6 +11,14 @@ internal class Stacktrace : JsonStream.Streamable {
     companion object {
         private const val STACKTRACE_TRIM_LENGTH = 200
 
+        /**
+         * Calculates whether a stackframe is 'in project' or not by checking its class against
+         * [Configuration.getProjectPackages].
+         *
+         * For example if the projectPackages included 'com.example', then
+         * the `com.example.Foo` class would be considered in project, but `org.example.Bar` would
+         * not.
+         */
         fun inProject(className: String, projectPackages: Collection<String>): Boolean? {
             for (packageName in projectPackages) {
                 if (className.startsWith(packageName)) {
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
@@ -10,6 +10,15 @@ internal class Stacktrace : JsonStream.Streamable {
 
     companion object {
         private const val STACKTRACE_TRIM_LENGTH = 200
+
+        fun inProject(className: String, projectPackages: Collection<String>): Boolean? {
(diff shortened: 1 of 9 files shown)


bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/NativeErrorDeserializer.java
```diff
@@ -0,0 +1,60 @@
+package com.bugsnag.android;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Deserializes an error from the 'nativeStack' property supplied by React Native
```
tomlongridge, line 9: Would be good to document the expected format here.

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/EventDeserializer.kt
```diff
@@ -44,6 +45,19 @@ internal class EventDeserializer(
         event.errors.clear()
         event.errors.addAll(errors.map(errorDeserializer::deserialize))
 
+        if (map.containsKey("nativeStack") && event.errors.isNotEmpty()) {
```
tomlongridge, line 48: Add some comments around what this will do to the errors list and why

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
```diff
@@ -10,6 +10,15 @@ internal class Stacktrace : JsonStream.Streamable {
 
     companion object {
         private const val STACKTRACE_TRIM_LENGTH = 200
+
+        fun inProject(className: String, projectPackages: Collection<String>): Boolean? {
```
tomlongridge, line 14: Can this be documented

tomlongridge (approved):
Change looks good.

Can you add some more commenting to the PR?

fractalwrench:
Updated changeset with some Javadoc/Kdoc

PR #939, opened 2020-09-22
Ensure device ID is set separately to the user ID

fractalwrench:
## Goal

Ensures that the device ID is set separately to the user ID. Previously if the user ID was set on `Configuration` then the device ID would be incorrectly overridden.

This fix alters the `Client` to load the device ID separately on initialization from the `UserRepository` class.

## Testing

Added unit and integration tests to verify that the `UserRepository` can read the device ID and that the Client value is set appropriately.

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
@@ -6,6 +6,7 @@ import static com.bugsnag.android.BugsnagTestUtils.getSharedPrefs;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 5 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/UserRepository.kt
```diff
@@ -46,4 +40,13 @@ internal class UserRepository(private val prefs: SharedPreferences, private val
         }
         editor.apply()
     }
+
+    fun getDeviceId(): String {
```
tomlongridge, line 44: Add some javadoc comments to explain the intent of this function.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
```diff
@@ -327,4 +328,12 @@ public void testUserNotCloned() {
...
         assertNull(user.getEmail());
     }
+
+    @Test
+    public void testDeviceIdNotUserId() {
+        config.setUser("123", "test@example.com", "Tess Derby");
+        client = new Client(context, config);
+        assertEquals("123", client.getUser().getId());
+        assertNotEquals("123", client.getDeviceDataCollector().generateDevice().getId());
+    }
```
tomlongridge, line 338: Do we have any tests that verify that the user ID is populated from the device ID if no user is set in configuration? If not, can you add one so that requirement is fully tested?

bugsnag-android-core/src/main/java/com/bugsnag/android/UserRepository.kt
fractalwrench: Added some KDoc

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
fractalwrench: Added an additional test case

tomlongridge (approved):

PR #944, opened 2020-10-01
Parameterize endpoints

twometresteve:
## Goal

Pass the Bugsnag endpoints in from the UI where possible (i.e. for all scenarios apart from those that load config from the manifest).

## Design

This approach has uses for manual testing, but also opens the door to Appium-driven tests setting the endpoints to custom URLs (especially useful for running automated tests locally during development).

## Changeset

e2e test fixture updates, together with the addition of soft fail for the Android 8.1 build step.  This is pending further investigations in the short-term and as a result of flaky infrastructure causing repeated failures and badly impacting productivity. 

## Testing

The e2e tests would completely fail if this wan't working correctly.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -122,6 +122,7 @@ steps:
         download: "build/fixture.apk"
       docker-compose#v3.3.0:
         run: android-maze-runner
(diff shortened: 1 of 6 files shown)


fractalwrench (approved):
LGTM pending CI

PR #947, opened 2020-10-07
Prevent ConcurrentModificationException when setting redactedKeys

fractalwrench:
## Goal

Prevents a `ConcurrentModificationException` from being thrown when `redactedKeys` is set on the `Metadata` of an `Event`. This can manifest as a crash when `notify()` is called concurrently.

## Changeset

- Converted `ObjectJsonStreamer` to use `var Set<String>` rather than `val MutableSet<String>`, as this avoids the problem of mutating a collection concurrently
- Converted `redactedKeys` to a property on `Metadata` rather than a property and function, as this is more consistent
- Removed `ObjectJsonStreamer` from the constructor as there is no need for it to be injectable

## Testing

Added a unit test to verify that when `redactedKeys` is set on `Metadata` a crash does not occur. Additionally added an integration test which calls `notify()` concurrently to confirm that no further crashes occur with the default configuration.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
@@ -146,6 +146,6 @@ internal data class Metadata @JvmOverloads constructor(
 
     fun copy(): Metadata {
         return this.copy(store = toMap())
-            .also { it.redactedKeys = redactedKeys }
+            .also { it.redactedKeys = redactedKeys.toSet() }
     }
 }
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
@@ -3,7 +3,6 @@
 package com.bugsnag.android
 
 import java.io.IOException
-import java.util.HashSet
 import java.util.concurrent.ConcurrentHashMap
 
 /**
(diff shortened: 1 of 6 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
```diff
@@ -145,9 +144,8 @@ internal data class Metadata @JvmOverloads constructor(
...
     }
 
-    fun copy() = this.copy(
-        store = toMap(),
-        jsonStreamer = jsonStreamer,
-        redactedKeys = redactedKeys
-    )
+    fun copy(): Metadata {
+        return this.copy(store = toMap())
+            .also { it.redactedKeys = redactedKeys }
```
tomlongridge, line 149: Doesn't this want to be a copy of the redactedKeys set?

bugsnag-android-core/src/test/java/com/bugsnag/android/MetadataConcurrentModificationTest.kt
```diff
@@ -18,4 +19,26 @@ internal class MetadataConcurrentModificationTest {
...
+        val orig = Metadata()
+        val executor = Executors.newSingleThreadExecutor()
+
+        repeat(100) {
+            orig.redactedKeys = keys
+            executor.execute {
+                keys.add("$it")
+            }
+        }
+    }
```
tomlongridge, line 35: Some comments would be useful here for future context as it's a non-obvious test.

fractalwrench: Added some KDoc

bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
fractalwrench: In practice a copy isn't required because nothing creates an iterator on the object anymore. For consistency with `objectJsonStreamer` however I've updated this to make a copy of the set.

bugsnag-android-core/src/test/java/com/bugsnag/android/MetadataConcurrentModificationTest.kt
```diff
@@ -18,4 +20,30 @@ internal class MetadataConcurrentModificationTest {
...
+
+    @Test()
+    fun testRedactedKeysCopy() {
+        val orig = Metadata()
+        orig.redactedKeys = mutableSetOf("alpha", "omega")
+        val copy = orig.copy()
+        assertNotSame(orig.redactedKeys, copy.redactedKeys)
```
tomlongridge, line 47: Would be good to have a positive assertion of the copy too: checking for "alpha" and "omega"?

tomlongridge (approved):
LGTM - one request for an extra assertion.

PR #953, opened 2020-10-15
Avoid crash when initializing bugsnag in attachBaseContext

fractalwrench:
## Goal

Avoids a crash when initializing Bugsnag in `Application#attachBaseContext()`. Previously Bugsnag always attempted to obtain the application context to avoid leaking short-lived activity context. However, at this point in time the application context will return null, resulting in an `IllegalArgumentException` within the `Client` constructor.

This changeset fixes this by adding a null check. If the context does return null then it is the base context, and its safe to initialize with as it won't leak memory unlike an `Activity`.

## Testing

Tested manually in an example app as this is not a suitable scenario for mazerunner. Added unit tests to confirm that the original crash does not occur.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+* Avoid crash when initializing bugsnag in attachBaseContext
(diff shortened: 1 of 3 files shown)


twometresteve (approved):
LGTM

PR #954, opened 2020-10-15
Increase breadcrumb time precision to milliseconds

fractalwrench:
## Goal

Where there are breadcrumbs in the second before an error they are all displayed with the same time difference (0s before). Although the order is correct, it would be useful to show more precision when breadcrumbs occur close to the error.

## Changeset

Altered the date format in `DateUtils` to include milliseconds as per the ISO 8601 standard. This does not fully update the NDK layer as `device.time` uses `time_t` in the public struct which is only accurate to milliseconds. The NDK layer will however use breadcrumbs and the sent-at header with millisecond precision.

## Testing

Updated unit tests to assert that milliseconds are included in the serialized timestamp. In addition, verified that the breadcrumb and device timestamp appear as expected for a handled JVM error and unhandled NDK error. Further verified that the NDK serialization change does not result in a crash loop by triggering an unhandled NDK error with v5.2.1, then installing a local artefact before relaunching the app and verifying that the report received by the dashboard was in the correct format.

Screenshot 2020-10-15 at 16 12 44


fractalwrench:
I've noticed that the precision is off for unhandled NDK errors so this will need some additional changes.

tomlongridge (approved):

PR #955, opened 2020-10-15
Bump gradle dependency versions

fractalwrench:
## Goal

Bumps Gradle dependency versions to 6.7 in the project so that we can remain up to date. This also gives us some substantial performance benefits from using later versions of Gradle which use more performant APIs such as Configuration Avoidance. 

## Testing

Ran the example app and generated an APK by hand. Relied on CI to run lint/test tasks.


twometresteve (approved):
LGTM

PR #956, opened 2020-10-16
bump bugsnag-android dependency versions

fractalwrench:
## Goal

Bumps Bugsnag dependency versions in the project so that we can remain up to date with the latest bugsnag. Additionally updated the detekt plugin for static analysis to use the latest version and updated baseline.xml to use new format.

## Testing

Ran the example app and generated an APK by hand. Relied on CI to run lint/test tasks.


twometresteve (approved):
LGTM

PR #960, opened 2020-10-16
Increase gradle's max memory

fractalwrench:
## Goal

Increases the maximum memory available for Gradle builds which should help reduce the amount of time taken to build.


twometresteve (approved):
LGTM

PR #962, opened 2020-10-16
Remove editor config

fractalwrench:
## Goal

The editor config seems to clash with Android Studio's settings and can be removed as it appears to be unused.


twometresteve (approved):
I'm pretty confident we can lose this relic.

PR #964, opened 2020-10-20
Support changing NDK Event's api key in OnErrorCallback

fractalwrench:
## Goal

Adds support for changing the apiKey on events from the NDK layer. #928 added equivalent functionality for JVM events.

## Changeset

The changeset consists mainly of five alterations:

1. Passing the api key to the NDK layer via a `StateEvent.Install` message
2. Populating the api key on the `bugsnag_event` struct, which requires a migration for legacy persisted structs
3. Adding accessors to the NDK API for altering the API key on an event
4. Decoding the apiKey from C struct in `deliverPendingReports()` and pass to the JVM layer along with the deserialized event payload

## Testing

Existing unit and integration tests have been updated to match the new assertions. New test cases have also been added for additional coverage and to verify the NDK migration.

Manual testing also covered:

- Unhandled NDK events are delivered
- API key can be changed on an Unhandled NDK event
- Unhandled NDK events using v3 of the `bugsnag_event` struct are delivered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 * Increase breadcrumb time precision to milliseconds
   [#954](https://github.com/bugsnag/bugsnag-android/pull/954)
 
+* Support changing NDK Event's api key in OnErrorCallback
+  [#964](https://github.com/bugsnag/bugsnag-android/pull/964)
+
 ## 5.2.2 (2020-10-19)
 
 ### Bug fixes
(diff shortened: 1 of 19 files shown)


fractalwrench:
Updated to avoid encoding the apiKey in the NDK filename. Retested by setting the apiKey in an example app and triggering an NDK crash.

tomlongridge (approved):
LGTM.

I think the changeset details in the PR description just need a slight amendment now the NDK file name format has not changed?

PR #967, opened 2020-10-21
Take maze runner image from releases repository

twometresteve:
## Goal

Take the Maze Runner image from its releases repository.

## Design

maze-runner-releases is a new ECR repository of ours that is dedicated to releases and separate to images built for development branches.  It means that releases will not get swept away as part of our housekeeping policy.

## Changeset

Maze Runner docker file(s).

## Testing

Covered by standard CI.


fractalwrench (approved):
LGTM pending CI

PR #968, opened 2020-10-24
Consolidate e2e test scenario classes

twometresteve:
## Goal

Moves the end-to-end test scenarios that were under `/mazerunner-scenarios` to beneath `features/fixtures/mazerunner`.

## Design

For historical reasons the repo had a `mazerunner-scenarios` folder which contained some of the scenarios, whereas the others lived in `features/fixtures/mazerunner`. This change moved the former to the latter so that everything is in one place.  No duplicate classes were found.

## Changeset

Files moved as described above.

## Testing

I've verified that the same number of scenarios are run before and after the refactor and believe I have moved the Detekt configuration across correctly.  A closer eye on the Detekt aspect may be a good idea to ensure it is still correct.


features/fixtures/mazerunner/build.gradle
```diff
@@ -111,3 +112,6 @@ bugsnag {
     reportBuilds = false
     failOnUploadError = false
 }
+
+apply from: "../../../gradle/detekt.gradle"
```
fractalwrench, line 116: I don't think detekt/ktlint/lint run when building the mazerunner fixture. It'd be good to raise a separate task to make that so.

fractalwrench (approved):

PR #971, opened 2020-10-27
Reinstate fast-failing e2e tests

twometresteve:
## Goal

Reinstates the `fail-fast` option, which was accidentally overridden when moving to Maze Runner v3.

## Design

Removed the `command` entry from `docker-compose.yml`, which is a bit hidden away as it is, and added to the command arrays in the build steps.  I have omitted the `--tags` option to reduce noise in the pipeline, as the use of `@ignore` appears to be historical.

## Changeset

Docker compose file and e2e testing steps in the pipeline.

## Testing

Covered by standard CI.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -248,9 +248,13 @@ steps:
         download: "build/fixture.apk"
       docker-compose#v3.3.0:
         run: android-maze-runner
-    env:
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):

twometresteve:
Force merging despite test failures on Android 6 and 7 due to current blocking device farm issue.

PR #973, opened 2020-10-28
Flush persisted sessions on launch and on connectivity changes

fractalwrench:
## Goal

Session payloads are persisted on disk if a request to bugsnag was unsuccessful. Delivery of these is only attempted currently when a new session is captured. To increase the speed with which requests are received, delivery should also be attempted on `Client` initialization` and whenever connectivity changes.

## Changeset

Called `flushAsync()` in the client constructor and connectivity change callback. Access to this method is guarded by a Semaphore so concurrent calls will be ignored.

## Testing

Relied on existing E2E tests, and manually verified that an undelivered session is now delivered when the client is launched or the connectivity is regained.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Bug fixes
 
+* Flush persisted sessions on launch and on connectivity changes
+  [#973](https://github.com/bugsnag/bugsnag-android/pull/973)
+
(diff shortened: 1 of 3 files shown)


fractalwrench:
Test failures on Android 6 and 7 due to current blocking device farm issue.

tomlongridge (approved):

PR #977, opened 2020-10-29
Refactor Stacktrace to not require logger

kstenerud:
## Goal

The new ANR native trace functionality will require native code to produce `Stacktrace` objects, however the constructors for `Stacktrace` require a logger, which is only used when converting JVM stack traces. Rather than passing a logger through native code, it's cleaner to move the "Java trace to Bugsnag trace format" code into a helper function.

## Design

This change moves the "java to bugsnag" constructor into a helper function in the object's companion. The `Stacktrace` object itself is now a POJO, containing data only.

## Testing

- Ran unit tests
- Ran example app, triggering crashes and verifying that the stack traces are correct in Java and native crashes.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
@@ -27,23 +27,44 @@ internal class Stacktrace : JsonStream.Streamable {
             }
             return null
         }
+
+        fun stacktraceFromJavaTrace(
+            stacktrace: Array<StackTraceElement>,
+            projectPackages: Collection<String>,
+            logger: Logger): Stacktrace {
+            val frames = stacktrace.mapNotNull { serializeStackframe(it, projectPackages, logger) }
+            return Stacktrace(frames)
+        }
+
+        private fun serializeStackframe(
+            el: StackTraceElement,
+            projectPackages: Collection<String>,
+            logger: Logger
+        ): Stackframe? {
+            try {
+                val methodName = when {
+                    el.className.isNotEmpty() -> el.className + "." + el.methodName
+                    else -> el.methodName
+                }
+
+                return Stackframe(
+                    methodName,
+                    if (el.fileName == null) "Unknown" else el.fileName,
+                    el.lineNumber,
+                    inProject(el.className, projectPackages)
+                )
+            } catch (lineEx: Exception) {
+                logger.w("Failed to serialize stacktrace", lineEx)
+                return null
+            }
+        }
+
     }
 
     val trace: List<Stackframe>
-    val logger: Logger
 
-    constructor(
-        stacktrace: Array<StackTraceElement>,
-        projectPackages: Collection<String>,
-        logger: Logger
-    ) {
-        trace = limitTraceLength(stacktrace.mapNotNull { serializeStackframe(it, projectPackages) })
-        this.logger = logger
-    }
-
-    constructor(frames: List<Stackframe>, logger: Logger) {
+    constructor(frames: List<Stackframe>) {
         trace = limitTraceLength(frames)
-        this.logger = logger
     }
 
     private fun <T> limitTraceLength(frames: List<T>): List<T> {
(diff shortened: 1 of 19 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
```diff
@@ -6,7 +6,7 @@ import java.io.IOException
  * Serialize an exception stacktrace and mark frames as "in-project"
  * where appropriate.
  */
-internal class Stacktrace : JsonStream.Streamable {
+class Stacktrace : JsonStream.Streamable {
```
fractalwrench, line 9: This should remain `internal` - the default visibility in Kotlin is public, so removing this modifier would expose `Stacktrace` as part of the public API.

bugsnag-android-core/src/test/java/com/bugsnag/android/StacktraceSerializationTest.kt
```diff
@@ -18,10 +18,10 @@ internal class StacktraceSerializationTest {
                 "stacktrace",
 
                 // empty stacktrace element ctor
-                Stacktrace(arrayOf(), emptySet(), NoopLogger),
+                    Stacktrace.stacktraceFromJavaTrace(arrayOf(), emptySet(), NoopLogger),
```
fractalwrench, line 21: style nit: the indentation is inconsistent with other parameters

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
```diff
@@ -27,31 +27,52 @@ internal class Stacktrace : JsonStream.Streamable {
...
         }
-    }
 
-    val trace: List<Stackframe>
-    val logger: Logger
+        fun stacktraceFromJavaTrace(
+            stacktrace: Array<StackTraceElement>,
+            projectPackages: Collection<String>,
+            logger: Logger): Stacktrace {
+            val frames = limitTraceLength(stacktrace.mapNotNull { serializeStackframe(it, projectPackages, logger) })
```
fractalwrench, line 35: The NDK will also need to limit the number of stackframes to 200 - this seems fine as long as the requirement isn't forgotten when implementing a similar method to achieve this via the JNI call.

fractalwrench (commented):
The general approach seems ok here. I added a few minor comments inline, and there also seem to be a couple of compilation errors in the tests + static analysis warnings that will need resolving before this can be merged.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
kstenerud: Hmm I had to expose it in order for the JNI plugin to see it. Is there a way to keep it internal but allow the JNI plugin visibility?

bugsnag-android-core/src/test/java/com/bugsnag/android/StacktraceSerializationTest.kt
kstenerud: For some reason Android Studio insists on this indentation, and backspacing just moves the whole thing to the previous line. I had to manually select 4 spaces and cut them to get back to the original indentation :/

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
kstenerud: Yes, it will use the same `BUGSNAG_FRAMES_MAX ` as https://github.com/bugsnag/bugsnag-android/blob/master/bugsnag-plugin-android-ndk/src/main/jni/event.h#L166

bugsnag-android-core/src/main/java/com/bugsnag/android/Stacktrace.kt
fractalwrench: Ah, I see. Kotlin decompiles down to Java bytecode and [mangles names](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#visibility) for methods in internal classes I believe. You could attempt using [`@JvmName`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-name/) or even making the method private to avoid name mangling, which is what the `AnrPlugin` does: https://github.com/bugsnag/bugsnag-android/blob/69aba9a05e5de94101a03ebb24bf49d7d6438085/bugsnag-plugin-android-anr/src/main/jni/anr_handler.c#L43

fractalwrench (approved):
LGTM pending CI

PR #978, opened 2020-10-29
Add integrity header to verify Error and Session payloads have not changed

fractalwrench:
## Goal

Adds the `Bugsnag-Integrity` header to Error API requests whose value is a SHA-1 digest of the payload. If the payload is altered in transit then it will be rejected by the server.

## Changeset

- Added method for computing SHA-1 digest from an `EventPayload` to `DeliveryHeaders` by using a `DigestOutputStream` with a no-op output
- Added `Bugsnag-Integrity` to the request headers for each Error API request
- Simplified `DefaultDelivery` implementation by using [kotlin stdlib](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) functions

## Testing

Added unit tests to cover generation of the SHA-1 digest, and updated E2E scenarios so that they assert the header is set on error report payloads. Additionally verified via Android Studio's Profiler that an error request contains the header and the payload is received by the dashboard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,9 @@
 # Changelog
 
-## TBD
+## 5.3.0 (TBD)
 
 * Add integrity header to verify Error API payloads have not changed
   [#978](https://github.com/bugsnag/bugsnag-android/pull/978)
-
-## 5.2.3 (2020-10-29)
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Add integrity header to verify Error API payloads have not changed
+  [#978](https://github.com/bugsnag/bugsnag-android/pull/978)
+
 ## 5.2.3 (2020-10-29)
 
 ### Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 5.3.0 (TBD)
 
-* Add integrity header to verify Error API payloads have not changed
+* Add integrity header to verify Error and Session API payloads have not changed
   [#978](https://github.com/bugsnag/bugsnag-android/pull/978)
 
 ### Bug fixes
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/DefaultDelivery.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/DefaultDelivery.kt
@@ -1,7 +1,7 @@
 package com.bugsnag.android
 
-import java.io.BufferedWriter
 import java.io.IOException
+import java.io.OutputStream
 import java.io.OutputStreamWriter
 import java.net.HttpURLConnection
 import java.net.HttpURLConnection.HTTP_BAD_REQUEST
@@ -9,6 +9,8 @@ import java.net.HttpURLConnection.HTTP_CLIENT_TIMEOUT
 import java.net.HttpURLConnection.HTTP_OK
 import java.net.URL
 import java.nio.charset.Charset
+import java.security.DigestOutputStream
+import java.security.MessageDigest
 
 internal class DefaultDelivery(private val connectivity: Connectivity?, val logger: Logger) : Delivery {
 
@@ -27,7 +29,7 @@ internal class DefaultDelivery(private val connectivity: Connectivity?, val logg
     fun deliver(
         urlString: String,
         streamable: JsonStream.Streamable,
-        headers: Map<String, String>
+        headers: Map<String, String?>
     ): DeliveryStatus {
 
         if (connectivity != null && !connectivity.hasNetworkConnection()) {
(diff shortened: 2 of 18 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryHeaders.kt
```diff
@@ -12,19 +16,44 @@ internal const val HEADER_INTERNAL_ERROR = "Bugsnag-Internal-Error"
...
+
+internal fun computeSha1Digest(payload: EventPayload): String? {
+    runCatching {
+        val shaDigest = MessageDigest.getInstance("SHA-1")
+        val builder = StringBuilder("sha1 ")
+
+        // Pipe the object through a no-op output stream
+        DigestOutputStream(NullOutputStream(), shaDigest).use { stream ->
+            stream.bufferedWriter().use { writer ->
+                payload.toStream(JsonStream(writer))
```
kstenerud, line 47: It looks like DefaultDelivery is already serialising to JSON. Can't we use that data instead of serialising a second time?

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryHeaders.kt
```diff
@@ -12,19 +16,44 @@ internal const val HEADER_INTERNAL_ERROR = "Bugsnag-Internal-Error"
...
  * Supplies the headers which must be used in any request sent to the Session Tracking API.
  *
  * @return the HTTP headers
  */
-internal fun sessionApiHeaders(apiKey: String) = mapOf(
+internal fun sessionApiHeaders(apiKey: String): Map<String, String?> = mapOf(
     HEADER_API_PAYLOAD_VERSION to "1.0",
     HEADER_API_KEY to apiKey,
     HEADER_BUGSNAG_SENT_AT to DateUtils.toIso8601(Date())
 )
```
kstenerud, line 37: Does this need `HEADER_BUGSNAG_INTEGRITY` as well?

bugsnag-android-core/src/main/java/com/bugsnag/android/DefaultDelivery.kt
```diff
@@ -1,14 +1,16 @@
...
+import java.io.OutputStream
 import java.io.OutputStreamWriter
 import java.net.HttpURLConnection
 import java.net.HttpURLConnection.HTTP_BAD_REQUEST
 import java.net.HttpURLConnection.HTTP_CLIENT_TIMEOUT
 import java.net.HttpURLConnection.HTTP_OK
 import java.net.URL
 import java.nio.charset.Charset
+import java.security.DigestOutputStream
+import java.security.MessageDigest
```
kstenerud, line 13: These classes don't seem to be used in this file?

fractalwrench: Removed unused imports

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryHeaders.kt
fractalwrench: As the request uses chunked transfer encoding to minimise memory footprint, the Java object/File is streamed directly into the HTTP request, meaning the payload size is not known ahead of time. I don't think it's possible to get around needing to serialize twice without sacrificing chunked transfer encoding - and this is what was chosen during the design phase.

twometresteve (commented):
I've comments primarily from the test perspective - happy for @kstenerud to provide the approval.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryHeaders.kt
fractalwrench: Yes - this was missing from the design, I'll add this in.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
```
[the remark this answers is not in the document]
fractalwrench: Updated and bumped version to 5.3.0 as this is an additive change and not a bug fix.

fractalwrench:
Updated PR so that session payloads now use the `Bugsnag-Integrity` header too, and manually reverified.

kstenerud (approved):

twometresteve:
Only thing I've now noticed is that we don't have a MazeRunner check in place for the header being present in Session requests.  However, I can see that it's caught up in the fact that Android has a local implementation for `the request is valid for the error reporting API version` but not `the request is valid for the session reporting API version` (which is in Maze Runner itself).

I think if we could get a check added to any of the session scenarios in the short term, that would be good.  Then longer term we can refactor the steps so that they are absorbed into MazeRunner and the check added to the session steps.

fractalwrench:
@twometresteve updated the session tracking assertions and created a ticket to eventually move the steps into the main mazerunner repo.

twometresteve (approved):
LGTM from the test perspective.

PR #980, opened 2020-11-03
Add configuration option to control maximum number of persisted events/sessions

fractalwrench:
## Goal

The bugsnag-android SDK persists events & sessions to disk if it is unable to deliver them to the API due to network failure. This provides a configuration option to control how many payloads should be stored on disk for both events and sessions.

## Changeset

- Added `maxPersistedEvents` and `maxPersistedSessions` to `Configuration` with a default value of 32
- Added `MAX_PERSISTED_EVENTS` and `MAX_PERSISTED_SESSIONS` for reading configuration options from the manifest
- Updated `EventStore` and `SessionStore` so that they respect the configuration options
- Fixed a bug in `discardOldestFileIfNeeded()`, where in the unlikely scenario that >32 payloads were persisted on disk, the function would delete all the payloads if another payload needed persisting

## Testing

Added integration tests which verify the event/session store classes only persist up to the configured limit, and that this limit can be changed. Unit tests verify that the max persisted events/sessions configuration options are validated as necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Add configuration option to control maximum number of persisted events/sessions
+  [#980](https://github.com/bugsnag/bugsnag-android/pull/980)
+
 ## 5.3.0 (TBD)
 
 * Add integrity header to verify Error and Session API payloads have not changed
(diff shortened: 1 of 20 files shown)


tomlongridge:
Do we definitely want to allow the value 0 for this configuration? Does that not effectively break the functionality as nothing would be delivered?

fractalwrench:
>Do we definitely want to allow the value 0 for this configuration? Does that not effectively break the functionality as nothing would be delivered?

I felt this was a reasonable, if a bit of an unusual, use-case. It would still allow users to send handled errors and sessions if they didn't want to persist them on disk if delivery was not possible. This is effectively what we do for Internal Error reporting already.

tomlongridge (approved):

PR #987, opened 2020-11-05
Assert that Bugsnag-Integrity header is valid

fractalwrench:
## Goal

Asserts that the `Bugsnag-Integrity` header is a valid SHA-1 digest, by hashing the payload using mazerunner.


twometresteve (approved):
I'm happy if CI is happy.  :-)

PR #989, opened 2020-11-06
Conditionally fill in native portion of ANR stack trace

kstenerud:
## Goal

ANRs can occur in JVM code and in native code. We only want to add the native trace if the ANR occurred in native code.

## Design

Check the JVM trace leading up to the ANR notify method. If the last method before it was native, assume that this is a native ANR and modify the JVM stack trace to include the native frames at the end before reporting the event.

## Changeset

All changes are in `notifyAnrDetected` in `AnrPlugin.kt`, with two new private methods.

## Testing

These changes were tested manually since the mazerunner tests will be done in a separate PR. I manually tested both native and JVM ANRs in a modified example app, and verified the stack traces that were reported to the backend.


bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -42,31 +43,70 @@ internal class AnrPlugin : Plugin {
...
+        val notifyIndex = getNotifyAnrDetectedIndex(stackTrace)
+        if (notifyIndex < 0) {
+            return false
+        }
+
+        return stackTrace[notifyIndex + 1].isNativeMethod
+    }
+
+    private fun mergeTraces(jvmTrace: Array<StackTraceElement>, nativeTrace: List<Stackframe>):
+        Array<StackTraceElement> {
```
fractalwrench, line 69: This function needs to return `com.bugsnag.android.Stackframe` - it appears as though it's currently discarding the information gathered in the NDK stacktrace by attempting to convert it into a `StacktraceElement`

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -42,31 +43,70 @@ internal class AnrPlugin : Plugin {
...
     /**
      * Notifies bugsnag that an ANR has occurred, by generating an Error report and populating it
      * with details of the ANR. Intended for internal use only.
      */
     @Suppress("UNUSED_PARAMETER")
     private fun notifyAnrDetected(info: Long, userContext: Long) {
-        val thread = Looper.getMainLooper().thread
-
         // generate a full report as soon as possible, then wait for extra process error info
-        val exc = RuntimeException()
```
fractalwrench, line 54: this shouldn't be moved unless there's a good reason to do so as it affects the accuracy of the stacktrace?

fractalwrench (changes requested):
Left a few comments inline - I'm not convinced that this approach will work as-is since it looks like the NDK stackframe information is being discarded, but I didn't run this in an example app so could be wrong.

Several of these functions are really good candidates for unit tests - in general at Bugsnag we'd aim to write unit tests for as much of the code as is practical. If you need a few pointers getting setup with JUnit and the tests targets we can spend some time going through it.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
kstenerud: The real stack trace is pulled from `Looper.getMainLooper().thread.stackTrace` at the beginning of the method. The `RuntimeException` is just a holder for this trace (its `stackTrace` var gets replaced with the trace grabbed from the main thread) so that the contrived exception can be converted to the internal trace format via the standard path.

fractalwrench: Ah, got it - this is fine then! :)

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
kstenerud: The NDK is only filling out filename, method, and line number. I didn't realise when writing the native portion that the JVM side is taking StackTraceElements and then converting to Stackframe. Maybe I should modify the native side to return StackTraceElement instead since the other Stackframe fields are used only for react native, or for project packages?

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -42,31 +43,70 @@ internal class AnrPlugin : Plugin {
...
+                stackTrace.add(
+                    StackTraceElement(
+                        "",
+                        frame.method,
+                        frame.file,
+                        frame.lineNumber?.toInt() ?: 0
+                    )
+                )
+            }
+            for (i in notifyIndex + 1..jvmTrace.size - 1) {
```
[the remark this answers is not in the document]
kstenerud: That would attempt to add into the destination at that index. I need to grab the source from that index to the end. I could change this to use `takeLast`, but that would create an intermediary list before adding to the destination.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeStackframe.kt
```diff
@@ -0,0 +1,52 @@
+package com.bugsnag.android
+
+import java.io.IOException
+
+/**
+ * Represents a single native stackframe
+ */
+class NativeStackframe internal constructor(
```
fractalwrench, line 8:
```suggestion
internal class NativeStackframe internal constructor(
```

Confusingly the default visibility in Kotlin [is public](origin/5317-conditional-ndk-anr-2). As there isn't any requirement to expose `NativeStackframe` as part of public API it can be hidden for now.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeStackframe.kt
```diff
@@ -0,0 +1,52 @@
...
+    var symbolAddress: Long?,
+
+    /**
+     * The address of the library where the event occurred.
+     */
+    var loadAddress: Long?
+) : JsonStream.Streamable {
+
+    @Throws(IOException::class)
+    override fun toStream(writer: JsonStream) {
```
fractalwrench, line 42: It'd be good to create a unit test for this - `StackframeSerializationTest` would be a good starting point. It's a parameterized unit test that generates JSON, then loads a test fixture file from `test/resources/stackframe.json` to compare the expected vs observed JSON.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -5,7 +5,7 @@ import java.io.IOException
 /**
  * Represents a single stackframe from a [Throwable]
  */
-class Stackframe @JvmOverloads internal constructor(
+class Stackframe @JvmOverloads constructor(
```
fractalwrench, line 8: this should remain internal - we never want to expose constructors on classes within the `Event` payload to users, as it's an implementation detail and will constrain future changes.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -46,27 +67,26 @@ internal class AnrPlugin : Plugin {
...
-        val thread = Looper.getMainLooper().thread
-
         // generate a full report as soon as possible, then wait for extra process error info
-        val exc = RuntimeException()
-        exc.stackTrace = thread.stackTrace
+        var stackTrace = Looper.getMainLooper().thread.stackTrace
+        var nativeTrace: List<NativeStackframe>? = null
 
         @Suppress("UNCHECKED_CAST")
         val clz = Class.forName("com.bugsnag.android.NdkPlugin") as Class<Plugin>
```
fractalwrench, line 76: I missed this on the last review, but it'd probably be worth wrapping the whole native stacktrace generation with a `try-catch`, as JVM reflection is brittle and prone to throwing exceptions. If an exception does occur at least we would still report something rather than terminating the process.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -40,8 +40,20 @@ class Stackframe @JvmOverloads internal constructor(
     var columnNumber: Number? = null
 ) : JsonStream.Streamable {
 
+    private var nativeFrame: NativeStackframe? = null
+
+    constructor(nativeFrame: NativeStackframe): this(nativeFrame.method, nativeFrame.file, nativeFrame.lineNumber, false, null) {
```
fractalwrench, line 45: I think the arguments should probably be null for this, as the design's requirement is that the properties would not be publicly accessible. Furthermore if a user edited the `method` with this approach, the change would not be reflected in the payload sent to bugsnag - which would be confusing behaviour

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeStackframe.kt
kstenerud: It's already in https://github.com/bugsnag/bugsnag-android/pull/989/files#diff-6e669e8f7bb6e07399cd1971023f9ce8b119db6c1806f13359434fa63e919af3 and https://github.com/bugsnag/bugsnag-android/pull/989/files#diff-b8e696c2d2762d03cd3678387602472c30f89842f1d8fafe47c46adc42da936eR18

fractalwrench: Nice, beat me to it! :D

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
kstenerud: I had to open it up so that the ANR tests could see it. Is there another way to make it visible?

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeStackframe.kt
kstenerud: This also needs to be accessible to AnrPlugin. How can I keep it "internal" but accessible to that plugin?

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
kstenerud: This is where the model starts to break down... I'm not so sure that Stackframe is the right holder for the native frame, since it's essentially lying to the outside world via its public interface.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
kstenerud: OK, I've wrapped the whole thing in a try-catch

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
kstenerud: I've made this constructor internal now. The other constructor is public, but only accepts a NativeStackframe, so it's pretty safe I think.

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
kstenerud: Updated to propagate the setters to the wrapped native trace if it's available.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeStackframe.kt
kstenerud: This one I couldn't set to internal because AnrPlugin needs to use it directly...

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -5,43 +5,86 @@ import java.io.IOException
...
-
+class Stackframe : JsonStream.Streamable {
     /**
      * The name of the method that was being executed
      */
-    var method: String?,
+    var method: String?
+    set(value) {
+        field = value
+        val frame = this.nativeFrame
```
fractalwrench, line 15: Kotlin has a safe call operator which performs this null check behind the scenes. Therefore this could be stripped down a bit to: `nativeFrame?.file = value`

fractalwrench (approved):
This LGTM pending one very minor comment, and fixing the E2E tests, which I'll help investigate shortly 👁️

PR #990, opened 2020-11-06
Improve logging in mazerunner fixture

fractalwrench:
## Goal

Increases logging in the mazerunner fixture to aid debugging. This now logs at several key points:

- When the fixture is launched
- When user data is cleared
- When a scenario is loaded
- When a scenario is executed


twometresteve (approved):
LGTM

PR #993, opened 2020-11-10
Karl check segv

kstenerud:
DELETEME


fractalwrench:
Did the segfault issue still require investigation or can this be closed?

PR #995, opened 2020-11-10
Separate static analysis into individual jobs

fractalwrench:
## Goal

Separates static analysis into individual jobs on CI. This ensures that if multiple static analysis tools are failing, the developer is always aware of the failures without having to push multiple times. Parallelising the steps should also reduce the time taken to get feedback on the static analysis tasks by ~1 minute.


twometresteve (approved):
LGTM

PR #996, opened 2020-11-10
Add test fixture that excludes the ANR/NDK plugins

fractalwrench:
## Goal

Adds a test fixture that excludes the ANR/NDK plugins to provide confidence that bugsnag is capable of functioning without the ANR/NDK plugins [as we document](https://docs.bugsnag.com/platforms/android/#reporting-ndk-errors).

## Changeset

The mazerunner test fixture required some Gradle restructuring in order to support this, but the overall functionality of the scenarios has not been altered.

- Created an `app` module which is capable of building the fixture APK
- Moved `app` module sources to the standard Gradle project location. This depends on the following modules:
  - `jvm-scenarios` which includes crashy JVM scenarios
  - `cxx-scenarios` which includes crashy CXX scenarios that don't require linking against the bugsnag SO file
  - `cxx-scenarios-bugsnag` which includes crashy scenarios that require linking against the bugsnag SO file (this makes up the majority of the diff)
- Removed unnecessary `loadLibrary` calls across the CXX scenarios as otherwise this would crash the fixtures with link errors
- Setup gradle to conditionally include `cxx-scenarios-bugsnag` through use of `MINIMAL_FIXTURE` envar, which is set to the appropriate value for each buildkite step
- Added a gradle task to extract SO files from the bugsnag AARs in order to get the `cxx-scenarios-bugsnag` module compiling. This is required as a workaround as the BAGP does not support library modules yet
- Bumped Gradle/AGP/Docker versions to latest available, removed unused cruft from the fixture project
- Added two jobs to buildkite pipeline:
  - One that builds the minimal fixture APK
  - One that runs it against the `unhandled_exception.feature`


features/native_crash_handling.feature
```diff
@@ -1,212 +1,5 @@
 Feature: Native crash reporting
 
-    Scenario: Dereference a null pointer
```
twometresteve, line 3: Did you mean to remove all of these?

fractalwrench: Nope, that was accidentally left in after speeding up a run of two scenarios with BrowserStack local. I've reinstated these test cases

twometresteve (approved):
LGTM, pending CI.

PR #997, opened 2020-11-12
5393 native anr scenarios

kstenerud:
## Goal

We have ANR scenarios, but none that are caused from native code

## Design

Add additional scenarios that are caused by native code.

## Testing

These are e2e tests.


fractalwrench:
Rebased branch to resolve conflicts - this adds the minimal test fixture which means it'll be possible to run scenarios against an app that doesn't have bugsnag's NDK plugin installed.

features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXAnrScenario.java
```diff
@@ -0,0 +1,28 @@
...
+import com.bugsnag.android.Configuration;
+
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+public class CXXAnrScenario extends Scenario {
+
+    static {
+        System.loadLibrary("monochrome");
```
fractalwrench, line 12: This can be removed - it's only necessary to load the `cxx-scenarios` SO library since the minimal test fixture has been added.

features/detect_anr.feature
```diff
@@ -11,6 +11,20 @@ Scenario: Sleeping the main thread with pending touch events when autoDetectAnrs
     And the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
     And the exception "errorClass" equals "ANR"
     And the exception "message" starts with " Input dispatching timed out"
+    And the exception "type" equals "android"
+
+@skip_android_8_1
+Scenario: Sleeping the main thread with pending touch events when autoDetectAnrs = true
```
fractalwrench, line 17: It would be good to rename this scenario and the previous one to indicate that one triggers an ANR in the NDK layer, and one triggers an ANR in the JVM layer.

features/detect_anr.feature
```diff
@@ -33,6 +61,15 @@ Scenario: Sleeping the main thread with pending touch events
...
     Then I should receive no requests
 
+@skip_android_8_1
+Scenario: Sleeping the native main thread with pending touch events with ANR disabled
+    When I run "CXXAnrWithNdk_AnrDisabledScenario"
+    And I wait for 2 seconds
+    And I tap the screen 3 times
+    And I wait for 4 seconds
+    And I clear any error dialogue
+    Then I should receive no requests
```
fractalwrench, line 71: This, along with `CXXAnrWithNdk_AnrDisabledScenario`, doesn't trigger an ANR so the test will always pass. As a general rule mazerunner scenarios work best when they always send a request, as otherwise the fixture could crash and we'd never know. I'd suggest calling `Bugsnag.notify` after the ANR has been triggered to achieve this.

fractalwrench:
Updated the changeset with the following to enable ANR scenarios:

- Moved scenarios to `features/full` and `features/minimal` depending on which feature they run against, as suggested by @twometresteve 
- Added `CXXAnrScenario/CXXAnrNdkDisabledScenario` which trigger native ANRs
- Renamed existing ANR scenario classes and converted them to trigger purely JVM ANRs
- Split `detect_anr.feature` into three separate features:
  - One for JVM scenarios
  - One for CXX scenarios
  - One for an ANR in a minimal test fixture

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/JvmAnrMinimalFixtureScenario.kt
```diff
@@ -5,23 +5,24 @@ import com.bugsnag.android.Configuration
...
-        config.enabledErrorTypes.ndkCrashes = false
     }
 
     override fun run() {
         super.run()
+        Bugsnag.notify(generateException())
         val main = Handler(Looper.getMainLooper())
         main.postDelayed(Runnable {
-            Thread.sleep(50000) // FOREVER
+            Thread.sleep(50000)
```
tomlongridge, line 25: Can you inline-comment the rationale for choosing `Thread.sleep` over a `while(true)`? I think that would help future selves read the test.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -103,6 +103,12 @@ internal class AnrPlugin : Plugin {
                 err.stacktrace.addAll(0, nativeTrace.map { Stackframe(it) })
             }
 
+            if (nativeTrace != null) {
+                for (error in event.errors) {
+                    error.type = ErrorType.C
+                }
+            }
+
```
tomlongridge, line 111: I don't think we want to do this - the frame type will (eventually) be set.

features/full/detect_anr_jvm.feature
```diff
@@ -0,0 +1,37 @@
...
+    And I wait for 2 seconds
+    And I tap the screen 3 times
+    And I wait for 4 seconds
+    And I clear any error dialogue
+    Then I wait to receive a request
+    And the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+    And the payload field "events" is an array with 1 elements
+    And the exception "errorClass" equals "java.lang.RuntimeException"
+    And the exception "message" equals "JvmAnrDisabledScenario"
+    And the payload field "events.0.device.cpuAbi" is a non-empty array
```
tomlongridge, line 28: Consider removing as it makes the test harder to understand.

features/full/detect_anr_cxx.feature
```diff
@@ -0,0 +1,27 @@
...
+    When I run "CXXAnrScenario"
+    And I wait for 2 seconds
+    And I tap the screen 3 times
+    And I wait for 4 seconds
+    And I clear any error dialogue
+    Then I wait to receive a request
+    And the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+    And the exception "errorClass" equals "ANR"
+    And the exception "message" starts with " Input dispatching timed out"
+    And the exception "type" equals "c"
```
tomlongridge, line 14: As per comment above - remove this and we'll add the native assertions once the frame-level functionality is merged.

features/full/detect_anr_jvm.feature
fractalwrench: Removed step

features/full/detect_anr_cxx.feature
fractalwrench: Removed assertion for now

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
fractalwrench: Removed this override of frame type.

tomlongridge (approved):

PR #998, opened 2020-11-12
Add persistenceDirectory config option for controlling event/session storage

fractalwrench:
## Goal

Adds a `persistenceDirectory` config option that controls where events/sessions are stored. By default this remains `context.cacheDir`. This gets us one step closer to supporting Clients in multiple processes.

## Changeset

- Added `persistenceDirectory` option to `Configuration` that allows users to set a custom File
- Updated `FileStore` constructor to take a `File` that specifies the directory where reports should be persisted. If it is not possible to write to the directory an error is logged
- Refactored existing storage code to use greater use of `File` rather than `String` for calculating paths
- Updated `NativeInterface.getReportPath()` to use `config.persistenceDirectory` when calculating storage directory

## Testing

- Added unit tests to assert that `persistenceDirectory` can have its values set on the public API as expected
- Assert that `persistenceDirectory` is set to the `context.cacheDir` by default
- Refactor `Event/SessionStoreMaxLimitTest` to be unit tests, as `FileStore` no longer requires a `Context` and can therefore run on the JVM (this is much faster)
- Updated tests to use a custom `persistenceDirectory` and Files rather than Strings where necessary


bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -45,32 +42,30 @@
...
         try {
-            path = appContext.getCacheDir().getAbsolutePath() + folder;
-
-            File outFile = new File(path);
-            outFile.mkdirs();
-            if (!outFile.exists()) {
-                this.logger.w("Could not prepare file storage directory");
-                path = null;
+            storageDir.mkdirs();
+            if (!storageDir.isDirectory() || !storageDir.canWrite()) {
```
tomlongridge, line 58: `storageDir` can no longer be `null`, so some of the guards in `FileStore` and `EventStore` need updating. 

Perhaps we should set `storageDir` to `null` here in the error case and guard against `null` everywhere? Either way, make sure all the operations on `storageDir` are suitably guarded/caught to make to avoid an exception.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -53,12 +51,14 @@ public int compare(File lhs, File rhs) {
     };
 
     EventStore(@NonNull ImmutableConfig config,
-               @NonNull Context appContext,
                @NonNull Logger logger,
                Notifier notifier,
                Delegate delegate) {
-        super(appContext, "/bugsnag-errors/", config.getMaxPersistedEvents(),
-                EVENT_COMPARATOR, logger, delegate);
+        super(new File(config.getPersistenceDirectory(), "bugsnag-errors"),
```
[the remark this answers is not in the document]
fractalwrench: This would break delivery of legacy payloads which was decided against in the design, so the old directory name still needs to be used

bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.java
fractalwrench: Updated to remove null guards. I think it's preferable to check whether the directory exists when writing to it, as some places were assuming that `storageDir` was non-null anyway - e.g. `SessionStore#getFilename()`.

tomlongridge: Can I propose that we go the other direction and set `storageDir` to `null` in the constructor if the value is not suitable (i.e. not a writable directory). It feels like allowing things to proceed with just a log message leaves the processing open to issues. For example, if the directory is not writable then it looks like the `write` and `enqueueContentForDelivery` methods would cause an internal report for something that is a configuration issue.

However, if all the file IO was skipped if there was no storageDir, we can ensure that no processing happens if validation in the constructor fails (rather than distributing the validation throughout the code).

To make this better encapsulated, I believe we could move the directory part of the filename out of the duty of `getFilename` for the Event- and SessionStores, so that it literally just returns the file name. The directory can then be prepended by FileStore (having performed the necessary null check).

bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.java
```diff
@@ -138,7 +130,6 @@ String write(@NonNull JsonStream.Streamable streamable) {
 
     void discardOldestFileIfNeeded() {
         // Limit number of saved payloads to prevent disk space issues
-        File storageDir = new File(storeDirectory);
         if (storageDir.isDirectory()) {
             List<File> files = new ArrayList<>(Arrays.asList(storageDir.listFiles()));
```
tomlongridge, line 134: Not a new issue, but I've noticed that `listFiles` can return `null` if there's an IO problem and this would throw an NPE.

fractalwrench: Updated to add a null check. Also refactored `getFilename()` to avoid requiring `storageDir` to be package private.

tomlongridge (approved):

PR #999, opened 2020-11-16
Disable Gradle Android plugin features

arturdryomov:


gradle.properties
```diff
@@ -30,3 +30,13 @@ ANDROID_TARGET_SDK_VERSION=28
 ANDROID_COMPILE_SDK_VERSION=28
 android.useAndroidX=true
 android.enableJetifier=true
+
+# Android: disable features
+# Reference: https://developer.android.com/studio/releases/gradle-plugin#4-0-0-new
+android.defaults.buildfeatures.aidl=false
+android.defaults.buildfeatures.buildconfig=false
```
arturdryomov, line 37: This one in particular should be the most useful one. Currently Bugsnag ships this `BuildConfig` file:

```java
public final class BuildConfig {
    public static final boolean DEBUG = false;
    public static final String APPLICATION_ID = "com.bugsnag.android.core";
    public static final String BUILD_TYPE = "release";
    public static final String FLAVOR = "";
    public static final int VERSION_CODE = -1;
    public static final String VERSION_NAME = "";

    public BuildConfig() {
    }
}
```

I don’t believe it’s used (and should be used). The generation was disabled in #343 and re-enabled in #491. I haven’t found reference points in the current implementation for the latter change so it should be safe enough to disable it once again.

johnkiely1:
Hi @arturdryomov,

Thanks for the PR. Unfortunately we wouldn't be able to accept this as is.  The build options you are suggesting, were introduced in AGP v4.0.0. [Bugsnag-android](https://github.com/bugsnag/bugsnag-android/blob/master/build.gradle#L12) does not yet use AGP 4. Having said that we would accept a PR that disables the generation of BuildConfig as long as it is applicable to the currently used version of AGP.  

Thanks

PR #1001, opened 2020-11-17
Support setting error type on individual stack frames

fractalwrench:
## Goal

Supports setting the error type on individual stackframes by setting the `Bugsnag-Stacktrace-Types` header. This allows the pipeline to distinguish events and process them accordingly. Supported values are `android, c, reactnativejs`. These are set as a comma delimited string.

## Changeset

- Updated `errorApiHeaders()` to set `Bugsnag-Stacktrace-Types` header if the value has been set
- Calculated value of `Bugsnag-Stacktrace-Types` from the event (by checking all the types set on every stackframe) and from the file (by checking encoded info in the filename)
- Updated `EventStore` to encode the error types in the event file
- The `AnrPlugin` now updates the `thread.0.stacktrace` when capturing an error - previously only `errors.0.stacktrace` was updated, which lead to a disparity in information on the dashboard
- Created `EventFilenameInfo` file which holds classes/functions responsible for encoding/decoding event information into an `Event`

## Testing

- Added unit tests to verify error API headers, the error type calculation, and the event store filename
- Updated existing stackframe serialization tests to verify `type` is serialized
- Enhanced existing E2E scenarios to include checks on `type` field
- Verified that ANRs set their error type accordingly via E2E scenarios


tomlongridge (changes requested):
The API spec, and the intent of the change, is for the field says that the frame type should be set where it doesn't match the error type. I don't think we should be setting the type for all NDK frames - only when it's been set explicitly either in a callback or by one of the new features.

fractalwrench:
Updated changeset so that the `type` is only set on individual stackframes when the error object is mixed (as with native ANRs). E2E scenarios have also been added to verify this.

This should be ready for another round of review, pending the results of CI.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -365,13 +365,24 @@ public static void notify(@NonNull final String name,
             public boolean onError(@NonNull Event event) {
                 event.updateSeverityInternal(severity);
                 List<Error> errors = event.getErrors();
+                Error error = event.getErrors().get(0);
+
+                // update the thread's stacktrace and type to C
+                for (Thread thread : event.getThreads()) {
+                    if (thread.getErrorReportingThread()) {
+                        thread.setStacktrace(error.getStacktrace());
```
tomlongridge, line 373: Why do we want to update the stacktrace, not just the type?

bugsnag-android-core/src/main/java/com/bugsnag/android/EventPayload.kt
```diff
@@ -30,6 +30,58 @@ class EventPayload : JsonStream.Streamable {
...
+            // validate that this only contains valid error type info
+            val valid = errorTypes.split(",").all { errorType ->
+                validValues.contains(errorType)
+            }
+            if (valid) {
+                return errorTypes
+            }
+        }
+        return null
+    }
```
tomlongridge, line 69: This doesn't feel like the correct place to put this logic - it's separated from the code in `EventStore` that writes the filename and requires that `EventStore` constructs an `EventPayload` just to call this method.

Can we abstract it into some sort of helper class that deals with encoding and decoding the filename so that it can be accessed easily in both contexts and the encoding and decoding (including `getApiKeyFromFilename`) is maintained together.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -23,7 +23,7 @@
  */
 class EventStore extends FileStore {
 
-    private static final String STARTUP_CRASH = "_startupcrash";
+    private static final String STARTUP_CRASH = "startupcrash";
```
tomlongridge, line 26: Does the presence or not of the underscore affect decoding of API key or event types? I notice that until now, the "not-jvm" part doesn't have an underscore. This changes seems good to make it consistent, but could it affect decoding the information at all?

I notice the javadoc (L222) needs updating for new format.

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
fractalwrench: After further consideration I don't think is required - we only alter the thread's stacktrace for unhandled errors. Reverting this change.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
fractalwrench: `isLaunchCrashReport` checks for the presence of `_startupcrash.json` and `not-jvm` is not explicitly checked anywhere. The underscore has been added in the `String.format` call for greater consistency so this hsouldn't affect decoding info.

Javadoc updated.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventPayload.kt
fractalwrench: I've extracted this into the `EventFilenameInfo` file, which is responsible for encoding/decoding event information from filenames. It additionally extracted several functions that were previously in the `EventStore`.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventFilenameInfo.kt
```diff
@@ -0,0 +1,165 @@
...
+                name.substring(start, end)
+            }
+            return apiKey ?: config.apiKey
+        }
+
+        /**
+         * Retrieves the error types encoded in the filename, or an empty string if this
+         * information is not encoded for the given event
+         */
+        internal fun findErrorTypesInFilename(eventFile: File): Set<ErrorType> {
```
tomlongridge, line 99: Is this made more accessible so that it can be called from `EventPayload` which doesn't have a config? IMO, keeping this method private to this class and maintaining the pattern is more important than avoiding adding a config field to `EventPayload`.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -219,45 +191,21 @@ String getApiKeyFromFilename(File file) {
 
     /**
      * Generates a filename for the Event in the format
-     * "[timestamp]_[apiKey]_[UUID][startupcrash|not-jvm].json"
+     * "[timestamp]_[apiKey]_[errorTypes]_[UUID]_[startupcrash|not-jvm].json"
```
tomlongridge, line 194: Remove this part of javadoc so the format isn't maintained here.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventFilenameInfo.kt
```diff
@@ -0,0 +1,165 @@
...
+
+    fun isLaunchCrashReport(): Boolean = suffix == STARTUP_CRASH
+
+    internal companion object {
+        private const val STARTUP_CRASH = "startupcrash"
+        private const val NON_JVM_CRASH = "not-jvm"
+
+        /**
+         * Generates a filename for the Event in the format
+         * "[timestamp]_[apiKey]_[errorTypes]_[UUID]_[startupcrash|not-jvm].json"
```
tomlongridge, line 41: Suggest moving this java doc to the encode method.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventFilenameInfo.kt
```diff
@@ -0,0 +1,165 @@
+package com.bugsnag.android
+
+import java.io.File
+import java.util.Locale
+import java.util.UUID
+
+/**
+ * Represents important information about an event which is encoded/decoded from a filename.
```
tomlongridge, line 8: This would be a useful place to detail the various things we're now encoding and the rationale behind them.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -219,45 +191,21 @@ String getApiKeyFromFilename(File file) {
...
-            if (apiKey.isEmpty()) {
-                apiKey = config.getApiKey();
-            }
-        }
-        return String.format(Locale.US, "%s%d_%s_%s%s.json",
-                storeDirectory, timestamp, apiKey, uuid, suffix);
+        EventFilenameInfo eventInfo
+                = EventFilenameInfo.Companion.fromEvent(object, null, config);
+        String encodedInfo = eventInfo.encode();
+        return String.format(Locale.US, "%s%s.json", storeDirectory, encodedInfo);
```
tomlongridge, line 202: The `.json` extension is currently outside the part controlled by `EventFilenameInfo` but is used internally by that class as part of the `findApiKeyInFilename` test. Suggest the extension becomes part of the suffix field so it's all controlled by `EventFilenameInfo`.

tomlongridge (commented):
Looks much cleaner 👍  I've made some suggestions from the point-of-view of someone reading this for the first time, to make the abstraction cleaner.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventFilenameInfo.kt
fractalwrench: Added some additional docs here

bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
fractalwrench: Updated so that the extension is set in `EventFilenameInfo`

bugsnag-android-core/src/main/java/com/bugsnag/android/EventPayload.kt
```diff
@@ -30,13 +31,28 @@ class EventPayload : JsonStream.Streamable {
...
+        }
+    }
+
+    private fun getErrorTypesFromStackframes(event: Event): Set<ErrorType> {
+        val errorTypes = event.errors.mapNotNull(Error::getType).toSet()
+        val frameOverrideTypes = event.errors
+            .map { it.stacktrace }
+            .flatMap { it.mapNotNull(Stackframe::type) }
+        return errorTypes.plus(frameOverrideTypes)
+    }
```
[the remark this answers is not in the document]
fractalwrench: Moved to `Event` - this required refactoring the unit test to avoid mocks

bugsnag-android-core/src/main/java/com/bugsnag/android/EventFilenameInfo.kt
```diff
@@ -0,0 +1,165 @@
...
+                }
+                else -> {
+                    NON_JVM_CRASH
+                }
+            }
+        }
+
+        private fun isStartupCrash(durationMs: Long, config: ImmutableConfig): Boolean {
+            return durationMs < config.launchCrashThresholdMs
+        }
```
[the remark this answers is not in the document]
fractalwrench: Agreed to leave as-is for now, as `isStartupCrash` is only used within the `EventFilenameInfo` source file and `shouldNotifyForReleaseStage` is not really a configuration option so doesn't wholly belong in `ImmutableConfig`.

bugsnag-android-core/src/main/java/com/bugsnag/android/EventFilenameInfo.kt
fractalwrench: changed visibility to private and added `config` to `event` constructor

tomlongridge (approved):

PR #1004, opened 2020-11-19
[PLAT-5161] Remove soft fail from Android 10/11

Cawllec:
Removes soft fail from Android 10/11 tests while skipping problematic scenarios.
This should allow us to get some test coverage on these platforms now, and these scenarios can be re-instated once appium issues have been investigated and resolved.


twometresteve (approved):
LGTM pending CI.

PR #1006, opened 2020-11-23
Encode react native promise rejections as a mixed stacktrace

fractalwrench:
## Goal

Alters the `nativeStack` property added in #937 so that it is deserialized as a list of `Stackframe` objects, which are set on an individual `Error`. This forms a mixed stacktrace rather than two separate error objects. The `getPlugin` method has also been updated to allow a raw class, so that both the `NdkPlugin` and `AnrPlugin` can be accessed from Java/Kotlin files as needed. This was originally changed [here](https://github.com/bugsnag/bugsnag-android/commit/239d12a613ef770d1bdfb70324dade17b3a9b8e7#diff-c952c857a72826421f5588f79a4f32113c19fb0860aa4f6571408dc6158d521a) but the solution didn't work for React Native - this covers both scenarios.

## Testing

Updated unit test coverage, and manually tested on React Native to confirm that the nativeStack is set:

Screenshot 2020-11-24 at 11 10 00

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -933,8 +933,9 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
     /**
      * Retrieves an instantiated plugin of the given type, or null if none has been created
      */
+    @SuppressWarnings("rawtypes")
     @Nullable
-    Plugin getPlugin(@NonNull Class<Plugin> clz) {
+    Plugin getPlugin(@NonNull Class clz) {
         Set<Plugin> plugins = pluginClient.getPlugins();
         for (Plugin plugin : plugins) {
             if (plugin.getClass().equals(clz)) {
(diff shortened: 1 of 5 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -933,8 +933,9 @@ Logger getLogger() {
     /**
      * Retrieves an instantiated plugin of the given type, or null if none has been created
      */
+    @SuppressWarnings("rawtypes")
     @Nullable
-    Plugin getPlugin(@NonNull Class<Plugin> clz) {
+    Plugin getPlugin(@NonNull Class clz) {
```
tomlongridge, line 938: Is this required? Worth noting on the PR changeset if so.

tomlongridge (approved):

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: Updated changeset description

PR #1007, opened 2020-11-24
Alter foreground/background scenarios to use activity lifecycle callbacks

fractalwrench:
## Goal

This should help avoid a source of flakiness in the E2E scenarios by making the foreground/background scenarios use [activity lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle) callbacks. These are invoked when the activity is put in the stopped state, which is interpreted by bugsnag as when the application is in the background.

Previously the `CXXDelayedNotifyScenario` used an arbitrary 3 second sleep before calling `bugsnag_notify_env` - this is not always long enough for the activity to be put into the background by the Android OS, leading to test failures.

## Testing

Ran the feature ~5 times on browserstack local.


twometresteve (approved):
LGTM

PR #1008, opened 2020-11-24
[PLAT-5291] Refactor pipeline to use multistage builds

Cawllec:
## Goal

Initial refactor to using dynamic pipelines to allow quick "smoke" tests to run vs "full" tests.

Smoke tests are triggered by default, or if the commit message contains `[quick ci]`
Full tests are triggered if the commit message contains `[full ci]`, or if the branch running is, or is merging into, `master` or `next`

Aiming this into an integration branch as I need to add more to the smoke tests before we have this present generally.

## Design

Most of the common steps are still present in `pipeline.yml`, with 3 main exceptions:
- Maze runner container build step
- Integration test steps
- E2E steps

These are triggered slightly differently depending on whether it's a smoke or full test set:
- The maze runner container has the `full_tests` or `smoke_tests` feature set added depending on the pipeline.
- Only NDK r16b integration test steps are run on the smoke tests
- The E2E maze-runner steps are run based on the maze-runner container previously built.  In the future we may decide which Android versions or Android/NDK combinations to target with smoke vs full test sets.

## Testing

[An example of a smoke test run](https://buildkite.com/bugsnag/bugsnag-android/builds/2668)
[An example of a full test run](https://buildkite.com/bugsnag/bugsnag-android/builds/2670#_)


dockerfiles/Dockerfile.android-maze-runner
```diff
@@ -1,3 +1,8 @@
 FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:latest-v3-cli as android-maze-runner
 
-COPY features features
+COPY features/scripts features/scripts
+COPY features/steps features/steps
+COPY features/support features/support
+
+ARG TESTS_DIR=features/full_tests
+COPY $TESTS_DIR $TESTS_DIR
```
fractalwrench, line 8: Is it worth just always copying all the tests, or would there be some side effect to that which is undesirable?

It would definitely be good to run the `_smoke_test.feature` on the full CI run either way - as that should get quick feedback about whether a feature is successful when running against `next/master`.

.buildkite/full_pipeline_trigger.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
+if [[ "$BUILDKITE_MESSAGE" == *"[QUICK CI]"* ]]; then
```
fractalwrench, line 3: Do we want this to be possible? It would allow developers to skip the full checks on next/master.

.buildkite/pipeline.smoke.yml
```diff
@@ -0,0 +1,239 @@
...
+  - label: ':docker: Build Android maze runner image'
+    key: "maze-runner"
+    timeout_in_minutes: 20
+    plugins:
+      - docker-compose#v3.3.0:
+          build: android-maze-runner
+          args:
+            - TESTS_DIR=features/smoke_tests
+
+  - label: ':android: NDK 16b SDK 4.4 Instrumentation tests'
```
fractalwrench, line 11: Running the instrumentation tests on just 16b + 4.4 would be fine for a quick CI run - there's no advantage from my perspective in running these tests 4x.

.buildkite/pipeline.full.yml
```diff
@@ -0,0 +1,341 @@
...
+    plugins:
+      - docker-compose#v3.3.0:
+          run: android-ci
+    env:
+      INSTRUMENTATION_DEVICES: '["Google Nexus 5-4.4"]'
+      NDK_VERSION: r16b
+    concurrency: 10
+    concurrency_group: 'browserstack-app'
+
+  - label: ':android: NDK 16b SDK 4.4 Instrumentation tests'
```
fractalwrench, line 61: An existing issue, but this appears to be a dupe job

.buildkite/pipeline.full.yml
```diff
@@ -0,0 +1,341 @@
+steps:
+  - label: ':docker: Build Android maze runner image'
+    key: "maze-runner"
+    timeout_in_minutes: 20
+    plugins:
+      - docker-compose#v3.3.0:
+          build: android-maze-runner
+
+  - label: ':android: NDK 12b SDK 4.4 Instrumentation tests'
```
fractalwrench, line 9: Would it be possible to tweak the ordering of jobs so that one mazerunner job runs after the first instrumentation test? That might help speed up feedback when Browserstack is saturated

fractalwrench (commented):
Looks like this will be fast :)

This looks like the right approach. I had a few queries which are mostly around what the process for quick/full CI runs should be - I've left these as inline comments.

twometresteve:
Please always name integration branches with a `integration/` prefix.  It's then clear to all of its purpose and it also allows us to use a single branch protection rule to cover all integration branches (by setting it on "integration/*".

dockerfiles/Dockerfile.android-maze-runner
Cawllec: We could have a single maze-runner container and leave the individual config to the step in the pipeline, I'm not opposed to that

.buildkite/full_pipeline_trigger.sh
Cawllec: Possibly not, just added it as a utility.

.buildkite/full_pipeline_trigger.sh
```diff
@@ -0,0 +1,14 @@
...
+  PIPELINE='pipeline.smoke.yml'
+elif [[ "$BUILDKITE_MESSAGE" == *"[FULL CI]"* ||
+  "$BUILDKITE_BRANCH" == "next" ||
+  "$BUILDKITE_BRANCH" == "master" ]]; then
```
twometresteve, line 7: We'll also want full ci if it's a PR into `next` or `master`.  I think you can use `BUILDKITE_PULL_REQUEST_BASE_BRANCH` for that.

.buildkite/pipeline.full.yml
```diff
@@ -0,0 +1,341 @@
...
+        - "--username=$BROWSER_STACK_USERNAME"
+        - "--access-key=$BROWSER_STACK_ACCESS_KEY"
+        - "--fail-fast"
+    concurrency: 10
+    concurrency_group: 'browserstack-app'
+    #
+    # TODO: Remove as soon as possible - working around BrowserStack network problem
+    #
+    soft_fail:
+      - exit_status: "*"
```
twometresteve, line 224: These should come out for Android 6 and 7 now.

dockerfiles/Dockerfile.android-maze-runner
```diff
@@ -1,3 +1,8 @@
 FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:latest-v3-cli as android-maze-runner
 
-COPY features features
+COPY features/scripts features/scripts
+COPY features/steps features/steps
+COPY features/support features/support
+
+ARG TESTS_DIR=features/full_tests
```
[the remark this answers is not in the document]
Cawllec: Ok, we can adjust to that when necessary

dockerfiles/Dockerfile.android-maze-runner
twometresteve: In the Cocoa pipeline we consume the Maze Runner image directly in `docker-compose.yml` and mount `features/` as a volume.  There is no build step in the pipeline at all - would that be an option here?  Although it takes next to no time to build the image, there is instance wait time and general use to consider.  We could have an annoying wait time to get a fresh agent, just for it to then sit idle for 30 minutes before being surrendered.

dockerfiles/Dockerfile.android-maze-runner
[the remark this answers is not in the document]
twometresteve: Not a strong preference by any means, but I like the one-word descriptions and matching the feature directory name to the commit message term would seem to make sense.  If I were to choose I'd probably say full, quick and minimal here.

.buildkite/full_pipeline_trigger.sh
twometresteve: Hmm, slightly in two minds here as Github status checks should cover us.  I'm not sure how much we'd actually want to use it and if there is a compelling case for it.

.buildkite/full_pipeline_trigger.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
+
+if [[ "$BUILDKITE_MESSAGE" == *"[QUICK CI]"* ]]; then
+  PIPELINE='pipeline.smoke.yml'
+elif [[ "$BUILDKITE_MESSAGE" == *"[FULL CI]"* ||
```
twometresteve, line 5: Could we make the commit string lower case, i.e. `[full ci]`?  It would then be consistent with Buidlkite's `[skip ci]` construct and I think a little easier to type.

twometresteve: ...or do a case-insensitive string comparison, of course.

dockerfiles/Dockerfile.android-maze-runner
Cawllec: Yes, that would make a lot more sense

.buildkite/full_pipeline_trigger.sh
Cawllec: Just thought we have cases where a changelog gets changed or typo gets fixed before a release and we often have to wait for a full CI run for that minor change.  Would it be acceptable to use quick-ci in these cases?

fractalwrench (approved):

.buildkite/full_pipeline_trigger.sh
twometresteve: With a new commit the Github status checks will be reset, so we'd need another full run _at some point_ before being able to merge - but perhaps I can see the use...  If there are a few review comments to action (including code changes) and the author wants to address them in separate commits with fast-feedback, it could be useful to be able to push them with `[quick ci]` in the knowledge that a full run will be needed before being able to actually merge (which can then be achieved by simply amending the last commit message).  Let's leave it in for now and see how it pans out.

twometresteve (approved):
LGTM for this stage of the work.

PR #1009, opened 2020-11-25
Update example fixture to support multi process sample code

fractalwrench:
## Goal

Updates the example fixture to support multi process sample code. As part of this changeset the example app has also been regenerated via Android Studio, so that it uses a more conventional project structure.

## Changeset

- Regenerated example app via Android Studio project creation wizard so that it uses a more conventional structure
- Changed package from `com.bugsnag.android.example` to `com.example.bugsnag.android`, as this means proguard runs on the example project normally and it isn't affected by pipeline grouping rules due to the use of `com.bugsnag`
- Created `BaseCrashyActivity`, which contains all the code for triggering a crash and displays UI.
- Added `MultiProcessActivity`. This can be started from `ExampleActivity` and runs in a separate process, and is capable of triggering all the same crashes.
- Configured the application to use a custom `persistenceDirectory` for each process (this needs to be commented out until the project ships, or it will break compilation of the example fixture)

**Note**: The `Application` subclass will run each time a multi process app launches a new process. This behaviour might affect how we want to document the `persistenceDirectory` should be set in multi process apps, particularly as getting the process name seems to be feasible and unique.

## Testing

Ran app manually and confirmed all errors are reported appropriately.


examples/sdk-app-example/app/src/main/java/com/example/bugsnag/android/ExampleApplication.kt
```diff
@@ -22,12 +22,22 @@ class ExampleApplication : Application() {
 
     private external fun performNativeBugsnagSetup()
 
+    private val multiProcessApp = true
```
tomlongridge, line 25: I don't think we really need this, but happy to tidy this up nearer the end of the project when we uncomment the `persistenceDirectory`.

tomlongridge (approved):

PR #1010, opened 2020-11-26
Create synchronized store for user information

fractalwrench:
## Goal

Creates a synchronized store which is used for reading & writing user information. This replaces `SharedPreferences` which is not safe for use in a multi process environment, and does not give control over the location where information is persisted.

## Changeset

- Created `SynchronizedStreamableStore`, which is capable of reading and writing `Streamable` objects
- Synchronized access to the `persist()` and `load()` methods with a `ReentrantReadWriteLock` so that concurrent reads of the information is possible when the file is not being written
- Added method to allow deserialization of `User`, along with `JsonReadable` interface

## Testing

Created instrumentation tests which verify that data can be persisted and loaded, and that the read lock allows concurrent access.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/JsonReadable.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/JsonReadable.kt
@@ -1,0 +1,14 @@
+package com.bugsnag.android
+
+import android.util.JsonReader
+
+/**
+ * Classes which implement this interface are capable of deserializing a JSON input.
+ */
+internal interface JsonReadable<T : JsonStream.Streamable> {
+
+    /**
+     * Constructs an object from a JSON input.
+     */
+    fun fromReader(reader: JsonReader): T
+}
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/SynchronizedStreamableStore.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/SynchronizedStreamableStore.kt
@@ -1,0 +1,33 @@
+package com.bugsnag.android
+
+import android.util.JsonReader
+import java.io.File
+import java.io.IOException
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.concurrent.withLock
+
+internal class SynchronizedStreamableStore<T : JsonStream.Streamable>(
+    private val file: File
+) {
+
+    private val lock = ReentrantReadWriteLock()
+
+    @Throws(IOException::class)
+    fun persist(streamable: T) {
+        lock.writeLock().withLock {
+            file.writer().use {
+                streamable.toStream(JsonStream(it))
+                true
+            }
+        }
+    }
+
+    @Throws(IOException::class)
+    fun load(loadCallback: (JsonReader) -> T): T {
+        lock.readLock().withLock {
+            return file.reader().use {
+                loadCallback(JsonReader(it))
+            }
+        }
+    }
+}
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/User.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/User.kt
@@ -1,5 +1,6 @@
 package com.bugsnag.android
 
+import android.util.JsonReader
 import java.io.IOException
 
 /**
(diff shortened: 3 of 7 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/SynchronizedStreamableStore.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    fun load(logger: Logger): T? {
+        lock.readLock().withLock {
+            return try {
+                val user = file.reader().use {
+                    User.fromReader(JsonReader(it))
+                }
+                @Suppress("UNCHECKED_CAST")
+                return user as T
```
tomlongridge, line 36: This rather breaks the use of generics as it only reads `User` types, regardless of the type used to instantiate the store. I assume this is done with generics to provide a way to reuse for device ID and any future store - if so how would you implement a second type here?

bugsnag-android-core/src/main/java/com/bugsnag/android/SynchronizedStreamableStore.kt
```diff
@@ -0,0 +1,44 @@
...
+            } catch (exc: Throwable) {
```
tomlongridge, line 37: This seems too generic for a catch statement considering we control the implementations - it seems reasonable for the eventual caller to catch an `IOException` and handle it, rather than handling a null response.

Same for the `persist` catch - I think this can be passed up for the caller to handle.

bugsnag-android-core/src/main/java/com/bugsnag/android/SynchronizedStreamableStore.kt
fractalwrench: Yep - the plan is to reuse this class for device ID and future stores if required. I've extracted the `User.fromReader` call into a callback parameter that uses generics - this avoids the cast and supports the use of generics.

bugsnag-android-core/src/main/java/com/bugsnag/android/SynchronizedStreamableStore.kt
fractalwrench: Updated to throw the exception and changed test assertions accordingly.

bugsnag-android-core/src/main/java/com/bugsnag/android/JsonReadable.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Classes which implement this interface are capable of deserializing a JSON input.
+ */
+internal interface JsonReadable<T : JsonStream.Streamable> {
+
+    /**
+     * Constructs an object from a JSON input.
+     */
+    fun fromReader(reader: JsonReader): T?
```
tomlongridge, line 13: This can now be non-optional I believe?

bugsnag-android-core/src/main/java/com/bugsnag/android/User.kt
```diff
@@ -25,12 +26,41 @@ class User @JvmOverloads internal constructor(
...
+                    val key = nextName()
+                    val value = nextString()
+                    when (key) {
+                        KEY_ID -> id = value
+                        KEY_EMAIL -> email = value
+                        KEY_NAME -> name = value
+                    }
+                    user = User(id, email, name)
+                }
+                endObject()
```
tomlongridge, line 58: This constructs a new User three times - any reason it cannot be done after the end of the object and just returned?

```suggestion
                while (hasNext()) {
                    val key = nextName()
                    val value = nextString()
                    when (key) {
                        KEY_ID -> id = value
                        KEY_EMAIL -> email = value
                        KEY_NAME -> name = value
                    }
                }
                endObject()
                return User(id, email, name)
```

fractalwrench: Nice catch - the intention was to create the object only once.

bugsnag-android-core/src/main/java/com/bugsnag/android/JsonReadable.kt
fractalwrench: Updated to make this non-optional

tomlongridge (approved):

PR #1013, opened 2020-11-27
Use consistent device ID for multi process apps

fractalwrench:
## Goal

This changeset makes the notifier use a consistent device ID for apps which initialize Bugsnag in multiple processes. This is achieved by storing the device ID in a single file location which can be accessed concurrently from each process and is protected by a [FileLock](https://developer.android.com/reference/java/nio/channels/FileLock).

## Changeset

- Created `DeviceIdStore` which is responsible for loading/saving the device ID to the file system
- Guarded the device ID file with a `FileLock` to prevent concurrent access in multiple processes, and a `ReadWriteLock` to prevent concurrent access from multiple threads.
- Updated the `Client` to persist a device ID to a file and to migrate any existing device ID from `SharedPreferences` or generating a random UUID if none already exists
- Created `SharedPrefMigrator` which reads legacy information from `SharedPreferences`. A separate PR will build on this by making this class read user information too, and delete the `SharedPreferences` file once the migration is complete
- Removed responsibility for storing the device ID from `UserRepository`
- Updated `SynchronizedStreamableStore` to use buffered IO

## Testing

Added unit test coverage for new classes, and to verify that a file lock is used. Relied on existing E2E test coverage for regression testing, and also manually tested to verify that the device ID is stored in the new location.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -18,8 +18,10 @@ import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 
 import kotlin.Unit;
+import kotlin.jvm.functions.Function0;
 import kotlin.jvm.functions.Function2;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
@@ -28,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Observer;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.RejectedExecutionException;
 
 /**
@@ -152,29 +155,23 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
                 sessionStore, logger);
         metadataState = copyMetadataState(configuration);
 
-        // Set up and collect constant app and device diagnostics
-        sharedPrefs = appContext.getSharedPreferences(SHARED_PREF_KEY, Context.MODE_PRIVATE);
-
         ActivityManager am =
                 (ActivityManager) appContext.getSystemService(Context.ACTIVITY_SERVICE);
 
         appDataCollector = new AppDataCollector(appContext, appContext.getPackageManager(),
                 immutableConfig, sessionTracker, am, logger);
 
-        UserRepository userRepository = new UserRepository(sharedPrefs,
-                immutableConfig.getPersistUser());
-        userState = new UserState(userRepository);
-        User user = configuration.getUser();
+        sharedPrefs = appContext.getSharedPreferences(SHARED_PREF_KEY, Context.MODE_PRIVATE);
 
-        if (user.getId() != null || user.getEmail() != null || user.getName() != null) {
-            userState.setUser(user.getId(), user.getEmail(), user.getName());
-        }
+        // load the device + user information
+        DeviceIdStore deviceIdStore = new DeviceIdStore(appContext, logger);
+        String deviceId = deviceIdStore.loadDeviceId();
+        userState = loadUserState(configuration, deviceId);
 
         DeviceBuildInfo info = DeviceBuildInfo.Companion.defaultInfo();
         Resources resources = appContext.getResources();
-        String id = userRepository.getDeviceId();
-        deviceDataCollector = new DeviceDataCollector(connectivity, appContext, resources, id, info,
-                Environment.getDataDirectory(), logger);
+        deviceDataCollector = new DeviceDataCollector(connectivity, appContext,
+                resources, deviceId, info, Environment.getDataDirectory(), logger);
 
 
         if (appContext instanceof Application) {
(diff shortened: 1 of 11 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceIdStore.kt
```diff
@@ -0,0 +1,179 @@
...
+            null
+        }
+    }
+
+    private fun persistNewDeviceIdWithLock(
+        channel: FileChannel,
+        logger: Logger,
+        uuidProvider: () -> UUID
+    ): String? {
+        val lock = waitForFileLock(channel) ?: return null
```
tomlongridge, line 108: What lock is the `synchronizedStreamableStore` providing on top of this? I thought that was the purpose of the generic store?

If it's because we need to check the device ID once inside the lock, perhaps the `synchronizedStreamableStore` could be extended to allow this lock to be taken by a client ahead of persisting?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -253,6 +251,18 @@ public void run() {
         leaveAutoBreadcrumb("Bugsnag loaded", BreadcrumbType.STATE, data);
     }
 
+    private UserState loadUserState(@NonNull Configuration configuration, String deviceId) {
+        boolean persistUser = immutableConfig.getPersistUser();
+        UserRepository userRepository = new UserRepository(sharedPrefs, persistUser, deviceId);
```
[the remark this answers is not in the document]
fractalwrench: Yep, that's the plan 👍

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceIdStore.kt
```diff
@@ -0,0 +1,179 @@
...
+     * be used. If no value is present then a random UUID will be generated and persisted.
+     */
+    fun loadDeviceId(context: Context): String? {
+        val sharedPrefMigrator = SharedPrefMigrator(context)
+        val legacyDeviceId = sharedPrefMigrator.loadDeviceId()
+
+        return loadDeviceId(logger) {
+            when (legacyDeviceId) {
+                null -> UUID.randomUUID()
+                else -> UUID.fromString(legacyDeviceId)
```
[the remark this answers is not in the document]
fractalwrench: It feels preferable to return a `UUID` here as it enforces that the device ID must be a valid UUID rather than just some string value.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceIdStore.kt
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+    }
+
+    internal fun loadDeviceId(logger: Logger, uuidProvider: () -> UUID): String? {
```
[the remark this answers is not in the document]
fractalwrench: Updated to use `logger` property instead of passing in as a parameter

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -152,29 +155,24 @@ public Unit invoke(Boolean hasConnection, String networkState) {
...
-        userState = new UserState(userRepository);
-        User user = configuration.getUser();
+        sharedPrefs = appContext.getSharedPreferences(SHARED_PREF_KEY, Context.MODE_PRIVATE);
 
-        if (user.getId() != null || user.getEmail() != null || user.getName() != null) {
-            userState.setUser(user.getId(), user.getEmail(), user.getName());
-        }
+        // load the device + user information
+        File deviceIdFile = new File(appContext.getFilesDir(), "device-id");
+        DeviceIdStore deviceIdStore = new DeviceIdStore(deviceIdFile, logger);
```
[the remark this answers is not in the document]
fractalwrench: Updated constructor to use this file by default.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceIdStore.kt
fractalwrench: There are two mechanisms in use: a [ReentrantReadWriteLock](https://developer.android.com/reference/java/util/concurrent/locks/ReentrantReadWriteLock) which is used by `SynchronizedStreamableStore` to synchronize across multiple threads, and a [FileLock](https://developer.android.com/reference/java/nio/channels/FileLock) which is used by `DeviceIdStore` to synchronize across multiple processes.

`SynchronizedStreamableStore` is effectively the replacement for `SharedPreferences`. When saving/loading the user information it isn't necessary to be multi process safe as each process will store this information in a separate file. Therefore as the device ID file is the only one that can be accessed across multiple processes, I've used the file locking in the `DeviceIdStore` only.

tomlongridge: OK - yes, I see. Noted that you've said this in the PR description, but can you add the above explanation to the method or class comments so it's obvious for future readers.

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceIdStore.kt
```diff
@@ -0,0 +1,179 @@
...
+            // optimistically read device ID without a lock - the majority of the time
+            // the device ID will already be present so no synchronization is required.
+            val deviceId = loadDeviceIdInternal()
+
+            if (deviceId?.id != null) {
+                deviceId.id
+            } else {
+                return persistNewDeviceUuid(uuidProvider)
+            }
+        } catch (exc: Throwable) {
```
tomlongridge, line 66: Is there good cause to catch `Throwable` here? I would be keen to avoid this and catch more specific errors as we don't catch throwables elsewhere in the codebase. (Also the catch in `persistNewDeviceUuid` is dong the same an so seem redundant.)

bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceIdStore.kt
fractalwrench: Updated docs to include some more information on this

tomlongridge (approved):

PR #1014, opened 2020-11-27
Integration branch for using multistage builds

twometresteve:


Cawllec:
Looks like this is running the [full pipeline](https://buildkite.com/bugsnag/bugsnag-android/builds/2760#_) due to the merge into `next`

PR #1015, opened 2020-11-27
Skip failing tests and prepare for MazeRunner 3.6.0

twometresteve:
## Goal

Skip understood failing tests pending ticketed fix and prepare for MazeRunner 3.6.0.

## Design

I am aware that this creates a minor conflict on one of the integration branches, but there are more scenarios across more Android versions that will start failing once MazeRunner 3.6.0 is released.  The fixes coming in MR 3.6.0 that cause this are:
- Waiting for a period of time before checking for no requests in the `I should receive no requests` step
- Detection of received but invalid requests (e.g. invalid JSON or missing body).

## Changeset

- Skip and annotate tests that will fail in the next MR release
- Disable enforcement of the Bugsnag-Integrity header, until the functionality is released.  MazeRunner 3.6.0 will fail tests if the header is not present, unless configured otherwise.
- Fix the `I clear any error dialogue` step to ignore exceptions raised by clicking on elements that disappear just after we check they are present.

## Testing

As part of developing this branch I changes the Dockerfile to use the `master-cli` version of MazeRunner.  This allowed me to flush out all of the failures that would otherwise hit us, before changing back to `latest-v3-cli`.

I'nm a bit concerned about the length of time that the Android 4.4 and 6 end-to-end tests are taking, but this does seem to be a general slow-down on those devices and not related to this change.  I suggest we monitor for now and seek support if necessary.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -2,8 +2,9 @@
 # Set this explicitly
 $api_key = "a35a2a72bd230ac0aa0f52715bbdc6aa"
 
-# TODO: Remove once the Bugsnag-Integrity header has been implemented
-AfterConfiguration do |config|
+AfterConfiguration do |_config|
+  MazeRunner.config.receive_no_requests_wait = 10 if MazeRunner.config.respond_to? :receive_no_requests_wait=
+  # TODO: Remove once the Bugsnag-Integrity header has been implemented
   MazeRunner.config.enforce_bugsnag_integrity = false if MazeRunner.config.respond_to? :enforce_bugsnag_integrity=
 end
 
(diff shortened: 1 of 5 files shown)


features/support/env.rb
```diff
@@ -2,6 +2,16 @@
 # Set this explicitly
 $api_key = "a35a2a72bd230ac0aa0f52715bbdc6aa"
 
+AfterConfiguration do |_config|
+  MazeRunner.config.receive_no_requests_wait = 10 if MazeRunner.config.respond_to? :receive_no_requests_wait=
+  # TODO: Remove once the Bugsnag-Integrity header has been implemented
```
fractalwrench, line 7: It'd be good to create a ticket to track this when this functionality is merged, so that this doesn't get left in.

fractalwrench (approved):

features/support/env.rb
twometresteve: PLAT-5510.

PR #1016, opened 2020-11-29
Multistage smoke tests

Cawllec:
## Goal

Update smoke tests to be as comprehensive as possible with as few test-runs as possible.

Validates in brief:
- Default config
- Overwritten config
- Handled Java/Kotlin/C++ errors
- Unhandled Java errors
- Signals
- Android not responding crashes
- Automatic/Manual app/device data
- Automatic/Manual sessions
- Breadcrumbs


features/fixtures/mazerunner/src/main/cpp/bugsnags.cpp
```diff
@@ -372,6 +372,24 @@ bool add_java_data(void *event_ptr) {
   return true;
 }
 
+bool add_smoke_data(void *event_ptr) {
+    bugsnag_event_add_metadata_string(event_ptr,
```
Cawllec, line 376: As mentioned below, this call doesn't appear to be adding the corresponding metadata in the `CXXNotifySmokeScenario`.

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the event "user.name" equals "Jack Mill"
+
+    # Breadcrumbs
+    And the event has a "log" breadcrumb named "Cold beans detected"
+
+    # MetaData
+    And the event "metaData.TestData.Source" equals "ClientCallback"
+
+    # Potential issue with missing data here
+    #And the event "metaData.TestData.C++" equals "present"
```
Cawllec, line 164: Note - This is commented out due to the above note - this data doesn't appear to be being added to the payload successfully.

Cawllec:
Note - at last look this had an issue running against Android 4 - The unhandled test was failing due to "seeing" the app-crashed button but it not being available when attempting to press it.  @twometresteve, have you seen this failure before?

fractalwrench:
>The unhandled test was failing due to "seeing" the app-crashed button but it not being available when attempting to press it.


Sounds like this part may be addressed by the changes in #1015?

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the event "device.manufacturer" is not null
+    And the event "device.model" is not null
+    And the event "device.osName" equals "android"
+    And the event "device.osVersion" is not null
+    And the event "device.runtimeVersions" is not null
+    And the event "device.runtimeVersions.androidApiLevel" is not null
+    And the event "device.runtimeVersions.osBuild" is not null
+    And the payload field "events.0.device.totalMemory" is greater than 0
+    And the payload field "events.0.device.freeDisk" is greater than 0
+    And the payload field "events.0.device.freeMemory" is greater than 0
```
fractalwrench, line 47: Is there any overall preference for `the event` vs `the payload field` as mazerunner steps? This probably doesn't require immediate action, but it'd be nice to get the naming consistent

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the event "metaData.device.batteryLevel" is not null
+
+    # User
+    And the event "user.id" is not null
+    And the payload field "events.0.user.id" equals the stored value "device_id"
+
+    # Breadcrumbs
+    And the event has a "state" breadcrumb named "Bugsnag loaded"
+    And the event has a "manual" breadcrumb named "HandledJavaSmokeScenario"
+    And the event "breadcrumbs.2.metaData.source" equals "BreadcrumbCallback"
```
fractalwrench, line 67: This will probably cause flakes because automatic breadcrumbs can be captured at any time, meaning there's no guarantee this breadcrumb will always be the element at index 2. The `event has a {string} breadcrumb named` step has an approach for this of iterating through to find a matching breadcrumb. An alternative would be to set a breadcrumb callback and discard anything we don't want.

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/HandledKotlinSmokeScenario.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.bugsnag.android.*
+import java.util.*
+
+/**
+ * Sends a handled exception to Bugsnag, which does not include session data.
+ */
+internal class HandledKotlinSmokeScenario(config: Configuration,
+                                        context: Context) : Scenario(config, context) {
+    init {
+        config.autoTrackSessions = false
```
fractalwrench, line 13: Kotlin has a pretty cool `with` statement where anything in the scope wouldn't have to call `config.`: https://kotlinlang.org/docs/reference/scope-functions.html#with

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the event "metaData.TestData.CallbackMetadata" is true
+
+Scenario: Notify Kotlin exception with overwritten configuration
+    When I run "HandledKotlinSmokeScenario"
+    Then I wait to receive a request
+    And the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+
+    # Exception details
+    And the payload field "events" is an array with 1 elements
+    And the exception "errorClass" equals "java.lang.RuntimeException"
```
fractalwrench, line 80: Could the scenario be changed to something like `IllegalStateException`? This would add extra assurance that Bugsnag isn't wrapping the captured exception with its own `Throwable` and that the original type is serialized correctly.

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the event "metaData.TestData.Source" equals "HandledKotlinSmokeScenario"
+    And the event "metaData.TestData.Callback" is true
+    And the event "metaData.TestData.redacted" equals "[REDACTED]"
+
+Scenario: Handled C functionality
+    When I run "CXXNotifySmokeScenario"
+    And I wait to receive a request
+
+    # Exception details
+    Then the request payload contains a completed handled native report
```
fractalwrench, line 113: It doesn't look like this step validates the headers etc to the same extent as `the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier`. It'd probably be worth using that here instead and raising a ticket to address the existing native scenarios which use this step, unless there's a reason this step is preferred.

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    When I run "HandledJavaSmokeScenario"
+    And I wait to receive a request
+    Then the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+
+    # Exception details
+    And the payload field "events" is an array with 1 elements
+    And the exception "errorClass" equals "java.lang.RuntimeException"
+    And the exception "message" equals "HandledJavaSmokeScenario"
+    And the exception "type" equals "android"
+    And the payload field "events.0.exceptions.0.stacktrace" is a non-empty array
```
fractalwrench, line 13: It'd be nice to have some more validation of the stacktrace for both JVM/native reports, as both are serialized quite differently. Stacktraces tend to be quite brittle and change when scenario code alters, so it would be fine for now just to check that fields aren't null..

twometresteve:
> > The unhandled test was failing due to "seeing" the app-crashed button but it not being available when attempting to press it.
> 
> Sounds like this part may be addressed by the changes in #1015?

Yeah, it should be.

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/AutoSessionSmokeScenario.kt
```diff
@@ -0,0 +1,27 @@
...
+internal class AutoSessionSmokeScenario(config: Configuration,
+                                        context: Context) : Scenario(config, context) {
+    override fun run() {
+        super.run()
+        config.autoTrackSessions = true
+        Bugsnag.start(context, config)
+        context.startActivity(Intent("com.bugsnag.android.mazerunner.UPDATE_CONTEXT"))
+
+        val main = Handler(Looper.getMainLooper())
+        main.postDelayed(Runnable {
```
fractalwrench, line 22: This feels like it could introduce a source of flakes - the Android system will need to launch the new activity in < 1s. Using the [activity lifecycle callbacks](https://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks#onActivityCreated(android.app.Activity,%20android.os.Bundle)) instead to detect when the new activity has launched would make this predictable and avoid the arbitrary wait - a couple of other scenarios have examples of how to achieve this.

features/smoke_tests/sessions.feature
```diff
@@ -0,0 +1,101 @@
...
+    Then the request is valid for the session reporting API version "1.0" for the "Android Bugsnag Notifier" notifier
+    And the payload field "notifier.name" equals "Android Bugsnag Notifier"
+    And the payload field "sessions" is an array with 1 elements
+    And the session "user.id" is not null
+    And the payload field "sessions.0.user.id" is stored as the value "automated_user_id"
+    And the session "id" is not null
+    And the payload field "sessions.0.id" is stored as the value "automated_session_id"
+    And the session "startedAt" is not null
+
+    # App data
```
fractalwrench, line 17: `binaryArch` is also serialized for sessions

features/smoke_tests/sessions.feature
```diff
@@ -0,0 +1,101 @@
...
+    And the payload field "device.cpuAbi" is a non-empty array
+    And the payload field "device.jailbroken" is false
+    And the payload field "device.id" is not null
+    And the payload field "device.id" equals the stored value "automated_user_id"
+    And the payload field "device.locale" is not null
+    And the payload field "device.manufacturer" is not null
+    And the payload field "device.model" is not null
+    And the payload field "device.osName" equals "android"
+    And the payload field "device.osVersion" is not null
+    And the payload field "device.runtimeVersions" is not null
```
fractalwrench, line 35: `totalMemory` is also serialized for sessions

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/ManualSessionSmokeScenario.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    override fun run() {
+        super.run()
+        val client = Bugsnag.getClient()
+        client.setUser("123", "ABC.CBA.CA", "ManualSessionSmokeScenario")
+        val thread = HandlerThread("HandlerThread")
+        thread.start()
+
+        Handler(thread.looper).post(Runnable {
```
fractalwrench, line 32: The sleeps are only necessary as we don't know when reports are delivered. It should be possible to get this information by implementing a custom delivery class, and then chaining requests one after the other. Something along these lines should do the trick:

```
val baseDelivery = config.delivery
config.delivery = InterceptingDelivery(baseDelivery)

class InterceptingDelivery(baseDelivery: Delivery, callback: () -> Unit): Delivery {
    override fun deliver() {
        baseDelivery.deliver()
        callback()
    } 
}
```

fractalwrench (commented):
Looking nice and quick on CI :)

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the event "severityReason.type" equals "handledException"
+
+    # App data
+    And the event "app.buildUUID" is not null
+    And the event "app.id" equals "com.bugsnag.android.mazerunner"
+    And the event "app.releaseStage" equals "production"
+    And the event "app.type" equals "android"
+    And the event "app.version" equals "1.1.14"
+    And the event "app.versionCode" equals 34
+    And the event "app.duration" is not null
```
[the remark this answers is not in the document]
Cawllec: Yep, will spin this out as a separate ticket as I think we'll need additional steps for type checking in the "event" (combined with the below)

features/smoke_tests/handled.feature
Cawllec: As with implementing type checks this will be spun out into a separate ticket.

features/smoke_tests/sessions.feature
Cawllec: I'm not seeing that present on the session payloads currently.

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/HandledKotlinSmokeScenario.kt
Cawllec: I'll add a ticket to refactor the Android test fixture, I'm sure there's a bunch of things in there which could be changed

features/smoke_tests/handled.feature
```diff
@@ -0,0 +1,164 @@
+Feature: Handled smoke tests
+
+Scenario: Notify caught Java exception with default configuration
+    When I run "HandledJavaSmokeScenario"
+    And I wait to receive a request
+    Then the request is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+
+    # Exception details
```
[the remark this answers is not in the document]
Cawllec: `context` doesn't appear to be present by default.  I've added some more checking around threads.

features/smoke_tests/sessions.feature
fractalwrench: Actually - I think this is only set for NDK errors IIRC.

features/smoke_tests/handled.feature
twometresteve: What was the outcome here?

features/smoke_tests/handled.feature
twometresteve: What was the outcome here?

Cawllec: This was addressed as suggested

features/smoke_tests/handled.feature
Cawllec: This was addressed as suggested

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/ManualSessionSmokeScenario.kt
```diff
@@ -0,0 +1,65 @@
...
+ * Sends an exception after pausing the session
+ */
+internal class ManualSessionSmokeScenario(config: Configuration,
+                                          context: Context) : Scenario(config, context) {
+
+    init {
+        config.autoTrackSessions = false
+        val baseDelivery = createDefaultDelivery()
+
+        class InterceptingDelivery(private val baseDelivery: Delivery,
```
fractalwrench, line 22: It'd be good if this could be a top-level class that just invoked a callback with zero params, and if the `state` variable was stored in the scenario. This would allow the class to be reused by other scenarios - which should help reduce flakes.

fractalwrench (approved):
LGTM pending CI

twometresteve (commented):
I'm happy if @fractalwrench is happy.

features/fixtures/mazerunner/src/main/java/com/bugsnag/android/mazerunner/scenarios/ManualSessionSmokeScenario.kt
Cawllec: Moved this out into a separate class, seems to work as expected.

Cawllec:
Assuming CI passes, will merge shortly.

PR #1017, opened 2020-12-01
Store user information in persistenceDirectory

fractalwrench:
## Goal

Stores user information in the `persistenceDirectory`. If legacy user information is set in `SharedPreferences`, it is migrated from there.

## Changeset

- Loaded device/user info in `Client` and then deleted legacy `SharedPreferences` file
- Updated `SharedPrefMigrator` to read user information, check whether a file exists on disk, and to provide a mechanism for deleting the legacy file
- Removed redundant `UserRepository` class
- Created `UserStore` class which loads/saves the user, and migrates the legacy value from `SharedPreferences`
- Refactored `UserState` so that it only emits an observable when its value is set, and is not responsible for loading the user from disk. The `UserStore` registers an observer on `UserState` which persists the changed user value if required
- Optimized `DeviceIdStore` so that it only calls `loadDeviceId` on the first launch, rather than each time
- Added `user` property to `ImmutableConfig` which mirrors `Configuration#user`

## Testing

- Extracted test cases from `ClientTest` into `ClientUserTest`
- Added unit and instrumentation tests that verify the user information is stored and loaded correctly


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/SynchronizedStreamableStoreTest.kt
```diff
@@ -133,9 +134,9 @@ internal class ThreadTestStreamable(
 }
 
 internal class CrashyStreamable : JsonStream.Streamable {
-    override fun toStream(stream: JsonStream) = TODO("I'll handle this later...")
+    override fun toStream(stream: JsonStream) = throw IllegalStateException()
```
fractalwrench, line 137: This prevents an unnecessary TODO from showing up in the Android Studio UI

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
```diff
@@ -30,7 +28,8 @@ internal data class ImmutableConfig(
     val maxBreadcrumbs: Int,
     val maxPersistedEvents: Int,
     val maxPersistedSessions: Int,
-    val persistenceDirectory: File
+    val persistenceDirectory: File,
+    val user: User
```
tomlongridge, line 32: Adding `user` to `ImmutableConfig` seems to break the pattern that the class was created for as the user can be changed in the client later, so is not one of the fields considered immutable configuration.

I can see it provides a useful mechanism to pass into the UserStore, but it seems like a bit of a compromise in the design to me. Can we pass a `User` into `UserStore.load` that represents the active user in the client?

bugsnag-android-core/src/main/java/com/bugsnag/android/UserStore.kt
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * This class is responsible for persisting and retrieving user information.
+ */
+internal class UserStore @JvmOverloads constructor(
+    private val config: ImmutableConfig,
+    private val deviceId: String?,
+    file: File = File(config.persistenceDirectory, "user-info"),
+    private val logger: Logger,
+    private val sharedPrefMigrator: SharedPrefMigrator
```
tomlongridge, line 15: Nit: these params are the other way around to the `DeviceIdStore`

bugsnag-android-core/src/main/java/com/bugsnag/android/UserStore.kt
```diff
@@ -0,0 +1,99 @@
...
+        return userState
+    }
+
+    /**
+     * Persists the user if [Configuration.getPersistUser] is true and the object is different
+     * from the previously persisted value.
+     */
+    fun save(user: User) {
+        val hashCode = user.hashCode()
+        if (persist && previousUserHash.getAndSet(hashCode) != hashCode) {
```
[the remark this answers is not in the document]
fractalwrench: Updated to use `equals`.

bugsnag-android-core/src/main/java/com/bugsnag/android/ImmutableConfig.kt
fractalwrench: Updated to pass the initial user object in via `UserStore#load`.

tomlongridge (approved):

PR #1019, opened 2020-12-02
Setup mazerunner fixture to use multiple processes

fractalwrench:
## Goal

Updates the mazerunner fixture so that it is capable of running scenarios in a background [Service](https://developer.android.com/guide/components/services) which belongs to another process. This allows for multi process support to be covered by E2E tests.

## Changeset

The majority of changeset is made up of 2 relatively trivial changes to the scenario code:

1. `eventMetadata` is now a constructor parameter so that it is provided up front, and does not need to be obtained from the activity context
2. `run()` has been renamed as `startScenario()`, and scenarios can override how `Bugsnag` is initialized in the new `startBugsnag()` method. This makes it clearer that `startScenario()` should only ever be responsible for running the crashy code - and not for initializing Bugsnag.

### Other changes

- Added `MultiProcessService`, a background `Service` which is started from the main activity by an [Intent](https://developer.android.com/reference/android/content/Intent?hl=en) and then runs a `Scenario`
- Extracted logging code to separate kotlin file
- Extracted code for running crashy code into `Scenario` so that it does not require an `Activity` to run. 
- Created `BugsnagIntentParams` which encodes/decodes  the information required to initialize Bugsnag to/from an `Intent`.
- Added a `MultiProcessHandledExceptionScenario` which runs a single test case. The process name is added to metadata to distinguish each request. More test cases will be added in subsequent changesets.
- Updated manifest config loading scenarios to use `startBugsnag()` to initialize Bugsnag, and removed `skipBugsnag` metadata
- Deleted unused `Wait` class


Cawllec (approved):
LGTM - I noticed a couple of test failures in the smoke tests, which seem unrelated to the actual change.  Going to make a couple of tweaks upstream to remedy this.

PR #1021, opened 2020-12-03
[PLAT-5291] Add multistage CI for faster builds

Cawllec:
## Goal

To allow faster "smoke" tests to be run on general pushes and small feature branches, while retaining the more comprehensive tests on integration branches, or when targeting `next` or `master`.

The triggers for a full test set include:
- The commit is included in a PR targeting `next` or `master`
- The commit is on the `next` or `master` branches
- The commit contains the text `[full ci]`

The triggers for a quick test set include:
- The commit contains the text `[quick ci]` (This is not advised to be used generally, or without previous discussion)
- Any other scenarios not covered by the above

The smoke tests have been changed to be a bit more comprehensive and cover as much functionality as possible without slowing the tests down.  As several different things are occurring in each smoke scenario these shouldn't be taken as anything more than basic functionality coverage, and the full ci tests should be used to verify that specific elements of functionality work correctly in isolation.


twometresteve (approved):
LGTM - will be great to see how this goes for engineering over the coming days.  
I tweaked a Buildkite setting so that pull request builds are no longer skipped if a push build is already in progress.  This will mean the (quick) build for the push will be cancelled and (in the case of a next/master PR) a full build will be kicked off.  This may be an annoyance for PRs into an integration branch, where a quick build is sufficient - but we can see how that works out and if there are further settings we can look at.

PR #1024, opened 2020-12-04
Reduce browserstack concurrency to 9

fractalwrench:
## Goal

Reduces the browserstack concurrency from 10 to 9. This makes one device slot will always be available for running local tests, which speeds up the feedback loop without dramatically slowing down CI times.


twometresteve (approved):
LGTM

PR #1025, opened 2020-12-07
Make event.unhandled overridable for JVM errors

fractalwrench:
## Goal

This makes it possible to override the `unhandled` property for JVM errors. This allows users to discard certain events from contributing to the stability score.

## Changeset

- Renamed `HandledState` to `SeverityReason` for greater alignment with notifier spec
- Added setter to `Event` which allows altering the `unhandled` state
- If the `unhandled` has been changed, `unhandledOverridden` is serialized in the JSON payload
- Updated `DeliveryDelegate` so that it caches unhandled errors even if a user has set their `unhandled` state to `false`

## Testing

Updated and added to existing unit test coverage for `unhandled` state serialization. Added new E2E tests to assert that `unhandled` can be overwritten both ways, which have been run locally.


bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
```diff
@@ -45,7 +45,10 @@ void deliver(@NonNull Event event) {
             }
         }
 
-        if (event.isUnhandled()) {
+        boolean unhandledOverridden = event.impl.getUnhandledOverridden();
+        boolean unhandled = event.isUnhandled() || unhandledOverridden;
```
tomlongridge, line 49: If the user has changed a handled event to unhandled, it should not be treated as unhandled.

Perhaps rather than expose `getUnhandledOverridden`, we should expose the original handled value so it can be easily queried here?

bugsnag-android-core/src/test/java/com/bugsnag/android/EventFacadeTest.java
```diff
@@ -183,4 +184,11 @@ public void setUserValid() {
         event.setUser(null, null, null);
         assertEquals(new User(null, null, null), event.getUser());
     }
+
+    @Test
+    public void unhandledValid() {
+        assertFalse(event.isUnhandled());
+        event.setUnhandled(true);
+        assertTrue(event.isUnhandled());
```
tomlongridge, line 192: For completeness, wouldn't we set it back to false to make sure you can toggle both ways?

features/smoke_tests/sessions.feature
```diff
@@ -49,6 +49,7 @@ Scenario: Automated sessions send
     And the payload field "events.0.session.id" equals the stored value "automated_session_id"
     And the event "session.events.handled" equals 1
     And the event "session.events.unhandled" equals 0
+    And the event "severityReason.unhandledOverridden" is false
```
tomlongridge, line 52: Is there a reason to assert this here and none of the other smoke tests?

tomlongridge (commented):
There seem to be a couple of MR changes - is that deliberate? (including some sort of change on features/fixtures/mazerunner/gradlew)

features/smoke_tests/sessions.feature
fractalwrench: Updated to assert this in other smoke tests

bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
fractalwrench: Added an internally visible method named `getOriginalUnhandled()` which is now used here.

fractalwrench:
Addressed inline review comments. There are two parts to the mazerunner changes which are intentional:

1. Update `Gemfile` and `Gemfile.lock` so that they point at the latest mazerunner when running tests locally
2. Make the gradle wrapper executable through `chmod +x` which allows for the mazerunner test fixture to be built from the command line

tomlongridge (approved):

PR #1029, opened 2020-12-08
Add test that sends unusual strings in request payload

fractalwrench:
## Goal

Adds two scenarios (a handled JVM error and unhandled NDK error) which test that reports handle unusual strings in their metadata. This helps confirm that the SHA-1 validation is not affected by unusual unicode characters, and adds test coverage to ensure that requests are delivered with unusual unicode characters.

Assertions haven't been added on the metadata itself (blocked by mazerunner bug PLAT-5593), but this should validate that the request hash is not incorrect.


Gemfile
```diff
@@ -1,10 +1,10 @@
 source "https://rubygems.org"
 
 # You can run against local Maze Runner branches and uncommitted changes:
-gem 'bugsnag-maze-runner', path: '../maze-runner'
+#gem 'bugsnag-maze-runner', path: '../maze-runner'
 
 # Or a specific release:
-#gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v3.2.0'
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v3.5.1'
```
twometresteve, line 7: If you wait until later this morning you can do straight to v3.6.0, which will enforce the presence of the Bugsnag-Integrity header on all requests (not just that the digest is correct if present)

PR #1030, opened 2020-12-08
Tweak testing docs

fractalwrench:
## Goal

Tweaks the testing docs so that the command always builds the test fixture, and so that the feature file being run exists.


twometresteve (approved):
LGTM

PR #1031, opened 2020-12-08
Tests: Update smoke tests to be less flakey

Cawllec:
## Goal

A couple of scenarios were consistently flaking out.  This change:
- Removes Automatic breadcrumb test from CXX signal smoke test, as synchronisation couldn't be guaranteed. 
- Instead of lifecycle callbacks when delivering sessions switch to use a callback after the initial delivery.
- Look for the `Connectivity changed` breadcrumb in certain scenarios where `Bugsnag Loaded` proved to be flakey.


fractalwrench (approved):
LGTM pending CI

PR #1034, opened 2020-12-09
Load bugsnag-ndk lib for multi process scenario

fractalwrench:
## Goal

Loads the `bugsnag-ndk` library for the unhandled multi process scenario. This manifested as a failure on the E2E tests as the scenario tries to call C++ code which has not been loaded: https://buildkite.com/bugsnag/bugsnag-android/builds/2807#b6514d4f-cf03-4550-b11f-02110c44b8ed


twometresteve (approved):
LGTM

PR #1037, opened 2020-12-09
Make event.unhandled overridable for NDK errors

fractalwrench:
## Goal

Makes `event.unhandled` overridable for NDK errors, which means users can avoid certain errors counting towards their stability score.

## Changeset

- Added `bugsnag_event_set_unhandled` mutator for altering `event.unhandled` in a `bsg_on_error` callback
- Incremented unhandled count after the `bsg_on_error` callback has run
- Renamed `handled_state` to `severity_reason` for greater alignment with notifier spec naming
- Serialized `severityReason.unhandledOverridden` which denotes whether the `unhandled` flag has been altered by the user

## Testing

Updated existing integration tests for serialization coverage, and added E2E scenarios to verify that NDK errors can be overridden.


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -417,12 +417,14 @@ void bsg_serialize_grouping_hash(const bugsnag_event *event, JSON_Object *event_
...
-void bsg_serialize_handled_state(const bugsnag_event *event, JSON_Object *event_obj) {
+void bsg_serialize_severity_reason(const bugsnag_event *event, JSON_Object *event_obj) {
   // FUTURE(dm): severityReason/unhandled attributes are currently
   // over-optimized for signal handling. in the future we may want to handle
   // C++ exceptions, etc as well.
   json_object_set_string(event_obj, "severity", bsg_severity_string(event->severity));
-  json_object_dotset_boolean(event_obj, "unhandled", event->unhandled);
+  bool unhandled = event->unhandled;
+  json_object_dotset_boolean(event_obj, "unhandled", unhandled);
+  json_object_dotset_boolean(event_obj, "severityReason.unhandledOverridden", !unhandled);
```
tomlongridge, line 427: Think it would be worth commenting here on why `!unhandled` is correct (due to lack of support of notify callbacks), as a helpful pointer in future. 

```suggestion
  json_object_dotset_boolean(event_obj, "severityReason.unhandledOverridden", !unhandled);
```

tomlongridge (approved):
LGTM - would be good to add a comment on why we don't track handled->unhandled.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: Added some comments to the method

PR #1038, opened 2020-12-09
Tests: Update pipeline logic for improved integration PRs

Cawllec:
## Goal

This creates 3 paths for our CI pipelines:
- Smoke tests - Now integrated into the pipeline and always run, these do not depend on any pipeline conditions
- Integration branches - Triggered when a PR is created targeting `next` or `master`, or `[integration ci]` is included in a commit message.  This will run the smoke tests, and enclose the remaining set of tests behind a manual trigger.  Until this trigger is activated the build will remain incomplete.
- Full tests - Triggered when a PR is based on `next` or `master`, or `[full ci]` is included in a commit message.  This will act as the above, but skips the manual trigger ensuring the whole test set is run every time.

## Changeset

- `pipeline.smoke.yml` pipeline removed with steps moved into the main `pipeline.yml` to run every time
- Full test runs in `pipeline.full.yml` now depend on the appropriate smoke tests passing first (i.e. Android 10 full tests require Android 10 smoke tests to have passed)
- Instrumentation tests in `pipeline.full.yml` now depend on the `NDK 16 4.4` instrumentation test run, to avoid clogging the concurrency groups before more important E2E tests can run
- `block.step.yml` added consisting of the manual trigger step and a command that adds `pipeline.full.yml` to the pipeline
- Pipeline triggering script now has 3 outcomes:
  - Doing nothing, in the case of the smoke tests pipeline
  - Appending `pipeline.full.yml` to the pipeline if the full tests are triggered
  - Appending `block.step.yml` to the pipeline if the integration branch pipeline is triggered.

## Testing

- [Example of the smoke test pipeline](https://buildkite.com/bugsnag/bugsnag-android/builds/2841)
- [Example of the integration branch pipeline](https://buildkite.com/bugsnag/bugsnag-android/builds/2838)
- [Example of the fully triggered pipeline](https://buildkite.com/bugsnag/bugsnag-android/builds/2839)
- [Example of integration branch pipeline created from this PR](https://buildkite.com/bugsnag/bugsnag-android/builds/2846)


twometresteve (commented):
This makes sense to me.  It stops merges to integration branches that have a draft PR from consuming loads of CI time, whilst still providing the flexibility for a full build.
I'm sure the pattern will continue to evolve, but I think this is a good step to make right now. 

Will let @fractalwrench give the approval, assuming you're happy.

fractalwrench:
This LGTM as a general approach & feels very promising for reducing CI usage. I think it'd be good to get this merged as there's a pressing need. 

As far as future multistage builds go I think the following changes would be worth making. Although this probably requires a conversation with you, me & @twometresteve:

1. Consider running the smoke test E2E jobs against the min/max Android API levels only, as it's fairly rare that API level behaviour breaks something. _This reduces browserstack slots used from 9 to 2._
2. Running the smoke test E2E jobs against all API levels on the integration/full run looks good.
3. This feature is probably redundant so can be removed to shave ~1 minute of build times: https://github.com/bugsnag/bugsnag-android/blob/tests/pipeline-update/features/full_tests/_smoke_test.feature
4. The test matrix for the instrumentation tests can be cut down without adding much risk as they don't explicitly test anything API specific. _This reduces browserstack slots used from 11 to 4:_
  - NDK 12, 4.4
  - NDK 16, 7.1
  - NDK 19, 9.0
  - NDK 21, 11
5. For the integration run I think 1 E2E job should run the full suite of tests by default. This might seem counterintuitive but running all the tests on one API level is usually sufficient to detect test failures on all of them. Doing this means that engineers will be less likely to trigger a full pipeline run, which uses 9 browserstack slots. This will probably be slow/flaky to start off with but we could investigate sharding the scenarios as a future optimization.

That's all the low-hanging fruit I can think of for multistage builds. The next area of low hanging fruit IMO is going through all the existing scenarios and removing redundant test cases, along with enhancing the assertions so that each scenario can test more. I think the smoke test scenarios have been really successful in that regard.

fractalwrench (approved):

PR #1040, opened 2020-12-10
Add test that sends unusual strings in request payload

fractalwrench:
## Goal

Adds two scenarios (a handled JVM error and unhandled NDK error) which test that reports handle unusual strings in their metadata. This helps confirm that the SHA-1 validation is not affected by unusual unicode characters, and adds test coverage to ensure that requests are delivered with unusual unicode characters.


twometresteve (approved):
LGTM pending CI

PR #1043, opened 2020-12-11
Prevent potential SHA-1 hash mismatch in Bugsnag-Integrity header for session requests

fractalwrench:
## Goal

Prevents a SHA-1 hash mismatch when the React Native notifier changes the `notifier` field on startup. Because the Android notifier assigns the event payload value by reference from `Client#notifier`, the request payload can change between generating a SHA-1 hash and making the request.

## Changeset

Copied the `notifier` parameter when constructing a `Session`, so that it does not hold a copy of global state.

## Testing

Added unit test to verify that the `notifier` property is always copied when constructing a `Session`.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
@@ -5,6 +5,7 @@ import androidx.annotation.Nullable;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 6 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
```diff
@@ -61,7 +60,9 @@ static Session copySession(Session session) {
     Session(File file, Notifier notifier, Logger logger) {
         this.file = file;
         this.logger = logger;
-        this.notifier = notifier;
+        Notifier copy = new Notifier(notifier.getName(), notifier.getVersion(), notifier.getUrl());
+        copy.setDependencies(new ArrayList<>(copy.getDependencies()));
```
tomlongridge, line 64: Copy-paste-o:

```suggestion
        copy.setDependencies(new ArrayList<>(notifier.getDependencies()));
```

This shows the unit test didn't cover as much as it could - it should check the behaviour with dependent notifiers set.

tomlongridge (changes requested):

bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
fractalwrench: Updated to test this with the dependent notifier set.

tomlongridge (approved):

PR #1046, opened 2020-12-15
Split native mazerunner scenarios into smaller feature files

fractalwrench:
## Goal

Splits up the native mazerunner scenarios into more specific feature files. This makes it easier to determine at a glance what each scenario is testing and makes it easier to run specific scenarios. For example, it's now much easier to run scenarios which test the native User API in isolation from other native scenarios, which would be helpful when making changes in this area.


features/full_tests/native_crash_handling.feature
```diff
@@ -105,80 +105,6 @@ Feature: Native crash reporting
         And the event "severity" equals "error"
         And the event "unhandled" is true
 
-    Scenario: Raise SIGILL
```
twometresteve, line 108: There are 142 scenarios running on next and only 141 here.  It look like this one has been dropped - was that intentional?

twometresteve (commented):
I've not checked that precisely the same scenarios are present before and after, but did notice that one scenario has been removed without mention on the front page.  If that's intentional then I'm happy to approve.

features/full_tests/native_crash_handling.feature
fractalwrench: Nice catch - that was a scenario I missed. This has now been reinstated in `native_signal_raise.feature`.

twometresteve (approved):
LGTM.  I've checked that the set of scenario names is the same before and after the change (but not that all scenarios are precisely the same).

PR #1048, opened 2020-12-15
Build example app on CI

fractalwrench:
## Goal

Ensures that the example app builds by adding a check to the full CI pipeline which assembles an APK.


twometresteve (approved):
LGTM.  I half-wondered if uploading the built app as a Buildkite artifact could be useful, but I think not without a valid apiKey.

PR #1049, opened 2020-12-15
Mazerunner fixture multiple apks

fractalwrench:
## Goal

Cherry picks the commits from the [ANR NDK stacktrace integration branch](https://github.com/bugsnag/bugsnag-android/pull/1003) which make it possible to build multiple mazerunner fixtures. Note that this changeset  has previously been reviewed and does not contain any changes to the ANR code - it purely restructures the test fixtures.

This changeset opens the way for building multiple fixture APKs that use different versions of the NDK. This will grant us better test coverage and also enable an update of the AGP version to 4.1.


twometresteve (approved):
LGTM, just keeping my fingers cross that everything comes out in the git wash when it comes to merging the integration branch.  It's probably worth merging `next` up into it after this PR is merged to help git on its way.

PR #1050, opened 2020-12-15
Increase kotlin dependency version to 1.3.72

fractalwrench:
## Goal

Increases the Kotlin dependency version to 1.3.72. This contains the [following changes](https://blog.jetbrains.com/kotlin/2020/03/kotlin-1-3-70-released/) and allows us to keep up to date.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Increase kotlin dependency version to 1.3.72
+  [#1050](https://github.com/bugsnag/bugsnag-android/pull/1050)
+
 ## 5.4.0 (2020-12-14)
 
 * Make `event.unhandled` overridable for NDK errors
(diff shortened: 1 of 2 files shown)


tomlongridge (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Increase kotlin dependency version to 1.3.72
```
tomlongridge, line 5: I'm not sure whether this needs a changelog entry though as it doesn't affect our users?

fractalwrench: There shouldn't be any effect on the API bugsnag uses, but the Kotlin compiler won't produce identical bytecode for 1.3.61 to 1.3.72, so I thought it'd be sensible to document as there is the small possibility it could introduce change. Adding a changelog entry is also consistent with what was done when upgrading previously

PR #1055, opened 2020-12-17
Soft fail Android 4, 6, and 7 E2E tests

fractalwrench:
## Goal

Adds a temporary soft fail to the Android 4, 6, and 7 E2E tests which are experiencing 400 status codes due to BrowserStack network issues.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -198,6 +198,9 @@ steps:
           - "--fail-fast"
     concurrency: 9
     concurrency_group: 'browserstack-app'
+    # temp workaround browserstack network issue where 400 status codes fail the tests
+    soft_fail:
+      - exit_status: "*"
 
   - label: ':android: Android 5 end-to-end tests'
     depends_on:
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.full.yml
```diff
@@ -198,6 +198,8 @@ steps:
           - "--fail-fast"
     concurrency: 9
     concurrency_group: 'browserstack-app'
+    soft_fail:
```
twometresteve, line 201: Although we initially saw problems on Android 4, I don't think there is an inherent infrastructure issue there that we're working around temporarily.  As it represents the minimum supported version, I'd rather not soft fail unless we really have to.

.buildkite/pipeline.full.yml
```diff
@@ -242,6 +244,8 @@ steps:
...
```
twometresteve, line 247: Please comment these to make it clear that we're working around a temporary browserstack issue (presumed) and that the soft fail should be removed as soon as possible once confirmed and fixed.

fractalwrench:
Removed soft fail from Android 4, added comment explaining the temporary soft fails

twometresteve (approved):
LGTM

PR #1056, opened 2020-12-17
Update CI pipeline to build multiple fixture APKs and use AGP 4.1

fractalwrench:
## Goal

Updates the CI pipeline to build multiple fixture APKs which are compiled using different versions of the NDK. This reduces the number of CI jobs and is a more comprehensive test of differences in our NDK code. It also opens the opportunity for altering other fields in future, such as changing the `compileSdkVersion` or creating fixtures that use bugsnag with other common Android libraries.

A necessary pre-requisite of this was to update the Android Gradle Plugin to 4.1.

### AGP
- Bump AGP version to [4.1.1](https://developer.android.com/studio/releases/gradle-plugin#4-1-0) from 3.4.2, which brings large performance improvements when building locally and enables several improvements in our backlog
- Set the [ndkVersion](https://developer.android.com/studio/releases/gradle-plugin#default-ndk-version) in the build script rather than through the use of the deprecated `ANDROID_NDK_HOME`. `sdkmanager` is now responsible for downloading the NDK bundles
- Enabled [parallel execution](https://developer.android.com/studio/releases/gradle-plugin#default-ndk-version) which should decrease Gradle build times
- Used the [publish plugin](https://developer.android.com/studio/releases/gradle-plugin#maven-publish) bundled with AGP as the old one is incompatible
- Include first SO file to avoid duplicate `liblog.so` causing [build errors](https://developer.android.com/studio/releases/gradle-plugin#cmake-imported-targets)

### Building multiple test fixtures

- The mazerunner test fixture uses `TEST_FIXTURE_NDK_VERSION` to control the NDK version used to build the fixture, and `TEST_FIXTURE_NAME` to control the filename of the generated APK
- The buildkite pipeline generates 4 fixture APKs: r16, r19, and r21, a minimal fixture with only bugsnag-android-core
- The E2E tests have been updated so that the older Android versions use the older NDK fixture APKs (the exact test matrix is open to future refinement)
- The instrumentation tests in the full pipeline have been cut down, so that they run against Android 4, 7, and 9. This is because the instrumentation tests were previously building bugsnag-android against different NDK versions which is not supported - adapting the E2E tests to do this instead is preferable as it's a more realistic test, and cuts down on the number of CI jobs

### Misc

- Explicitly set [the sourceCompatibility](https://developer.android.com/studio/write/java8-support#supported_features) for bugsnag-android modules rather than relying on the default
- Bumped [compileSdkVersion](https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd) to 30 for test/example code which reduces the number of Android platforms required on CI from 3 to 2. The `compileSdkVersion` used for generating bugsnag-android artefacts remains unchanged for now
- Example apps & test fixture now use most up-to-date versions of Bugsnag deps
- Commented out some steps on Cxx scenarios which were failing due to different information being received when running against different versions of the NDK, and created a ticket to address these
- Verified that artefacts are uploaded to sonatype & bintray, and that the mazerunner fixture compiles when installing the artefact from mavenLocal


features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
```diff
@@ -84,7 +84,7 @@ abstract class Scenario(
     override fun onActivityResumed(activity: Activity) {}
     override fun onActivityPaused(activity: Activity) {}
     override fun onActivityStopped(activity: Activity) {}
-    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle?) {}
+    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) {}
```
fractalwrench, line 87: The nullability annotation for this method changed between API 29 and 30 allowing for better Kotlin support

dockerfiles/Dockerfile.android-ci-base
```diff
@@ -30,7 +27,7 @@ COPY build.gradle settings.gradle /app/
 
 # Build the example app - needed for Espresso tests
 COPY examples/ examples/
-RUN ./gradlew -p examples/sdk-app-example assembleRelease
+RUN cd examples/sdk-app-example && ./gradlew assembleRelease
```
fractalwrench, line 30: This is a better test that the example app builds - if something was wrong with the example app's gradle wrapper the old command would not have caught this.

features/fixtures/mazerunner/cxx-scenarios/build.gradle
```diff
@@ -35,7 +35,6 @@ android {
 dependencies {
     implementation project(":jvm-scenarios")
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation "com.bugsnag:bugsnag-android:9.9.9"
```
fractalwrench, line 38: This fixes a bug where the minimal fixture was bringing in the NDK/ANR plugins

Cawllec (approved):
LGTM, only ask would be to change the title of the e2e test steps to include the NDK version.

fractalwrench:
Rebased and updated E2E tests to include NDK versions.

PR #1057, opened 2020-12-22
Reorganize dockerfiles to reduce CI build times

fractalwrench:
## Goal

Reorganizes the dockerfiles with the aim of decreasing CI build times.

## Changeset

- Android Size Reporting job only runs on a full CI run
- Removed `android-jvm` container. This was redundant as the `android-ci` container is capable of performing the same job, and removing it avoids having to store an extra image on AWS (and waiting for it to download)
- Updated labels for `android-common` and `android-ci` to make it clearer what the responsibilities are for each
- Updated `minimalapp` test fixture to use latest gradle and [parallel execution](https://docs.gradle.org/nightly/userguide/performance.html#parallel_execution), which decreases job time

### android-common and android-ci

`android-common` is now responsible for all the dependencies which are required to build an Android application, such as the Android platform image, the JDK, etc. These are installed up-front and are not anticipated to change except in the rare case where the dependency graph changes.

`android-ci` is based on `android-common` but builds the `bugsnag-android` code, so is anticipated to change on each branch.
 
The following improvements have been made to these containers:

- Fresh build of 'build android base image' now takes 5 minutes rather than 20, and takes up substantially less storage on the docker registry. This previously took a long time because the container had 4 NDK toolchains - the download for these has been moved to the fixture APKs, which increases the build time by around a minute (it is anticipated that on a quick CI run only the r16b fixture needs building, so this will not negatively affect speed)
- General version bump of build tool dependencies
- Migrated image from openjdk-8 to openjdk-11, which should bring an [easy performance improvement](https://docs.gradle.org/nightly/userguide/performance.html#use_latest_gradle_and_jvm_versions) for Gradle
- The Android toolchain is now installed using a [command line tools](https://developer.android.com/studio/command-line) package which installs the required dependencies

## Future work

The Android CI image seems to build itself from scratch for each branch. As its dependencies rarely change it might be better to manually tag & push to the docker registry when things change.


dockerfiles/Dockerfile.android-sizer
```diff
@@ -1,23 +1,14 @@
...
 
-# Build and upload to the local maven as version 9.9.9
-RUN sed -i -e 's/VERSION_NAME=.*/VERSION_NAME=9.9.9/g' gradle.properties
-
-RUN ./gradlew assembleRelease publishToMavenLocal -PABI_FILTERS=arm64-v8a,armeabi,armeabi-v7a,x86,x86_64
-
-COPY features features
-
 # Download the bundle tool
 RUN apt-get install -y ruby-full
```
fractalwrench, line 5: This has been kept in the sizer container due to its size and the fact it's only used here.

dockerfiles/Dockerfile.android-ci-base
```diff
@@ -14,14 +9,6 @@ COPY gradle/ /app/gradle/
 ENV GRADLE_OPTS="-Dorg.gradle.daemon=false"
 RUN ./gradlew
 
-# Setup signing
```
fractalwrench, line 17: Signing now takes place in the `android-common` image as this improves caching of the generated key

dockerfiles/Dockerfile.android-ci-base
```diff
@@ -37,5 +24,7 @@ COPY bugsnag-android-core/ bugsnag-android-core/
 COPY bugsnag-plugin-android-ndk/ bugsnag-plugin-android-ndk/
 COPY bugsnag-plugin-react-native/ bugsnag-plugin-react-native/
 COPY scripts/ scripts/
+COPY config/ config/
 
-RUN ./gradlew
+# Build and upload to the local maven as version 9.9.9
+RUN ./gradlew assembleRelease publishToMavenLocal -PVERSION_NAME=9.9.9 -PABI_FILTERS=arm64-v8a,armeabi,armeabi-v7a,x86,x86_64
```
fractalwrench, line 30: Assembling the bugsnag-android AAR here avoids the previous situation where it needed to be assembled twice in different containers

dockerfiles/Dockerfile.android-common
```diff
@@ -1,25 +1,44 @@
...
 
 RUN apt-get update
-RUN apt-get install -y gradle jq
+RUN apt-get install -y wget
+RUN apt-get install -y maven
+RUN apt-get install -y gnupg1
+RUN apt-get install -y cppcheck
+RUN apt-get install -y libncurses5
+RUN apt-get install -y jq
+RUN apt-get clean
```
twometresteve, line 10: As I understand it, it's generally best to chain commands together in Docker files where reasonable to reduce the number of image layers and overall size.  E.g. something like:
RUN apt-get update && apt-get install -y wget maven gnupg1.... && apt-get clean

twometresteve (approved):
Made one suggestion, but LGTM.

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,5 @@
 steps:
-  - label: ':docker: Build Android base image'
+  - label: ':docker: Build Android base image (downloads the Android toolchain)'
```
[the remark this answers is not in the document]
fractalwrench: Updated so that a comment has been added to the pipeline yaml

PR #1059, opened 2020-12-22
Use correct assertion for native throw scenarios

fractalwrench:
## Goal

#1049 allowed the CI pipeline to build multiple fixture APKs so that a 'minimal' test fixture which only uses `bugsnag-android-core` could be tested. A bad merge with #1055 resulted in two scenarios failing in `native_throw_crash.feature`, due to the old assertions being used. This changeset adds the correct assertions to the test cases.

As part of #1049 the two scenarios were moved to a separate `cxx-scenarios` module to allow the building of a minimal test fixture that doesn't rely on bugsnag's NDK layer. Both two scenarios throw a C++ exception, and were previously asserting that information about the top stackframes were available. Because the scenario code has been migrated to another module they are now in a separate SO file, so the app is unable to obtain the previous information through the C++ exception handling APIs.

### DIFF
--- a/features/full_tests/native_throw_crash.feature
+++ b/features/full_tests/native_throw_crash.feature
@@ -8,12 +8,8 @@ Feature: Native crash reporting with thrown objects
         And the request payload contains a completed unhandled native report
         And the event "severity" equals "error"
         And the event "unhandled" is true
-        And the exception "errorClass" equals "PSt13runtime_error"
-        And the exception "message" equals "How about NO"
-        And the first significant stack frame methods and files should match:
-            | run_away(bool)       | | libentrypoint.so |
-            | trigger_an_exception | | libentrypoint.so |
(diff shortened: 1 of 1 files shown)


tomlongridge:
As noted, the change reduces the level of testing. Are the equivalent checks elsewhere in the e2e suite? If so, would be good to note them on the PR, so we can be sure we're not losing anything overall.

fractalwrench:
I don’t think it is possible to capture the equivalent information from an uncaught C++ exception when it’s built in a separate SO file. This happened when the `cxx-scenario` module was created to allow running scenarios against a minimal test fixture. There are a couple of other scenarios which validate NDK stacktraces are captured: https://github.com/bugsnag/bugsnag-android/blob/next/features/full_tests/native_crash_handling.feature#L134

tomlongridge (approved):

PR #1060, opened 2020-12-23
Update example app's to only use different directory for background process

fractalwrench:
## Goal

Most users will want to keep the default directory for their main app. When using a second process it is necessary to set the `persistenceDirectory`, therefore the example app has been configured to set this value only for any process which does not match the main process.

Tested by running the example app and confirming that the file structure appeared as expected in the Device File explorer.

### DIFF
--- a/examples/sdk-app-example/app/src/main/java/com/example/bugsnag/android/ExampleApplication.kt
+++ b/examples/sdk-app-example/app/src/main/java/com/example/bugsnag/android/ExampleApplication.kt
@@ -22,21 +22,19 @@ class ExampleApplication : Application() {
 
     private external fun performNativeBugsnagSetup()
 
-    private val multiProcessApp = true
-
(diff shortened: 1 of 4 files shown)


examples/sdk-app-example/app/src/main/java/com/example/bugsnag/android/ExampleApplication.kt
```diff
@@ -22,21 +22,21 @@ class ExampleApplication : Application() {
 
     private external fun performNativeBugsnagSetup()
 
-    private val multiProcessApp = true
+    private val MAIN_PROCESS_NAME = "com_example_bugsnag_android"
```
tomlongridge, line 25: Can be removed.

tomlongridge (approved):
LGTM - just an extra variable to remove.

PR #1064, opened 2021-01-05
Remove temporary soft fails

twometresteve:
Remove the temporary soft fail put in place on Android 4,6 and 7 end-to-end tests now that we have determined the cause (enabled network logs in on BrowserStack) and deployed a workaround in MazeRunner (no not enable the logs!).


fractalwrench (approved):

twometresteve:
Now overtaken by #1066

PR #1065, opened 2021-01-06
Add git to docker image

fractalwrench:
## Goal

`git` isn't included in the ubuntu 20.04 image which was causing the Android Size reporting step to fail. No other CI jobs are affected.


twometresteve (approved):
LGTM

PR #1066, opened 2021-01-06
[PLAT-5613] Add barebones stage to CI that runs a minimal amount of tests

Cawllec:
## Goal

Refactors the CI pipeline to:
- Replace the "quick" CI build with a "barebones" CI build that only runs smoke tests against Android 4.4 and Android 11
- Enables the traditional "quick" ci build to be run by adding `[quick ci]` to the commit message
- Re-names the "integration" CI build to "gated-full", which reflects the type of build more accurately
- Adds further feedback to the pipeline trigger


.buildkite/pipeline_trigger.sh
```diff
@@ -0,0 +1,37 @@
...
+  buildkite-agent pipeline upload .buildkite/pipeline.quick.yml
+  buildkite-agent pipeline upload .buildkite/block.step.yml
+elif [[ "$BUILDKITE_MESSAGE" == *"[quick ci]"* ]]; then
+  echo "Running quick build"
+  echo "Unit and static tests will be run"
+  echo "Minimal instrumentation tests will be run"
+  echo "End-to-end smoke tests will be run on all supported Android versions"
+  buildkite-agent pipeline upload .buildkite/pipeline.quick.yml
+else
+  echo "Running quick build"
```
twometresteve, line 35: I think this should say barebones now?

twometresteve (commented):
Nice.  Just a couple of small comments and I'll let @fractalwrench give the approval.

fractalwrench (approved):
LGTM, one comment would be that the 'Build Fixture r19 APK' step looks like it runs on the barebones stage, when it could be moved to the gated/full builds.

.buildkite/pipeline_trigger.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "Minimal instrumentation tests will be run"
+  echo "End-to-end smoke tests will be run on minimum and maximum supported Android versions"
+elif [[ "$BUILDKITE_MESSAGE" == *"[full ci]"* ||
+  "$BUILDKITE_BRANCH" == "next" ||
+  "$BUILDKITE_BRANCH" == "master" ]]; then
+  echo "Running full build"
+  echo "Unit and static tests will be run"
+  echo "All instrumentation tests will be run"
+  echo "All end-to-end tests will be run on all supported Android versions"
+  buildkite-agent pipeline upload .buildkite/pipeline.full.yml
```
twometresteve, line 15: I've just realised (having asked for a `[full ci]` that the post-merge build will not complete, because the full pipeline has dependencies on steps that have now moved to the quick file (which isn't loaded now).

Can we simply add the quick pipeline here for now and create a fresh ticket to have another look?

PR #1070, opened 2021-01-07
Add one full set of e2e tests to quick build

twometresteve:
## Goal

Move the full set of e2e tests for Android 9 from the Full to Quick build.  

## Design

The theory is that a single full run is generally sufficient to realise any issues and may reduce the likelihood of an engineer triggering a Full build, consuming a lot of test device bandwidth.

## Changeset

As well as moving the build step, I have also made a few improvements generally now that MazeRunner itself supports specific environment variables (the previous variables came only from a commonly used convention).

## Testing

Covered by CI and I checked the pipeline that was generated for the PR, before opening the gate.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -54,8 +54,6 @@ steps:
           - "--app=/app/build/release/fixture-minimal.apk"
           - "--farm=bs"
           - "--device=ANDROID_9_0"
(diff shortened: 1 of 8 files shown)


fractalwrench (approved):
LGTM pending CI

twometresteve:
Overtaken by #1074 !

PR #1071, opened 2021-01-08
Fix failing multi process scenario

fractalwrench:
## Goal

Fixes a multi process E2E scenario which failed on Android <6 due to SIGILL rather than SIGTRAP. This appears to be caused by a difference in platform behaviour by the crashing code - this has been simplified by just calling `abort()` so that it's consistent across API levels. The scenario was also failing on 4.4 due to the service finishing before the error could be thrown - running the crashy code on the main thread has resolved this issue.

Additionally this changeset bumps mazerunner to the latest version.


twometresteve (approved):
LGTM

PR #1072, opened 2021-01-08
Separate steps for launching and relaunching the app

twometresteve:
## Goal

Separate `I relaunch the app` step into its two separate uses.

## Design

`I relaunch the app` was being used for two subtly different purposes:
- Launching the app after a crash
- Closing the app whilst it is running and then launching it

This PR separates the two concepts into more explicitly worded steps.  With the extra delay before launching the app after a crash, it's possible that some flakes will resolve, because MazeRunner won't be forcefully closing or opening the app while it is writing to file.

## Changeset

Cucumber step split into two.

## Testing

Covered by CI.


twometresteve:
Warning: This PR is being overtaken!

twometresteve:
Overtaken by #1074

PR #1074, opened 2021-01-11
Various pipeline enhancements and tagging of flaky scenarios

twometresteve:
## Goal

This PR has ended up making a number of changes:

1. Add a single full e2e test run (for Android 9) to the “quick”/“gated” runs
2. (Overtaking #1072) separates the two concepts of “relaunching” the app into 2 steps
3. Separates e2e feature files into two batches
4. Addition of `flaky.feature`, currently containing apparently flaky scenarios.

Also the following incidental improvement:
- Set device farm creds via the environment rather than CL options

## Design

1. This will hopefully reducing the likelihood of engineerings triggering a full build and using more test device slots than really needed.
2. One steps is for closing and opening a running app, the other for opening the app after if has crashed.  The hope here is that flakes will be reduced because there will be more control over how apps are closed and (re)opened).
3.  This makes retries quicker and potentially reduces elapsed wait time if enough devices are available.
4. These will be investigated in more detail, ticketed and original locations commented as appropriate (immediately following this PR).

## Testing

Covered by CI, noting there is one all-present issue where a 400 response could be returned from the BS proxy, causing any scenario to fail.  These do no necessarily constitute a notifier issue and investigatory and remedial work on hardening the test setup continues.


.buildkite/pipeline_trigger.sh
```diff
@@ -1,19 +1,23 @@
...
 elif [[ "$BUILDKITE_MESSAGE" == *"[full ci]"* ||
   "$BUILDKITE_BRANCH" == "next" ||
   "$BUILDKITE_BRANCH" == "master" ]]; then
   echo "Running full build"
   echo "Unit and static tests will be run"
   echo "All instrumentation tests will be run"
   echo "All end-to-end tests will be run on all supported Android versions"
-  buildkite-agent pipeline upload .buildkite/pipeline.quick.yml
   buildkite-agent pipeline upload .buildkite/pipeline.full.yml
+  buildkite-agent pipeline upload .buildkite/pipeline.quick.yml
```
twometresteve, line 20: This was needed because the steps are become reversed in the pipeline (because Buildkite inserts then in place).  As there are steps in full that depend on those in quick, full builds ended up being blocked from finishing.  I'll add a comment to that effect.

docker-compose.yml
```diff
@@ -44,8 +44,11 @@ services:
   android-maze-runner:
     image: 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner-releases:latest-v3-cli
     environment:
-      VERBOSE:
       DEBUG:
+      BUILDKITE:
```
twometresteve, line 48: Means that MazeRunner will be able to detect it being run from BK and use ANSI logging for the BrowserStack dashboard link.

.buildkite/pipeline.full.yml
```diff
@@ -268,12 +346,198 @@ steps:
...
+    timeout_in_minutes: 60
+    plugins:
+      artifacts#v1.2.0:
+        download: "build/release/fixture-r21.apk"
+      docker-compose#v3.7.0:
+        run: android-maze-runner
+        command:
+          - "features/full_tests/flaky.feature"
+          - "--retry=2"
+          - "--strict-undefined"
```
fractalwrench, line 537: Perhaps a future enhancement, but if there's any way of specifying this (and other flags) via mazerunner config/in one place it would help make the yaml a bit more readable

.buildkite/pipeline.full.yml
```diff
@@ -268,12 +346,198 @@ steps:
...
+          - "--strict-pending"
+          - "--app=/app/build/release/fixture-r16.apk"
+          - "--farm=bs"
+          - "--device=ANDROID_4_4"
+    concurrency: 9
+    concurrency_group: 'browserstack-app'
+
+  - label: ':hankey: Flaky Android 5 NDK r16 end-to-end tests'
```
fractalwrench, line 399: Probably a future enhancement, but is there any way to reduce the length of these yaml files? A lot of the jobs are doing the same thing but only altering the fixture APK/device, so perhaps there's a case for generating the yaml from some sort of script that has a test matrix?

Makefile
```diff
@@ -12,18 +12,18 @@ test:
...
+ifeq ($(MAZE_DEVICE_FARM_ACCESS_KEY),)
+	@$(error MAZE_DEVICE_FARM_ACCESS_KEY is not defined)
 endif
 	@APP_LOCATION=/app/bugsnag-android-core/build/outputs/apk/androidTest/debug/bugsnag-android-core-debug-androidTest.apk \
 	 INSTRUMENTATION_DEVICES='["Google Nexus 5-4.4", "Google Pixel-7.1", "Google Pixel 3-9.0"]' \
 	 docker-compose up --build android-instrumentation-tests
 
 TEST_FIXTURE_NDK_VERSION ?= 16.1.4479499
-test-fixture:
+test-fixtures:
```
fractalwrench, line 26: The name should stay as-is IMO as only one APK is generated

TESTING.md
```diff
@@ -44,38 +44,34 @@ __Your session will periodically expire__, so you'll need to run this command to
...
+* `MAZE_DEVICE_FARM_ACCESS_KEY`: The BrowserStack App Automate Access Key
 
 Run `make remote-test`
 
 ### End-to-end tests
 
 Ensure that the following environment variables are set:
-* `BROWSER_STACK_USERNAME`: Your BrowserStack App Automate Username
-* `BROWSER_STACK_ACCESS_KEY`: You BrowserStack App Automate Access Key
+* `MAZE_DEVICE_FARM_USERNAME`: Your BrowserStack App Automate Username
```
fractalwrench, line 63: Will this break things for existing users or as is there backwards compatibility?

.buildkite/pipeline.full.yml
```diff
@@ -83,7 +81,7 @@ steps:
     concurrency: 9
     concurrency_group: 'browserstack-app'
 
-  - label: ':android: Android 4.4 NDK r16 end-to-end tests'
+  - label: ':android: Android 4.4 NDK r16 end-to-end tests - batch 1'
```
fractalwrench, line 84: I'd propose a logical way of splitting scenarios would be to run all the NDK/ANR scenarios in one job, and everything else in another. There are roughly ~150 scenarios in total and this approximately splits them in half, _and_ logically orders them in a directory structure.

features/full_tests/async_error_flush.feature
```diff
@@ -1,20 +0,0 @@
-Feature: Flushing errors does not send duplicates
```
fractalwrench, line 1: It seems preferable to use a `@FlakyTest` tag as that would result in less merge conflicts and would be easy to add if new flakes are discovered

fractalwrench (commented):
The general approach of splitting E2E tests into two jobs LGTM. Left a few comments inline - in addition there are a couple more CI flakes it'd be good to add to the flaky job.

I also wondered whether the buildkite UI allows grouping jobs together using something [like this](https://buildkite.com/changelog/42-parallel-jobs-now-show-as-a-single-group) - obviously a nice to have, but might be useful now that there are quite a few jobs on CI.

.buildkite/pipeline.full.yml
twometresteve: Yes, I agree and already an idea in the back of my mind and an enhancement (maybe more than one) to MR.

.buildkite/pipeline.full.yml
twometresteve: We've discussed a bit recently and I think @imjoehaines may be trialling an approach on our Ruby notifier.  Certainly one to keep in mind and consideration once the pipeline has settled down fully.

TESTING.md
twometresteve: The `--username` and `--access-key` command line options are still available, so this won't break people still using the old approach.  They will only need to change the environment variables they use if they want to take advantage of MazeRunner's ability to read config from the environment.  It's maybe worth noting that `BROWSER_STACK_USERNAME` was never a feature of MazeRunner itself - it was just very commonly used in the `env.rb` files - closely associated with MazeRunner, of course, but not a part of the tool itself.

.buildkite/pipeline.full.yml
twometresteve: Sounds good to me - could you raise an internal ticket and send it my way?  If it can just makes the separation clear so there's no margin for error and we'll make it happen in a forthcoming sprint.

features/full_tests/async_error_flush.feature
twometresteve: Yes, that's a good idea and occurred to me after creating the PR.  I needed to raise tickets for and comment the individual tests anyway, so can factor this into that - in a fast-following PR.

twometresteve: Actually, I think I'll do it on this PR.

Makefile
twometresteve: Ah, I had thought (without checking) that it was now building multiple, for the minimal test fixture.  There was also a driver to make it consistent with Cocoa so I don't keep getting it wrong when moving between repos, but I'm happy to change it back.

.buildkite/pipeline.full.yml
fractalwrench: Raised an internal ticket

features/full_tests/async_error_flush.feature
twometresteve: Now done.

fractalwrench (approved):
LGTM

Issue #1026, opened 2020-12-07
Attempting to perform JNI actions synchronously during a signal handler is not safe


allight:
### Describe the bug

The code here: https://github.com/bugsnag/bugsnag-android/blob/ff3c8b66c08dfe1f68360b22b67613137be00e8b/bugsnag-plugin-android-anr/src/main/jni/anr_handler.c#L48-L62 violates JNI invariants and is not safe.

During handling of SIGQUIT signals your library can attempt to call JNI functions directly on the current thread. Safe use of JNI code requires that the thread it is called on is in the native thread state. This can only be accomplished by either (1) the (previously unattached) thread calling the AttachCurrentThread method or (2) the most recent java language function call being a native method (and control having been passed to the native method implementation). Performing JNI calls with the thread in an incorrect state can cause many issues including SEGV crashes as the GC or stack-tracing attempts to walk an invalid stack.

Depending on the requirements of bugsnag there are a few ways you could solve this. The easiest would be to simply have a single signal-handling thread that the signal-handler wakes up with a semaphore. Note that this handling cannot be synchronous because waiting for the handling to finish before allowing the signal-handler to continue could cause deadlocks, depending on where the runtime was during the signal.

As another more minor issue this entirely disables the ART internal SIGQUIT/ANR dumping system since it relies on a [signal-catcher thread](https://cs.android.com/android/platform/superproject/+/master:art/runtime/signal_catcher.cc;l=171;bpv=0;bpt=1) being [woken up by sigwait](https://cs.android.com/android/platform/superproject/+/master:art/runtime/signal_set.h;drc=master;bpv=0;bpt=1;l=46) which your unblocking of the SIGQUIT mask will break. Leaving this as just an FYI though since I'm not sure if this is intentional or not.

### Environment
* Android version: All
* Bugsnag version: Unknown
* Emulator or physical device: All

yousif-bugsnag:
Hi @allight,

Thanks for the detailed report. We are investigating this and will update you as soon as we know more.

phillipsam:
A fix for this has now been released in [v5.5.1](https://github.com/bugsnag/bugsnag-android/releases/tag/v5.5.1).

PR #1078, opened 2021-01-13
Switch to Google-friendly ANR handler.

kstenerud:
## Goal

PLAT-4956 and PLAT-5607

## Design

- Re-block SIGQUIT and then signal the Google ANR thread directly.
- Call Java from an async-safe thread.

## Testing

Manual testing on multiple OS and device types, checking that both Bugsnag and Google get ANR reports.

### DIFF
--- a/bugsnag-plugin-android-anr/src/main/jni/anr_google.c
+++ b/bugsnag-plugin-android-anr/src/main/jni/anr_google.c
@@ -1,0 +1,155 @@
+// File:   anr_google.c
+// Author: Karl Stenerud
+// Date:   January 13, 2021
+//
+// Before starting an app, the Android runtime library first sets up its own SIGQUIT handler
+// (art/runtime/signal_catcher.cc, art/runtime/signal_set.h), which listens for SIGQUIT messages
+// that the OS sends to it when it detects an ANR. The handler uses sigwait() instead of the more
+// common sigaction() mechanism, which complicates things:
+//
+// - sigwait() only triggers when a signal becomes PENDING, which can only happen if the signal is
+//   blocked on all threads. As soon as a signal handler gets called, the signal is no longer
+//   pending. Also, an unblocked signal never becomes pending.
+//
+// - Blocked signals will never reach a signal handler registered via sigaction(), which is why the
+//   old BSG implementation had to unblock SIGQUIT to work (while at the same time breaking the
+//   Google handler).
+//
+// - Registering multiple handlers via sigwait() doesn't work because only the first registered
+//   handler gets called. Once the first handler runs, the signal is no longer pending. And since
+//   the first handler in this case is registered by the runtime library before our code even has
+//   a chance to run, we'd never get called.
+//
+// - Even though the runtime library has callback hooks for ANRs (art/runtime/runtime.cc), it's
+//   not a public API, so we can't use it.
+//
+// So we need a workaround. Calling raise() and signal() won't work, because the signal ends up
+// getting routed incorrectly. The only way to successfully signal the runtime handler is by use
+// of a syscall() of SYS_tgkill with SIGQUIT that targets the Google handler thread ID directly.
+//
+//
+// How to work around it:
+//
+// - Unblock SIGQUIT (breaking the Google handler)
+//
+// - Register a SIGQUIT handler via sigaction()
+//
+// - Our handler callback blocks SIGQUIT (allowing the Google handler to work again) and sets a
+//   flag like "should_report_anr"
+//
+// - We have another thread waiting on the "should_report_anr" flag, which then calls
+//   bsg_google_anr_call() after a short delay to raise a SIGQUIT on Google's handler thread.
+//   Re-raising the signal on the signal handler thread won't work because the signaling system
+//   in the OS won't finish updating the new blocking state until the current signal handler
+//   returns.
+//
+// - Do our handler stuff, delay a short while (like 2 seconds) for Google's handlers to finish,
+//   then exit our watchdog thread. It's important that our watchdog thread stops before the
+//   runtime's timeout (currently 20 seconds), or else we'll be force-killed, which breaks Google
+//   reporting and the ANR popup in some cases.
+//
+//
+// Functionality in this file:
+//
+// - bsg_google_anr_init (NOT async-safe): Find and save the Google ANR handler thread ID and the
+//   process ID, which are needed to make the syscall.
+//
+// - bsg_google_anr_call (async-safe): Make the syscall to send a SIGQUIT directly to the runtime
+//   library's ANR handler thread.
+
+#include "anr_google.h"
+
+#include <ctype.h>
+#include <dirent.h>
+#include <inttypes.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+static pid_t process_id = -1;
+static pid_t google_thread_id = -1;
+
+static bool is_thread_named_signal_catcher(pid_t tid) {
+  static const char*const SIGNAL_CATCHER_THREAD_NAME = "Signal Catcher";
+
+  bool success = false;
+  char buff[256];
+
+  snprintf(buff, sizeof(buff), "/proc/%d/comm", tid);
+  FILE* fp = fopen(buff, "r");
+  if(fp == NULL) {
+    return false;
+  }
+
+  if(fgets(buff, sizeof(buff), fp) != NULL) {
+    success = strncmp(buff, SIGNAL_CATCHER_THREAD_NAME, strlen(SIGNAL_CATCHER_THREAD_NAME)) == 0;
+  }
+
+  fclose(fp);
+  return success;
+}
+
+static bool is_thread_signal_catcher_sigblk(pid_t tid) {
+  static const uint64_t SIGNAL_CATCHER_THREAD_SIGBLK = 0x1000;
+  static const char* SIGBLK_HEADER = "SigBlk:\t";
+  const size_t SIGBLK_HEADER_LENGTH = strlen(SIGBLK_HEADER);
+
+  char buff[256];
+  uint64_t sigblk = 0;
+
+  snprintf(buff, sizeof(buff), "/proc/%d/status", tid);
+  FILE* fp = fopen(buff, "r");
+  if(fp == NULL) {
+    return false;
+  }
+
+  while(fgets(buff, sizeof(buff), fp) != NULL) {
+    if (strncmp(buff, SIGBLK_HEADER, SIGBLK_HEADER_LENGTH) == 0) {
+      sigblk = strtoull(buff + SIGBLK_HEADER_LENGTH, NULL, 16);
+      break;
+    }
+  }
+  fclose(fp);
+  return sigblk == SIGNAL_CATCHER_THREAD_SIGBLK;
+}
+
+bool bsg_google_anr_init() {
+  pid_t pid = getpid();
+  pid_t tid = -1;
+
+  char path[256];
+  snprintf(path, sizeof(path), "/proc/%d/task", pid);
+  DIR* dir = opendir(path);
+  if(dir == NULL) {
+    return false;
+  }
+
+  struct dirent *dent;
+  while((dent = readdir(dir)) != NULL) {
+    if(dent->d_name[0] < '0' || dent->d_name[0] > '9') {
+      continue;
+    }
+
+    tid = strtol(dent->d_name, NULL, 10);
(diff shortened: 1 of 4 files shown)


bugsnag-plugin-android-anr/src/main/jni/anr_google.c
```diff
@@ -0,0 +1,96 @@
...
+
+  snprintf(buff, sizeof(buff), "/proc/%d/comm", tid);
+  FILE* fp = fopen(buff, "r");
+  if(fp == NULL) {
+    return false;
+  }
+
+  if(fgets(buff, sizeof(buff), fp) != NULL) {
+    trim_string_right(buff);
+    success = strcmp(buff, SIGNAL_CATCHER_THREAD_NAME) == 0;
```
fractalwrench, line 37: Would `strncmp` work here instead for comparing the two strings? I don't quite follow why the string needs whitespace trimmed off - is that returned as part of the `fgets` call?

bugsnag-plugin-android-anr/src/main/jni/anr_google.c
```diff
@@ -0,0 +1,96 @@
...
+  uint64_t sigblk = 0;
+
+  snprintf(buff, sizeof(buff), "/proc/%d/status", tid);
+  FILE* fp = fopen(buff, "r");
+  if(fp == NULL) {
+    return false;
+  }
+
+  while(fgets(buff, sizeof(buff), fp) != NULL) {
+    if(sscanf(buff, "SigBlk: %" SCNx64, &sigblk) == 1) {
```
fractalwrench, line 57: The following warning shows up when the project is opened in Android Studio, is this anything to be concerned about?

>Clang-Tidy: 'sscanf' used to convert a string to an unsigned integer value, but function will not report conversion errors; consider using 'strtoull' instead

bugsnag-plugin-android-anr/src/main/jni/anr_google.c
```diff
@@ -0,0 +1,96 @@
...
+    return;
+  }
+
+  struct dirent *dent;
+  while((dent = readdir(dir)) != NULL) {
+    if(dent->d_name[0] < '0' || dent->d_name[0] > '9') {
+      continue;
+    }
+
+    tid = atoi(dent->d_name);
```
fractalwrench, line 82: >Clang-Tidy: 'atoi' used to convert a string to an integer value, but function will not report conversion errors; consider using 'strtol' instead

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -5,118 +5,115 @@
...
-      previous.sa_sigaction(SIGQUIT, info, user_context);
-    } else if (previous.sa_handler == SIG_DFL) {
-      // Do nothing, the default action is nothing
-    } else if (previous.sa_handler != SIG_IGN) {
-      void (*previous_handler)(int) = previous.sa_handler;
-      previous_handler(signum);
-    }
+static void* sigquit_thread_main(void* _) {
+  while(!should_report_anr) {
+    usleep(100000);
```
fractalwrench, line 62: It would be good to extract this as a constant or add some documentation about why it's been selected - the context is missing for future readers of a 10s gap being required to avoid process termination.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -5,118 +5,115 @@
...
+  sleep(2);
+  return NULL;
 }
 
-/**
- * Configure ANR detection using a signal handler listening for SIGQUIT
- */
-void *bsg_monitor_anrs(void *_arg) {
+static void handle_sigquit(int signum, siginfo_t* info, void* user_context) {
+  // TODO: Save stack trace
```
fractalwrench, line 78: Does this still need to be addressed?

fractalwrench (commented):
This appears to work well from a functional perspective which is great 🎉 

It'd be really helpful to have some more documentation in the form of inline comments, and possibly a preamble somewhere in a header comment about the general architecture of this approach. Part of this is probably down to me not having much of a C background, so it'd be good to have someone familiar with the ins & outs of that review the PR too.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
kstenerud: Thats where the direct native-to-native call to stack unwind that I've been working on in the other integration PR will go. Right now it's not tracking native traces since that's not in the next branch.

tomlongridge: In the case the TODO should be removed so as not to indicate a partially complete job in the this release.

Is it just this TODO-comment that relates to the native ANR work or is there functionality that should be parred down for this release?

tomlongridge (commented):
Can you add some further detail to the PR description and code comments to explain the rationale behind this change, and the various parts, for reviewers and future readers?

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
kstenerud: It's the place in the code where the call to get a native stack trace will eventually go. I've updated the TODO to be more explicit about that, but I could remove it altogether if it's still subject to confusion...

kstenerud:
I've put a big text blob in the code itself now, since the reasoning behind much of the code is not very clear. Do we need more descriptions in the PR or is that okay?

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
tomlongridge: I don't see that it serves any purpose, as you don't need it to remind you long term, and makes the release look incomplete. I think it should be removed.

fractalwrench (approved):
LGTM, the added docs are really great and will be very useful when looking at this in future. It'd be good to get another review from someone else with more C experience so I've provisionally added Delisa as a reviewer

kattrali (approved):

bugsnag-plugin-android-anr/src/main/jni/anr_google.c
```diff
@@ -0,0 +1,155 @@
...
+// - Blocked signals will never reach a signal handler registered via sigaction(), which is why the
+//   old BSG implementation had to unblock SIGQUIT to work (while at the same time breaking the
+//   Google handler).
+//
+// - Registering multiple handlers via sigwait() doesn't work because only the first registered
+//   handler gets called. Once the first handler runs, the signal is no longer pending. And since
+//   the first handler in this case is registered by the runtime library before our code even has
+//   a chance to run, we'd never get called.
+//
+// - Even though the runtime library has callback hooks for ANRs (art/runtime/runtime.cc), it's
```
allight, line 23: If you file a bug issuetracker.google.com we might be able to add a public API of some sort for this (though it would not be available until 'T' at the earliest). cc me (allight@google.com) on the bug if you do end up filing one.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -5,118 +5,128 @@
...
+static void notify_anr_detected() {
   if (enabled) {
-    // invoke a JNI call from the SIGQUIT handler
     JNIEnv *env;
-    int result = (*bsg_jvm)->GetEnv(bsg_jvm, (void **)&env, JNI_VERSION_1_4);
+    int result = (*bsg_jvm)->GetEnv(bsg_jvm, (void **) &env, JNI_VERSION_1_4);
 
     if (result == JNI_OK) { // already attached
-      bsg_notify_anr_detected(env);
+      (*env)->CallVoidMethod(env, obj_plugin, mthd_notify_anr_detected);
```
allight, line 49: This is dead code (and illegal if it ever becomes live).

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -5,118 +5,128 @@
...
+static void handle_sigquit(int signum, siginfo_t* info, void* user_context) {
+  // Re-block SIGQUIT so that the Google handler can trigger
+  sigset_t sigmask;
+  sigemptyset(&sigmask);
+  sigaddset(&sigmask, SIGQUIT);
+  pthread_sigmask(SIG_BLOCK, &sigmask, NULL);
+  sigaction(SIGQUIT, &original_sigquit_handler, NULL);
+
+  // Instruct our watchdog thread to report the ANR and also call Google
+  should_report_anr = true;
```
allight, line 94: pthread_cond_broadcast ?

bugsnag-plugin-android-anr/src/main/jni/anr_google.c
kstenerud: Awesome! We'll raise an issue about it :)

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
kstenerud: Could you please explain more about what's wrong? This is following recommendations from https://stackoverflow.com/a/12901159 to handle all edge cases of the environment's attached state.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -5,118 +5,128 @@
...
-      previous.sa_sigaction(SIGQUIT, info, user_context);
-    } else if (previous.sa_handler == SIG_DFL) {
-      // Do nothing, the default action is nothing
-    } else if (previous.sa_handler != SIG_IGN) {
-      void (*previous_handler)(int) = previous.sa_handler;
-      previous_handler(signum);
-    }
+  // Wait until our SIGQUIT handler is ready for us to start
+  while(!should_report_anr) {
+    usleep(delay_100ms);
```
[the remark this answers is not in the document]
kstenerud: pthread functions are unfortunately not async-safe, so we can't use them to signal from the SIGQUIT handler.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
allight: As I said in bug https://github.com/bugsnag/bugsnag-android/issues/1026 attempting to make any JNI calls from a signal handler when the thread was running java or runtime code can put the runtime into an illegal state.

kstenerud: Yes, thanks for that, we took it to heart! This code gets called from the thread running `sigquit_watchdog_thread_main`, which waits on `should_report_anr` and then calls `notify_anr_detected()`. The SIGQUIT handler sets `should_report_anr` to true and then exits (primitive, but async-safe). So no Java code is called from the signal handler's thread. Unfortunately the diff here is horribly messy; the finished code is a lot easier to follow: https://github.com/bugsnag/bugsnag-android/blob/1552f7d58396fca8b99630516487641a901bfb7a/bugsnag-plugin-android-anr/src/main/jni/anr_handler.c#L43

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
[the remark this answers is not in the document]
allight: Yeah you're right I misread the signal block you have. You could still use sem_wait and either be careful about checking EINTR or use the same condition loop, the way bionic [implements semaphores](https://cs.android.com/android/platform/superproject/+/master:bionic/libc/bionic/semaphore.cpp;l=214?q=sem_wait&ss=android%2Fplatform%2Fsuperproject&start=11) means it will be safe in practice on android. Up to you though. This sort of wait loop will likely cause apps to use significantly more power and possibly put them at greater risk of being killed in the background for waking up the CPU.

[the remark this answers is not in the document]
kstenerud: Good to know! Better to not busy wait if it can be avoided :)

bugsnag-plugin-android-anr/src/main/jni/anr_google.c
kstenerud: I've filed an issue here: https://issuetracker.google.com/issues/178107277

Unfortunately it won't let me CC you.

allight: No problem. Thanks for filing it.

kstenerud: Looks like they're not interested in fixing it? https://issuetracker.google.com/issues/178107277#comment3

allight: Odd, the internal bug is still open. I think it was closed as not going into S.

PR #1079, opened 2021-01-14
Disable failing test on Android 5

fractalwrench:
## Goal

Disables a test which was failing on Android 5, which was missed due to the presence of a soft fail. This will remain disabled until PLAT-5606 is addressed. This seems to be the only genuine failure in a [full CI run](https://buildkite.com/bugsnag/bugsnag-android/builds/3197#_) at the moment.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -433,7 +433,7 @@ steps:
           - "--app=/app/build/release/fixture-r16.apk"
           - "--farm=bs"
           - "--device=ANDROID_4_4"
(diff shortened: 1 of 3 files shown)


twometresteve:
The change here looks fine, but it seems that I made a mistake in an earlier PR that's led to no scenarios being run in the "Flaky" steps.  The single quotes in the pipeline yaml: `- "--tags='@Flaky'"` should not have been there, as they are interpreted as part of the `--tags` option themselves.

In the interest of making progress, I've pushed a small commit to correct that here.

twometresteve (approved):

PR #1081, opened 2021-01-17
Review scenarios considered to be flaky

twometresteve:
## Goal

Takes a fresh look at the scenarios we consider to be flaky, annotating them appropriately.

The aim is to ring-fence all flaky scenarios, avoiding the need for soft build failures and the repetitive coverage of Android 8.0 and 8.1.  However, it is possible that further flakes will show themselves.

## Changeset

In addition to reworking the tags, I've set MazeRunner to use its resilient appium driver, which often has the ability to rescue tests that suffer from an appium session loss.

## Testing

Cover by CI.


.buildkite/pipeline.full.yml
```diff
@@ -213,11 +212,9 @@ steps:
           - "--fail-fast"
           - "--tags"
           - "not @Flaky"
+          - "--resilient"
```
fractalwrench, line 215: Is there any way this could be the default setting in mazerunner? Otherwise there'll be a divergence in behaviour when running with BrowserStackLocal & flakes will be more likely

features/full_tests/batch_1/in_foreground.feature
```diff
@@ -7,7 +7,8 @@ Scenario: Test handled exception after delay
     And the event "app.inForeground" is true
 
 # Skip due to an issue on later Android platforms - [PLAT-5464]
-@Flaky @skip_android_11 @skip_android_10
+# Possible @Flaky
```
fractalwrench, line 10: Is this scenario flaky?

fractalwrench (commented):
This mostly LGTM - for approval it'd be interesting to know how many times these changes were run on CI. Although the changes look good I'm a bit reticent to merge until there's evidence that all the scenarios are unlikely to flake (e.g. by passing 3 times in a row)

.buildkite/pipeline.full.yml
twometresteve: I had the same thought myself and was probably waiting to see how it performs on Android before making that leap.  It's generally much better having in than not on Cocoa, but logged errors can be harder to interpret when things go wrong.  Can we see how it runs for a bit and go from there?  We might decide we don't like the option at all on Android.

features/full_tests/batch_1/in_foreground.feature
twometresteve: Possibly.  😁  I think I left that in by mistake - now removed.

twometresteve:
> This mostly LGTM - for approval it'd be interesting to know how many times these changes were run on CI. Although the changes look good I'm a bit reticent to merge until there's evidence that all the scenarios are unlikely to flake (e.g. by passing 3 times in a row)

I had done this during development of the changes, but I've just pushed a temporary commit to disable the fast fails, in order to perform some soak testing.  I've also set up a schedule to (full) build this branch every two hours (overnight), so we should get a good picture by tomorrow.

fractalwrench (approved):

PR #1086, opened 2021-01-21
Do not use the resilient appium driver

twometresteve:
## Goal

Remove use of the Resilient Appium driver.

## Design

There's a bug in MazeRunner's Resilient Appium Driver that makes it incompatible with the way that the Android test run.  It erroneously treats a `NoSuchElement` exception, which is "normal" in the Android setup, as a broken Appium session.  The simplest thing to do at this stage is to just stop using the flag.

## Changeset

- Stop using the resilient drive
- Use the official MazeRunner release image
- Remove `@Flaky` tag from scenario, as it was down to the resilient driver and not the scenario.

## Testing

Covered by running a full CI.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -58,7 +58,6 @@ steps:
   - label: ':android: Android 7 Instrumentation tests'
     depends_on:
       - "android-ci"
-      - "NDK-smoke-test"
     timeout_in_minutes: 30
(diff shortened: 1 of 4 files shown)


fractalwrench (approved):
LGTM pending CI

twometresteve:
Review re-requested having removed several commits for the ANR related changes.  These will be put into the ANR integration branch we have instead.

fractalwrench (approved):
LGTM pending CI

PR #1088, opened 2021-01-22
Check internal JNI calls for pending exceptions and no-op

fractalwrench:
## Goal

Makes safe all invocations of `GetMethodId`, `GetStaticMethodId`, and `FindClass` within the NDK module. This is achieved by calling [ExceptionClear](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#exceptionclear) to remove any pending JVM exceptions that could lead to app termination. The return value of all these invocations is checked for `NULL`, and no-ops if this is the case.

## Changeset

- Replaced `FindClass` with `bsg_safe_find_class` (and equivalents for `GetMethodId/GetStaticMethodId`)
- When loading a class, check if there was an exception in the JVM, and if so clear it and return `NULL`. The following errors can be thrown. As these are not recoverable, Bugsnag should no-op rather than crashing the process:
  - `NoSuchMethodError`
  - `ExceptionInInitializerError`
  - `OutOfMemoryError`
  - `ClassFormatError`
  - `ClassCircularityError`
  - `NoClassDefFoundError`
- Alter all invocations to check the return value for `NULL` and no-op in this case
- Added inline comments to explain what each JNI call is doing, which should improve readability (this comes from a previous review comment and it seemed an opportune moment)

## Testing

## Remaining work

The ANR plugin has not been converted to use these safe calls yet as it cannot directly access the methods. There are also several other JNI calls which can throw errors that Bugsnag does not currently handle. Both these will be addressed in separate changesets.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/metadata.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/metadata.c
@@ -1,4 +1,5 @@
 #include "metadata.h"
+#include "safejni.h"
 #include "utils/string.h"
 #include <malloc.h>
 #include <string.h>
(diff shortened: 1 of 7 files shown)


bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -40,58 +41,207 @@ void bsg_populate_metadata_value(JNIEnv *env, bugsnag_metadata *dst,
...
+  // lookup NativeInterface.getMetadata()
+  jni_cache->get_metadata = bsg_safe_get_static_method_id(
       env, jni_cache->native_interface, "getMetadata", "()Ljava/util/Map;");
-  jni_cache->get_context = (*env)->GetStaticMethodID(
+  if (jni_cache->get_metadata == NULL) {
+    return NULL;
+  }
+
+  // lookup NativeInterface.getContext()
+  jni_cache->get_context = bsg_safe_get_static_method_id(
```
kstenerud, line 236: As a future improvement, we could implement the cache as a C++ map, with `bsg_safe_get_static_method_id` and friends checking the cache, and performing the lookup + adding to the cache if not found. Then we wouldn't need all of this initialization code. Classes and methods are forever objects so we're not taking up any extra memory really. We could even put in warning log messages at the lookup function level so that we get logs whenever something like this fails (because it should never fail in properly working code, and without logs we wouldn't know except indirectly when things that should happen, don't).

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -40,58 +41,207 @@ void bsg_populate_metadata_value(JNIEnv *env, bugsnag_metadata *dst,
...
       env, jni_cache->native_interface, "getContext", "()Ljava/lang/String;");
+  if (jni_cache->get_context == NULL) {
+    return NULL;
+  }
   return jni_cache;
 }
 
+// FIXME problematic global ref?
```
kstenerud, line 244: What is the issue? From the looks of it, it creates a key, performs a lookup, and returns the result. It's on the caller to not attempt to store the result beyond the NDK function call.

kstenerud (approved):

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
fractalwrench: I don't think there is an issue - this was down to a misunderstanding that only functions marked with `JNI_EXPORT` create local references to objects.

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
fractalwrench: Created PLAT-5813 to track this.

PR #1090, opened 2021-01-25
Re-add the ANR NDK traces that were removed in the last merge

kstenerud:
## Goal

PLAT-5732

The recent merge of next into the ANR integration branch necessarily clobbered some of the newer native ANR code. This PR corrects the issue.

## Changeset

Native stack traces are now generated entirely in native code via a direct call to the NDK plugin from the ANR plugin (provided it is available). The function pointer is injected if the NDK plugin is loaded. The native trace is passed as a Java list of entries.

## Testing

The JVM ANR e2e test has been split into two: One that stalls the thread using Thread.sleep(), and one that uses an endless loop. They generate very different looking stack traces due to sleep calling a system-generated JNI bridge, and need to both be tested.


bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -9,31 +9,56 @@ internal class AnrPlugin : Plugin {
...
+            // There are two known telltale signs in a native trace to indicate that only Java code
+            // is involved and the native trace is a red herring:
+            //
+            // - If it's in 100% Java land, the first native frame is "empty" in that every field
+            //   is blank except for frame address.
+            //
+            // - If it's in a system-generated native bridge such as from Thread.sleep(), the trace
+            //   will contain reserved methods starting with "_Z", for example
+            //   "_ZN3art17ConditionVariable9TimedWaitEPNS_6ThreadEli"
+            //
```
kstenerud, line 49: Actually, I think I may have a better way to detect real native traces... Stay tuned!

bugsnag-plugin-android-anr/src/test/java/com/bugsnag/android/AnrPluginTest.kt
```diff
@@ -1,37 +0,0 @@
-package com.bugsnag.android
-
-import org.junit.Assert.assertEquals
-import org.junit.Test
-
-class AnrPluginTest {
```
fractalwrench, line 6: Has equivalent test coverage been moved elsewhere, or does this lose unit test coverage? Ideally it'd be good to maintain or increase test coverage unless it's absolutely unavoidable.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -228,3 +228,8 @@ void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message,
   (*env)->DeleteLocalRef(env, type_class);
   (*env)->DeleteLocalRef(env, interface_class);
 }
+
+ssize_t bsg_unwind_stack_default(bugsnag_stackframe stacktrace[BUGSNAG_FRAMES_MAX],
```
fractalwrench, line 232: Would be nice to add some quick docs here to explain how the function is accessed for any future travellers

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -9,31 +9,53 @@ internal class AnrPlugin : Plugin {
...
+            return stackTrace.first().isNativeMethod
+        }
 
-            val nativeAnrIndex = notifyIndex + 1
-            if (nativeAnrIndex >= stackTrace.size || !stackTrace[nativeAnrIndex].isNativeMethod) {
-                return -1
+        internal fun hasReservedNativeFrames(stackTrace: List<NativeStackframe>): Boolean {
+            val index = stackTrace.indexOfFirst { frame ->
+                val method = frame.method
+                method != null && method.startsWith("_Z")
```
fractalwrench, line 22: May be worth extracting "_Z" into a constant variable and adding a quick comment about its significance - I initially missed the longer comment below and got confused about what this was.

You can create a constant in Kotlin using either of the following approaches: https://stackoverflow.com/a/46729715

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -46,6 +68,16 @@ internal class AnrPlugin : Plugin {
         }
 
         if (loaded) {
+            @Suppress("UNCHECKED_CAST")
+            val clz = Class.forName("com.bugsnag.android.NdkPlugin") as Class<Plugin>
```
fractalwrench, line 72: This needs to catch a `ClassNotFoundException` if the `NdkPlugin` is not present - otherwise the ANR plugin will fail to load

bugsnag-plugin-android-anr/src/test/java/com/bugsnag/android/AnrPluginTest.kt
kstenerud: It's covered in the native and JVM ANR e2e tests. These tests were testing against a particular implementation, which made them very brittle to change. The e2e test checks against external results, which is resilient.

fractalwrench (approved):
LGTM pending CI

PR #1091, opened 2021-01-26
Check internal JNI calls for pending exceptions and no-op

fractalwrench:
## Goal

Checks additional JNI calls in the NDK module for pending exceptions and no-ops, building on #1088. This is achieved by calling [ExceptionClear](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#exceptionclear) to remove any pending JVM exceptions that could lead to app termination.

## Changeset

Added safe versions of the following methods and updated code to check the return value where appropriate:

```
NewStringUTF
CallBooleanMethod
CallDoubleMethod
CallFloatMethod
CallIntMethod
NewByteArray
SetByteArrayRegion
NewObjectArray
SetObjectArrayElement
GetObjectArrayElement
```

Additionally fixes a (unlikely) deadlock introduced in #1088 in `Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath`

## Testing

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -148,45 +148,47 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
   pthread_mutex_lock(&bsg_native_delivery_mutex);
   const char *event_path = (*env)->GetStringUTFChars(env, _report_path, 0);
   bugsnag_event *event = bsg_deserialize_event_from_file((char *)event_path);
+  jbyteArray jpayload = NULL;
+  jbyteArray jstage = NULL;
+  char *payload = NULL;
 
   if (event != NULL) {
-    char *payload = bsg_serialize_event_to_json_string(event);
+    payload = bsg_serialize_event_to_json_string(event);
     if (payload != NULL) {
 
       // lookup com/bugsnag/android/NativeInterface
       jclass interface_class =
           bsg_safe_find_class(env, "com/bugsnag/android/NativeInterface");
       if (interface_class == NULL) {
-        return;
+        goto exit;
       }
 
       // lookup NativeInterface.deliverReport()
       jmethodID jdeliver_method = bsg_safe_get_static_method_id(
           env, interface_class, "deliverReport", "([B[BLjava/lang/String;)V");
       if (jdeliver_method == NULL) {
-        return;
+        goto exit;
       }
 
-      size_t payload_length = bsg_strlen(payload);
-      jbyteArray jpayload = (*env)->NewByteArray(env, payload_length);
-      (*env)->SetByteArrayRegion(env, jpayload, 0, payload_length,
-                                 (jbyte *)payload);
+      // generate payload bytearray
+      jpayload = bsg_byte_ary_from_string(env, payload);
+      if (jpayload == NULL) {
+        goto exit;
+      }
 
-      size_t stage_length = bsg_strlen(event->app.release_stage);
-      jbyteArray jstage = (*env)->NewByteArray(env, stage_length);
-      (*env)->SetByteArrayRegion(env, jstage, 0, stage_length,
-                                 (jbyte *)event->app.release_stage);
+      // generate releaseStage bytearray
+      jstage = bsg_byte_ary_from_string(env, event->app.release_stage);
+      if (jstage == NULL) {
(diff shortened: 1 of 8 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -157,30 +157,41 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
...
       if (interface_class == NULL) {
+        pthread_mutex_unlock(&bsg_native_delivery_mutex);
         return;
       }
 
       // lookup NativeInterface.deliverReport()
       jmethodID jdeliver_method = bsg_safe_get_static_method_id(
           env, interface_class, "deliverReport", "([B[BLjava/lang/String;)V");
       if (jdeliver_method == NULL) {
+        pthread_mutex_unlock(&bsg_native_delivery_mutex);
```
kstenerud, line 168: All of these mutex unlocks make me nervous. All it takes is a future modification that fails to cover this and we get deadlocks :/

Can we do a C-style RAII setup (i.e. `goto success` and `goto fail` which handle cleanup)

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -548,12 +568,14 @@ void bsg_populate_metadata_value(JNIEnv *env, bugsnag_metadata *dst,
                                  bsg_jni_cache *jni_cache, char *section,
                                  char *name, jobject _value) {
   if ((*env)->IsInstanceOf(env, _value, jni_cache->number)) {
-    double value =
-        (*env)->CallDoubleMethod(env, _value, jni_cache->number_double_value);
+    // add a boolean metadata value
+    double value = bsg_safe_call_double_method(env, _value,
```
kstenerud, line 572: This is double, not boolean

kstenerud (changes requested):

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
fractalwrench: Updated to use a RAII setup for cleanup.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -157,30 +157,37 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
...
-                                   jstage, jpayload, japi_key);
+      // call NativeInterface.deliverReport()
+      jstring japi_key = bsg_safe_new_string_utf(env, event->api_key);
+      if (japi_key != NULL) {
+        (*env)->CallStaticVoidMethod(env, interface_class, jdeliver_method,
+                                     jstage, jpayload, japi_key);
+      }
       (*env)->DeleteLocalRef(env, japi_key);
+
       (*env)->ReleaseByteArrayElements(env, jpayload, (jbyte *)payload,
```
kstenerud, line 191: These should also be in the cleanup code at the end. They could be live objects when the code fails later in the function.

kstenerud (changes requested):

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -196,7 +198,20 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
...
+    pthread_mutex_unlock(&bsg_native_delivery_mutex);
+    if (jpayload != NULL) {
+      (*env)->ReleaseByteArrayElements(env, jpayload, (jbyte *)payload,
+                                       0); // <-- frees payload
+    }
+    if (jstage != NULL) {
+      (*env)->ReleaseByteArrayElements(
+          env, jstage, (jbyte *)event->app.release_stage, JNI_COMMIT);
+    }
+    (*env)->DeleteLocalRef(env, jpayload);
```
kstenerud, line 213: will this succeed if `jpayload` is NULL?

fractalwrench: Yep. The call has no effect if the reference is NULL: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#deletelocalref

kstenerud (approved):

PR #1092, opened 2021-01-26
Handle remaining JNI calls error handling

fractalwrench:
## Goal

Checks additional JNI calls in the NDK module for pending exceptions and no-ops, building on #1088. This is achieved by calling [ExceptionClear](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#exceptionclear) to remove any pending JVM exceptions that could lead to app termination.

## Changeset

Added safe versions of the following methods and updated code to check the return value where appropriate:

```
GetStaticFieldID
GetStaticObjectField
NewObject
CallObjectMethod
CallStaticVoidMethod
CallStaticObjectMethod
```

## Testing

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -54,12 +54,14 @@ jfieldID bsg_parse_jseverity(JNIEnv *env, bugsnag_severity severity,
                              jclass severity_class) {
   const char *severity_sig = "Lcom/bugsnag/android/Severity;";
   if (severity == BSG_SEVERITY_ERR) {
-    return (*env)->GetStaticFieldID(env, severity_class, "ERROR", severity_sig);
+    return bsg_safe_get_static_field_id(env, severity_class, "ERROR",
+                                        severity_sig);
   } else if (severity == BSG_SEVERITY_WARN) {
-    return (*env)->GetStaticFieldID(env, severity_class, "WARNING",
-                                    severity_sig);
+    return bsg_safe_get_static_field_id(env, severity_class, "WARNING",
+                                        severity_sig);
   } else {
-    return (*env)->GetStaticFieldID(env, severity_class, "INFO", severity_sig);
+    return bsg_safe_get_static_field_id(env, severity_class, "INFO",
+                                        severity_sig);
   }
 }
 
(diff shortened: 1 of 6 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -274,11 +291,20 @@ void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message,
...
+  // get breadcrumb type fieldID
+  jfieldID crumb_type = bsg_parse_jcrumb_type(env, type, type_class);
+  if (crumb_type == NULL) {
+    return;
+  }
+
+  // get the breadcrumb type
+  jobject jtype = bsg_safe_get_static_object_field(env, type_class, crumb_type);
+  if (jtype == NULL) {
+    return;
```
kstenerud, line 303: Will `crumb_type` leak here?

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -137,28 +139,41 @@ void bugsnag_notify_env(JNIEnv *env, char *name, char *message,
...
     }
+
+    // create StackTraceElement object
     jobject jframe =
-        (*env)->NewObject(env, trace_class, trace_constructor, class, method,
-                          filename, frame.line_number);
+        bsg_safe_new_object(env, trace_class, trace_constructor, class, method,
+                            filename, frame.line_number);
+    if (jframe == NULL) {
+      return;
```
kstenerud, line 148: `trace` will leak here

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -137,28 +139,41 @@ void bugsnag_notify_env(JNIEnv *env, char *name, char *message,
...
+  // get the severity field
+  jfieldID severity_field = bsg_parse_jseverity(env, severity, severity_class);
+  if (severity_field == NULL) {
+    return;
+  }
+  // get the error severity object
+  jobject jseverity =
+      bsg_safe_get_static_object_field(env, severity_class, severity_field);
+  if (jseverity == NULL) {
+    return;
```
kstenerud, line 166: These will leak things as well. This should probably be a c-style RAII

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -360,16 +360,23 @@ void bsg_populate_crumb_metadata(JNIEnv *env, bugsnag_breadcrumb *crumb,
...
-      (*env)->CallObjectMethod(env, metadata, jni_cache->map_key_set);
-  jobject keylist = (*env)->NewObject(
+      bsg_safe_call_object_method(env, metadata, jni_cache->map_key_set);
+  if (keyset == NULL) {
+    return;
+  }
+  jobject keylist = bsg_safe_new_object(
       env, jni_cache->arraylist, jni_cache->arraylist_init_with_obj, keyset);
+  if (keylist == NULL) {
+    return;
```
kstenerud, line 372: These will leak things. Should be c-style RAII

kstenerud (changes requested):
There are many functions that have existing/new leak issues that can be fixed with c-style RAII. We should probably just convert everything over once and for all...

fractalwrench:
Thanks @kstenerud - I've updated the changeset so that the functions use RAII where it seemed appropriate.

kstenerud (approved):
LGTM!

PR #1098, opened 2021-01-28
Update ANR tests to run on Samsung devices

twometresteve:
## Goal

Update the mechanism by which ANRs are generated in tests, in order that they run successfully on Samsung devices (which the previous mechanism did not).

## Design

The idea and code for generating ANRs using a deadlock is taken directly from [this post](https://stackoverflow.com/a/45587343/15097391).  With the scenarios running on Samsung devices we no longer need to skip Android 8.1 (it was only coincidence that our 8.1 device is a Samsung).  It also opens the door to running scenarios on alternative Android 5/6 devices, increasing our diversity and resilience to things like unavailable devices.

## Changeset

- New `AnrKelper.kt` to centralize the ANR mechanism.
- Tests no longer skipped on Android 8.1
- Clearing of error popups removed - it's unnecessary and I found that trying to press buttons during an ANR can cause Appium failures and flaking tests.

## Testing

Covered by a full CI run and I've tested it locally several times on Android 5 and 8 Samsung devices.

### DIFF
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/JvmAnrSleepScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/JvmAnrSleepScenario.kt
@@ -1,9 +1,8 @@
 package com.bugsnag.android.mazerunner.scenarios
 
 import android.content.Context
-import android.os.Handler
-import android.os.Looper
 import com.bugsnag.android.Configuration
+import com.bugsnag.android.mazerunner.createDeadlock
 
 /**
  * Stops the app from responding for a time period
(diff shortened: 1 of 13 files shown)


features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/JvmAnrSleepScenario.kt
```diff
@@ -21,12 +20,6 @@ internal class JvmAnrSleepScenario(
 
     override fun startScenario() {
         super.startScenario()
-        val main = Handler(Looper.getMainLooper())
-        main.postDelayed(
-            Runnable {
-                Thread.sleep(100000)
```
kstenerud, line 27: This scenario is designed specifically to test how we deal with stack traces of syscall-generating Java methods like `Thread.sleep()`. Changing it to something else will render it useless, unfortunately.

twometresteve: Ah, ok - good spot.  I'll change it back and add a comment to that effect so I don't swoop in another day and try to put it back!  If it's the only such scenario it's far from the end of the world and I'll try to put some smarter logic in for the `skip`.

kstenerud (approved):

PR #1101, opened 2021-02-01
Only attempt to clear error dialogs if element is not found

twometresteve:
## Goal

React to system dialogs popping up in end-to-end tests rather than always trying to dismiss them.

## Design

There were a few problems with the previous approach of always attempting to dismiss dialogs, whether they were there or not:
- The 3s delay designed to wait for dialogs to appear adds a lot of time to the overall run when it's not always necessary.
- Scenarios could still flake when dialogs appeared after the 3s.
- The mechanism appears completely unnecessary on Android 7+, where the `Intent` broadcast on app start can be seen to successfully dismiss dialogs.
- Attempting to click button that aren't there generates a lot of noise in the Appium logs.

If flakes still occasionally occur, we can tweak the parameters and retries. With further analysis we can also optimise the delays further - for example, rather than a 5s delay with a single retry is would probably be better to timeout after 1s and try up to 5 times.

## Changeset

New cucumber step introduced that assumes a failure to find an element is due to the presence of a system dialog hogging focus.  If will attempt to close the system dialog by pressing the button, before making a single retry.

## Testing

In the course of analysing the behaviour and developing the change I performed several runs locally and full builds on CI.

### DIFF
--- a/features/steps/android_steps.rb
+++ b/features/steps/android_steps.rb
@@ -1,6 +1,24 @@
+# Waits 5s for an element to be present.  If it isn't assume a system error dialog is
+# blocking its view and dismiss it before trying once more.
+#
+# @step_input element_id [String] The element to wait for
+When('any dialog is cleared and the element {string} is present') do |element_id|
+  present = Maze.driver.wait_for_element(element_id, timeout = 5)
+  next if present
+
+  clicked = click_if_present('android:id/button1') ||
+            click_if_present('android:id/aerr_close') ||
+            click_if_present('android:id/aerr_restart')
+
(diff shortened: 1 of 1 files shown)


fractalwrench (approved):
LGTM pending CI. It'd be good to create a ticket to come up with the more sophisticated mechanism of retrying ~5 times, as that'd help cut down wait times even further.

twometresteve:
> LGTM pending CI. It'd be good to create a ticket to come up with the more sophisticated mechanism of retrying ~5 times, as that'd help cut down wait times even further.

Done - PLAT-5902.

PR #1104, opened 2021-02-02
Only use isNative for detecting native stack transitions

kstenerud:
## Goal

The current method of native trace detection doesn't work anymore on Android 11.

## Design

Use only the built-in `isNative` call to detect native frames.

## Testing

Re-ran e2e tests

### DIFF
--- a/bugsnag-plugin-android-anr/src/main/jni/bsg_unwind.h
+++ b/bugsnag-plugin-android-anr/src/main/jni/bsg_unwind.h
@@ -1,1 +1,16 @@
-../../../../bugsnag-plugin-android-ndk/src/main/jni/bsg_unwind.h
+#ifndef BSG_UNWIND_H
+#define BSG_UNWIND_H
+
+#include "../assets/include/event.h"
+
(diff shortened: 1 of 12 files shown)


bugsnag-plugin-android-anr/src/main/jni/bsg_unwind.h
```diff
@@ -1 +0,0 @@
-../../../../bugsnag-plugin-android-ndk/src/main/jni/bsg_unwind.h
```
fractalwrench, line 1: is this intentional? I thought this was required to access restricted functions in the ANR module.

fractalwrench (approved):

bugsnag-plugin-android-anr/src/main/jni/bsg_unwind.h
kstenerud: It was the only way to keep the definition in one file, because it's used by the internal API for stack unwind.

I had to link it, otherwise we'd have to make two copies of it and remember to keep them in sync.

PR #1106, opened 2021-02-02
Improve handling of system dialogs in end-to-end tests

twometresteve:
## Goal

Reduce end-to-end test run times by handling system error dialogs smarter.

## Design

Reworks the Cucumber step that's run after a crash so that it times out much quicker (1s), making the assumption that if an element is not visible in that time it's because of a system error dialog taking the focus.  This process will be repeated up to 15 times, equivalent to the default timeout for `find_element`.

## Changeset

Cucumber step reworked, MazeRunner version bumped for local running.

## Testing

I ran 6 full builds overnight and clearing of the error dialog can be seen working well.  Looking at a couple of Android 5/6 runs, the dialog was dismissed ~40 times, saving at least a couple of minutes on each.

It's worth noting that the Android 5 tests are currently set to `soft_fail` and there is a strange (but existing) issue where the app appears to crash at points where we could not expect (e.g. simply by entering text into a field).  This could be an issue with Appium, but needs further investigation for which I have raised an internal ticket.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source "https://rubygems.org"
 #gem 'bugsnag-maze-runner', path: '../maze-runner'
 
 # Or a specific release:
-gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v4.3.0'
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM pending CI

PR #1108, opened 2021-02-02
Fix flaky session smoke test

fractalwrench:
## Goal

Removes a source of flakiness from the `ManualSessionSmokeScenario`, which suddenly started failing on Android 4.4 devices only.

The source of flakiness was introduced by setting an `InterceptingDelivery` in the constructor parameter. The `Scenario` is always constructed before Bugsnag is initialized, which meant on the second launch that although the crashy code should not be invoked, the `InterceptingDelivery` will notify/start sessions when cached payloads are sent from disk.

The fix has been to remove the wait from the mazerunner steps and to disable delivery interception on the second launch.

It's unclear why this has suddenly started failing on Android 4 - I'd hypothesise that these older devices are more sluggish and therefore take longer to respond to Appium commands, giving enough time for extra network requests to make it through.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
@@ -180,6 +180,8 @@ class SessionTracker extends BaseObservable {
                                     logger.w("Dropping invalid session tracking payload");
                                     break;
                                 case DELIVERED:
+                                    logger.d("Sent 1 new session to Bugsnag");
+                                    break;
                                 default:
                                     break;
                             }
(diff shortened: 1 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -280,6 +282,7 @@ void flushStoredSession(File storedFile) {
         switch (deliveryStatus) {
             case DELIVERED:
                 sessionStore.deleteStoredFiles(Collections.singletonList(storedFile));
+                logger.d("Sent 1 new session to Bugsnag");
```
twometresteve, line 285: Will it always be precisely 1 session sent here?

fractalwrench: yes - `Collections.singletonList()` can only hold one object.

twometresteve: D'oh, sorry - I missed that!

twometresteve (approved):
LGTM pending CI

PR #1111, opened 2021-02-03
Enable full obfuscation in test fixture

fractalwrench:
## Goal

Enables obfuscation in the test fixture and updates assertions around lineNumbers where necessary. This ensures that the tests run against a minified and obfuscated bugsnag SDK, which should highlight any failures in ProGuard keep rules for the SDK.

## Testing

This can be verified by inspecting the classes.dex of the fixture APK on Buildkite in Android Studio's APK analyzer.

### DIFF
--- a/features/smoke_tests/handled.feature
+++ b/features/smoke_tests/handled.feature
@@ -19,7 +19,8 @@ Scenario: Notify caught Java exception with default configuration
     And the error payload field "events.0.exceptions.0.stacktrace" is a non-empty array
     And the event "exceptions.0.stacktrace.0.method" ends with "HandledJavaSmokeScenario.startScenario"
     And the exception "stacktrace.0.file" equals "HandledJavaSmokeScenario.java"
-    And the event "exceptions.0.stacktrace.0.lineNumber" equals 51
+    # R8 minification alters the lineNumber, see the mapping file/source code for the original value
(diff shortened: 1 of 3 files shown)


features/smoke_tests/handled.feature
```diff
@@ -19,7 +19,7 @@ Scenario: Notify caught Java exception with default configuration
     And the error payload field "events.0.exceptions.0.stacktrace" is a non-empty array
     And the event "exceptions.0.stacktrace.0.method" ends with "HandledJavaSmokeScenario.startScenario"
     And the exception "stacktrace.0.file" equals "HandledJavaSmokeScenario.java"
-    And the event "exceptions.0.stacktrace.0.lineNumber" equals 51
+    And the event "exceptions.0.stacktrace.0.lineNumber" equals 6
```
twometresteve, line 22: I wonder if a comment above each step, or just at the top of each file might help people not in the R8 know understand where the line number comes from.  E.g. "See mapping.txt for R8 line number mappings"?

twometresteve (approved):
LGTM.  My only suggest is a comment to reduce the chance of people bothering you with questions about line number mappings.

features/smoke_tests/handled.feature
fractalwrench: Updated with a comment

twometresteve (approved):
LGTM

PR #1112, opened 2021-02-03
Refactor shared build logic into internal plugin

fractalwrench:
## Goal

Refactors shared build logic to use an internal gradle plugin, to avoid the use of `afterEvaulate` when configuring the build.

## Changeset

Refactored shared build logic so that it is supplied from a build plugin in the `buildSrc` directory, as [recommended by Gradle](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html) for multi-module projects. This stops the app from applying shared build logic in an `afterEvaluate` block in `subprojects {}`, which is [discouraged](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration).

This [buildSrc plugin](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) is responsible for:

- Applying 3rd party plugins used by every module (such as kotlin/license checking/static analysis)
- Adding 3rd party dependencies to each module
- Configuring compilation, test, and static analysis defaults
- Configures 3rd party plugins using Groovy build scripts stored under the `gradle` directory (in future these can be migrated to take advantage of the typesafety of the Kotlin DSL)

The `BugsnagBuildPluginExtension` allows for this internal build plugin to be configured as appropriate for each module. For example, configuration related to the NDK can be switched off by setting `usesNdk` to false. This has greatly cut down on the amount of duplication in module build.gradle files and fixed some inconsistencies in how the project was building.

### Related changes

- Removed redundant NDK version check from release script now that this is enforced via [android.ndkVersion](https://developer.android.com/studio/projects/install-ndk#apply-specific-version)
- Replaced single-quote string with double-quotes for consistency, and in preparation for any future migration of build scripts from Groovy to Kotlin
- Extracted project dependency versions to `Versions` object, which allows any version to be updated from one central location
- Applied the ktlint plugin to the bugsnag-android-core module and temporarily disabled the task (previously the plugin was not applied so there are multiple style violations)

## Testing

Relied on existing E2E test coverage. Manually confirmed that the AAR contains ProGuard rules, the POM is formed correctly, and that the expected class files/SO files are present.


build.gradle
```diff
@@ -39,59 +35,8 @@ allprojects {
...
-    defaultConfig.testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
-
-    proj.android.testOptions.unitTests.all {
-        testLogging {
-            events "passed", "skipped", "failed", "standardOut", "standardError"
-            outputs.upToDateWhen {false}
-            showStandardStreams = true
-            info.events = ["passed", "skipped", "failed", "standardOut", "standardError"]
-        }
-    }
```
tomlongridge, line 71: This configuration doesn't seem to be present in the new plugin - is that intentional?

fractalwrench: Yes this is intentional. The logs now only log when something has failed which reduces console spam and also seems to have sped up the task. Setting `outputs.upToDateWhen` to `false` also invalidates Gradle build caching which slows things down further. IMO none of this is required anymore.

tomlongridge (approved):

PR #1115, opened 2021-02-04
Trim stack gathering frames out of native ANR traces

kstenerud:
## Goal

We don't want to have the stack unwinding functions in the native portion of ANR traces; we want it to point to the actual culprit.

## Design

Trim the top 5 entries from the native trace (there are 5 stack levels in the unwind functions).

## Testing

Manual testing, and also ran e2e tests.

### DIFF
--- a/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
+++ b/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
@@ -18,6 +18,13 @@ internal class AnrPlugin : Plugin {
             // The only check that will work across all Android versions is the isNativeMethod call.
             return javaTrace.first().isNativeMethod
         }
+
+        private const val stackUnwindOverhead = 5
+
+        // Trim the stack trace generation components from the native trace
+        internal fun trimNativeTrace(nativeTrace: List<NativeStackframe>): List<NativeStackframe> {
+            return nativeTrace.subList(stackUnwindOverhead, nativeTrace.size)
+        }
     }
 
     private val loader = LibraryLoader()
@@ -97,7 +104,7 @@ internal class AnrPlugin : Plugin {
             // append native stackframes to error/thread stacktrace
             if (hasNativeComponent) {
                 // update error stacktrace
-                val nativeFrames = nativeTrace.map { Stackframe(it) }
+                val nativeFrames = trimNativeTrace(nativeTrace).map { Stackframe(it) }
                 err.stacktrace.addAll(0, nativeFrames)
 
                 // update thread stacktrace
--- a/features/full_tests/batch_1/detect_anr_jvm.feature
+++ b/features/full_tests/batch_1/detect_anr_jvm.feature
@@ -14,25 +14,6 @@ Scenario: ANR triggered in JVM loop code is captured
     And the error payload field "events.0.threads.0.type" equals "android"
     And the error payload field "events.0.threads.0.stacktrace.0.type" is null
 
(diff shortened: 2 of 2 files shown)


features/full_tests/batch_1/detect_anr_jvm.feature
```diff
@@ -14,25 +14,6 @@ Scenario: ANR triggered in JVM loop code is captured
     And the error payload field "events.0.threads.0.type" equals "android"
     And the error payload field "events.0.threads.0.stacktrace.0.type" is null
 
-# Other scenarios use a deadlock to generate an ANR, which works on Samsung devices. This scenario remains skipped
```
kstenerud, line 17: Note: This is being removed because the new detection code can no longer tell the difference between Java functions that syscall, and actual native functions.

features/full_tests/batch_1/detect_anr_jvm.feature
```diff
@@ -14,25 +14,6 @@ Scenario: ANR triggered in JVM loop code is captured
...
-# on Samsung as it is explicitly design to test ANRs caused by a sleeping thread.
-@skip_samsung
-Scenario: ANR triggered in JVM sleep code is captured
-    When I run "JvmAnrSleepScenario"
```
fractalwrench, line 21: There still need to be two ANR scenarios if it's at all possible IMO, giving coverage of:

1. An ANR triggered where the method is native (e.g. Thread.sleep)
2. An ANR triggered where the method originates in the JVM

Would it be possible to trigger an ANR by something like `while (true) {}` to maintain this coverage? If not I have a couple of ideas such as using a `BroadcastReceiver` to trigger an ANR.

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
```diff
@@ -18,6 +18,13 @@ internal class AnrPlugin : Plugin {
             // The only check that will work across all Android versions is the isNativeMethod call.
             return javaTrace.first().isNativeMethod
         }
+
+        private const val stackUnwindOverhead = 5
```
fractalwrench, line 22: This raises a few questions for me:

- Is the number of frames consistent for all versions of Android?
- Would it make sense to consider trimming frames for handled errors from the NDK too?
- How will this affect the grouping of errors on the dashboard, as it will affect the top frame of the stacktrace?

If this has been considered then this changeset looks fine to me (once the other comment has been resolved)

kstenerud: I think we need to revisit this in general. The original requirements included trimming the top of the stack trace for ANRs, and in fact the entire purpose of this particular work order was to bring it in line with those requirements.

But in general we now have an issue with what should be present in what stack traces:

- Should NDK traces be trimmed the same as for ANRs?
- Should Java traces be trimmed in the same manner when the android project is part of React or Unity? (basically the same problem, one level higher).
- Maybe all traces should have the top trimmed regardless, so that the top is the actual culprit rather than our trace gathering code?

We should definitely have consistency one way or the other with this.

In this particular case, however, we have direct control over the calls leading to the stack trace, so it will always be 5 unless we change our trace gathering code in the NDK plugin.

sss545:
Trace mobile A10s

bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
sss545: Send me link deposits

tomlongridge:
Agreed to do this on the backend - closing PR.

PR #1124, opened 2021-02-09
Tidy up various mazerunner scenarios

fractalwrench:
## Goal

Resolves issues with several mazerunner scenarios which should speed up the test run without losing coverage.

## Changeset

- Adding logging to `Client` when an error is discarded due to class name or release stage
- Refactored mazerunner code for zero log message assertions into `getZeroEventsLogMessages()`
- Combined error/session scenarios in `custom_http_client.feature` into two scenarios now that separate endpoints are available
- Updatd `ignored_reports.feature` and `release_stages.feature` to use log assertions to speed up run, and to use `Bugsnag.notify` rather than an unhandled crash (functionally this is testing the same codepath but is quicker for testing purposes
- Removed redundant scenario from `in_foreground.feature` which is covered by the smoke tests
- Converted scenarios to Kotlin (no effect on test coverage)


twometresteve (approved):
LGTM

PR #1133, opened 2021-02-12
Check additional JNI calls for pending exceptions and no-op

fractalwrench:
## Goal

Performs additional checking of JNI calls for pending exceptions, and no-ops if these are detected. This changeset also wraps all JNI calls in a `bsg_safe` wrapper, which checks for disallowed parameter values such as `NULL` and no-ops.

## Changeset

The changeset is roughly broken up into individual commits for each affected JNI method:

```
DeleteLocalRef
ReleaseByteArrayElements
ReleaseStringUTFChars
GetStringUTFChars
IsInstanceOf
GetArrayLength
```

Each affected method has been added to `safejni.h` and its parameters checked according to the [JNI docs](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html). For `GetStringUTFChars` the calling method has been updated to handle a `NULL` return value where necessary.

JNI calls in the ANR plugin have also been updated to perform equivalent safety checks.

Some internal method signatures have been updated where possible to use `const char *` rather than `char *`, as this is what `GetStringUTFChars` actually returns - a lot of invocations have been incorrectly casting it to `char *`.

## Testing

Relied on existing E2E test coverage, also performed smoke testing with example app on emulator to cover x86 architectures.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -111,7 +111,7 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_install(
   bugsnag_env->report_header.version = BUGSNAG_EVENT_VERSION;
   const char *event_path = (*env)->GetStringUTFChars(env, _event_path, 0);
   sprintf(bugsnag_env->next_event_path, "%s", event_path);
-  (*env)->ReleaseStringUTFChars(env, _event_path, event_path);
+  bsg_safe_release_string_utf_chars(env, _event_path, event_path);
 
   if ((bool)auto_detect_ndk_crashes) {
     bsg_handler_install_signal(bugsnag_env);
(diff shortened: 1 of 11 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -400,168 +421,196 @@ JNIEXPORT void JNICALL
...
-  bugsnag_event_add_metadata_string(&bsg_global_env->next_event, tab, key,
-                                    value);
-  bsg_release_env_write_lock();
-  (*env)->ReleaseStringUTFChars(env, tab_, tab);
-  (*env)->ReleaseStringUTFChars(env, key_, key);
-  (*env)->ReleaseStringUTFChars(env, value_, value);
+  }
+  char *tab = (char *)bsg_safe_get_string_utf_chars(env, tab_);
+  char *key = (char *)bsg_safe_get_string_utf_chars(env, key_);
+  char *value = (char *)bsg_safe_get_string_utf_chars(env, value_);
```
kstenerud, line 525: We should definitely eliminate these casts away from const in a later PR...

kstenerud (approved):

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
fractalwrench: Raised PLAT-6035 to do this. The problem is that a lot of our public API specifies `char *` rather than `const char *` - I'm assuming that altering this would count as a breaking change. I can create a separate ticket to address the internal usages where possible.

PR #1136, opened 2021-02-14
Remove use of "non-crashy" metadata in test scenarios

twometresteve:
## Goal

Remove use of "non-crashy" metadata from e2e test scenarios, simplifying the test code base and removing several Appium interactions (approx 0.5s each).

## Design

Rather than than passing "non-crashy" through as metadata for the scenario, knowledge of whether the app is "crashy" or reopening after a crash is now built into the test fixture buttons.  Pressing "Start Bugsnag" will, by default, only start Bugsnag and is typically used to send any errors after relaunching a crashed app.  "Run Scenario" will start Bugsnag and perform any notify/crashes coded into the scenario.  

This default behaviour is sufficient for most scenarios, but can be overridden as necessary.  When it is, the `startBugsnag` function is overridden with conditional logic based on the `startBugsnagOnly` parameter (calling `super.StartBugsnag` at the end).

The Cucumber steps of interest are:
- `I configure Bugsnag for {string}` ("Start Bugsnag only")
- `I run {string} and relaunch the app` 

## Changeset

Whilst refactoring the test fixture, I took to opportunity to update the layout to be more consistent with our other notifiers and in line with internally published conventions.

## Testing

Covered by a full CI run.  A few scenarios are currently `@skip`ped and it's harder to see if these have been updated completely correctly.  However, any errors or omissions should naturally come out in the wash when the relevant ticket is processed.

### DIFF
--- a/.buildkite/pipeline.quick.yml
+++ b/.buildkite/pipeline.quick.yml
@@ -110,9 +110,6 @@ steps:
           - "--app=/app/build/release/fixture-r21.apk"
           - "--farm=bs"
           - "--device=ANDROID_9_0"
-          - "--retry=1"
-          - "--strict-undefined"
-          - "--strict-pending"
           - "--tags"
           - "not @Flaky"
     concurrency: 9
@@ -134,9 +131,6 @@ steps:
           - "--app=/app/build/release/fixture-r21.apk"
           - "--farm=bs"
           - "--device=ANDROID_9_0"
-          - "--retry=1"
-          - "--strict-undefined"
-          - "--strict-pending"
           - "--tags"
           - "not @Flaky"
     concurrency: 9
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
@@ -11,23 +11,15 @@ internal class EmptyReportScenario(
     eventMetadata: String
 ) : Scenario(config, context, eventMetadata) {
 
-    init {
-        config.autoTrackSessions = false
-        val errDir = File(context.cacheDir, "bugsnag-errors")
-
-        if (eventMetadata != "non-crashy") {
+    override fun startBugsnag(startBugsnagOnly: Boolean) {
+        if (startBugsnagOnly) {
             disableAllDelivery(config)
-        } else {
-            val files = errDir.listFiles()
-            files.forEach { it.writeText("") }
         }
     }
 
     override fun startScenario() {
         super.startScenario()
 
-        if (eventMetadata != "non-crashy") {
-            Bugsnag.notify(java.lang.RuntimeException("Whoops"))
-        }
+        Bugsnag.notify(java.lang.RuntimeException("Whoops"))
     }
 }
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
@@ -3,7 +3,6 @@ package com.bugsnag.android.mazerunner.scenarios
 import android.content.Context
 import com.bugsnag.android.Bugsnag
 import com.bugsnag.android.Configuration
-import java.io.File
 
 internal class EmptyReportScenario(
     config: Configuration,
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
@@ -3,6 +3,7 @@ package com.bugsnag.android.mazerunner.scenarios
 import android.content.Context
 import com.bugsnag.android.Bugsnag
 import com.bugsnag.android.Configuration
+import java.io.File
 
 internal class EmptyReportScenario(
     config: Configuration,
@@ -11,9 +12,16 @@ internal class EmptyReportScenario(
 ) : Scenario(config, context, eventMetadata) {
 
     override fun startBugsnag(startBugsnagOnly: Boolean) {
+        config.autoTrackSessions = false
+
         if (startBugsnagOnly) {
             disableAllDelivery(config)
+        } else {
+            val errDir = File(context.cacheDir, "bugsnag-errors")
+            val files = errDir.listFiles()
+            files.forEach { it.writeText("") }
         }
+        super.startBugsnag(startBugsnagOnly)
     }
 
     override fun startScenario() {
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptySessionScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptySessionScenario.kt
@@ -15,26 +15,24 @@ internal class EmptySessionScenario(
     eventMetadata: String
 ) : Scenario(config, context, eventMetadata) {
 
-    init {
+    override fun startBugsnag(startBugsnagOnly: Boolean) {
         config.autoTrackSessions = false
 
         val dir = File(context.cacheDir, "bugsnag-sessions")
 
-        if (eventMetadata != "non-crashy") {
-            disableAllDelivery(config)
-        } else {
+        if (startBugsnagOnly) {
             val files = dir.listFiles()
             Log.d("Bugsnag", "Empty sessions: $files")
             files.forEach { it.writeText("") }
+        } else {
+            disableAllDelivery(config)
         }
     }
 
     override fun startScenario() {
         super.startScenario()
 
-        if (eventMetadata != "non-crashy") {
-            Bugsnag.startSession()
-        }
+        Bugsnag.startSession()
 
         val thread = HandlerThread("HandlerThread")
         thread.start()
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptySessionScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptySessionScenario.kt
@@ -27,6 +27,7 @@ internal class EmptySessionScenario(
         } else {
             disableAllDelivery(config)
         }
+        super.startBugsnag(startBugsnagOnly)
     }
 
     override fun startScenario() {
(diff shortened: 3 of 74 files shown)


.buildkite/pipeline.quick.yml
```diff
@@ -134,9 +131,6 @@ steps:
           - "--app=/app/build/release/fixture-r21.apk"
           - "--farm=bs"
           - "--device=ANDROID_9_0"
-          - "--retry=1"
-          - "--strict-undefined"
-          - "--strict-pending"
```
fractalwrench, line 139: What is the effect of removing these flags?

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
```diff
@@ -29,26 +30,33 @@ abstract class Scenario(
...
+    protected var startBugsnagOnly: Boolean = false
+
     /**
      * Determines what log messages should be intercepted from Bugsnag and sent to Mazerunner
      * using a HTTP requests to the /logs endpoint. This is used to assert that Bugsnag is
      * behaving correctly in situations where sending an error/session payload is not
      * possible.
      */
-    open fun getInterceptedLogMessages() = emptyList<String>()
+    open fun getInterceptedLogMessages() = getZeroEventsLogMessages(startBugsnagOnly)
```
fractalwrench, line 44: The default for this should be an empty list, because otherwise mazerunner will send a HTTP request for _every_ scenario that matches these log messages. It should be overridden in the appropriate scenarios instead to minimize HTTP traffic.

features/full_tests/batch_1/cached_error_reports.feature
```diff
@@ -2,7 +2,6 @@ Feature: Cached Error Reports
 
 Scenario: If an empty file is in the cache directory then zero requests should be made
     When I run "EmptyReportScenario" and relaunch the app
-    And I configure the app to run in the "non-crashy" state
     And I run "EmptyReportScenario"
```
fractalwrench, line 5: It'd be good to have more of a distinction between the different run steps IMO. I can see developers unfamiliar with mazerunner getting confused about the difference of appending `and relaunch the app` onto the run step.

It also seems to be somewhat inconsistent across the feature files - sometimes it's `I run X`, and at other times it's `I configure X`.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/InternalReportScenario.kt
```diff
@@ -27,20 +27,12 @@ internal class InternalReportScenario(
             storageManager.setCacheBehaviorGroup(errDir, true)
             storageManager.setCacheBehaviorTombstone(errDir, true)
         }
-
-        if (eventMetadata != "non-crashy") {
-            disableAllDelivery(config)
-        } else {
-            val files = errDir.listFiles()
-            files.forEach { it.writeText("{[]}") }
```
fractalwrench, line 35: This block needs to be reinstated for the scenario to work

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
```diff
@@ -29,26 +30,33 @@ abstract class Scenario(
     protected val eventMetadata: String?
 ) : Application.ActivityLifecycleCallbacks {
 
+    /**
+     * Tracks whether the scenario is starting Bugsnag only, or running the scenario.
+     */
+    protected var startBugsnagOnly: Boolean = false
```
fractalwrench, line 36: This property doesn't seem to be read anywhere, so could be refactored to just a parameter.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
```diff
@@ -29,26 +30,33 @@ abstract class Scenario(
...
 
     /**
      * Initializes Bugsnag. It is possible to override this method if the scenario requires
      * it - e.g., if the config needs to be loaded from the manifest.
      */
-    open fun startBugsnag() {
+    open fun startBugsnag(startBugsnagOnly: Boolean) {
+        this.startBugsnagOnly = startBugsnagOnly
         Bugsnag.start(context, config)
     }
```
fractalwrench, line 54: I have another proposal here: rather than altering `Configuration` in the constructor or in `startBugsnag`, there should be a separate overriddable method called `prepareConfig(startBugsnagOnly: Boolean)`.

This would have a few advantages:

- Creates a consistent place for constructing configuration object (currently it's split between startBugsnag/init blocks)
- Allows for sensible defaults in `Scenario.kt` (such as setting `autoTrackSessions` to false)
- Reduces responsibilities of `startBugsnag`

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptySessionScenario.kt
```diff
@@ -15,26 +15,24 @@ internal class EmptySessionScenario(
...
-            disableAllDelivery(config)
-        } else {
+        if (startBugsnagOnly) {
             val files = dir.listFiles()
             Log.d("Bugsnag", "Empty sessions: $files")
             files.forEach { it.writeText("") }
+        } else {
+            disableAllDelivery(config)
         }
     }
```
fractalwrench, line 30: Existing problem, but this method doesn't call its super meaning nothing would ever be delivered

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
```diff
@@ -3,31 +3,22 @@ package com.bugsnag.android.mazerunner.scenarios
...
-        config.autoTrackSessions = false
-        val errDir = File(context.cacheDir, "bugsnag-errors")
-
-        if (eventMetadata != "non-crashy") {
+    override fun startBugsnag(startBugsnagOnly: Boolean) {
+        if (startBugsnagOnly) {
             disableAllDelivery(config)
-        } else {
-            val files = errDir.listFiles()
-            files.forEach { it.writeText("") }
```
fractalwrench, line 22: This block needs reinstating for the scenario to be testing anything.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
```diff
@@ -3,31 +3,22 @@ package com.bugsnag.android.mazerunner.scenarios
...
         }
```
fractalwrench, line 16: existing problem, but this method does not call its super

fractalwrench (commented):
This looks like the right direction and makes me very glad I'll never have to see `"non-crashy"` in the codebase again.

I had a few comments on the implementation which I've left inline. It'd also be good to see a full CI run, as I suspect the range of these changes has the potential to break a few scenarios.

.buildkite/pipeline.quick.yml
twometresteve: It means that failing scenarios won't retry.  I don't think these should have been in the Android 9 run at all in the first instance and appear to have been a copy/paste or perhaps merge error.

fractalwrench: Are the `--strict-` flags part of that or something else?

features/full_tests/batch_1/cached_error_reports.feature
twometresteve: I think the inconsistency is actually an omission on my part - it should be using `I configure Bugsnag for X` consistently after a crash.

I'm happy to change the wording of the run steps to aid clarity, it probably just needs a bit of further thought and analysis to make sure we capture the logical situations correctly.  I've taken a note to look at that, in a view to raising a separate ticket for it.

fractalwrench: Splitting this out sounds sensible.

.buildkite/pipeline.quick.yml
twometresteve: Yeah, they are part of it.  There are three flavours of "strict" - undefined, pending and flaky.  MazeRunner makes them all strict by default, but flaky is not compatible with retry (if you want the retry to pass the test).... so the two strict flags were added with retry to make it as strict as it could be.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
twometresteve: It's currently being passed through to `getZeroEventsLogMessages` below, but perhaps that will be overtaken by the comment there...

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
twometresteve: Yes, I think that would make sense - I'll raise a separate ticket to cover it, though, as it will end up touching quite a lot of files.

twometresteve: PLAT-6049.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
twometresteve: Reinstated, within the new format.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptyReportScenario.kt
twometresteve: Now added,

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/EmptySessionScenario.kt
twometresteve: Now added

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/InternalReportScenario.kt
twometresteve: Added back in, in a new `startBugsnag` method.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
twometresteve: Yeah, having refactored `getInterceptedLogMessages` for another comment it's now used.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/Scenario.kt
twometresteve: Done.

fractalwrench (approved):
LGTM.

I'll create a separate ticket to create a new method for preparing `Configuration`, which will provide some separation from always using the constructor - and hopefully make things a bit clearer.

PR #1143, opened 2021-02-16
Plat 5957/additional backports

fractalwrench:
Backport of #1133 to the v4 branch for bugsnag-unity..

## Goal

Performs additional checking of JNI calls for pending exceptions, and no-ops if these are detected. This changeset also wraps all JNI calls in a `bsg_safe` wrapper, which checks for disallowed parameter values such as `NULL` and no-ops.

## Changeset

The changeset is roughly broken up into individual commits for each affected JNI method:

```
DeleteLocalRef
ReleaseByteArrayElements
ReleaseStringUTFChars
GetStringUTFChars
IsInstanceOf
GetArrayLength
```

Each affected method has been added to `safejni.h` and its parameters checked according to the [JNI docs](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html). For `GetStringUTFChars` the calling method has been updated to handle a `NULL` return value where necessary.

JNI calls in the ANR plugin have also been updated to perform equivalent safety checks.

Some internal method signatures have been updated where possible to use `const char *` rather than `char *`, as this is what `GetStringUTFChars` actually returns - a lot of invocations have been incorrectly casting it to `char *`.

## Testing

Relied on existing E2E test coverage, also performed smoke testing with example app on emulator to cover x86 architectures.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -270,8 +270,8 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_addBreadcrumb(
 
   if (name != NULL && type != NULL && timestamp != NULL) {
     bugsnag_breadcrumb *crumb = calloc(1, sizeof(bugsnag_breadcrumb));
-    strncpy(crumb->name, name, sizeof(crumb->name));
-    strncpy(crumb->timestamp, timestamp, sizeof(crumb->timestamp));
+    bsg_strncpy_safe(crumb->name, name, sizeof(crumb->name));
+    bsg_strncpy_safe(crumb->timestamp, timestamp, sizeof(crumb->timestamp));
     if (strcmp(type, "user") == 0) {
       crumb->type = BSG_CRUMB_USER;
     } else if (strcmp(type, "error") == 0) {
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -85,8 +85,13 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_install(
   bugsnag_env->report_header.big_endian =
       htonl(47) == 47; // potentially too clever, see man 3 htonl
   bugsnag_env->report_header.version = BUGSNAG_REPORT_VERSION;
-  const char *report_path = (*env)->GetStringUTFChars(env, _report_path, 0);
+  const char *report_path = bsg_safe_get_string_utf_chars(env, _report_path);
+  if (report_path == NULL) {
+    return;
+  }
+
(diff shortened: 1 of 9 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -257,94 +264,102 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_addBreadcrumb(
...
-  } else {
-    crumb->type = BSG_CRUMB_MANUAL;
-  }
+  const char *name = bsg_safe_get_string_utf_chars(env, name_);
+  const char *type = bsg_safe_get_string_utf_chars(env, crumb_type);
+  const char *timestamp = bsg_safe_get_string_utf_chars(env, timestamp_);
+
+  if (name != NULL && type != NULL && timestamp != NULL) {
+    bugsnag_breadcrumb *crumb = calloc(1, sizeof(bugsnag_breadcrumb));
+    strncpy(crumb->name, name, sizeof(crumb->name));
```
kstenerud, line 273:
```suggestion
    bsg_strncpy_safe(crumb->name, name, sizeof(crumb->name));
```

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -257,94 +264,102 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_addBreadcrumb(
...
+    strncpy(crumb->timestamp, timestamp, sizeof(crumb->timestamp));
```
kstenerud, line 274:
```suggestion
    bsg_strncpy_safe(crumb->timestamp, timestamp, sizeof(crumb->timestamp));
```

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -530,17 +541,19 @@ void bsg_populate_user_data(JNIEnv *env, bsg_jni_cache *jni_cache,
...
                           bugsnag_report *report) {
   jstring _context = bsg_safe_call_static_object_method(
       env, jni_cache->native_interface, jni_cache->get_context);
   if (_context != NULL) {
-    const char *value = (*env)->GetStringUTFChars(env, (jstring)_context, 0);
-    strncpy(report->context, value, sizeof(report->context) - 1);
-    (*env)->ReleaseStringUTFChars(env, _context, value);
+    const char *value = bsg_safe_get_string_utf_chars(env, (jstring)_context);
+    if (value != NULL) {
+      strncpy(report->context, value, sizeof(report->context) - 1);
```
kstenerud, line 554:
```suggestion
      bsg_strncpy_safe(report->context, value, sizeof(report->context) - 1);
```

kstenerud (changes requested):

fractalwrench:
Nice catch! I've updated usage of `strncpy` to `bsg_strncpy_safe` and will raise a PR to do so against `next` too.

kstenerud (approved):

PR #1144, opened 2021-02-16
Avoid sleep() in session tracking E2E scenarios

fractalwrench:
## Goal

Alters some of the session tracking E2E scenarios so that they don't call `Thread.sleep()`, and reorganizes some so that they are combined into one scenario.

## Changeset

- Removed `sleep()` and `flushAllSessions()` calls which seem to be unnecessary for the scenarios to pass
- Combined `session_stopping.feature` into one scenario, removing sleeps by using a state machine to enforce request order
- Opportunistically converted scenarios to Kotlin


features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXPausedSessionScenario.kt
```diff
@@ -0,0 +1,28 @@
...
+    init {
+        System.loadLibrary("cxx-scenarios")
+        config.autoTrackSessions = false
+    }
+
+    external fun crash(counter: Int): Int
+
+    override fun startScenario() {
+        super.startScenario()
+        if ("non-crashy" != eventMetadata) {
```
twometresteve, line 22: "non-crashy" should never be mentioned following the previous PR.  It's probably as simple as just removed all the guards though.

features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXSessionInfoCrashScenario.kt
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    external fun crash(value: Int): Int
+
+    override fun startScenario() {
+        super.startScenario()
+        if (eventMetadata!! != "non-crashy") {
+            Bugsnag.startSession()
+            Bugsnag.notify(Exception("For the first"))
+            Bugsnag.notify(Exception("For the second"))
```
twometresteve, line 25: Are calls to `notify` synchronous?  I.e. crashing immediately afterwards will not cause flakes?

features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXPausedSessionScenario.kt
fractalwrench: Removed guards.

features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXSessionInfoCrashScenario.kt
fractalwrench: The error capture is synchronous but the delivery is asynchronous. In practice I don't think it will be an issue as the JNI call should take some time to execute, by which time the delivery should be done.

If it is an issue on the full CI run, or if it crops up as a failure, then it should be a easy enough fix of adding an `InterceptingDelivery` implementation that only calls `crash()` once the two error reports are sent.

fractalwrench: Updated to use `InterceptingDelivery` as this was indeed causing flakiness.

twometresteve (approved):
LGTM

PR #1145, opened 2021-02-17
Remove additional unnecessary mazerunner sleeps

fractalwrench:
## Goal

Removes additional unnecessary calls to `sleep()` and `flushAllSessions()`. These were once required in older versions of bugsnag-android but have been surpassed now that the `SessionTracker` automatically flushes stored sessions.

The cached error/session report tests were also obsolete - this alters them to validate that an internal Bugsnag error report is received when an error cannot be persisted.

## Changeset

- Removed unused scenario classes
- Removed `sleep()` and `flushAllSessions()` from scenarios where possible
- Validated that an internal report is sent to Bugsnag when an error cannot be persisted on disk


features/full_tests/batch_1/internal_error_reports.feature
```diff
@@ -0,0 +1,73 @@
...
+    And the event "user.id" is null
+
+    # App data
+    And the event "app.buildUUID" is not null
+    And the event "app.id" equals "com.bugsnag.android.mazerunner"
+    And the event "app.releaseStage" equals "production"
+    And the event "app.type" equals "android"
+    And the event "app.version" equals "1.1.14"
+    And the event "app.versionCode" equals 34
+    And the error payload field "events.0.app.duration" is an integer
```
twometresteve, line 42: Presumably we just don't have a steps for `the event {string} is an integer` (and `greater than 0` etc)?  Feel free to raise a ticket for that and we'll get them added for consistency.

twometresteve (approved):
LGTM, subject to what happens when we get an e2e test run for the PR into `next` (unless you want to push another commit with either `[quick ci]` or `[gated-full ci]`in it to force it).

features/full_tests/batch_1/internal_error_reports.feature
fractalwrench: Created a ticket to make these steps.

fractalwrench:
Merging into target branch, will run full CI against that.

PR #1146, opened 2021-02-17
Remove unnecessary null check for inForeground update

fractalwrench:
## Goal

The `activity` can be `NULL` - if this is the case then the in foreground state should be updated, rather than returning early.


twometresteve (approved):
LGTM

PR #1147, opened 2021-02-17
Ensure ANR handler calls JNI code safely

fractalwrench:
## Goal

Adds safe wrapper calls to the ANR handler code which invokes the JNI.

Because the ANR handler is in a separate module to the NDK code, the relevant methods have been temporarily duplicated until the project can be restructured to the point where it can share C code between modules. A ticket has been raised to do this but it will likely require a major version bump.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,5 +1,7 @@
-PATH
-  remote: ../maze-runner
+GIT
+  remote: https://github.com/bugsnag/maze-runner
+  revision: b12f25a65267a99162a40abf28fe02c3048a097f
+  tag: v4.10.1
   specs:
     bugsnag-maze-runner (4.10.1)
       appium_lib (~> 11.2.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -1,5 +1,7 @@
-PATH
-  remote: ../maze-runner
+GIT
+  remote: https://github.com/bugsnag/maze-runner
+  revision: b12f25a65267a99162a40abf28fe02c3048a097f
+  tag: v4.10.1
```
kstenerud, line 4: Is this supposed to change?

kstenerud (approved):

Gemfile.lock
fractalwrench: I believe so - the `Gemfile` was bumped to `v4.10.1` but the lockfile wasn't updated. This is generated as a result of running `bundle install` which is a prerequisite of running the local E2E tests.

PR #1153, opened 2021-02-18
String copy and JNI fixes

kstenerud:
## Goal

JNI-to-JVM calls could return garbage data if an exception is thrown by the JVM.
String copies were not all uniform, and in some cases would not be null terminated.
Some string functions (such as length calculation) are better handled by the standard library.

## Design

* Fix missing null termination, and consoliate string copy functionality into one function.
* Remove custom string length functions.
* Return NULL or default values from JNI-to-JVM calls when an exception occurs.

## Testing

Re-ran unit tests and e2e tests


bugsnag-plugin-android-ndk/src/main/jni/utils/string.c
```diff
@@ -1,40 +1,14 @@
...
-
-size_t bsg_strlen(const char *str) {
-  size_t i = 0;
-  while (true) {
-    char current = str[i];
-    if (current == '\0') {
-      return i;
-    } else if (i == INT_MAX) {
-      return INT_MAX;
+char *bsg_strncpy_safe(char *dst, const char *src, size_t dst_size) {
```
fractalwrench, line 4: Perhaps overly paranoid, but what will happen here if `dst` is `NULL`?

fractalwrench (approved):
LGTM pending a changelog entry, and resolution of one query.

bugsnag-plugin-android-ndk/src/main/jni/utils/string.c
kstenerud: It crashes, like any copy function would :)

It's a dangerous path once you start putting guardrails over function params, as it masks out coding errors and pushes them further and further from the source of the problem. I'm even nervous about the guard on `src`, but it's old code so I didn't touch it.

fractalwrench: Happy enough to leave this as is - I haven't spotted anywhere where we are passing it `NULL`, and it's something we can control so could fix if it ever did go wrong.

PR #1155, opened 2021-02-22
Improve build stability for v4 back-ports

twometresteve:
## Goal

Improve stability of builds for v4 back-ported changes.

## Design

Applies some techniques that we've learnt since v4, as well as allowing retries on e2e test scenarios (for these v4-basd changes only).

## Changeset

- Update Buildkite pipeline pipeline to allow retries
- Use latest v2 Maze Runner
- Dismiss error dialogs when app launches

## Testing

Covered by CI.


fractalwrench (approved):
LGTM pending CI

PR #1156, opened 2021-02-22
Improve stability of Android 5 testing

twometresteve:
## Goal

Attempts to improve the stability of Android 5 testing by increasing the timeout when waiting for elements and upping the Appium version to the latest possible at this time. 

## Design

We were seeing some failures to find the `scenario_name` element, despite the field being visible in the screenshot that was taken for the failure.  This change seems to improve the situation, but it's not perfect and we should explore using the very latest version of Appium now available.  Trying this now uncovered another problem, in that at least one of our scenarios fails consistently with a newer version (more investigation needed and internal ticket raised).

## Changeset

Use Appium 1.8.0 on Android 5 (1.9.1 causes scenario failures at present) and increase timeout when finding elements.

## Testing

Covered by CI.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -101,6 +101,7 @@ steps:
           - "--fail-fast"
           - "--tags"
           - "not @Flaky"
+          - "--appium-version=1.8.0"
     concurrency: 9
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM pending CI

PR #1158, opened 2021-02-22
Global JNI cache

kstenerud:
## Goal

https://bugsnag.atlassian.net/browse/PLAT-6090

## Design

The JNI cache used by the metadata and breadcrumb system is now initialised once at app start, and is made available as a global.

## Testing

Manual testing with the example app, and e2e tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/jnicache.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/jnicache.c
@@ -1,0 +1,249 @@
+//
+// Created by Karl Stenerud on 22.02.21.
+//
+
+#include "jnicache.h"
+#include <stdatomic.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <bugsnag_ndk.h>
+#include "safejni.h"
+
+static bsg_jni_cache jni_cache;
+static bool is_cache_valid = false;
+
+static inline jclass safe_find_class(JNIEnv *env, const char *name) {
+  jclass cls = (*env)->NewGlobalRef(env, bsg_safe_find_class(env, name));
+  if (cls == NULL) {
+    BUGSNAG_LOG("Could not load class %s", name);
+  }
+  return cls;
+}
+
+static inline jclass safe_get_method_id(JNIEnv *env, jclass cls, const char *name, const char *sig) {
+  jmethodID method = bsg_safe_get_method_id(env, cls, name, sig);
+  if (method == NULL) {
+    BUGSNAG_LOG("Could not load method %s with signature %s", name, sig);
+  }
+
+  return method;
+}
+
+static inline jmethodID safe_get_static_method_id(JNIEnv *env, jclass cls, const char *name, const char *sig) {
+  jmethodID method = bsg_safe_get_static_method_id(env, cls, name, sig);
+  if (method == NULL) {
+    BUGSNAG_LOG("Could not load static method %s with signature %s", name, sig);
+  }
+
+  return method;
+}
+
+bool bsg_init_jni_cache(JNIEnv *env) {
+  static atomic_flag initialized = ATOMIC_FLAG_INIT;
+  if (atomic_flag_test_and_set(&initialized)) {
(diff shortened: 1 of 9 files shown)


bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -606,15 +287,16 @@ void bsg_populate_metadata_obj(JNIEnv *env, bugsnag_report *dst,
...
   }
   bsg_safe_delete_local_ref(env, _value);
 }
 
-void bsg_populate_metadata_section(JNIEnv *env, bugsnag_report *report,
-                                   jobject metadata, bsg_jni_cache *jni_cache,
-                                   jobject keylist, int i) {
+static void populate_metadata_section(JNIEnv *env, bugsnag_report *report,
+                                      jobject metadata, jobject keylist, int i) {
+  type_bsg_jni_cache *jni_cache = bsg_get_jni_cache();
```
fractalwrench, line 298: IMO it's not particularly clear which methods have checked `bsg_is_jni_cache_valid`. I think it'd be safer to either always call `bsg_is_jni_cache_valid`, or to create a convention where `type_bsg_jni_cache` is passed as a parameter when the cache has already been checked.

bugsnag-plugin-android-ndk/src/main/jni/jnicache.c
```diff
@@ -0,0 +1,249 @@
...
+#include <stdatomic.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <bugsnag_ndk.h>
+#include "safejni.h"
+
+static type_bsg_jni_cache jni_cache;
+static bool is_cache_valid = false;
+
+static inline jclass safe_find_class(JNIEnv *env, const char *name) {
```
fractalwrench, line 15: methods should be prefixed with `bsg` throughout the changeset to avoid any potential for symbol clashes

bugsnag-plugin-android-ndk/src/main/jni/jnicache.c
```diff
@@ -0,0 +1,249 @@
...
+    BUGSNAG_LOG("Could not load method %s with signature %s", name, sig);
+  }
+
+  return method;
+}
+
+static inline jmethodID safe_get_static_method_id(JNIEnv *env, jclass cls, const char *name, const char *sig) {
+  jmethodID method = bsg_safe_get_static_method_id(env, cls, name, sig);
+  if (method == NULL) {
+    BUGSNAG_LOG("Could not load static method %s with signature %s", name, sig);
```
fractalwrench, line 35: Could this logging just be moved to `bsg_safe_get_static_method_id` (and other functions) instead?

bugsnag-plugin-android-ndk/src/main/jni/utils/jni_utils.h
```diff
@@ -0,0 +1,28 @@
+//
+// Created by Karl Stenerud on 22.02.21.
+//
+
+// Utils to deal with common JNI situations.
+//
+// These tools depend on the JNI cache. DO NOT call them unless
+// bsg_is_jni_cache_valid() returns true!
```
fractalwrench, line 8: As mentioned previously, these should either check whether the cache is valid or have it passed in as a parameter - it'd be easy for someone to assume that they already perform the check since they're utils functions.

bugsnag-plugin-android-ndk/src/main/jni/jnicache.c
kstenerud: It's a static method, so it won't clash.

bugsnag-plugin-android-ndk/src/main/jni/jnicache.c
kstenerud: I have another commit on the mainline that would clash with it. I think we should eventually put the logging in the safe_jni section, but not right now for this fix.

fractalwrench (approved):

PR #1159, opened 2021-02-22
Add public API for crash-on-launch detection

fractalwrench:
## Goal

Adds the public API for detecting whether an app crashed on launch.

## Changeset

Made the following interfaces are now available as part of the public API, adding JavaDoc as appropriate. Note that this simply adds the API - this will be hooked up to the actual functionality in separate changesets.

```
class Bugsnag {
    static LastRunInfo getLastRunInfo();
    static void markLaunchCompleted();
}

class Client {
    LastRunInfo getLastRunInfo();
    void markLaunchCompleted();
}

class Configuration {
    Long launchDurationMillis;
    Boolean sendLaunchCrashesSynchronously;
}

class AppWithState {
    Boolean isLaunching = null;
}

final class LastRunInfo {
    final int consecutiveLaunchCrashes;
    final boolean crashed;
    final boolean crashedDuringLaunch;
}
```

Additionally loaded `sendLaunchCrashesSynchronously` from the manifest, and verified that `app.isLaunching` is serialized to JSON.

## Testing

Added unit test coverage


bugsnag-android-core/src/main/java/com/bugsnag/android/Bugsnag.java
```diff
@@ -379,6 +379,31 @@ public static void pauseSession() {
         return getClient().getBreadcrumbs();
     }
 
+    /**
+     * Retrieves information about the last launch of the application, if there was a launch.
```
tomlongridge, line 383:
```suggestion
     * Retrieves information about the last launch of the application, if it has been run before.
```

tomlongridge (approved):
LGTM - one small wording suggestion

PR #1160, opened 2021-02-23
Make sure all strings get fully released

kstenerud:
## Goal

Tighten up the JNI code even more to ensure that the JNI local reference table cannot be overrun.

## Testing

Manual testing and re-ran e2e tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -51,7 +51,7 @@ void bugsnag_leave_breadcrumb(char *name, bsg_breadcrumb_t type) {
   }
 }
 
-jfieldID bsg_parse_jseverity(JNIEnv *env, bsg_severity_t severity,
+static jfieldID bsg_parse_jseverity(JNIEnv *env, bsg_severity_t severity,
                              jclass severity_class) {
   const char *severity_sig = "Lcom/bugsnag/android/Severity;";
   if (severity == BSG_SEVERITY_ERR) {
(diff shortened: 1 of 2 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -273,14 +290,13 @@ void bugsnag_set_user_env(JNIEnv *env, char *id, char *email, char *name) {
   bsg_safe_delete_local_ref(env, jemail);
   bsg_safe_delete_local_ref(env, jname);
 
-  goto exit;
-
   exit:
   bsg_safe_delete_local_ref(env, interface_class);
+  bsg_safe_delete_local_ref(env, set_user_method);
```
fractalwrench, line 295: This can result in a crash

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -347,12 +363,12 @@ void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message,
   bsg_safe_call_static_void_method(env, interface_class,
                                    leave_breadcrumb_method, jmessage, jtype);
 
-  goto exit;
-
   exit:
   bsg_safe_release_byte_array_elements(env, jmessage, (jbyte *)message);
+  bsg_safe_delete_local_ref(env, jmessage);
   bsg_safe_delete_local_ref(env, interface_class);
+  bsg_safe_delete_local_ref(env, leave_breadcrumb_method);
```
fractalwrench, line 370: This can result in a crash

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -347,12 +363,12 @@ void bugsnag_leave_breadcrumb_env(JNIEnv *env, char *message,
...
   bsg_safe_delete_local_ref(env, type_class);
+  bsg_safe_delete_local_ref(env, crumb_type);
```
fractalwrench, line 372: This can result in a crash

fractalwrench (changes requested):
Noticed a few places where jmethodID and jfieldID are passed to `DeleteLocalRef` - this can lead to crashes so should be avoided. It should be possible to detect when this is occurring by running the changeset on an emulator and sending a handled C++ error from the example app.

fractalwrench (approved):

PR #1161, opened 2021-02-23
Setup gwp-asan for CI fixture

fractalwrench:
## Goal

Sets up [GWP-ASan](https://developer.android.com/ndk/guides/gwp-asan) so that is enabled in our mazerunner test fixture. If memory is misused in the NDK layer this tool should detect this by randomly sampling memory usage, and raise SIGABRT. 

This should provide us insight into whether there are use-after free/buffer overflows problems in the use of our NDK code when the Android 11 tests are run. By extension this should find any latent issues in the implementation of v4 too.

### DIFF
--- a/features/fixtures/mazerunner/app/src/main/AndroidManifest.xml
+++ b/features/fixtures/mazerunner/app/src/main/AndroidManifest.xml
@@ -2,9 +2,12 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.bugsnag.android.mazerunner">
 
+    <!-- always enable GWP ASAN tool to detect native memory bugs. see
+        https://developer.android.com/ndk/guides/gwp-asan-->
     <application
         android:label="MazeRunner"
         android:networkSecurityConfig="@xml/network_security_config"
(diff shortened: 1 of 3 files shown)


features/fixtures/mazerunner/app/src/main/AndroidManifest.xml
```diff
@@ -2,9 +2,12 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.bugsnag.android.mazerunner">
 
+    <!-- always enable GWP ASAN tool to detect native memory bugs. see
```
twometresteve, line 5: I'd suggest adding a similar comment above the Android 11 steps in the pipeline, to make it stand out that we're using gwp-asan and that failures could occur as a result, but that the device logs need to be inspected to determine why.

twometresteve (approved):

features/fixtures/mazerunner/app/src/main/AndroidManifest.xml
fractalwrench: Added a comment to the Android 11 steps in the pipeline yml.

PR #1163, opened 2021-02-25
Update Unity branch to use Maze Runner v3

twometresteve:
## Goal

Update the branch used by the Unity notifier to use Maze Runner v3, allowing e2e test scenarios to be easily executed locally.

## Design

The headline number of changes file looks horrendous, but I have carefully performed a series of commits to make reviewing straightforward:
1. Removes the entire `features` directory, which consisted of the Maze v1 test fixture and scenarios
2. Promotes the Maze v2 tests from `tests/features` to `features` (simple git mv operation only)
3. Makes the path changes needed for point 2

The remaining commits then make the changes necessary to move from v2 to v3.  Shift-clicking all but the first 2 commits on Github's Files Changed tab will produce a meaningful diff for review.

## Changeset

The Android sizer steps are now skipped because they got broken slightly and it's not really worth spending the time fixing them right now based on what we're going to use the branch for.

## Testing

Covered by CI and I've run several different scenarios locally.


fractalwrench (approved):
LGTM

PR #1165, opened 2021-02-26
Persist LastRunInfo and populate app.isLaunching for JVM errors

fractalwrench:
## Goal

Persists the `LastRunInfo` so that previous launch information can be retrieved by the user. This also sets the value of `app.isLaunching` on JVM events.

The implementation of this behaviour for NDK errors will be done in a separate changeset to ease review.

## Changeset

- Added `LaunchCrashTracker`, which tracks the state of whether the application is currently launching or not
- Created `LastRunInfoStore`, which loads and saves the `LastRunInfo` object from disk in the key-value format
- Reset the `LastRunInfo` to the defaults on `Client` initialization
- Set the `LastRunInfo` whenever an unhandled exception occurred so that the next launch allows the user to check that the app crashed

## Testing

Added unit tests for new classes. Updated JVM events to assert that the `app.isLaunching` field is set to the correct default.

E2E tests for the NDK and when `markLaunchCompleted()` are called will be added in a separate PR.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -648,7 +648,7 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
         int consecutiveLaunchCrashes = lastRunInfo == null ? 0
                 : lastRunInfo.getConsecutiveLaunchCrashes();
         boolean launching = launchCrashTracker.isLaunching();
-        LastRunInfo runInfo = new LastRunInfo(consecutiveLaunchCrashes, true, launching);
+        LastRunInfo runInfo = new LastRunInfo(consecutiveLaunchCrashes + 1, true, launching);
         lastRunInfoStore.persist(runInfo);
     }
 


bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -82,6 +82,9 @@
     private PluginClient pluginClient;
 
     final Notifier notifier = new Notifier();
+    LastRunInfo lastRunInfo;
```
tomlongridge, line 85: This jumps out as having no modifier – any reason it can't be private and final?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -283,6 +292,25 @@ public void run() {
...
+        try {
+            Async.run(new Runnable() {
+                @Override
+                public void run() {
+                    lastRunInfo = lastRunInfoStore.load();
+                    lastRunInfoStore.persist(new LastRunInfo(0, false, false));
+                }
+            });
+        } catch (RejectedExecutionException ignored) {
+            // swallow this case, unlikely to happen outside of tests
```
tomlongridge, line 312: Shouldn't this at least be logged?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -623,6 +651,12 @@ void notifyUnhandledException(@NonNull Throwable exc, Metadata metadata,
         Metadata data = Metadata.Companion.merge(metadataState.getMetadata(), metadata);
         Event event = new Event(exc, immutableConfig, handledState, data, logger);
         populateAndNotifyAndroidEvent(event, null);
+
+        // persist LastRunInfo so that on relaunch users can check the app crashed
+        int consecutiveLaunchCrashes = lastRunInfo == null ? 0
+                : lastRunInfo.getConsecutiveLaunchCrashes();
+        new LastRunInfo(consecutiveLaunchCrashes, true, false);
+        lastRunInfoStore.persist(lastRunInfo);
```
tomlongridge, line 659: Unless I'm missing something here, you're not updating `lastRunInfo` so it's just being re-persisted with the old values.

bugsnag-android-core/src/main/java/com/bugsnag/android/LastRunInfoStore.kt
```diff
@@ -0,0 +1,90 @@
...
+            return null
+        }
+
+        return try {
+            val consecutiveLaunchCrashes = lines[0].asIntValue(KEY_CONSECUTIVE_LAUNCH_CRASHES)
+            val crashed = lines[1].asBooleanValue(KEY_CRASHED)
+            val crashedDuringLaunch = lines[2].asBooleanValue(KEY_CRASHED_DURING_LAUNCH)
+            LastRunInfo(consecutiveLaunchCrashes, crashed, crashedDuringLaunch)
+        } catch (ignored: NumberFormatException) {
+            null // unlikely case where information was serialized incorrectly
```
tomlongridge, line 70: Why not also log just in case.

bugsnag-android-core/src/main/java/com/bugsnag/android/LaunchCrashTracker.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    init {
+        val delay = config.launchDurationMillis
+
+        if (delay > 0) {
+            executor.executeExistingDelayedTasksAfterShutdownPolicy = false
+            try {
+                executor.schedule({ markLaunchCompleted() }, delay, TimeUnit.MILLISECONDS)
+            } catch (ignored: RejectedExecutionException) {
+                // ignore v. unlikely rejection failure on new executor
```
tomlongridge, line 26: We have access to the logger via config – we may as well log

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -623,6 +651,12 @@ void notifyUnhandledException(@NonNull Throwable exc, Metadata metadata,
         Metadata data = Metadata.Companion.merge(metadataState.getMetadata(), metadata);
         Event event = new Event(exc, immutableConfig, handledState, data, logger);
         populateAndNotifyAndroidEvent(event, null);
+
+        // persist LastRunInfo so that on relaunch users can check the app crashed
+        int consecutiveLaunchCrashes = lastRunInfo == null ? 0
+                : lastRunInfo.getConsecutiveLaunchCrashes();
+        new LastRunInfo(consecutiveLaunchCrashes, true, false);
```
tomlongridge, line 658: Is the setting of the data coming in a future PR or should this have computed values for the count (incremented) and whether it's at launch?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: This would generate a [synthetic accessor](https://medium.com/thoughts-overflow/effects-of-javas-synthetic-accessor-methods-in-android-bb67b3bac22e) as it's accessed from an inner class, which increases the library size unnecessarily. Our static analysis would fail if this were declared as private. The loading also needs to be asynchronous to avoid unnecessary work on the main thread, so it's not possible to declare it as final.

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: Updated to add log message

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
fractalwrench: Updated implementation so that it now persists the correct values.

fractalwrench:
Updated in response to review comments.

tomlongridge (approved):

PR #1166, opened 2021-03-01
Move the mutex unlock back to the end of deliverReportAtPath

kstenerud:
## Goal

v5.6.0 JNI fixes moved the mutex unlock code of deliverReportAtPath to before the function's resources were freed. This PR moves it back to the end of the function, and also fixes a potential deadlock.

## Testing

Re-ran e2e tests

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -151,75 +151,81 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
     JNIEnv *env, jobject _this, jstring _report_path) {
   static pthread_mutex_t bsg_native_delivery_mutex = PTHREAD_MUTEX_INITIALIZER;
   pthread_mutex_lock(&bsg_native_delivery_mutex);
-  const char *event_path = bsg_safe_get_string_utf_chars(env, _report_path);
-  if (event_path == NULL) {
-    return;
-  }
-  bugsnag_event *event = bsg_deserialize_event_from_file((char *)event_path);
+
+  bugsnag_event *event = NULL;
   jbyteArray jpayload = NULL;
+  jclass interface_class = NULL;
   jbyteArray jstage = NULL;
+  jstring japi_key = NULL;
   char *payload = NULL;
+  const char *event_path = NULL;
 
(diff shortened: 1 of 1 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -151,75 +151,79 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
...
     BUGSNAG_LOG("Failed to read event at file: %s", event_path);
+    goto exit;
+  }
+  payload = bsg_serialize_event_to_json_string(event);
+  if (payload == NULL) {
+    BUGSNAG_LOG("Failed to serialize event as JSON: %s", event_path);
+    goto exit;
+  }
+
+  jclass interface_class =
```
fractalwrench, line 178: Should this be cleaned up in the exit block too with `DeleteLocalRef()`?

fractalwrench (approved):

fractalwrench (approved):

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -151,75 +151,81 @@ Java_com_bugsnag_android_ndk_NativeBridge_deliverReportAtPath(
...
     free(payload);
   }
-  bsg_safe_release_byte_array_elements(env, jstage,
-                                       (jbyte *)event->app.release_stage);
+  bsg_safe_delete_local_ref(env, japi_key);
+  bsg_safe_delete_local_ref(env, interface_class);
   bsg_safe_delete_local_ref(env, jpayload);
   bsg_safe_delete_local_ref(env, jstage);
   bsg_safe_release_string_utf_chars(env, _report_path, event_path);
+  pthread_mutex_unlock(&bsg_native_delivery_mutex);
```
kattrali, line 228: The mutex was intended to avoid any possibility of attempting to deserialize and send the same report file twice, in the unlikely case that bugsnag clients on multiple threads initialized. What additional protection does moving it provide? As far as I can see, all of the references being cleaned up are local, and the position of the lock makes no difference.

kattrali:
There seems to be a lot of style-related changes here that could wait. As far as I can tell, there are 2-3 lines here that address a potential use-after-free and null deref.

PR #1176, opened 2021-03-08
[PLAT-5874] Add assertions for secondary error handlers using logs

Cawllec:
## Goal

Adds an assertion that a secondary error handler will be called correctly, using the log assertions added recently

### DIFF
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
@@ -1,7 +1,6 @@
 package com.bugsnag.android.mazerunner.scenarios
 
 import android.content.Context
-import android.util.Log
 import com.bugsnag.android.Configuration
 
 /**
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
@@ -13,8 +13,6 @@ internal class CrashHandlerScenario(
     eventMetadata: String
 ) : Scenario(config, context, eventMetadata) {
 
-    private var loggedMessages: MutableList<String> = mutableListOf()
-
     init {
         config.autoTrackSessions = false
     }
@@ -23,14 +21,13 @@ internal class CrashHandlerScenario(
         super.startScenario()
         val previousHandler = Thread.getDefaultUncaughtExceptionHandler()
 
-        Thread.setDefaultUncaughtExceptionHandler({ t, e ->
-            Log.d("Bugsnag", "CrashHandlerScenario: Intercepted uncaught exception")
-            loggedMessages.add("CrashHandlerScenario: Intercepted uncaught exception")
+        Thread.setDefaultUncaughtExceptionHandler { t, e ->
+            config.logger.d("Bugsnag", "CrashHandlerScenario: Intercepted uncaught exception")
             previousHandler.uncaughtException(t, e)
-        })
+        }
         throw RuntimeException("CrashHandlerScenario")
     }
 
     override fun getInterceptedLogMessages() =
-        loggedMessages
+        listOf("CrashHandlerScenario: Intercepted uncaught exception")
 }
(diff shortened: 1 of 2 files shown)


features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
```diff
@@ -22,9 +24,13 @@ internal class CrashHandlerScenario(
...
         Thread.setDefaultUncaughtExceptionHandler({ t, e ->
-            Log.d("Bugsnag", "Intercepted uncaught exception")
+            Log.d("Bugsnag", "CrashHandlerScenario: Intercepted uncaught exception")
+            loggedMessages.add("CrashHandlerScenario: Intercepted uncaught exception")
             previousHandler.uncaughtException(t, e)
         })
         throw RuntimeException("CrashHandlerScenario")
     }
+
+    override fun getInterceptedLogMessages() =
```
fractalwrench, line 34: While this should work, there are a couple of enhancements I'd propose which are generally useful when writing Kotlin:

1. `val` can be preferred over `var` because it means the property won't change from its initial value, resulting in less moving parts
2. `List` should be preferred over `MutableList` as it guarantees the elements won't change after initialization

With that in mind I'd suggest the following approach:

```suggestion
    override fun getInterceptedLogMessages() = listOf("CrashHandlerScenario: Intercepted uncaught exception")
```

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
```diff
@@ -22,9 +24,13 @@ internal class CrashHandlerScenario(
         val previousHandler = Thread.getDefaultUncaughtExceptionHandler()
 
         Thread.setDefaultUncaughtExceptionHandler({ t, e ->
-            Log.d("Bugsnag", "Intercepted uncaught exception")
+            Log.d("Bugsnag", "CrashHandlerScenario: Intercepted uncaught exception")
```
fractalwrench, line 27: Unfortunately the mazerunner fixture can only intercept messages routed through `config.logger`, so the E2E scenario won't detect this log message. You could create a property from `config.logger` and then log the message using that instead, which should resolve the issue.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
```diff
@@ -22,9 +24,13 @@ internal class CrashHandlerScenario(
         val previousHandler = Thread.getDefaultUncaughtExceptionHandler()
 
         Thread.setDefaultUncaughtExceptionHandler({ t, e ->
```
fractalwrench, line 26: Also a (pre-existing) pet peeve is that the lambda can be taken out of the parentheses, which is the equivalent functionality but a simpler syntax:

```
Thread.setDefaultUncaughtExceptionHandler { t, e ->
    Log.d("Bugsnag", "Intercepted uncaught exception")
    previousHandler.uncaughtException(t, e)
}
```

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
Cawllec: I'm not sure I understand this approach, as it feels like it's shortcutting the actual test.
In this case only adding the log to the list once the error handler fires would be the whole point?

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
Cawllec: This isn't for logging to Maze-runner, just for debugging in general.  I just updated the message

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CrashHandlerScenario.kt
Cawllec: Ah, this was a misunderstanding on my part, I'll be resolving this shortly.

fractalwrench (approved):

PR #1179, opened 2021-03-09
6178 backport anr fixes

kstenerud:
## Goal

We're seeing the same kinds of JVM lockups in the ANR handler in the unity branch, so backport the ANR fix.

## Testing

Re-run e2e tests


bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -1,119 +1,224 @@
...
+        BUGSNAG_LOG("Failed to call JNIEnv->AttachCurrentThread(): %d", result);
+        return;
       }
-    } // All other results are error codes
-  }
-
-  // Invoke previous handler, if a custom handler has been set
-  // This can be conditionally switched off on certain platforms (e.g. Unity) where
-  // this behaviour is not desirable due to Unity's implementation failing with a SIGSEGV
-  if (invokePrevHandler) {
```
fractalwrench, line 67: Does this trigger a SIGSEGV on Unity when an ANR happens?

fractalwrench (commented):
The changes look good but I have a concern based off the comments that this could trigger a SIGSEGV in a Unity app when ANRs occur.

It'd be good to test out ANRs in the Unity example app to confirm that they work as expected and that no untoward behaviour is happening.

bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
kstenerud: Verified that it doesn't trigger a segfault.

Since we block `SIGQUIT` now, there's no way for any handler Unity installs to get called.

fractalwrench (approved):
LGTM pending CI

PR #1180, opened 2021-03-09
Propagate changes in app.isLaunching to the NDK layer

fractalwrench:
## Goal

Currently `app.isLaunching` is only set on JVM error reports. This changeset adds it to NDK error reports and adds verification that it has been set for both in the form of E2E tests.

## Changeset

- `LaunchCrashTracker` extends `BaseObservable` and notifies its observers when `isLaunching` is set to false
- The `NativeBridge` which communicates with the NDK layer invokes a JNI function which updates `bsg_event->app.is_launching` to false
- Fixed JVM serialization bug where `inForeground` rather than `isLaunching` value was serialized
- Added `is_launching` to `bsg_app_info` struct
- Added `bugsnag_app_get_is_launching/bugsnag_app_set_is_launching` functions to allow users to alter the value in an on_error callback.
 
### Struct migration

When a fatal NDK error occurs, the memory contents of `bsg_event` struct is persisted on disk and then loaded again the next launch. Because the struct can change between releases, old payloads need to be migrated to the new struct. This has been achieved by the following changes:

- Check whether `BUGSNAG_EVENT_VERSION` is 4 and load it into `bugsnag_report_v4`, which is a copy of the previous struct definition
- Check the `BUGSNAG_EVENT_VERSION` in the serializer and migrated the `bsg_app_info` values
- Set a default value for `app.is_launching` to false if no information was available in the struct

## Testing

- Added/updated unit test coverage where appropriate, particularly for migrating the struct
- Verified that `app.isLaunching` is set appropriately in JVM/NDK E2E tests

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.c
@@ -324,6 +324,16 @@ void bugsnag_app_set_in_foreground(void *event_ptr, bool value) {
   event->app.in_foreground = value;
 }
 
+bool bugsnag_app_get_is_launching(void *event_ptr) {
+  bugsnag_event *event = (bugsnag_event *)event_ptr;
+  return event->app.is_launching;
+}
+
+void bugsnag_app_set_is_launching(void *event_ptr, bool value) {
+  bugsnag_event *event = (bugsnag_event *)event_ptr;
+  event->app.is_launching = value;
+}
+
 /* Accessors for event.device */
 
 bool bugsnag_device_get_jailbroken(void *event_ptr) {
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
@@ -18,13 +18,14 @@ extern "C" {
 bool bsg_event_write(bsg_report_header *header, bugsnag_event *event, int fd);
 
 bugsnag_event *bsg_event_read(int fd);
-bugsnag_event *bsg_report_v4_read(int fd);
 bsg_report_header *bsg_report_header_read(int fd);
+bugsnag_event *bsg_map_v4_to_report(bugsnag_report_v4 *report_v4);
 bugsnag_event *bsg_map_v3_to_report(bugsnag_report_v3 *report_v3);
 bugsnag_event *bsg_map_v2_to_report(bugsnag_report_v2 *report_v2);
 bugsnag_event *bsg_map_v1_to_report(bugsnag_report_v1 *report_v1);
 
 void migrate_app_v1(bugsnag_report_v2 *report_v2, bugsnag_report_v3 *event);
+void migrate_app_v2(bugsnag_report_v4 *report_v4, bugsnag_event *event);
 void migrate_device_v1(bugsnag_report_v2 *report_v2, bugsnag_report_v3 *event);
 void migrate_breadcrumb_v1(bugsnag_report_v2 *report_v2,
                            bugsnag_report_v3 *event);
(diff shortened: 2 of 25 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -324,6 +324,16 @@ void bugsnag_app_set_in_foreground(void *event_ptr, bool value) {
   event->app.in_foreground = value;
 }
 
+bool bugsnag_app_get_is_launching(void *event_ptr) {
+  bugsnag_event *event = (bugsnag_event *)event_ptr;
+  return event->app.in_foreground;
```
twometresteve, line 329: Copy/paste error, replicating the bug you fixed it seems!  Same below.

features/full_tests/batch_1/identify_crashes_on_launch.feature
```diff
@@ -0,0 +1,31 @@
...
+        And the exception "message" equals "JvmMarkLaunchCompletedScenario"
+        And the event "app.isLaunching" is false
+
+    Scenario: An NDK error captured after markLaunchComplete() is false for app.isLaunching
+        When I run "CXXMarkLaunchCompletedScenario" and relaunch the app
+        And I configure Bugsnag for "CXXMarkLaunchCompletedScenario"
+        Then I wait to receive an error
+        And the error is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+        And the exception "message" equals "Abort program"
+        And the event "app.isLaunching" is false
```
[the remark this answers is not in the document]
fractalwrench: Updated E2E tests to always check two reports.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -118,8 +131,42 @@ bugsnag_event *bsg_event_read(int fd) {
...
+    strcpy(event->session_id, report_v4->session_id);
+    strcpy(event->session_start, report_v4->session_start);
+    event->handled_events = report_v4->handled_events;
+    event->unhandled_events = report_v4->unhandled_events;
+    strcpy(event->grouping_hash, report_v4->grouping_hash);
+    event->unhandled = report_v4->unhandled;
+    bsg_strncpy_safe(event->api_key, report_v4->api_key,
+                     sizeof(event->api_key));
+    migrate_app_v2(report_v4, event);
+    free(report_v4);
```
kstenerud, line 169: It feels jarring having the free inside here, since it's not obvious to the caller that their param will get freed as part of the operation...

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -118,8 +131,42 @@ bugsnag_event *bsg_event_read(int fd) {
...
+    event->user = report_v4->user;
+    event->error = report_v4->error;
+    event->metadata = report_v4->metadata;
+    event->crumb_count = report_v4->crumb_count;
+    event->crumb_first_index = report_v4->crumb_first_index;
+    memcpy(event->breadcrumbs, report_v4->breadcrumbs,
+           sizeof(event->breadcrumbs));
+    event->severity = report_v4->severity;
+    strcpy(event->session_id, report_v4->session_id);
+    strcpy(event->session_start, report_v4->session_start);
```
kstenerud, line 161: If we ever shorten the size of any `event->xyz`, these `strcpy` calls will overwrite memory.

kstenerud (changes requested):

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: Updated to use `bsg_strncpy_safe` instead

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
fractalwrench: Added some comments around this after in-person discussion.

twometresteve (approved):
LGTM so long as @kstenerud is now happy.

kstenerud (approved):

PR #1181, opened 2021-03-10
Add E2E tests for crash loop scenarios

fractalwrench:
## Goal

Adds E2E scenarios where Bugsnag allows an application to recover from a crash loop by checking the value of `Bugsnag.getLastRunInfo()`. Along with #1180 this adds the required E2E test coverage for the project.

## Changeset

- Added scenario code which triggers a fatal JVM + NDK crash loop by crashing as soon as the scenario is run
- If `Bugsnag.getLastRunInfo().crashedDuringLaunch` is true, the application enables a 'safe-mode' which avoids the crashy path
- Confirmed that the values of `LastRunInfo` are populated as expected.

Note: the NDK `LastRunInfo` functionality hasn't been implemented yet so the scenario fails. This will be addressed in PLAT-5978 which will uncomment the E2E test.

### DIFF
--- a/features/fixtures/mazerunner/cxx-scenarios-bugsnag/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXCrashLoopScenario.kt
+++ b/features/fixtures/mazerunner/cxx-scenarios-bugsnag/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXCrashLoopScenario.kt
@@ -1,0 +1,51 @@
+package com.bugsnag.android.mazerunner.scenarios
+
+import android.content.Context
+import com.bugsnag.android.Bugsnag
+import com.bugsnag.android.Configuration
+import com.bugsnag.android.OnErrorCallback
+
+/**
+ * Triggers a crash loop which Bugsnag allows recovery from.
+ */
+internal class CXXCrashLoopScenario(
+    config: Configuration,
+    context: Context,
+    eventMetadata: String?
+) : Scenario(config, context, eventMetadata) {
+
+    init {
+        config.autoTrackSessions = false
+        System.loadLibrary("bugsnag-ndk")
+        System.loadLibrary("cxx-scenarios-bugsnag")
+        config.addOnError(
+            OnErrorCallback { event ->
+                Bugsnag.getLastRunInfo()?.let {
+                    event.addMetadata("LastRunInfo", "crashed", it.crashed)
+                    event.addMetadata("LastRunInfo", "crashedDuringLaunch", it.crashedDuringLaunch)
+                    event.addMetadata(
+                        "LastRunInfo",
+                        "consecutiveLaunchCrashes",
+                        it.consecutiveLaunchCrashes
+                    )
+                }
+                true
+            }
+        )
+    }
+
+    external fun crash()
+
+    override fun startScenario() {
+        super.startScenario()
+        val lastRunInfo = Bugsnag.getLastRunInfo()
+
+        // the last run info allows the scenario to escape from what would otherwise be
+        // a crash loop, by conditionally entering a 'safe mode'.
+        if (lastRunInfo?.crashedDuringLaunch == true) {
+            Bugsnag.notify(IllegalArgumentException("Safe mode enabled"))
+        } else {
+            crash()
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


features/fixtures/mazerunner/cxx-scenarios-bugsnag/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXCrashLoopScenario.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    external fun crash()
+
+    override fun startScenario() {
+        super.startScenario()
+        val lastRunInfo = Bugsnag.getLastRunInfo()
+
+        // the last run info allows the scenario to escape from what would otherwise be
+        // a crash loop, by conditionally entering a 'safe mode'.
+        if (lastRunInfo?.crashedDuringLaunch == true) {
```
twometresteve, line 45: Is the `== true` necessary in Kotlin?

features/full_tests/batch_1/identifying_crashes_on_launch.feature
```diff
@@ -0,0 +1,36 @@
+Feature: Identifying crashes on launch
+
+    Scenario: Escaping from a crash loop by reading LastRunInfo in a JVM error
+        When I run "JvmCrashLoopScenario" and relaunch the app
+        And I configure Bugsnag for "JvmCrashLoopScenario"
```
twometresteve, line 5: I think this step can be removed?  Although it typically goes with the previous, the following step will do the same thing again (I believe).  Same in the NDK scenario.

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/JvmCrashLoopScenario.kt
```diff
@@ -0,0 +1,47 @@
...
+        )
+    }
+
+    override fun startScenario() {
+        super.startScenario()
+        val lastRunInfo = Bugsnag.getLastRunInfo()
+
+        // the last run info allows the scenario to escape from what would otherwise be
+        // a crash loop, by conditionally entering a 'safe mode'.
+        if (lastRunInfo?.crashedDuringLaunch == true) {
```
twometresteve, line 41: Same question.

features/full_tests/batch_1/identifying_crashes_on_launch.feature
```diff
@@ -0,0 +1,36 @@
...
+        And the event "metaData.LastRunInfo.consecutiveLaunchCrashes" is null
+        And I discard the oldest error
+        Then the error is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
+        And the exception "message" equals "Safe mode enabled"
+        And the event "metaData.LastRunInfo.crashed" is true
+        And the event "metaData.LastRunInfo.crashedDuringLaunch" is true
+        And the event "metaData.LastRunInfo.consecutiveLaunchCrashes" equals 1
+
+#    Awaiting implementation of PLAT-5978
+#    Scenario: Escaping from a crash loop by reading LastRunInfo in an NDK error
```
twometresteve, line 21: When the time comes to enable this, I think you could just add some steps to the above scenario, rather than creating a new one.  I.e. JVM crash -> NDK crash -> Safe mode.  This would save a bit of test run time and also verify that consecutiveLaunchCrashes is incremented beyond 1.

twometresteve (approved):

features/fixtures/mazerunner/cxx-scenarios-bugsnag/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXCrashLoopScenario.kt
fractalwrench: In this case yes, as the expression is of type `Boolean?` because `lastRunInfo` can be null. Therefore it's necessary to do the explicit comparison.

features/full_tests/batch_1/identifying_crashes_on_launch.feature
fractalwrench: Added a note to do this onto the ticket, so that this can get merged down quickly.

PR #1182, opened 2021-03-10
Add BackgroundTaskService to replace Async.java

fractalwrench:
## Goal

Adds `BackgroundTaskService` with the intent of replacing the `Async` class. For general motivations of why this is necessary, please read [the following](https://github.com/bugsnag/bugsnag-android/pull/1172#issue-584912344).

## Changeset

- Added `BackgroundTaskService` class which will have the responsibility of controlling how all lengthy tasks are run in the background
- Created `submitTask` method to which callers can submit a `Runnable` or `Callable` and receive a `Future` in return
- Created a `TaskType` enumeration that determines what single-threaded executor the task is executed on
- Added a more nuanced shutdown strategy that only waits for IO/Error/Session requests to complete on process termination
- Diverged from the `Async` implementation by ensuring that each executor is single-threaded (this gives a deterministic request order) and only executes a specific task type (short-running operations aren't blocked by long running ones)
- Created a single-threaded executor with a max queue size of 128. These are essentially the same defaults as were used in `Async`.

## Testing

Added unit tests to verify that the service behaves as expected. Additionally the service will have general E2E/integration test coverage when the `Client` is converted to use the service in future PRs.

Manual/stress testing of the delivery mechanism will be performed before UAT which should additionally cover some of this functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   [#1159](https://github.com/bugsnag/bugsnag-android/pull/1159)
   [#1165](https://github.com/bugsnag/bugsnag-android/pull/1165)
   [#1164](https://github.com/bugsnag/bugsnag-android/pull/1164)
+  [#1182](https://github.com/bugsnag/bugsnag-android/pull/1182)
 
 ## 5.7.0 (2021-02-18)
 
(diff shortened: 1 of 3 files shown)


twometresteve (commented):
Seems reasonable to me.

kstenerud (approved):

PR #1184, opened 2021-03-10
Update SessionTracker to use BackgroundTaskService

fractalwrench:
## Goal

Updates `SessionTracker` to use `BackgroundTaskService` rather than `Async`. This allows for better control over asynchronous operations and negates the need for a `Semaphore` which was a source of non-deterministic behaviour that could lead to some session requests not immediately being sent. This is based on changes made in #1182.

## Changeset

- Create a `BackgroundTaskService` in the `Client`
- Updated `SessionTracker` to require a `BackgroundTaskService` constructor param
- Refactored `SessionTracker` to submit two types of task to the session request executor: flushing stored sessions from disk, and flushing in-memory sessions

## Testing

Added a unit test to confirm that requests on the session tracker are confined to a thread, otherwise relied on existing E2E test coverage for sessions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   [#1165](https://github.com/bugsnag/bugsnag-android/pull/1165)
   [#1164](https://github.com/bugsnag/bugsnag-android/pull/1164)
   [#1182](https://github.com/bugsnag/bugsnag-android/pull/1182)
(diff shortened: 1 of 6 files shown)


kstenerud (approved):

twometresteve (approved):
LGTM

PR #1185, opened 2021-03-10
Deliver events using BackgroundTaskService

fractalwrench:
## Goal

Updates the event delivery mechanism to use `BackgroundTaskService` rather than `Async`. This allows for better control over asynchronous operations and negates the need for a `Semaphore` which was a source of non-deterministic behaviour that could lead to some session requests not immediately being sent. This is based on changes made in #1182.

## Changeset

- Converted `DeliveryDelegate` and `EventStore` to use `BackgroundTaskService`
- Altered `ManualSessionSmokeTest` to account for deterministic behaviour, and avoid flaky delivery limitations.
- Called `bgTaskService.shutdown()` after a fatal JVM error, which avoids dropping in-flight requests

## Testing

Added a unit test to confirm that event requests are confined to a thread, otherwise relied on existing E2E test coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   [#1164](https://github.com/bugsnag/bugsnag-android/pull/1164)
   [#1182](https://github.com/bugsnag/bugsnag-android/pull/1182)
   [#1184](https://github.com/bugsnag/bugsnag-android/pull/1184)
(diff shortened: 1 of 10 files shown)


kstenerud (approved):

twometresteve (approved):
LGTM

PR #1186, opened 2021-03-10
Replace Async with BackgroundTaskService

fractalwrench:
## Goal

Replaces `Async` with `BackgroundTaskService` completely, which concludes the changes started in #1182.

## Changeset

- Removed `Async` class
- Removed obsolete `TestRunListener` class
- Bumped `LaunchCrashTrackerTest` sleep threshold to 20ms to stop timeout on CI observed once
- Internal error reports are now sent on their own executor
- `SystemBroadcastReceiver` is now registered on its own executor
- Refactored `Client` to load the `LastRunInfo` on the IO executor rather than `Async`

## Testing

Relied on existing E2E test coverage.


kstenerud (approved):

twometresteve (approved):
LGTM

PR #1188, opened 2021-03-11
Update synchronous crash reporting logic

fractalwrench:
## Goal

Updates the existing synchronous reporting of launch crashes so that they fit the behaviour of the new spec. Specifically, synchronous delivery should only be attempted for the most recent startup crash.

## Changeset

- "_startupcrash" is appended to the filename based on whether `app.isLaunching` is set to true, rather than the previous `app.duration` exceeding a threshold
- Refactored `flushOnLaunch()` so that it creates a `Future` and blocks for 2 seconds calling `get()`. This avoids a busy wait that the previous implementation used
- Altered `EventStore` so that it only looks for the most recent startupcrash, as determined by the timestamp and "_startupcrash_ suffix encoded in the filename
- Fixed a bug where .json was encoded in the filename twice (this does not affect any functionality)
- `flushOnLaunch()` no longer calls `flushAsync()` to improve testability

## Testing

- Updated existing test coverage where necessary
- Added new unit tests for logic around finding a launch crash
- Created tests that validate what thread event delivery occurs on, and confirms that `flushOnLaunch()` blocks the main thread for a short duration

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
@@ -1,16 +1,20 @@
 package com.bugsnag.android;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import java.io.File;
-import java.lang.Thread;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /**
  * Store and flush Event reports which couldn't be sent immediately due to
@@ -19,13 +23,11 @@ import java.util.concurrent.RejectedExecutionException;
 class EventStore extends FileStore {
 
     private static final long LAUNCH_CRASH_TIMEOUT_MS = 2000;
-    private static final int LAUNCH_CRASH_POLL_MS = 50;
 
-    volatile boolean flushOnLaunchCompleted = false;
     private final ImmutableConfig config;
     private final Delegate delegate;
     private final Notifier notifier;
-    private final BackgroundTaskService backgroundTaskService;
+    private final BackgroundTaskService bgTaskSevice;
     final Logger logger;
 
     static final Comparator<File> EVENT_COMPARATOR = new Comparator<File>() {
@@ -47,7 +49,7 @@ class EventStore extends FileStore {
     EventStore(@NonNull ImmutableConfig config,
                @NonNull Logger logger,
                Notifier notifier,
-               BackgroundTaskService backgroundTaskService,
+               BackgroundTaskService bgTaskSevice,
                Delegate delegate) {
         super(new File(config.getPersistenceDirectory(), "bugsnag-errors"),
                 config.getMaxPersistedEvents(),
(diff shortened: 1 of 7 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java
```diff
@@ -58,64 +60,78 @@ public int compare(File lhs, File rhs) {
...
-            } else {
-                logger.d("No startupcrash events to flush to Bugsnag.");
+        // cancel non-launch crash reports
+        if (launchCrashReport != null) {
+            storedFiles.remove(launchCrashReport);
+        }
+        cancelQueuedFiles(storedFiles);
+
+        if (launchCrashReport != null) {
+            logger.i("Attempting to send launch crash reports");
```
kattrali, line 105: This message should reflect the new intent of sending only one crash report. This change should avoid any confusion around what happens when there is more than one on disk.

```suggestion
            logger.i("Attempting to send the most recent launch crash report");
```

bugsnag-android-core/src/test/java/com/bugsnag/android/EventFilenameTest.kt
```diff
@@ -49,6 +49,66 @@ internal class EventFilenameTest {
...
+
+        // if multiple crashes exist, pick the most recent one
+        val expected = File("1664219155431_042c6195-a32c-2f84-11ae-77086509ca1e_startupcrash.json")
+        assertEquals(
+            expected,
+            eventStore.findLaunchCrashReport(
+                listOf(
+                    File("1504255147933_30b7e350-dcd1-4032-969e-98d30be62bbc_startupcrash.json"),
+                    File("1404205127135_683c6b92-b325-4987-80ad-77086509ca1e_startupcrash.json"),
+                    expected
```
kattrali, line 106: Since the sorting and selection logic grabs an item from an end of the array, I'd put `expected` in the middle to ensure that it worked correctly. For example, if the following was mistakenly commented out, this test would still pass as-is:

https://github.com/bugsnag/bugsnag-android/blob/c20f08c3ee31db3f7aa77d25d828bcbbb62e9a6d/bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java#L125

```suggestion
                    File("1504255147933_30b7e350-dcd1-4032-969e-98d30be62bbc_startupcrash.json"),
                    expected,
                    File("1404205127135_683c6b92-b325-4987-80ad-77086509ca1e_startupcrash.json")
```

bugsnag-android-core/src/test/java/com/bugsnag/android/LaunchCrashDeliveryTest.kt
```diff
@@ -0,0 +1,136 @@
...
+ */
+class LaunchCrashDeliveryTest {
+
+    private lateinit var storageDir: File
+    private lateinit var errorDir: File
+
+    @Before
+    fun setUp() {
+        storageDir = Files.createTempDirectory("tmp").toFile()
+        storageDir.deleteRecursively()
```
kattrali, line 26: Since [`Files.createTempDirectory()`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory(java.lang.String,%20java.nio.file.attribute.FileAttribute...)) creates a new directory with a different suffix every run, it doesn't look like this line does anything?

bugsnag-android-core/src/test/java/com/bugsnag/android/LaunchCrashDeliveryTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        assertEquals(1, delivery.count.get())
+        assertEquals("Bugsnag Error thread", delivery.threadName)
+
+        // non-launch crashes are not delivered in flushOnLaunch()
+        event.app.isLaunching = false
+        eventStore.write(event)
+        eventStore.flushOnLaunch()
+        assertEquals(1, delivery.count.get())
+
+        // non-launch crashes are not delivered in flushAsync() instead
```
kattrali, line 79:
```suggestion
        // non-launch crashes are delivered in flushAsync() instead
```
?

bugsnag-android-core/src/test/java/com/bugsnag/android/LaunchCrashDeliveryTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        assertTrue(now - baseline > 1500)
+        backgroundTaskService.shutdown()
+        assertEquals(1, delivery.count.get())
+    }
+
+    /**
+     * Crashes on launch are delivered synchronously
+     */
+    @Test
+    fun flushOnLaunchSync() {
```
kattrali, line 60: It looks like there's a missing test case here for generating two launch events and asserting that the correct one is delivered. Perhaps `TestDelivery` could keep the most recent `errorClass` sent or something to test against.

kattrali (commented):
Looks good, added a few questions/small improvements.

bugsnag-android-core/src/test/java/com/bugsnag/android/LaunchCrashDeliveryTest.kt
fractalwrench: Yep, this doesn't seem to be necessary - I've removed this line.

bugsnag-android-core/src/test/java/com/bugsnag/android/LaunchCrashDeliveryTest.kt
fractalwrench: Added an extra test case that asserts this.

kattrali (approved):

PR #1190, opened 2021-03-12
Fix deletion of JvmCrashLoopScenario

fractalwrench:
## Goal

Removes the unused `CXXCrashLoopScenario` and replaced it with `JvmCrashLoopScenario`, which was removed from version control due to a bad merge.


twometresteve (approved):
LGTM

PR #1193, opened 2021-03-17
Refactor test to avoid request flakes

fractalwrench:
## Goal

The `CxxMarkLaunchCompletedScenario` triggers an NDK crash immediately after calling `Bugsnag.notify()`. This triggers a race where the NDK crash can terminate the process before the JVM error has a chance to make a request.

This fixes the scenario to only trigger the NDK crash after the first request has been delivered.


twometresteve (approved):
LGTM.  Tested locally and passed 15 times in a row and no failures.

PR #1194, opened 2021-03-17
Root detection: check whether su binary exists

fractalwrench:
## Goal

Adds code to check whether the su binary exists. If the binary does exist then it's a good indication that the device has been rooted.

## Changeset

Added code to run `which su` and check the command output is non-empty.

## Testing

Added unit tests to verify the function handles edge cases around IOExceptions etc. Additionally tested this check on a few devices to confirm the results are useful:

- Returns false for an API 26 rooted emulator (this is expected because the su binary was not actually installed on the emulator)
- Returns false for an API 29 non-rooted emulator
- Returns true for a Pixel 3a rooted with Magisk, even when Magisk hide (root detection countermeasures) is enabled


twometresteve (approved):
LGTM

PR #1198, opened 2021-03-18
Check for presence of dangerous build properties

fractalwrench:
## Goal

Checks the contents of `/system/build.prop` to see whether it contains dangerous properties that might indicate the phone is using a custom ROM and is therefore rooted.

## Changeset

Checked whether `/system/build.prop` contains `ro.debuggable=[1]` or `ro.secure=[0]`. These indicate that the OS has been built by a 3rd party developer and is a good indication that the device has been rooted.

## Testing

Added unit test coverage for new method. Additionally tested the method behaves as expected on the following devices:

- API 21 emulator, no root: false
- API 28 emulator, root: false
- Pixel 3a rooted with Magisk: false


twometresteve:
> ## Testing
> Added unit test coverage for new method. Additionally tested the method behaves as expected on the following devices:
> 
> * API 21 emulator, no root: false
> * API 28 emulator, root: false
> * Pixel 3a rooted with Magisk: false

Did you mean false in all of these cases?  If so, in which scenario would you expect it to be true?

bugsnag-android-core/src/main/java/com/bugsnag/android/RootDetector.kt
```diff
@@ -39,7 +42,13 @@ internal class RootDetector(
     /**
      * Determines whether the device is rooted or not.
      */
-    fun isRooted() = checkBuildTags() || checkSuExists() || checkRootBinaries()
+    fun isRooted(): Boolean {
+        return try {
+            checkBuildTags() || checkSuExists() || checkBuildProps() || checkRootBinaries()
+        } catch (exc: Throwable) {
```
twometresteve, line 48: Yes, seems sensible.

fractalwrench:
It's expected that the detection via `system/build.prop` would be false for all of the above 3 test cases, as they use stock OS images which don't have these flags enabled. I'd expect to see the method return true in a custom ROM which has those flags enabled, which isn't currently available for our testing (but I believe should be in UAT).

bugsnag-android-core/src/test/java/com/bugsnag/android/RootDetectorTest.kt
```diff
@@ -96,4 +97,133 @@ class RootDetectorTest {
...
+     */
+    @Test
+    fun checkBuildPropsNonRooted() {
+        val tmpFile = Files.createTempFile("regular", ".prop").toFile()
+        tmpFile.writeText(SAMPLE_BUILD_PROPS)
+        assertFalse(RootDetector(buildProps = tmpFile).checkBuildProps())
+    }
+
+    /**
+     * Verifies that the ro.secure build property triggers root detection
```
twometresteve, line 130: Looks like a copy/paste error - should say root detection is not triggered.

bugsnag-android-core/src/main/java/com/bugsnag/android/RootDetector.kt
```diff
@@ -69,6 +78,22 @@ internal class RootDetector(
...
+    /**
+     * Checks the contents of /system/build.prop to see whether it contains dangerous properties.
+     * These properties give a good indication that a phone might be using a custom
+     * ROM and is therefore rooted.
+     */
+    fun checkBuildProps(): Boolean {
+        runCatching {
+            return buildProps.bufferedReader().useLines { lines ->
+                lines.filter { line ->
+                    line.contains("ro.debuggable=[1]") || line.contains("ro.secure=[0]")
```
twometresteve, line 90: How do we rate the risk of a false positive here?  E.g. what if the line was commented out or just the end of a completely unrelated property?  Both are arguably unlikely, but I wonder if we should tighten it up a little?

twometresteve (commented):
Looks good, just some minor comments/questions.

bugsnag-android-core/src/test/java/com/bugsnag/android/RootDetectorTest.kt
fractalwrench: Updated docs to address this

bugsnag-android-core/src/main/java/com/bugsnag/android/RootDetector.kt
fractalwrench: Tweaked detection so that it trims whitespace and checks that the start of the line begins with these properties (which discounts the case where the line is commented out)

twometresteve (approved):
LGTM

PR #1200, opened 2021-03-18
Perform root detection on a background thread

fractalwrench:
## Goal

The implementation should detect root on a background thread when Bugsnag initialises to reduce impact on the main thread. If root detection is halfway complete when an error is captured, then the notifier will block using a `Future` until the root detection is complete.

## Changeset

Invoked root detection on a background thread which returns a `Future<Boolean>`. This will execute immediately in the background, but if an error occurs before root detection can complete, the `Future` will block using `get()` until the value is available for use in the `Device` payload.

If the task can't be scheduled or is interrupted with an exception then `false` will be used as a default value. (This is a highly improbable scenarios as the executor is very unlikely to be saturated with tasks during initialization.) 

## Testing

Manually verified that the root detection occurs on a non-main thread.


bugsnag-android-core/src/main/java/com/bugsnag/android/DeviceDataCollector.kt
```diff
@@ -81,6 +97,17 @@ internal class DeviceDataCollector(
...
 
+    private fun checkIsRooted(): Boolean {
+        if (rootedFuture != null) {
+            return try {
+                rootedFuture.get()
+            } catch (exc: Exception) {
+                false
+            }
+        }
+        return false
```
twometresteve, line 108:
```suggestion
        return try {
            rootedFuture != null && rootedFuture.get()
        } catch (exc: Exception) {
            false
        }
```

twometresteve (approved):

PR #1202, opened 2021-03-19
Bump compileSdkVersion to apiLevel 30

fractalwrench:
## Goal

Bumps the [compileSdkVersion](https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd) to apiLevel 30 from 28. This ensures that our SDK is compiled against the latest version of Android while retaining compatibility for previously supported versions.

## Changeset

- Bumped compileSdkVersion to 30
- Removed unused API 28 system image from CI, reducing docker image size
- Suppressed new deprecations where necessary
- Altered internal code and tests where nullability annotations have been changed between 28/30, e.g. `fun onAvailable(network: Network?)` was altered to `fun onAvailable(network: Network)`. This should have no effect on our public API or customer behaviour.

## Testing

Relied on existing test coverage. Additionally read the [Android 10](https://developer.android.com/about/versions/10/behavior-changes-all#non-sdk-restrictions) and [Android 11](https://developer.android.com/about/versions/11/behavior-changes-all#non-sdk-restrictions) behaviour change list for any differences that might affect us.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 
 ### Enhancements
 
+* Bump compileSdkVersion to apiLevel 30
+  [#1202](https://github.com/bugsnag/bugsnag-android/pull/1202)
+
 * Add public API for crash-on-launch detection
   [#1157](https://github.com/bugsnag/bugsnag-android/pull/1157)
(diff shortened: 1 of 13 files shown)


twometresteve (approved):
LGTM

PR #1203, opened 2021-03-19
Detect non-sdk usages in E2E tests

fractalwrench:
## Goal

Android 9 and onwards [added restrictions](https://developer.android.com/guide/app-compatibility/restrictions-non-sdk-interfaces) on the use of hidden Android APIs, which can trigger warnings/crashes if they are used within an app. This changeset alters the test fixture so that it will throw an exception if any non-sdk API is used by Bugsnag.

## Changeset

Setup `StrictMode` to throw an exception if any non-SDK API is used in the E2E tests.

It's worth noting that I also investigated [veridex](https://developer.android.com/guide/app-compatibility/restrictions-non-sdk-interfaces#test-veridex-tool) for setting up on CI. I don't think it's necessary as `StrictMode` should cover this functionality (I also encountered a segfault when trying to use the latest version of the tool).

## Testing

### DIFF
--- a/features/fixtures/mazerunner/app/src/main/AndroidManifest.xml
+++ b/features/fixtures/mazerunner/app/src/main/AndroidManifest.xml
@@ -8,6 +8,7 @@
         android:label="MazeRunner"
         android:networkSecurityConfig="@xml/network_security_config"
         android:gwpAsanMode="always"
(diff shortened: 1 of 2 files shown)


twometresteve (approved):
LGTM

PR #1209, opened 2021-03-22
Root files detection

kstenerud:
## Goal

https://bugsnag.atlassian.net/browse/PLAT-6106

## Design

Use `lstat` to check for file existence and mode. We can't safely use syscalls since later versions of Android lock down certain functions, and some syscalls aren't directly accessible without guessing their numbers (which could break if they ever change).

## Changeset

Check for existence of a number of files, and writability of other files, per https://github.com/scottyab/rootbeer

## Testing

Ran on jailbroken and non-jailbroken devices (emulated and real)

### DIFF
--- a/bugsnag-android-core/src/main/jni/root_detection.c
+++ b/bugsnag-android-core/src/main/jni/root_detection.c
@@ -5,10 +5,84 @@
 extern "C" {
 #endif
 
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+
+static const char *su_paths[] = {
+        // Common binaries
+        "/system/xbin/su",
+        "/system/bin/su",
+        // < Android 5.0
+        "/system/app/Superuser.apk",
+        "/system/app/SuperSU.apk",
+        // >= Android 5.0
+        "/system/app/Superuser",
+        "/system/app/SuperSU",
+        // Fallback
+        "/system/xbin/daemonsu",
+        // Systemless root
+        "/su/bin"
+};
+static const int su_paths_count = sizeof(su_paths) / sizeof(*su_paths);
+
+static const char *should_not_be_writable[] = {
+        "/sbin",
+};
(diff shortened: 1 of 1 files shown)


bugsnag-android-core/src/main/jni/root_detection.c
```diff
@@ -5,10 +5,88 @@
 extern "C" {
 #endif
 
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+
+static const char *su_paths[] = {
```
fractalwrench, line 12: >If the native library cannot be loaded then these checks should be performed on the JVM.

It might be good to alter the `RootDetector.kt` class so that if the native library isn't loaded, it falls back to the JVM root binary checks, but otherwise skips them. This would avoid redundant work - as currently the JVM/NDK are doing the same thing twice.

I think this could be done by updating `isRooted()` and also this else branch: https://github.com/bugsnag/bugsnag-android/blob/6106-root-file-detection2/bugsnag-android-core/src/main/java/com/bugsnag/android/RootDetector.kt#L139

fractalwrench (approved):
LGTM pending one tweak and CI.

bugsnag-android-core/src/main/jni/root_detection.c
kstenerud: Isn't it doing this already?

```
    fun isRooted(): Boolean {
        return try {
            checkBuildTags() || checkSuExists() || checkBuildProps() || checkRootBinaries() || nativeCheckRoot()
        } catch (exc: Throwable) {
            logger.w("Root detection failed", exc)
            false
        }
    }
```
Since the native library can be blocked, we'd have to do some redundant checks; otherwise it would be all-or-nothing native library loading.

fractalwrench: I'd suggest that the implementation should check if `libraryLoaded.get() == false`, and only call `checkRootBinaries()` if the NDK checks can't occur. This would avoid duplicating the same work.

PR #1210, opened 2021-03-23
Root check: See if we can create a file in /system

kstenerud:
## Goal

PLAT-6108

## Design

Check if we can create the file `/system/bugsnag_test_file.txt`

## Testing

Ran manually, and reran tests


fractalwrench (approved):
LGTM

PR #1211, opened 2021-03-25
Collect whether the system has restricted background work for the app

fractalwrench:
## Goal

Collects whether the system has restricted background work for the app, as reported by [`isBackgroundRestricted()`](https://developer.android.com/reference/android/app/ActivityManager#isBackgroundRestricted()), which is available on Android 9+.

This is useful information as some users do enable battery optimizations that throttle the amount of work an app can do in the background. For example, scheduled services and alarms may not execute. It seems probable that this could lead to bugs in some apps so this would be useful information to collect.

Screenshot 2021-03-25 at 12 08 24

## Changeset

- Added `backgroundWorkRestricted` to the `app` metadata (only when restrictions are enabled, otherwise the field is not added)
- Added unit tests to confirm the functionality works as expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 * Bump compileSdkVersion to apiLevel 30
   [#1202](https://github.com/bugsnag/bugsnag-android/pull/1202)
 
+* Collect whether the system has restricted background work for the app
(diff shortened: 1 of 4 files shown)


twometresteve (approved):
LGTM

PR #1213, opened 2021-03-26
Set sensible default for autoTrackSessions in E2E tests

fractalwrench:
## Goal

Sets a sensible default for our E2E tests by ensuring autoTrackSessions is false by default, since most of the time we only assert on error requests. This reduces some duplication in the test code.


twometresteve (approved):
LGTM.  181 fewer lines of code to maintain. :-)

PR #1217, opened 2021-04-02
Relax checking of releaseStage

twometresteve:
## Goal

Cater for device farms that instrument APKs updated for end-to-end tests.

## Design

Sauce Labs behaves differently than BrowserStack in that it instruments the APKs we upload for e2e tests, making them "debuggable" and causing our notifier to report the `releaseStage` as `development` rather than `production`.

## Changeset

This change relaxes the e2e test expectations to accept either `releaseStage` reported by the notifier, whilst also adding a unit test to ensure the notifier logic is properly tested.

## Testing

Covered by CI and the new unit test.

### DIFF
--- a/features/full_tests/batch_1/internal_error_reports.feature
+++ b/features/full_tests/batch_1/internal_error_reports.feature
@@ -33,7 +33,7 @@ Scenario: If an exception is thrown when sending errors/sessions then internal e
     # App data
     And the event "app.buildUUID" is not null
     And the event "app.id" equals "com.bugsnag.android.mazerunner"
-    And the event "app.releaseStage" matches "(production|development)"
+    And the event "app.releaseStage" equals "mazerunner"
     And the event "app.type" equals "android"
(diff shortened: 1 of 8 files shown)


features/full_tests/batch_1/internal_error_reports.feature
```diff
@@ -33,7 +33,7 @@ Scenario: If an exception is thrown when sending errors/sessions then internal e
     # App data
     And the event "app.buildUUID" is not null
     And the event "app.id" equals "com.bugsnag.android.mazerunner"
-    And the event "app.releaseStage" equals "production"
+    And the event "app.releaseStage" matches "(production|development)"
```
fractalwrench, line 36: I'd be inclined to set `config.releaseStage` to something like 'mazerunner' in `prepareConfig()` so that all scenarios always have the same releaseStage. As the automatic setting of releaseStage is now covered by unit tests, this feels like the most consistent way of ensuring the releaseStage is populated correctly.

PR #1218, opened 2021-04-06
Use MazeRunner v5

twometresteve:
## Goal

Use the Maze Runner v5, which provides the benefit of running the latest version of Appium available on each Android version (1.20.2 apart from on Android 4.4 which is limited to 1.9.1 on BrowserStack).

## Changeset

A couple of changes were needed in response to differences in Appium behavior:
- Catching and ignoring exceptions raised when attempting to push buttons during an ANR
- Ensuring the orientation is portrait for the scenario that asserts on it (the previous scenario can leave it landscape for some reason).

## Testing

Covered by a `[full ci]` run.

### DIFF
--- a/features/full_tests/batch_1/internal_error_reports.feature
+++ b/features/full_tests/batch_1/internal_error_reports.feature
@@ -1,7 +1,8 @@
 Feature: Cached Error Reports
 
 Scenario: If an exception is thrown when sending errors/sessions then internal error reports should be sent
-    When I run "InternalErrorScenario"
+    When I set the screen orientation to portrait
+    And I run "InternalErrorScenario"
     And I wait to receive 1 errors
 
     # Validate internal error report for error serialization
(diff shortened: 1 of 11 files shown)


features/full_tests/batch_1/internal_error_reports.feature
```diff
@@ -1,7 +1,8 @@
 Feature: Cached Error Reports
 
 Scenario: If an exception is thrown when sending errors/sessions then internal error reports should be sent
-    When I run "InternalErrorScenario"
+    When I set the screen orientation to portrait
```
fractalwrench, line 4: I think ideally it'd be good to get to the bottom of why the previous scenario is left as landscape. If this test gets removed or skipped then I can see this causing all sorts of head scratching about what's going on

fractalwrench (commented):
One query - also I noticed the CI appears to be missing a full run.

features/full_tests/batch_1/internal_error_reports.feature
twometresteve: I tracked this down to the behavior of the Appium background_app routine changing between 1.8.0 and 1.9.1.  The app appears landscape when coming back to the foreground after the timeout period.

I've updated the tests now so that the orientation is correcting in the scenario that it goes awry in and added a comment.  I'll also raise a an issue on Appium's github.

fractalwrench (approved):

PR #1220, opened 2021-04-09
Fix deserialization bug in persisted NDK errors

fractalwrench:
## Goal

The memory offset for `bsg_app_info_v2` is wrong by one due to the unintentional inclusion of `is_launching`. This means that the first byte of `binary_arch` for persisted v4 structs is loaded into `is_launching`, causing the `binary_arch` to be incorrectly trimmed to '86' or 'rm64'.

## Changeset

Removed `is_launching` from `bsg_app_info_v2` as it was not originally present.

## Testing

Added additional unit test coverage of the migration function. Manually tested by triggering a fatal crash in v5.7.0, then updating to v5.9.0 to reproduce the behaviour. Confirmed that a local artefact with these changes fixed the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Fix deserialization bug in persisted NDK errors
+  [#1220](https://github.com/bugsnag/bugsnag-android/pull/1220)
+
 ## 5.9.0 (2021-03-30)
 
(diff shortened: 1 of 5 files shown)


twometresteve (commented):
No comments from me.

kstenerud (approved):

fractalwrench:
Tweaked release date to TBD.

twometresteve (approved):
LGTM

PR #1221, opened 2021-04-15
Allow publishing from master branch

fractalwrench:
## Goal

Allows publishing from the master branch rather than requiring a tag as this fits into our existing publishing process.

Tested by verifying that setting the `build.branch` to `PLAT-6241/publish-on-master` caused the publish step to appear - it's assumed that the publish step itself works as expected.


twometresteve (approved):
LGTM

PR #1224, opened 2021-04-16
Fix flaky mazerunner tests

fractalwrench:
## Goal

Fixes a couple of flaky mazerunner tests identified via scheduled builds.

## Changeset

`CrashHandlerScenario` was inherently racy as the log endpoint makes a network request which may or may not be successful before the uncaught exception terminates the process. It has been rewritten so the custom crash handler alters metadata, and the scenario verifies the handler was invoked.

`ManualSmokeSessionScenario` was flaky in very specific circumstances: if the last handled error + unhandled error are written to disk in the same millisecond, and the generated UUID meant the [request order](https://github.com/bugsnag/bugsnag-android/blob/next/bugsnag-android-core/src/main/java/com/bugsnag/android/EventStore.java#L124) was wrong. The scenario has been updated to ensure that the handled errors are always delivered first.


twometresteve (approved):
LGTM... pending CI.

PR #1225, opened 2021-04-19
Alter delay in E2E test

fractalwrench:
## Goal

Fixes a race in `CxxDelayedErrorScenario` between the NDK crash and handled JVM error delivery. Previously the fatal NDK error could terminates the JVM delivery halfway through, which would result in only one error being sent.

The scenario has been improved so that the NDK crash only happens once the first error has been delivered.


twometresteve (approved):
LGTM

PR #1228, opened 2021-04-23
Move migration code from serializer.c to its own file

kstenerud:
## Goal

serializer.c was getting too big. The migration code can be self-contained in a separate file.

## Testing

Re-ran unit tests and e2e tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
@@ -15,7 +15,7 @@ extern "C" {
 bool bsg_event_requires_migration(int event_version);
 
 /**
- * Read an event of the specified version from an FD and migrate it to the
+ * Read an event of the specified version from a FD and migrate it to the
  * latest structure.
  * @param event_version The version decoded from the event file header
  * @param fd The file descriptor after reading the header.
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
@@ -3,225 +3,25 @@
 
 #include "event.h"
 
-#ifndef V1_BUGSNAG_CRUMBS_MAX
-#define V1_BUGSNAG_CRUMBS_MAX 30
-#endif
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 /**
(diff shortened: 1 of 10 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/migrate.h
```diff
@@ -3,225 +3,25 @@
...
-  char session_id[33];
-  char session_start[33];
-  int handled_events;
-  int unhandled_events;
-  char grouping_hash[64];
-  bool unhandled;
-  char api_key[64];
-} bugsnag_report_v4;
+/**
+ * Read an event of the specified version from an FD and migrate it to the
```
fractalwrench, line 18:
```suggestion
 * Read an event of the specified version from a FD and migrate it to the
```

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -12,27 +12,10 @@
...
-void migrate_app_v1(bugsnag_report_v2 *report_v2, bugsnag_report_v3 *event);
-void migrate_app_v2(bugsnag_report_v4 *report_v4, bugsnag_event *event);
-void migrate_device_v1(bugsnag_report_v2 *report_v2, bugsnag_report_v3 *event);
-void migrate_breadcrumb_v1(bugsnag_report_v2 *report_v2,
-                           bugsnag_report_v3 *event);
-
-#ifdef __cplusplus
-}
-#endif
+static bool event_write(bsg_report_header *header, bugsnag_event *event,
```
fractalwrench, line 15: As a general rule all symbols use a `bsg_` prefix to avoid name clashes - we've been burnt by this in the past.

fractalwrench (commented):
This approach looks good - it's nice to see the file split out into two logical components.

It might be worth separating the `test_utils_serialize.c` unit test in a similar way, as currently that tests both serialization + migration. Otherwise I think this is nearly ready to be merged.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
kstenerud: These are static functions, so they're by definition safe from symbol name clashes.

fractalwrench (approved):

PR #1229, opened 2021-04-27
6250 new struct limits

kstenerud:
## Goal

PLAT-6250: Increase internal event structure sizes

## Testing

Re-ran e2e tests

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.h
@@ -27,7 +27,7 @@
 /**
  * Version of the bugsnag_event struct. Serialized to report header.
  */
-#define BUGSNAG_EVENT_VERSION 5
+#define BUGSNAG_EVENT_VERSION 6
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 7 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -128,7 +128,7 @@ typedef struct {
   /**
    * Value if type is BSG_CHAR_VALUE
    */
-  char char_value[64];
+  char char_value[128];
```
fractalwrench, line 131: This should be 256 according to the design

bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -162,7 +162,7 @@ typedef struct {
 } bsg_error;
 
 typedef struct {
-  char name[64];
+  char name[128];
```
fractalwrench, line 165: This should be 256 according to the design

fractalwrench (approved):
A couple of comments on a couple of allocation sizes, otherwise this looks good.

PR #1231, opened 2021-04-28
Add dependency license audit to CI

Cawllec:
## Goal

Verifies the dependencies of gradle and npm modules have suitable licenses for use with this notifier.

In the case of `libunwindstack`, it has been determined that the files reference the apache 2.0 license, so our list will be manually updated.


fractalwrench (approved):

scripts/audit-dependency-licenses.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if test -f "decisions.yml"; then
+    decision="--decisions-file=decisions.yml"
+else
+    decision=""
```
twometresteve, line 6: Why would you want to run without a decision file?  I think it the file isn't there then the tool just carries on with nothing anyway.

dockerfiles/Dockerfile.android-license-audit
```diff
@@ -0,0 +1,7 @@
+ARG BRANCH_NAME
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/android:ci-${BRANCH_NAME} as android
+
+RUN apt-get install -y ruby-full npm
+RUN gem install license_finder
+
+RUN curl https://raw.githubusercontent.com/bugsnag/license-audit/master/config/decision_files/global.yml -o decisions.yml
```
twometresteve, line 7: Any reason not to put this into the shell script?

twometresteve (approved):
LGTM

PR #1232, opened 2021-04-30
Debounce activity lifecycle callbacks in tests

fractalwrench:
## Goal

Prevents the possibility of multiple requests being sent in mazerunner scenarios when only one is expected, leading to test flakes as seen here: https://buildkite.com/bugsnag/bugsnag-android/builds/4294

## Changeset

- Removed dead code `AutoContextScenario`
- Added a flag that tracks whether the activity lifecycle has been called before, and only invokes the error once if so


twometresteve (approved):
LGTM

PR #1245, opened 2021-05-11
Fix NegativeArraySizeException during crash report deserialization

lemnik:
### Goal
Prevent a possible NegativeArraySizeException which could occur when Matcher.groupSize returned a negative value.

### Testing


fractalwrench (approved):
LGTM pending CI

PR #1246, opened 2021-05-12
Fix flaky test by verifying scheduled executor is called

fractalwrench:
## Goal

Alters the `LaunchCrashTrackerTest` as the `isLaunching` state might not be updated in the timeout required, as a `Runnable` is submitted to an `Executor`. Therefore the unit test just asserts that the necessary method on the mock is called.


twometresteve (approved):
LGTM

PR #1249, opened 2021-05-12
fix: avoid capturing thread stacktraces when not required to

lemnik:
## Goal
Avoid capturing all of the thread stacktraces when they won't be reported.

## Changeset

Instead of capturing stacktraces as a default in the constructor, the argument is now nullable and lazy-captured in the `init` block if required.

## Testing

Existing tests were altered to use the default null value, and a new test was added to ensure that the  `ThreadSendPolicy.NEVER` causes any stacktraces passed to `ThreadState` are discarded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Avoid unnecessary collection of Thread stacktraces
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM!

PR #1251, opened 2021-05-13
Improve fallbacks for unavailable system services

lemnik:
## Goal
Ensure we have fallbacks for our uses of Android system services where it's feasible to improve reliability in rare cases where some services are not available.

## Design
Added a new `UnkownConnectivity` object to handle cases where the system ConnectionManager is not available. The new facade is always connected, and doesn't fire any events.

Marked our diagnostic use of `StorageManager` as nullable, and added a null-check. There should be no normal conditions that we need this check, but the StorageManager use was already conditional - so this is an additional safety layer.

## Testing

Added new tests for the `ConnectivityCompat` and improved the test coverage of the `Client` init process.


bugsnag-android-core/src/test/java/com/bugsnag/android/ConnectivityCompatTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.Mockito
+import org.mockito.junit.MockitoJUnitRunner
+
+@Suppress("DEPRECATION")
```
fractalwrench, line 15: nit: could the suppression be added to the specific statements/methods using deprecated code?

fractalwrench (approved):
LGTM pending resolution of one minor comment.

fractalwrench (commented):
I noticed a few extra places that need fallbacks:

https://github.com/bugsnag/bugsnag-android/blob/v5.9.2/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java#L146
https://github.com/bugsnag/bugsnag-android/blob/v5.9.2/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java#L156
https://github.com/bugsnag/bugsnag-android/blob/v5.9.2/bugsnag-android-core/src/main/java/com/bugsnag/android/ForegroundDetector.java#L16
https://github.com/bugsnag/bugsnag-android/blob/v5.9.2/bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrDetailsCollector.kt#L26

fractalwrench (approved):

PR #1255, opened 2021-05-17
fix: avoid recursive Throwable cause chains

lemnik:
## Goal
Avoid errors caused by Throwable.cause chains that are recursive (self-referencing).

## Design
Added a method to unroll the cause-chain into a flat List and handle any recursion. 

## Testing
Added RecursiveThrowableCauseTest to test code that previously had recursion problems, and UnrollThrowableCausesTest to test the contract of the new unroll method.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/StrictModeHandler.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/StrictModeHandler.java
@@ -89,7 +89,7 @@ class StrictModeHandler {
      * @return the root cause of the throwable
      */
     private Throwable getRootCause(Throwable throwable) {
-        List<Throwable> causes = ThrowableExtensionsKt.safeUnrollCauses(throwable);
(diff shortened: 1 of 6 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/StrictModeHandler.java
```diff
@@ -87,12 +89,7 @@ String getViolationDescription(String exceptionMessage) {
...
      */
     private Throwable getRootCause(Throwable throwable) {
-        Throwable cause = throwable.getCause();
-
-        if (cause == null) {
-            return throwable;
-        } else {
-            return getRootCause(cause);
-        }
+        List<Throwable> causes = ThrowableExtensionsKt.safeUnrollCauses(throwable);
```
fractalwrench, line 92: I think your suggestion of using `@file:JvmName("ThrowableUtils")` to rename the file would work well here.

fractalwrench (approved):
LGTM pending one minor comment.

fractalwrench (approved):

PR #1258, opened 2021-05-19
Plat 6092/improve device data gathering performance

lemnik:
## Goal
Improve the performance of data gathering by only capturing the foreground and battery status once (where we were capturing it twice).

## Testing
Added a unit-test to cover the different possible outputs of `foregroundDuration`

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
@@ -6,7 +6,6 @@ import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.SystemClock
-import java.util.HashMap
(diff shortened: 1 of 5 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
```diff
@@ -102,9 +107,19 @@ internal class AppDataCollector(
      *
      * @return the duration in ms
      */
-    internal fun calculateDurationInForeground(): Long? {
+    internal fun calculateDurationInForeground(inForeground: Boolean? = sessionTracker.isInForeground): Long? {
+        if (inForeground == null) return null
```
fractalwrench, line 111: nit: the general (implicit) style rule at Bugsnag is to always use braces for single-line conditionals

fractalwrench (approved):

PR #1261, opened 2021-05-21
Android changes to support Unity uplift

fractalwrench:
## Goal

This is the integration branch for the Android Unity changes - the contents of this PR have been reviewed previously so it's just a case of approval to merge into `next`.


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## TBD
 
+### Enhancements
```
tomlongridge, line 5: I don't think I'd class either of these as an enhancement for Android - patch release seems fine.

PR #1262, opened 2021-05-24
fix: relay onLowMemory events to client Observers

lemnik:
## Goal
Relay onLowMemory events to the NDK layer through the client Observable so that it will be included in all NDK reports.

## Testing
Included a heavily mocked unit test to ensure that the events are relayed to the client Observables. This task also requires manual testing.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/StateEvent.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/StateEvent.kt
@@ -42,4 +42,6 @@ sealed class StateEvent {
     class UpdateOrientation(val orientation: String?) : StateEvent()
 
     class UpdateUser(val user: User) : StateEvent()
+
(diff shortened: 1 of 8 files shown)


fractalwrench (approved):

bugsnag-android-core/src/main/java/com/bugsnag/android/StateEvent.kt
```diff
@@ -42,4 +42,6 @@ sealed class StateEvent {
     class UpdateOrientation(val orientation: String?) : StateEvent()
 
     class UpdateUser(val user: User) : StateEvent()
+
+    class UpdateMemoryTrimEvent(val isLowMemory: Boolean) : StateEvent()
```
lemnik, line 46: Note: `StateEvent` is a public API and is a `sealed class`

kattrali (approved):

PR #1263, opened 2021-05-24
fix(ndk): link directly to liblog without find_library

lemnik:
## Goal
Stop including the `liblog.so` file in the `aar` files during a build on Linux.

## Testing
Affects the build process and has been checked on MacOS and Linux


kattrali (approved):

fractalwrench (approved):
LGTM. To resolve the ticket (and get publishing our artefacts via CI) it'd be good to check everything works end-to-end by pushing an artefact to Sonatype, testing it out via the staging repository, and then dropping the artefact so it's not published. I can assist with this if that'd be helpful

PR #1265, opened 2021-05-27
Fix disabling/re-enabling ANR after initialization

kstenerud:
## Goal

PLAT-6620

ANRs were not being re-enabled after being disabled via the API.

## Design

Don't exit the watchdog thread anymore. Just loop to the next thread wait point and then after being triggered, only do ANR processing if ANRs are currently enabled.

## Testing

Ran in a unity project that disables and enables error and ANR processing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Unity: Properly handle ANRs after multiple calls to autoNotify and autoDetectAnrs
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM pending CI

fractalwrench:
@kstenerud A changelog entry also needs adding

kstenerud:
OK, I wasn't sure if we needed an entry here or in Unity since the Android library has no public API for enabling/disabling.

PR #1272, opened 2021-06-15
Add performance benchmarks with microbenchmark library

fractalwrench:
## Goal

Adds performance benchmarks for crucial method calls using the [Microbenchmark library](https://developer.android.com/studio/profile/benchmark). This gives us a baseline idea of what performance is like and ensures we don't regress in future.

## Changeset

- Added 'bugsnag-benchmarks' library module that contains all the performance benchmark tests
- Altered build script so that only shippable modules target a 1.6 JVM for Kotlin, as the microbenchmark library requires 1.8
- Added step to buildkite that runs performance tests against a browserstack device
- Added one performance benchmark (others to follow in another PR)

The benchmark results can be read by inspecting the device logs of the Browserstack session - in future we could consider adding execution thresholds or grabbing the JSON data, but for now this felt sufficient.


lemnik (approved):
LGMT 👍

PR #1273, opened 2021-06-15
Add performance benchmark tests

fractalwrench:
## Goal

Adds performance tests using the [microbenchmark library](https://developer.android.com/studio/profile/benchmark) which provide a baseline insight into Bugsnag's performance.

These tests focus primarily on Bugsnag's critical path (notify/metadata/breadcrumbs) but has added some additional test cases to measure some other APIs - although this is not exhaustive.

I've avoided benchmarking `Bugsnag.start()` as this led to OOMs - the macrobenchmark library may be a better fit for this, which will be investigated separately.

## Changeset

Added benchmark tests for the following:

- Starting/pausing/resuming sessions
- Creating an `Event`
- Setting the user and context
- Constructing `Configuration`
- Calling `Bugsng.notify()`
- Leaving breadcrumbs on the client
- Leaving metadata on the client


lemnik (approved):
LGTM

lemnik:
It would be good to add a benchmark covering the JSON serialization for a typical event / session payload.

fractalwrench:
Updated to include benchmarks for JSON serialization.

lemnik (approved):
LGTM

PR #1274, opened 2021-06-16
Optimize implementation of internal state change observers

fractalwrench:
## Goal

Optimizes the implementation of the internal system used for notifying observers of state changes.

## Changeset

This new solution avoids the performance disadvantages of the previous solution:

- The `StateEvent` message is lazily created which avoids object instantiation if no observers are registered
- `BaseObservable` avoids creating a `Vector` by replacing `update()` with `onStateChange()` (this also grants additional type safety)
- `BaseObservable` uses `CopyOnWriteArrayList` which avoids the need for synchronization when adding/removing observers. This was seen as acceptable as observers will be added/removed very rarely, whereas the collection will be iterated over many times

The remaining changes entail updating the rest of the code and the tests to use the new method signatures.

## Testing

Updated existing unit test coverage and relied on existing tests.

## Performance

Profiled app to confirm there was no performance regression, and also checked CI stats.

changes_flame

baseline_flame

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -84,21 +84,6 @@ steps:
     concurrency: 9
     concurrency_group: 'browserstack-app'
 
-  - label: ':android: Performance benchmarks'
-    key: 'perf-benchmarks'
-    depends_on:
-      - "android-ci"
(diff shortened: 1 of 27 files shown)


lemnik (approved):
LGTM

PR #1275, opened 2021-06-17
Cache value of app.backgroundWorkRestricted

fractalwrench:
## Goal

Caches the value of `app.backgroundWorkRestricted` as it is unlikely to change over a session. This should improve the performance of subsequent `Bugsnag.notify()` calls as the API call took ~3ms on a OnePlus 7.

## Testing

Ran the example app with the artefact and confirmed the field is reported correctly.


lemnik (approved):
LGTM

PR #1276, opened 2021-06-17
Optimize execution of callbacks

fractalwrench:
## Goal

Bugsnag supplies the ability to set callbacks for when breadcrumbs are added, and when an error or session are captured. Breadcrumbs/errors in particular can be captured many times per second.

By default no callbacks are set. This changeset optimizes the default behaviour by performing an `isEmpty()` check before running the callbacks, which avoids construction of an iterator each time. For the addition of ~1000 breadcrumbs this saves ~3ms in the default case and has negligible impact when callbacks are set.

## Testing

Verified by profiling the changes in an example app with no callbacks set and observing `runOnBreadcrumbTasks()`:

baseline-no-cb

changes-no-cb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Cache value of app.backgroundWorkRestricted
   [#1275](https://github.com/bugsnag/bugsnag-android/pull/1275)
 
+* Optimize execution of callbacks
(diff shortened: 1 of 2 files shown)


lemnik (approved):
LGTM

PR #1277, opened 2021-06-17
Optimize metadata implementation by reducing type casts

fractalwrench:
## Goal

Bugsnag's stores metadata using a nested `ConcurrentHashMap`. The implementation uses unnecessary type casts, which resulted in unnecessary calls to `TypeIntrinsics.isMutableMap()` that added ~1ms for every ~10000 metadata values added.

## Changeset

Metadata was previously backed by `ConcurrentHashMap<String, Any>`. This has been changed to `MutableMap<String, MutableMap<String, Any>>`.

This more accurately reflects the metadata structure and has allowed for a net reduction in type casts throughout `Metadata.kt`. This has necessitated the addition of some type casts (mainly to maintain backwards compatibility with method signatures that should not change until the next major version).

## Testing

Relied on existing unit test coverage.

The performance impact of this change ended up being fairly minimal - typically `TypeIntrinstics` executed in <1ms, so its removal has a benefit around that amount. However, this still feels advantageous in that it simplifies the implementation and increases its type safety.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
@@ -64,7 +64,7 @@ internal data class Metadata @JvmOverloads constructor(
         val tab = store[section]
         tab?.remove(key)
 
-        if (tab?.isEmpty() == true) {
+        if (tab.isNullOrEmpty()) {
             store.remove(section)
         }
     }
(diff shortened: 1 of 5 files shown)


lemnik (approved):
LGTM

bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
```diff
@@ -65,49 +62,41 @@ internal data class Metadata @JvmOverloads constructor(
...
         val tab = store[section]
+        tab?.remove(key)
 
-        if (tab is MutableMap<*, *>) {
-            tab.remove(key)
-
-            if (tab.isEmpty()) {
-                store.remove(section)
-            }
+        if (tab?.isEmpty() == true) {
```
lemnik, line 67:
```suggestion
        if (tab.isNullOrEmpty()) {
```

lemnik (approved):

PR #1278, opened 2021-06-18
Basic journal functionality

kstenerud:
## Goal

This is the high-level journal functionality based on PLAT-6353 that supports:
- Journal serialization to a stream
- Journal deserialization from a byte buffer
- Executing a journal's commands to modify a document

Future PRs will handle:
- Journal metadata entry
- Low level operations and shared memory
- File management
- Async-safe journal management
- Crash-time journal additions
- Integration into Bugsnag's event system

## Design

See PLAT-6353

## Testing

Each of the new components has unit tests. There is also a very basic performance test to make sure the JSON codec performance isn't terrible.

### DIFF
--- a/bugsnag-android-core/build.gradle
+++ b/bugsnag-android-core/build.gradle
@@ -11,6 +11,10 @@ bugsnagBuildOptions {
 apply plugin: "com.android.library"
 apply plugin: "org.jetbrains.dokka"
 
+dependencies {
+    implementation 'com.dslplatform:dsl-json:1.9.8'
+}
+
 dokka {
     outputFormat = "html"
     outputDirectory = "$buildDir/dokka"
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BugsnagTestUtils.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BugsnagTestUtils.java
@@ -14,8 +14,12 @@ import org.json.JSONObject;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 final class BugsnagTestUtils {
 
@@ -131,5 +135,84 @@ final class BugsnagTestUtils {
 
     public static App generateApp() {
         return new App(generateImmutableConfig(), null, null, null, null, null);
+    }
+
+    /**
+     * "Normalize" a map by changing all numeric types to their largest forms.
+     * This is necessary for comparing the results of serialization/deserialization
+     * operations because we have no control over what types the codec will choose,
+     * and equals() takes into account the underlying type.
+     *
+     * @param map The map to normalize
+     * @param <K> The key type
+     * @param <V> The value type
+     * @return The normalized map
+     */
+    @SuppressWarnings("unchecked")
+    public static <K, V> Map<K, V> normalizedMap(Map<K, V> map) {
+        Map<K, V> newMap = new HashMap<>(map.size());
+        Set<Map.Entry<K, V>> set = map.entrySet();
+        for (Map.Entry<K, V> entry: set) {
+            K key = entry.getKey();
+            K normalizedKey = (K)normalized(key);
+            if (!key.equals(normalizedKey)) {
+                key = normalizedKey;
+            }
+            newMap.put(key, (V)normalized(entry.getValue()));
+        }
+        return newMap;
+    }
+
+    /**
+     * "Normalize" a list by changing all numeric types to their largest forms.
+     * This is necessary for comparing the results of serialization/deserialization
+     * operations because we have no control over what types the codec will choose,
+     * and equals() takes into account the underlying type.
+     *
+     * @param list The list to normalize
+     * @param <T> The element type
+     * @return The normalized list
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> List<T> normalizedList(List<T> list) {
+        List<T> newList = new ArrayList<>(list.size());
+        for (T entry: list) {
+            newList.add((T)normalized(entry));
+        }
+        return newList;
+    }
+
+    /**
+     * "Normalize" an unknown value by changing all numeric types to their largest forms.
+     * This is necessary for comparing the results of serialization/deserialization
+     * operations because we have no control over what types the codec will choose,
+     * and equals() takes into account the underlying type.
+     *
+     * This function normalizes integers, floats, lists, and maps and their subobjects.
+     *
+     * @param obj The object to normalize.
+     * @return The normalized object (may be the same object passed in)
+     */
+    @SuppressWarnings("unchecked")
+    public static Object normalized(Object obj) {
+        if (obj instanceof Byte) {
+            return ((Byte)obj).longValue();
+        }
+        if (obj instanceof Short) {
+            return ((Short)obj).longValue();
+        }
+        if (obj instanceof Integer) {
+            return ((Integer)obj).longValue();
+        }
+        if (obj instanceof Float) {
+            return ((Float)obj).doubleValue();
+        }
+        if (obj instanceof Map) {
+            return normalizedMap((Map<Object, Object>)obj);
+        }
+        if (obj instanceof List) {
+            return normalizedList((List<Object>)obj);
+        }
+        return obj;
     }
 }
--- a/bugsnag-benchmarks/src/androidTest/java/com/bugsnag/android/benchmark/JournalBenchmarkTest.kt
+++ b/bugsnag-benchmarks/src/androidTest/java/com/bugsnag/android/benchmark/JournalBenchmarkTest.kt
@@ -1,0 +1,50 @@
+package com.bugsnag.android.benchmark
+
+import android.app.Application
+import android.content.Context
+import androidx.benchmark.junit4.BenchmarkRule
+import androidx.benchmark.junit4.measureRepeated
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.bugsnag.android.*
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.io.ByteArrayOutputStream
+
+@RunWith(AndroidJUnit4::class)
+class JournalBenchmarkTest {
+    @get:Rule
+    val benchmarkRule = BenchmarkRule()
+
+    lateinit var client: Client
+    lateinit var ctx: Context
+
+    @Before
+    fun setUp() {
+        ctx = ApplicationProvider.getApplicationContext<Application>()
+        client = generateClient(ctx)
+    }
+
+    @Test
+    fun serializeEventPayload() {
+        val journal = Journal()
+        val entries = listOf<Journal.Command>(
+            Journal.Command("a", "b")
+        )
+
+        benchmarkRule.measureRepeated {
+            journal.clear()
+            val stream = benchmarkRule.scope.runWithTimingDisabled {
+                ByteArrayOutputStream()
+            }
+            stream.use {
+                for(entry in entries) {
+                    journal.add(entry)
+                }
+                journal.serialize(stream)
+            }
+        }
+    }
+}
--- a/buildSrc/src/main/kotlin/com/bugsnag/android/BugsnagBuildPlugin.kt
+++ b/buildSrc/src/main/kotlin/com/bugsnag/android/BugsnagBuildPlugin.kt
@@ -115,6 +115,7 @@ class BugsnagBuildPlugin : Plugin<Project> {
             isCheckAllWarnings = true
             baseline(File(project.projectDir, "lint-baseline.xml"))
             disable("GradleDependency", "NewerVersionAvailable")
+            ignore("InvalidPackage")
         }
     }
 
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/InterceptingLogger.kt
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/InterceptingLogger.kt
@@ -1,0 +1,22 @@
+package com.bugsnag.android
+
+class InterceptingLogger : Logger {
+
+    var msg: String? = null
+
+    override fun w(msg: String) {
+        this.msg = msg
+    }
+
+    override fun w(msg: String, throwable: Throwable) {
+        this.msg = msg
+    }
+
+    override fun e(msg: String) {
+        this.msg = msg
+    }
+
+    override fun e(msg: String, throwable: Throwable) {
+        this.msg = msg
+    }
+}
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/JournalCommandTest.kt
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/JournalCommandTest.kt
@@ -1,0 +1,60 @@
+package com.bugsnag.android
+
+import com.dslplatform.json.DslJson
+import org.junit.Assert
+import org.junit.Test
+import java.io.ByteArrayOutputStream
+
+class JournalCommandTest {
+    @Test
+    fun testSerializeBasic() {
+        assertSerializeDeserialize(
+            "{\"a\":1}",
+            Journal.Command("a", 1)
+        )
+        assertSerializeDeserialize(
+            "{\"a\":1.5}",
+            Journal.Command("a", 1.5)
+        )
+        assertSerializeDeserialize(
+            "{\"a\":\"x\"}",
+            Journal.Command("a", "x")
+        )
+        assertSerializeDeserialize(
+            "{\"a\":true}",
+            Journal.Command("a", true)
+        )
+        assertSerializeDeserialize(
+            "{\"a\":null}",
+            Journal.Command("a", null)
+        )
+    }
+
+    private fun serializeJournalEntry(entry: Journal.Command): String {
+        val os = ByteArrayOutputStream()
+        entry.serialize(os)
+        return os.toString(Charsets.UTF_8.name())
+    }
+
+    private fun deserializeJournalEntry(json: String): Journal.Command {
+        val dslJson = DslJson<Any>()
+        val document = json.toByteArray(Charsets.UTF_8)
+        return Journal.deserializeCommand(dslJson, document)
+    }
+
+    private fun assertSerializeDeserialize(json: String, entry: Journal.Command) {
+        val expectedJson = json
+        val actualJson = serializeJournalEntry(entry)
+        Assert.assertEquals(expectedJson, actualJson)
+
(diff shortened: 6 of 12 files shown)


bugsnag-benchmarks/src/androidTest/java/com/bugsnag/android/benchmark/JournalBenchmarkTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    lateinit var ctx: Context
+
+    @Before
+    fun setUp() {
+        ctx = ApplicationProvider.getApplicationContext<Application>()
+        client = generateClient(ctx)
+    }
+
+    @Test
+    fun serializeEventPayload() {
```
fractalwrench, line 31: This takes ~12,786 ns according to the CI benchmark - how does that roughly compare against the previous solution? It'd be interesting to see what the results are like when I/O is involved.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A journal represents a series of commands that will manipulate a document.
+ * Sometimes it's more efficient to store a base document and a list of modifications
+ * to me made to it rather than re-serialize the entire document after every modification.
```
fractalwrench, line 19:
```suggestion
 * to be made to it rather than re-serialize the entire document after every modification.
```

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+ *
+ * The document to be modified must be serializable into JSON. The following types are supported:
+ * - null
+ * - boolean
+ * - integer (max 15 digits)
+ * - float
+ * - string
+ * - List(Object)
+ * - Map(String, Object)
+ */
```
fractalwrench, line 58: great docs 👍

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+public class DocumentPath {
```
fractalwrench, line 59: Something to consider is that for new code in the Android notifier we're generally aiming to implement it in Kotlin, unless it's part of the public-facing API.

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
+package com.bugsnag.android;
+
+import android.icu.lang.UCharacterEnums;
```
fractalwrench, line 3:
```suggestion
```

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+        int strBegin = 0;
+        for (int i = 0; i < path.length(); i++) {
+            char ch = path.charAt(i);
+            switch (ch) {
+                case '\\':
+                    requiresEscaping = true;
+                    i++;
+                    break;
+                case '.':
+                    String pathComponent = null;
```
fractalwrench, line 138:
```suggestion
                    String pathComponent;
```

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+                    directives.add(makeDirectiveFromPathComponent(pathComponent));
+                    strBegin = i + 1;
+                    requiresEscaping = false;
+                    break;
+                default:
+                    break;
+            }
+        }
+        if (strBegin < path.length()) {
+            String pathComponent = null;
```
fractalwrench, line 153:
```suggestion
            String pathComponent;
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+ */
+public class Journal {
```
fractalwrench, line 21: Should these classes be exposed as part of our public API? This feels like something to avoid unless it's absolutely necessary. If this is unavoidable, then this could be moved to the `com.bugsnag.android.internal` package which be convention is not part of the API.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+    @NonNull
+    public final List<Command> commands;
+
+    public Journal() {
+        commands = new LinkedList<>();
```
fractalwrench, line 26: Will this class be called from multiple threads? What is the strategy used to avoid [concurrent modification exceptions](https://docs.oracle.com/javase/7/docs/api/java/util/ConcurrentModificationException.html) if so?

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+     * This method goes through the journal entries one-by-one, applying their modification
+     * commands to the document. This is a destructive operation that will modify the document.
+     *
+     * Note: If a journal command replaces the root object, the resulting document will be a
+     * different object from the one passed in.
+     *
+     * @param document The document to apply to.
+     * @return The finished document (may be a different object from the one passed in).
+     */
+    public @NonNull Object applyTo(@NonNull Object document) {
```
fractalwrench, line 41: I'm not clear what type is expected to be passed in and return here. All the examples use `Map`, which would be preferable to `Object` as it's more precise. Unless there are good reasons not to, I think these methods need to use types, as it gives type safety and will presumably reduce the number of expensive type checks by callers of this API.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+     * Serialize the journal to a stream.
+     *
+     * @param out The stream to serialize to.
+     * @throws IOException if an IO exception occurs.
+     */
+    public void serialize(@NonNull OutputStream out) throws IOException {
+        for (Command command : commands) {
+            command.serialize(out);
+            out.write(0);
+            out.flush();
```
fractalwrench, line 74: Should this be using a BufferedOutputStream? This would hopefully avoid the need to flush after each command (unless I'm really missing something)

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+    public static @NonNull Journal deserialize(@NonNull byte[] bytes, @NonNull Logger logger) {
+        Journal journal = new Journal();
+        int currentStart = 0;
+        for (int i = 0; i < bytes.length; i++) {
+            if (bytes[i] == 0) {
+                try {
+                    byte[] document = Arrays.copyOfRange(bytes, currentStart, i);
+                    journal.add(Command.deserialize(document));
+                    currentStart = i + 1;
+                } catch (Exception exception) {
```
fractalwrench, line 93: What operations specifically can fail here, and what would the ramifications be if it wasn't possible to deserialize a journal entry?

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return journal;
+    }
+
+    /**
+     * Represents a single journal command, consisting of a document path,
+     * and a value to be applied at that path level.
+     */
+    public static class Command {
+        private static final DslJson<Object> dslJson = new DslJson<>();
```
fractalwrench, line 108: Is this object guaranteed to be thread-safe?

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+         *
+         * @param document The document to deserialize from
+         * @return The journal command
+         * @throws IOException If an IO exception occurs
+         */
+        @SuppressWarnings("unchecked")
+        public static @NonNull
+        Command deserialize(@NonNull byte[] document)
+                throws IOException {
+            Map<String, Object> map = dslJson.deserialize(Map.class, document, document.length);
```
fractalwrench, line 162:
```suggestion
            @SuppressWarnings("unchecked") Map<String, Object> map = dslJson.deserialize(Map.class, document, document.length);
```

Keeping this suppression local will avoid accidentally suppressing any problems that might get introduced in future.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        /**
+         * Deserialize from a byte buffer.
+         *
+         * @param document The document to deserialize from
+         * @return The journal command
+         * @throws IOException If an IO exception occurs
+         */
+        @SuppressWarnings("unchecked")
```
fractalwrench, line 158:
```suggestion
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
```diff
@@ -0,0 +1,170 @@
...
+     * Represents a single journal command, consisting of a document path,
+     * and a value to be applied at that path level.
+     */
+    public static class Command {
+        private static final DslJson<Object> dslJson = new DslJson<>();
+
+        @NonNull
+        public final DocumentPath path;
+        @NonNull
+        public final String pathAsString;
```
fractalwrench, line 113: nit: this might be more useful to access the path as a string via some method on `DocumentPath` instead

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BugsnagTestUtils.java
```diff
@@ -132,4 +136,83 @@ public static AppWithState generateAppWithState() {
     public static App generateApp() {
         return new App(generateImmutableConfig(), null, null, null, null, null);
     }
+
+    /**
+     * "Normalize" a map by changing all numeric types to their largest forms.
+     * This is necessary for comparing the results of serialization/deserialization
+     * operations because we have no control over what types the codec will choose,
```
fractalwrench, line 143: Will the output not be predictable with the codec? I feel I'm missing something over why these methods are necessary

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+ * is separated by a dot '.' If a literal dot is required, the path supports an escaping
+ * mechanism using the backslash character:
+ *
+ *     \. = literal dot
+ *     \\ = literal backslash
+ *
+ * Path components are either integers or non-integers. Integers refer to the previous path
+ * component as a list, providing the index to look up. Non-integers refer to the previous path
+ * component as a map, providing the key to look up (for example, in a.b.3, a is treated as a
+ * map and b is treated as a list, so the path refers to document root with map key a, which is
```
fractalwrench, line 32: It would be incredibly helpful to define 'a', 'b', etc as constants so that it's easier to understand what's going on in the tests.

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+                case '.':
+                    String pathComponent = null;
+                    if (requiresEscaping) {
+                        pathComponent = unescape(path, strBegin, i);
+                    } else {
+                        pathComponent = path.substring(strBegin, i);
+                    }
+                    directives.add(makeDirectiveFromPathComponent(pathComponent));
+                    strBegin = i + 1;
+                    requiresEscaping = false;
```
fractalwrench, line 146:
```suggestion
```

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+        }
+        return true;
+    }
+
+    /**
+     * A directive determines how document path commands are interpreted. Depending on the
+     * current context, making a container, getting from a container, or setting to a container
+     * will require different operations to complete.
+     */
+    private interface Directive {
```
fractalwrench, line 244: Consider moving this to a separate source file as the file is currently >300 lines, and will probably grow over time

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
```diff
@@ -0,0 +1,385 @@
...
+         * Setting null removes the value.
+         */
+        class MapKeyDirective implements Directive {
+            private final String key;
+
+            public MapKeyDirective(@NotNull String key) {
+                this.key = key;
+            }
+
+            @NotNull
```
fractalwrench, line 279: `@NonNull` is the preferred annotation (but frustratingly android studio sometimes picks `@NotNull`)

fractalwrench (commented):
I gave a stab at an initial review but think it'd be really helpful if the changeset could be broken into more manageable chunks, preferably <200 lines of diff. As it stands I couldn't fully grok what was going on - perhaps that's partly down to Friday afternoon syndrome :)

bugsnag-android-core/build.gradle
```diff
@@ -11,6 +11,10 @@ bugsnagBuildOptions {
 apply plugin: "com.android.library"
 apply plugin: "org.jetbrains.dokka"
 
+dependencies {
+    implementation 'com.dslplatform:dsl-json:1.9.8'
```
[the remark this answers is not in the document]
kstenerud: I used this benchmark as a guide: https://github.com/fabienrenaud/java-json-benchmark

Most Java-based JSON packages are very slow because they convert to strings, which then have to be converted to UTF-8. I chose this one because it has the fastest performance, and it also can read/write directly from byte arrays (which we need for the native-java shared memory region).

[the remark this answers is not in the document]
kstenerud: It tells me `compile` is deprecated and I should use `implementation` instead.

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BugsnagTestUtils.java
kstenerud: When deserializing, I encountered issues where it would parse a number as a long, and in other cases as an int. This would break naive equality checks.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: There will be an outer class that manages the shared memory region between java and native side. This will already have to handle concurrency, so it will protect the journal as well since it has to add entries to the journal anyway.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: We need to flush immediately because this will be written to the shared memory region. The actual serialization of the journal will be done in a worker thread, but the result needs to be deterministic so that I can update the status fields for the native side to know what's going on in case of a crash.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: Bad JSON from a partial write will be the most common cause. If deserialization of a journal entry fails, we toss this one out and continue on. You end up with partial data, but it's better than no data.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: No, but it will only be called from a thread-safe environment.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: Its only purpose is for serializing the journal entry. Since DocumentPath doesn't get serialized or care about the serialized path form, I put it here.

bugsnag-android-core/src/main/java/com/bugsnag/android/DocumentPath.java
kstenerud: I'm not sure what you mean... 'a', 'b', etc are just examples to help explain how paths work. They don't actually exist.

bugsnag-android-core/build.gradle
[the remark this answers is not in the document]
fractalwrench: That warning can be ignored for now - we're stuck using `compile` for a bit longer due to the license plugin that we're using, so it's ok to use in this project.

bugsnag-benchmarks/src/androidTest/java/com/bugsnag/android/benchmark/JournalBenchmarkTest.kt
kstenerud: The previous solution was around 40k.

buildSrc/src/main/kotlin/com/bugsnag/android/BugsnagBuildPlugin.kt
```diff
@@ -115,6 +115,7 @@ class BugsnagBuildPlugin : Plugin<Project> {
             isCheckAllWarnings = true
             baseline(File(project.projectDir, "lint-baseline.xml"))
             disable("GradleDependency", "NewerVersionAvailable")
+            ignore("InvalidPackage")
```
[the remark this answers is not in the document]
kstenerud: Unfortunately, removing this line reintroduces the problems

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: Actually I just re-checked this, and it is actually threa-safe.

bugsnag-android-core/src/main/java/com/bugsnag/android/Journal.java
kstenerud: I've now moved them to internal, and changed the implementation to Kotlin.

bugsnag-android-core/build.gradle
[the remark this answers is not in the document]
kstenerud: If I use compile, it fails the lint check...

kstenerud:
This is handled in https://github.com/bugsnag/bugsnag-android/pull/1307 and https://github.com/bugsnag/bugsnag-android/pull/1306

PR #1279, opened 2021-06-18
Use direct field access when adding breadcrumbs and state updates

fractalwrench:
## Goal

The `Breadcrumb` and `StateEvent` implementation use getters rather than direct field access. While this is generally considered good practice, accessing the field directly is faster as it avoids the overhead of a method call - and as these methods could be called thousands of times through an application's lifecycle they should be optimized.

This has the added benefit of reducing bugsnag's dex count.

## Changeset

- Altered all `StateEvent` classes to use `@JvmField` rather than generating a getter.
- Updated `Breadcrumb` to use direct field access

## Testing

Relied on existing test coverage to confirm functionality works. Profiled the example app and ran benchmark tests to confirm there was no performance regression.


kstenerud (approved):

lemnik (approved):
LGTM

PR #1280, opened 2021-06-18
Discard unwanted automatic data earlier where possible

fractalwrench:
## Goal

Improves performance by discarding unwanted automatic data as early as possible. In the happy case (where an error/session is discarded) then this will remove any overhead from gathering stacktraces/metadata, which should be a significant saving.

## Changeset

The changeset has a lot of altered imports/reorganized tests so may be easier to view via the individual commits.

- Moved `ImmutableConfig` to `com.bugsnag.android.internal` and consolidated all logic for discarding automatically captured data. This involved renaming and documenting some existing methods.
- Made `ImmutableConfig` public (note it's under the `internal` package)
- Moved existing checks to _before_ the instantiation of the `Event` object. This now takes place at the point of error capture, and checks whether the `Throwable` itself is within the `discardClasses` or the `errorClass` for ANRs/JS/NDK errors
- Moved checks to _before_ the `Session` is instantiated, which avoids the collection of app/device metadata in `SessionTracker`
- Avoided logging when events/sessions are discarded as this can be a hot path and writing to Logcat could become a bottleneck - E2E test assertions have been updated accordingly

## Testing

Enhanced unit test coverage for discard logic.

This also relies on the existing E2E tests, specifically `ignored_reports` and `release_stage`. It's worth noting that these tests aren't exhaustive but are good enough to confirm that the functionality is working correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@
 * Avoid using regex to validate api key
   [#1282](https://github.com/bugsnag/bugsnag-android/pull/1282)
 
+* Discard unwanted automatic data earlier where possible
+  [#1280](https://github.com/bugsnag/bugsnag-android/pull/1280)
+
 ## 5.9.4 (2021-05-26)
 
 * Unity: add methods for setting autoNotify and autoDetectAnrs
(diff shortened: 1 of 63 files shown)


lemnik (approved):
LGTM

PR #1281, opened 2021-06-21
Trim stacktraces to <200 frames before attempting to construct POJOs

fractalwrench:
## Goal

The notifier trims stacktraces so that they contain a maximum of 200 frames. This means that very large stacktraces such as `StackOverflowError` do not exceed the API's payload limit.

This change ensures that the traces are trimmed _before_ constructing POJOs, which reduces unnecessary object instantiation in this scenario.

## Changeset

- Converted `stacktraceFromJavaTrace` companion object function into a constructor
- Ensured that both constructors in `Stacktrace` trim traces to a maximum of 200 frames

## Testing

Added unit test for new constructor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
 * Optimize metadata implementation by reducing type casts
   [#1277](https://github.com/bugsnag/bugsnag-android/pull/1277)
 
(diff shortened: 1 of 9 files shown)


lemnik (approved):
LGTM

PR #1282, opened 2021-06-21
Avoid using regex to validate api key

fractalwrench:
## Goal

The API key is currently validated using a regex, which was identified through profiling as taking ~3ms to compile and match. This alters the implementation so that it checks the string length and iterates through the characters instead. Additionally an expensive call to `String.format` has been replaced with concatenation.

## Testing

Increased unit test coverage and profiled example app to confirm that API key validation does not add on a large amount of time.


lemnik (approved):
LGTM

PR #1284, opened 2021-06-21
Enable AnrPlugin immediately if called on main thread

lemnik:
## Goal
Enable AnrPlugin immediately if called on main thread so that Application Not Responding errors can be caught within the application startup.

## Changeset
In the `AnrPlugin` is started on the main-thread `Looper` it starts immediately. If the current thread is not the main, it dispatches the startup logic using `Handler.postAtFrontOfQueue` to ensure we miss as few startup events as possible.

## Testing
Added new test scenario to the Mazerunner fixture, and modified the `MazerunnerApp.onCreate` to allow startup ANRs to be simulated.

The Mazerunner scenario creates a `SharedPreferences` option (picked up in `MazerunnerApp.onCreate`) and forces the application to terminate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
 * Discard unwanted automatic data earlier where possible
   [#1280](https://github.com/bugsnag/bugsnag-android/pull/1280)
 
+* Enable ANR handling on immediately if started from the main thread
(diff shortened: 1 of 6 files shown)


fractalwrench (approved):
LGTM

PR #1286, opened 2021-06-22
Use ring buffer to store breadcrumbs

fractalwrench:
## Goal

Breadcrumbs are stored in a `ConcurrentLinkedQueue` and then trimmed as they are added. This was identified as an area where performance could potentially be improved as over an application's lifecycle thousands of breadcrumbs might be added to Bugsnag.

## Changeset

`BreadcrumbState` has been altered to use a ring buffer instead of a `ConcurrentLinkedQueue`. Every time a breadcrumb is added an index is incremented by one, and the breadcrumb is stored in a backing array. If the number of breadcrumbs logged is greater than `config.maxBreadcrumbs` then the implementation wraps around to zero and overwrites the oldest breadcrumb added.

When an error is captured breadcrumbs must be added to an `Event` in the order in which they were captured. Previously this was a case of transforming a `Queue` into a `List` but using a ring buffer means that the breadcrumbs can be out of order. The solution chosen here is to create two sublists based on the position of the most recent breadcrumb and then combine them together.

## Performance improvements

Adding 1000 breadcrumbs to the store took ~9ms with these changes, and ~56ms without, a difference of ~45ms.

The `BreadcrumbState#copy()` implementation has added on ~0.5ms to the `Bugsnag.notify()` call.

changes_add_crumbs

baseline_crumbs

## Tradeoffs of this changeset

There are a couple of tradeoffs made with this changeset:

- `BreadcrumbState#copy()` does not sort elements or use synchronisation, so in rare cases this could lead to some breadcrumbs being ordered incorrectly. Sorting the collection was profiled as being expensive, but this is something that could potentially be added in if needed.
- This increases the time taken to execute `Bugsnag.notify()`, by the equivalent of adding ~11 breadcrumbs. This feels like a reasonable trade off as breadcrumbs should typically be added much more frequently than `Bugsnag.notify()` is called.

## Testing

Updated existing test coverage to match new method signatures, and added additional unit test cases for the ring buffer wrapping round. Additionally added a benchmark test for copying breadcrumbs (which happens whenever an event is generated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@
 * Cache results from PackageManager
   [#1288](https://github.com/bugsnag/bugsnag-android/pull/1288)
 
+* Use ring buffer to store breadcrumbs
+  [#1286](https://github.com/bugsnag/bugsnag-android/pull/1286)
+
 ## 5.9.4 (2021-05-26)
 
 * Unity: add methods for setting autoNotify and autoDetectAnrs
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbFilterTest.kt
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/BreadcrumbFilterTest.kt
@@ -24,8 +24,7 @@ class BreadcrumbFilterTest {
         client = generateClient(configuration)
 
         client.leaveBreadcrumb("Hello World")
-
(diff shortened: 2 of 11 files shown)


lemnik (approved):
LGTM

fractalwrench:
@kstenerud would you mind giving a secondary review for this changeset, as it has altered the core breadcrumb functionality?

kstenerud (approved):

PR #1288, opened 2021-06-23
Cache results from PackageManager

fractalwrench:
## Goal

`AppDataCollector` and `Configuration` both get the `PackageInfo` and `ApplicationInfo` from the `PackageManager`. This takes ~3.5ms on a OnePlus Pro 7 so the result should be cached to speed up initialization.

## Testing

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
@@ -23,8 +23,8 @@ internal class AppDataCollector(
     var codeBundleId: String? = null
 
     private val packageName: String = appContext.packageName
-    private var packageInfo = config.packageInfo
-    private var appInfo: ApplicationInfo? = config.appInfo
+    private var packageInfo get() = config.packageInfo
+    private var appInfo: ApplicationInfo? get() = config.appInfo
     private val bgWorkRestricted = isBackgroundWorkRestricted()
 
     private var binaryArch: String? = null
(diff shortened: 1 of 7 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
```diff
@@ -23,8 +23,8 @@ internal class AppDataCollector(
     var codeBundleId: String? = null
 
     private val packageName: String = appContext.packageName
-    private var packageInfo = packageManager?.getPackageInfo(packageName, 0)
-    private var appInfo: ApplicationInfo? = packageManager?.getApplicationInfo(packageName, 0)
+    private var packageInfo = config.packageInfo
+    private var appInfo: ApplicationInfo? = config.appInfo
```
lemnik, line 27:
```suggestion
    private var appInfo: ApplicationInfo? get() = config.appInfo
```

bugsnag-android-core/src/main/java/com/bugsnag/android/AppDataCollector.kt
```diff
@@ -23,8 +23,8 @@ internal class AppDataCollector(
     var codeBundleId: String? = null
 
     private val packageName: String = appContext.packageName
-    private var packageInfo = packageManager?.getPackageInfo(packageName, 0)
-    private var appInfo: ApplicationInfo? = packageManager?.getApplicationInfo(packageName, 0)
+    private var packageInfo = config.packageInfo
```
lemnik, line 26:
```suggestion
    private var packageInfo get() = config.packageInfo
```

lemnik (approved):
LGTM pending resolution of suggestions

PR #1289, opened 2021-06-24
Avoid expensive set construction in Config constructor

fractalwrench:
## Goal

Initializing `Configuration` was taking more time than it needed to due to the construction of some unnecessary sets. The `toSet()` implementation in Kotlin seems like it can be particularly slow in some circumstances, as it constructs a `LinkedHashSet`. This is unnecessary for the purposes of `Configuration`.

## Changeset

- Avoided creation of unnecessary field for `redactedKeys`
- Initialized `plugins` with a `HashSet` rather than a `LinkedHashSet`
- Set `enabledBreadcrumbTypes` to `null` rather than constructing a set - this has the same behaviour as setting all the values

Without these optimizations `Configuration` took ~38ms to initialize, whereas with them it took ~9ms. It's worth noting that this only seems to be on a cold start, suggesting that there is some one-off work required when the JVM is initialized. Therefore the differences don't show up in the automated benchmarks and have been manually profiled instead.

before_changes
after_changes

## Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@
 * Use ring buffer to store breadcrumbs
   [#1286](https://github.com/bugsnag/bugsnag-android/pull/1286)
 
+* Avoid expensive set construction in Config constructor
+  [#1289](https://github.com/bugsnag/bugsnag-android/pull/1289)
+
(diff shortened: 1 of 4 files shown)


lemnik (approved):
LGTM

PR #1290, opened 2021-06-24
Replace LinkedHashSet with LinkedHashSet where possible

fractalwrench:
## Goal

Replaces `LinkedHashSet` with `HashSet` where possible, as this was identified as quicker through performance profiling. As an example, using `HashSet` in `sanitiseConfiguration()` reduced the execution time from ~11ms to ~8ms.

## Testing

Relied on existing test coverage and profiled example app.


lemnik (approved):
LGTM

bugsnag-android-core/src/main/java/com/bugsnag/android/EventInternal.kt
```diff
@@ -55,7 +54,7 @@ internal class EventInternal @JvmOverloads internal constructor(
     protected fun shouldDiscardClass(): Boolean {
         return when {
             errors.isEmpty() -> true
-            else -> errors.any { discardClasses.contains(it.errorClass) }
+            else -> errors.any { config.discardClasses.toHashSet().contains(it.errorClass) }
```
lemnik, line 57: This will create a new `HashSet` for each `Error` I don't think we want that overhead?

bugsnag-android-core/src/main/java/com/bugsnag/android/Metadata.kt
```diff
@@ -97,7 +97,7 @@ internal data class Metadata @JvmOverloads constructor(
         }
 
         internal fun mergeMaps(data: List<Map<String, Any>>): MutableMap<String, Any> {
-            val keys = data.flatMap { it.keys }.toSet()
+            val keys = data.flatMap { it.keys }.toHashSet()
```
lemnik, line 100:
```suggestion
            val keys = data.flatMapTo(HashSet()) { it.keys }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/EventInternal.kt
fractalwrench: True - I've reverted this change.

PR #1291, opened 2021-06-24
Skip the startup_anr test on Android 10

lemnik:
## Goal
Skip the startup ANR test on Android 10 until we can find a work-around or solution.


fractalwrench (approved):
LGTM pending CI

twometresteve (approved):
LGTM - thanks

PR #1292, opened 2021-06-24
Register system callbacks on background thread

fractalwrench:
## Goal

Registers callbacks on a background thread to avoid unnecessary work in the `Client` constructor. This saves approximately ~10ms when calling `Bugsnag.start()`.

## Changeset

Registered the following on the background thread:

- `SystemBroadcastReceiver`, which receives intents that are turned into breadcrumbs
- Network connectivity changes
- Orientation changes
- Memory trim callbacks

Converted `SystemBroadcastReceiver` to kotlin while making changes. It's also worth noting that some methods are package visible to prevent the creation of synthetic accessors.

## Testing

Relied on existing test coverage, and ran an example app to test orientation/connectivity/system broadcast changes


bugsnag-android-core/src/main/java/com/bugsnag/android/SystemBroadcastReceiver.kt
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+
+        fun isAndroidKey(actionName: String): Boolean {
+            return actionName.startsWith("android.")
+        }
+
+        fun shortenActionNameIfNeeded(action: String): String {
+            return if (isAndroidKey(action)) {
+                action.substring(action.lastIndexOf(".") + 1)
```
lemnik, line 38:
```suggestion
                action.substringAfterLast('.')
```

bugsnag-android-core/src/main/java/com/bugsnag/android/SystemBroadcastReceiver.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private fun addExtrasToMetadata(
+        intent: Intent,
+        meta: MutableMap<String, Any>,
+        shortAction: String
+    ) {
+        val extras = intent.extras
+        extras?.keySet()?.forEach { key ->
+            val valObj = extras[key] ?: return@forEach
```
lemnik, line 69:
```suggestion
        extras?.forEach { (key, valObj) ->
           if(valObj == null) return@forEach
```

lemnik (approved):
LGTM - With some minor suggestions

PR #1293, opened 2021-06-24
Replace calls to String.format() with concatenation

fractalwrench:
## Goal

Replaces calls to `String.format` with concatenation. This was specifically found to be a problem in `DeviceDataCollector` which added ~1ms onto `Bugsnag.start()`, and profiling shows that using concatenation takes ~0.1ms instead.

As `String.format` isn't required in most cases throughout the codebase, it has been replaced with concatenation instead.

## Testing


lemnik (approved):
LGTM - With some suggestions

PR #1294, opened 2021-06-25
Upload logs for failed scenarios

twometresteve:
## Goal

MazeRunner now logs all received requests to file.  Upload these for failed scenarios to aid debugging.

## Changeset

A couple of other test improvements are also included:
- Remove running of "Flaky" scenarios - they are no longer needed
- Ignore Appium errors when clearing error dialogs (e.g. during an ANR)

## Testing

Covered by CI and the uploads can be seen working in the soft failing steps.


fractalwrench (approved):
I can't see the failed uploads in the soft-fail step, but think that would be expected as no requests are sent. Otherwise LGTM

PR #1295, opened 2021-06-25
fix: remove double-applied address offsets

kattrali:
## Goal

Fix a case where C/C++ events could be resolved to be 2-3 lines off from the actual source location, due to memory address offsets being applied twice. This change relies on the offset adjustment being performed server-side, using the program counter value to determine how much adjustment to apply.

PLAT-6861

## Changeset

Add a program counter indicator to native frames as needed

## Testing

- manual testing of events to determine the correct resolution of file/line info
- added new step to verify program counter flag is set

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
@@ -694,13 +694,20 @@ void bsg_serialize_error(bsg_error exc, JSON_Object *exception,
   json_object_set_string(exception, "errorClass", exc.errorClass);
   json_object_set_string(exception, "message", exc.errorMessage);
   json_object_set_string(exception, "type", "c");
-  for (int findex = 0; findex < exc.frame_count; findex++) {
+  // assuming that the initial frame is the program counter. This logic will
+  // need to be revisited if (for example) we add more intelligent processing
+  // for stack overflow-type errors, like discarding the top frames, which
+  // would mean no stored frame is the program counter.
+  if (exc.frame_count > 0) {
+    bsg_serialize_stackframe(&(exc.stacktrace[0]), true, stacktrace);
+  }
+  for (int findex = 1; findex < exc.frame_count; findex++) {
     bugsnag_stackframe stackframe = exc.stacktrace[findex];
-    bsg_serialize_stackframe(&stackframe, stacktrace);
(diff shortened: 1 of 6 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -694,20 +694,32 @@ void bsg_serialize_error(bsg_error exc, JSON_Object *exception,
   json_object_set_string(exception, "errorClass", exc.errorClass);
   json_object_set_string(exception, "message", exc.errorMessage);
   json_object_set_string(exception, "type", "c");
-  for (int findex = 0; findex < exc.frame_count; findex++) {
+  // assuming that the initial frame is the program counter. This logic will
+  // need to be revisited if (for example) we add more intelligent processing
+  // for stack overflow-type errors, like discarding the top frames, which
+  // would mean no stored frame is the program counter.
```
fractalwrench, line 700: Is there a ticket that tracks future improvements to this functionality?

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -694,20 +694,32 @@ void bsg_serialize_error(bsg_error exc, JSON_Object *exception,
...
+  if (exc.frame_count > 0) {
+    bsg_serialize_stackframe(&(exc.stacktrace[0]), true, stacktrace);
+  }
+  for (int findex = 1; findex < exc.frame_count; findex++) {
```
fractalwrench, line 704: nit: this for loop could be simplified by using `bool isPc = findex == 0`

fractalwrench (approved):
LGTM pending resolution of comments

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
kattrali: No, and no plans to do so. Just noted in here in case we ever do it.

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
kattrali: Correct, and the initial implementation worked that way, though this opted for perf over doing the index check every loop

PR #1297, opened 2021-06-25
Explain why the startup_anr.feature is skipped on Android 10

lemnik:
## Goal
Explain why we skip the startup ANR Mazerunner test on Android 10 devices


fractalwrench (approved):

twometresteve (approved):
LGTM

PR #1299, opened 2021-06-25
6711 journal metadata

kstenerud:
## Goal

Part of PLAT-6711. This PR adds the metadata field to the journal, to support backwards compatibility when changes occur.

## Design

This PR only affects the Journal portion.

```
```

## Testing

Added more unit tests to the journal tests for the metadata portion.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/Journal.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/Journal.kt
@@ -9,7 +9,7 @@ import java.io.OutputStream
  * Sometimes it's more efficient to serialize a base document and a list of modifications
  * to be made to it rather than to re-serialize the entire document after every modification.
  */
-class Journal {
+class Journal(private val type: String, private val version: Int) {
     private val commands: MutableList<Command> = mutableListOf()
 
     /**
@@ -52,6 +52,15 @@ class Journal {
      */
     @Throws(IOException::class)
     fun serialize(out: OutputStream) {
+        val journalInfoEntry = mutableMapOf<String, Any>(
+            JOURNAL_INFO_PATH to mutableMapOf(
+                TYPE_KEY to type,
+                VERSION_KEY to version
+            )
+        )
+        dslJson.serialize(journalInfoEntry, out)
+        out.write(0)
+
         for (command in commands) {
             command.serialize(out)
         }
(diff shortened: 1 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/Journal.kt
```diff
@@ -110,9 +141,28 @@ class Journal {
...
+            val length = bytes.size
+
+            // Decode the metadata entry
+            for (i in 0 until length) {
+                if (bytes[i] == 0.toByte()) {
+                    currentStart = i + 1
+                    break
+                }
+            }
+            require(currentStart > 0) { "Serialized journal doesn't contain a metadata entry" }
```
fractalwrench, line 154: nit: these messages should be updated to reword 'metadata' to 'version information' or whatever naming convention is settled upon

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/Journal.kt
```diff
@@ -9,9 +9,31 @@ import java.io.OutputStream
  * Sometimes it's more efficient to serialize a base document and a list of modifications
  * to be made to it rather than to re-serialize the entire document after every modification.
  */
-class Journal {
+class Journal(val metadata: MutableMap<String, Any>) {
     private val commands: MutableList<Command> = mutableListOf()
```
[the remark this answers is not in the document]
kstenerud: Hmm actually the more I think about this, the more it feels over-engineered. The original problem to make sure that older versions of the file can be migrated easily (thus the "type" and "version" fields). The "extra" metadata was just an offshoot from the fact that this data gets stored in a map.

But we actually have no need to store extra data, so YAGNI. Probably better to just store the fields individually in the Journal object and only play with maps when serializing/deserializing.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/Journal.kt
```diff
@@ -9,7 +9,7 @@ import java.io.OutputStream
  * Sometimes it's more efficient to serialize a base document and a list of modifications
  * to be made to it rather than to re-serialize the entire document after every modification.
  */
-class Journal {
+class Journal(val type: String, val version: Int) {
```
fractalwrench, line 12:
```suggestion
class Journal(private val type: String, private val version: Int) {
```

fractalwrench (approved):
LGTM pending resolution of one minor comment

PR #1300, opened 2021-06-25
Optimize capture of thread traces

fractalwrench:
## Goal

Adds some additional optimizations for capturing thread traces. Namely:

- Use `any()` rather than a for loop, which saves on iterator creation when the collection is empty and marginally improves performance
- Store the result of `getClassName()` to avoid multiple lookups
- Avoid calling `Thread.currentThread()` when `sendThreads` should not collect this information

## Testing


lemnik (approved):
LGTM

PR #1302, opened 2021-06-28
Fix flaky instrumentation test

fractalwrench:
## Goal

Fixes an instrumentation test that was failing on CI. This could occur because connectivity changes (which trigger automatic breadcrumbs) could happen at anytime, which alters the breadcrumbs reported by `Client`. It seems that this has been exacerbated by changes to how breadcrumbs are copied in #1286. As the test isn't asserting anything specific about automatic breadcrumbs, this disables them.

### DIFF
--- a/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
+++ b/bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
@@ -127,7 +127,7 @@ public class ClientTest {
     @Test
     public void testClientBreadcrumbRetrieval() {
         Configuration config = new Configuration("api-key");
-        config.setEnabledBreadcrumbTypes(Collections.EMPTY_SET);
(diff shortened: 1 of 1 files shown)


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ClientTest.java
```diff
@@ -126,7 +126,9 @@ public void testClientUser() {
 
     @Test
     public void testClientBreadcrumbRetrieval() {
-        client = generateClient();
+        Configuration config = new Configuration("api-key");
+        config.setEnabledBreadcrumbTypes(Collections.EMPTY_SET);
```
lemnik, line 130:
```suggestion
        config.setEnabledBreadcrumbTypes(Collections.emptySet());
```

lemnik (approved):
LGTM - With a suggestion

PR #1303, opened 2021-06-28
Avoid unnecessary BroadcastReceiver registration for monitoring device orientation

fractalwrench:
## Goal

Avoids unnecessary registration of a `BroadcastReceiver` that is used to monitor the device's orientation. The changeset implements the `ComponentCallbacks` instead to track when the orientation changes, and updates `DeviceDataCollector` so that it holds this value as state.

## Testing

Added to existing unit tests and verified manually that breadcrumbs are still automatically recorded for orientation changes.


lemnik (approved):
LGTM

PR #1305, opened 2021-06-29
Add note to release instructions

fractalwrench:
## Goal

Clarifies that the release should involve updating the bugsnag-js and bugsnag-unity libraries.


lemnik (approved):
LGTM

PR #1308, opened 2021-06-30
Revert "Replace Set implementations with HashSet where possible"

fractalwrench:
## Goal

Reverts bugsnag/bugsnag-android#1290 as this seems to have introduced an E2E test flake.


lemnik (approved):
LGTM

PR #1310, opened 2021-06-30
Respect manual setting of context

fractalwrench:
## Goal

If the manual context was set as null or an empty string, the automatic context was set on the `Event` instead. For JVM crashes this was acceptable as it's possible to set an `OnErrorCallback` and redact the context, but this is not possible on unhandled NDK crashes.

This changeset resolves this bug by always using the manual context if it has been set.

## Changeset

- Updated `ContextState` to track automatic and manual context separately
- Updated `ContextState` so that once manual context has been set, it is always used in precedence of automatic context
- Updated `Client`, `SessionTracker`, and `AppDataCollector` to use `ContextState` instead to hold information about the current context
- Updated native serializer to always use the context that has been set, rather than looking at the `activeScreenClass`

It's worth noting that because `Configuration.context` is `null` by default it's only possible to overwrite the manual context to null via `Bugsnag.setContext()`. Setting `Configuration.context` to any string however will have the desired effect.

## Testing

Updated existing test cases and added additional unit tests. Tested setting the manual context to null/an empty string/a populated value via configuration and also `Bugsnag.setContext()`, confirming that the filename is displayed instead of context where appropriate:

Screenshot 2021-06-30 at 15 48 08


lemnik (approved):
LGTM - Pending CI

PR #1311, opened 2021-06-30
Handle NullPointerExceptions from getActiveNetworkInfo

lemnik:
## Goal
Handle cases where `ConnectivityManager.getActiveNetworkInfo` throws a NullPointerException instead of returning `null` (as expected).

## Design
Treat the `NullPointerException`s as `null` values, and leave all other exceptions to propagate as normal to avoid breaking existing error-handling logic.

## Testing
Tested using mocking in unit tests as this has only been seen in very limited real-world situations.


fractalwrench (approved):
LGTM pending CI

PR #1312, opened 2021-07-08
Memory mapped stream

kstenerud:
## Goal

Part of PLAT-6961

This adds an OutputStream that writes to a memory mapped file. Memory mapped files are guaranteed to be consistent with the last memory write, even if the app crashes (because the memory mapping is maintained by the OS, not the app).

This allows us to always serialize the latest document changes, regardless of app conditions, and recover all journal entries on app relaunch.

## Testing

Added unit tests for the memory mapped stream.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
@@ -1,0 +1,63 @@
+package com.bugsnag.android.internal
+
+import java.io.File
+import java.io.OutputStream
+import java.io.RandomAccessFile
+import java.nio.BufferOverflowException
+import java.nio.MappedByteBuffer
+import java.nio.channels.FileChannel
+
+/**
+ * OutputStream wrapper around a shared memory file.
+ *
+ * Memory-mapped files are maintained by the operating system, and so they are guaranteed to match
+ * the last data written to the memory region, even if the app crashes.
+ *
+ * The memory mapped file is fixed to the length specified at instantiation, meaning that
+ * write() will throw java.nio.BufferOverflowException if the file is too full.
+ */
+class MemoryMappedOutputStream(
+    file: File,
+    private val bufferSize: Long,
+    private val clearedByteValue: Byte = 0
+) : OutputStream() {
+
+    private val raf: RandomAccessFile = RandomAccessFile(file, "rw")
+    private val memory: MappedByteBuffer =
+        raf.channel.map(FileChannel.MapMode.READ_WRITE, 0, bufferSize)
+
+    init {
+        raf.setLength(bufferSize)
+    }
+
+    fun clear() {
+        memory.rewind()
+        for (i in 1..bufferSize) {
+            memory.put(clearedByteValue)
+        }
+        memory.rewind()
+    }
+
+    override fun close() {
+        memory.force()
+        raf.close()
+    }
+
+    /**
+     * Write a byte value to the file.
+     * @throws java.nio.BufferOverflowException if the file is too full.
+     */
+    @Throws(BufferOverflowException::class)
+    override fun write(i: Int) {
+        memory.put(i.toByte())
+    }
+
+    /**
+     * Write a byte array to the file.
+     * @throws java.nio.BufferOverflowException if the file is too full.
+     */
+    @Throws(BufferOverflowException::class)
+    override fun write(b: ByteArray, off: Int, len: Int) {
+        memory.put(b, off, len)
+    }
+}
--- a/bugsnag-android-core/src/test/java/com/bugsnag/android/MemoryMappedOutputStreamTest.kt
+++ b/bugsnag-android-core/src/test/java/com/bugsnag/android/MemoryMappedOutputStreamTest.kt
@@ -1,0 +1,124 @@
+package com.bugsnag.android
+
+import com.bugsnag.android.internal.MemoryMappedOutputStream
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import java.io.File
+import java.nio.BufferOverflowException
+
+class MemoryMappedOutputStreamTest {
+    @Rule @JvmField
+    var folder: TemporaryFolder = TemporaryFolder()
+
+    @Test
+    fun testOpenClose() {
+        val file = folder.newFile("stream.memmapped")
+        val mm = MemoryMappedOutputStream(file, 4, clearedByteValue)
+        mm.close()
+        assertFileContents(file, byteArrayOf(0, 0, 0, 0))
+    }
+
+    @Test
+    fun testClear() {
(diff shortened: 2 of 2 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.channels.FileChannel
+
+/**
+ * OutputStream wrapper around a shared memory file.
+ *
+ * Memory-mapped files are maintained by the operating system, and so they are guaranteed to match
+ * the last data written to the memory region, even if the app crashes.
+ *
+ * The memory mapped file is fixed to the length specified at instantiation, meaning that
+ * write() will throw java.nio.BufferOverflowException if the file is too full.
```
fractalwrench, line 16: Could the relevant methods be annotated with [`@Throws`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/) to document this?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,61 @@
...
+ */
+class MemoryMappedOutputStream(
+    file: File,
+    private val bufferSize: Int,
+    private val clearedByteValue: Byte
+    ) : OutputStream() {
+
+    private val raf: RandomAccessFile = RandomAccessFile(file, "rw")
```
fractalwrench, line 24: The Javadoc for this class states that the rwd/rws mode ensures all updates are made synchronously to the storage device, and specifically mentions a system crash as a use-case. Is this something we'd want to consider or is rw sufficient?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val memory: MappedByteBuffer =
+        raf.channel.map(FileChannel.MapMode.READ_WRITE, 0, bufferSize.toLong())
+
+    init {
+        raf.setLength(bufferSize.toLong())
```
fractalwrench, line 29: Why can't `bufferSize` be declared as a `Long`?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun write(i: Int) {
+        memory.put(i.toByte())
+    }
+
+    /**
+     * Write a byte array to the file.
+     * @throws java.nio.BufferOverflowException if the file is too full.
+     */
+    override fun write(b: ByteArray?, off: Int, len: Int) {
+        requireNotNull(b)
```
fractalwrench, line 58:
```suggestion
    override fun write(b: ByteArray, off: Int, len: Int) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,61 @@
...
+ * Memory-mapped files are maintained by the operating system, and so they are guaranteed to match
+ * the last data written to the memory region, even if the app crashes.
+ *
+ * The memory mapped file is fixed to the length specified at instantiation, meaning that
+ * write() will throw java.nio.BufferOverflowException if the file is too full.
+ */
+class MemoryMappedOutputStream(
+    file: File,
+    private val bufferSize: Int,
+    private val clearedByteValue: Byte
```
fractalwrench, line 21: Could this use a default argument of 0x00, or some arbitrary value, if it doesn't matter too much what the value is?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override fun close() {
+        memory.force()
+        raf.close()
+    }
+
+    /**
+     * Write a byte value to the file.
+     * @throws java.nio.BufferOverflowException if the file is too full.
```
fractalwrench, line 47: What sort of strategy will the journal changes use to avoid the file getting too full? Presumably handling will take place higher up away from this class?

fractalwrench (commented):
Had a few queries which I've left as inline comments

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
kstenerud: I think it's better to not have synchronous, since OS crashes are extremely rare, and synchronous writes take a lot of time.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
kstenerud: They calling object will catch this exception and then do some housekeeping before retrying.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val memory: MappedByteBuffer =
+        raf.channel.map(FileChannel.MapMode.READ_WRITE, 0, bufferSize.toLong())
+
+    init {
+        raf.setLength(bufferSize.toLong())
+    }
+
+    fun clear() {
+        memory.rewind()
+        for (i in 1..bufferSize) {
```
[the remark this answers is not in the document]
kstenerud: This won't work with type Long, unfortunately.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
kstenerud: Added

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/MemoryMappedOutputStream.kt
kstenerud:
```kotlin
    fun addCommand(command: Journal.Command) {
        if (!isOpen) throw IllegalStateException("Cannot add commands to a closed document")
        try {
            command.serialize(journalStream)
        } catch (ex: BufferOverflowException) {
            snapshot()
            command.serialize(journalStream)
        }
        command.apply(document)
        journal.add(command)
    }
```

lemnik (approved):
LGTM

PR #1314, opened 2021-07-08
Add JournaledDocument

kstenerud:
## Goal

JournaledDocument uses Journal and MemoryMappedOutputStream to build a journal-backed document.

## Design

Document changes are made via journal commands, which are serialized to the memory-mapped file that MemoryMappedOutputStream writes to. Thus, the document is recoverable no matter when the app crashes.

The document contents can be fetched the same as any `Map<String, Any>`.

## Testing

Added unit tests for JournaledDocument

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
@@ -1,0 +1,178 @@
+package com.bugsnag.android.internal
+
+import android.os.Build
+import androidx.annotation.RequiresApi
+import java.io.Closeable
+import java.io.File
+import java.io.IOException
+import java.lang.IllegalStateException
+import java.nio.BufferOverflowException
+import java.util.function.BiConsumer
+
+/**
+ * A document with journal backing.
+ * Document snapshots are persisted to files, and journal entries are persisted to a memory-mapped
+ * file.
+ *
+ * Changes to the document must me made only via journal commands. Do not modify the document or any
+ * of its sub-components directly!
+ */
+class JournaledDocument
+/**
+ * Constructor.
+ * Upon construction, this document will have associated snapshot and journal files.
+ *
+ * @param baseDocumentPath The base path + filename to derive snapshot and journal names from.
+ * @param journalType A string identifying the type of journal (used for verification when the
+ *                    journal is reloaded)
+ * @param version The journal type version (to support migration)
+ * @param bufferSize The size of the shared memory buffer (and thus the maximum size of the journal)
+ * @param initialDocument The initial document contents (will be shallow copied)
+ */
+constructor(
+    baseDocumentPath: File,
+    journalType: String,
+    version: Int,
+    bufferSize: Long,
+    initialDocument: MutableMap<String, Any>
+) : Map<String, Any>, Closeable {
+    private val journalPath = getJournalPath(baseDocumentPath)
+    private val snapshotPath = getSnapshotPath(baseDocumentPath)
+    private val newSnapshotPath = getNewSnapshotPath(baseDocumentPath)
+
+    private val document = initialDocument.toMutableMap()
+    private val journal = Journal(journalType, version)
+    private val journalStream = MemoryMappedOutputStream(journalPath, bufferSize, clearedByteValue)
+    private var isOpen = true
+
+    init {
+        // Make sure any leftover files from the last instance at this path are gone.
+        newSnapshotPath.delete()
+        snapshot()
+    }
+
+    /**
+     * Add a journal command.
+     *
+     * This will first serialize the command to shared memory. If the memory is too full,
+     * it will create a snapshot to free up some room and try a second time.
+     * Next, it will apply the command to the document, and then add it to the journal itself.
+     */
+    fun addCommand(command: Journal.Command) {
+        if (!isOpen) {
+            throw IllegalStateException("Cannot add commands to a closed document")
+        }
+        try {
+            command.serialize(journalStream)
+        } catch (ex: BufferOverflowException) {
+            snapshot()
+            command.serialize(journalStream)
+        }
+        command.apply(document)
+        journal.add(command)
+    }
+
+    /**
+     * Save a current document snapshot and clear the journal
+     */
+    fun snapshot() {
+        if (!isOpen) {
+            throw IllegalStateException("Cannot snapshot a closed document")
+        }
+        JsonHelper.serialize(document, newSnapshotPath)
+        journal.clear()
+        journalStream.clear()
+        journal.serialize(journalStream)
+        newSnapshotPath.renameTo(snapshotPath)
+    }
+
+    /**
+     * Close the journal backing this document. All further modifications will throw exceptions.
+     */
+    override fun close() {
+        journalStream.close()
+        isOpen = false
+    }
+
+    override val entries: Set<Map.Entry<String, Any>> get() = document.entries
+    override val keys: Set<String> get() = document.keys
+    override val size: Int get() = document.size
+    override val values: Collection<Any> get() = document.values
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JsonHelper.kt
```diff
@@ -32,14 +32,14 @@ class JsonHelper private constructor() {
                 bytes,
                 bytes.size
             )
-            require(!document.isNullOrEmpty()) { "JSON document is invalid" }
+            require(document != null) { "JSON document is invalid" }
```
fractalwrench, line 35: nit: using `requireNotNull` in this class would be the idiomatic Kotlin way of doing things

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -0,0 +1,171 @@
...
+    private var isOpen = true
+
+    init {
+        // Make sure any leftover files from the last instance at this path are gone.
+        newSnapshotPath.delete()
+        snapshot()
+    }
+
+    val journalType: String get() = this.journal.type
+    val version: Int get() = this.journal.version
```
fractalwrench, line 54: nit: these could be `private` unless explicitly needed outside of this class, and could possibly also be declared in the constructor

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -0,0 +1,171 @@
...
+        }
+        command.apply(document)
+        journal.add(command)
+    }
+
+    /**
+     * Save a current document snapshot and clear the journal
+     */
+    fun snapshot() {
+        if (!isOpen) throw IllegalStateException("Cannot snapshot a closed document")
```
fractalwrench, line 79: style: the general approach at Bugsnag across all repos is to always use braces for single-line statements

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -0,0 +1,171 @@
...
+ * @param bufferSize The size of the shared memory buffer (and thus the maximum size of the journal)
+ * @param initialDocument The initial document contents (will be shallow copied)
+ */
+constructor(
+    baseDocumentPath: File,
+    journalType: String,
+    version: Int,
+    bufferSize: Long,
+    initialDocument: MutableMap<String, Any>
+) : Map<String, Any> {
```
fractalwrench, line 37: Is this inheritance hierarchy necessary? Methods from `Map` are unused in the test code and it seems like it'd be preferable to use composition rather than inheritance as this would hide the implementation detail.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -0,0 +1,171 @@
...
+     * it will create a snapshot to free up some room and try a second time.
+     * Next, it will apply the command to the document, and then add it to the journal itself.
+     */
+    fun addCommand(command: Journal.Command) {
+        if (!isOpen) throw IllegalStateException("Cannot add commands to a closed document")
+        try {
+            command.serialize(journalStream)
+        } catch (ex: BufferOverflowException) {
+            snapshot()
+            command.serialize(journalStream)
```
fractalwrench, line 69: This could (theoretically) fail with the same exception which would then be uncaught and terminate the process. Although it's unlikely for common cases it'd be good to handle this scenario.

The `serialize` method also throws an `IOException` - should that be handled here or higher up?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -0,0 +1,171 @@
...
+         *   - Attempt to load the ".journal" file, and apply it if it's valid.
+         *   - Return the finished document.
+         *
+         * @param baseDocumentPath The base path + filename to derive snapshot and journal names from.
+         * @return The document.
+         */
+        fun loadDocumentContents(baseDocumentPath: File): MutableMap<in String, out Any> {
+            val journalPath = File(baseDocumentPath.path + journalExtension)
+            val snapshotPath = File(baseDocumentPath.path + snapshotExtension)
+            val newSnapshotPath = File(baseDocumentPath.path + newSnapshotExtension)
```
fractalwrench, line 146: nit: there's similar code to calculate these paths in the constructor. It might be worth thinking about some abstraction that makes sure the path is only calculated in one place, in case it changes in future.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: It's an interface implementation. I want the document to be usable like a map (i.e. `myDocument["somekey"]`)

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: If it throws again, there's a serious problem. Any exceptions that escape this function must be handled by the caller (which will be the wrapping journal API that handles errors, enforces thread safety, etc).

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: The map interface is checked here: https://github.com/bugsnag/bugsnag-android/pull/1314/files#diff-b1235e74d61e2d17abb375db5d7c87ffc872a3908e1a01d2a7d1ea12fdb20f50R84

fractalwrench (approved):
LGTM - @lemnik will have a second look at things.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentTest.kt
```diff
@@ -0,0 +1,212 @@
...
+    private fun assertFileContents(file: File, expectedContents: ByteArray) {
+        val observedContents = file.readBytes()
+        Assert.assertArrayEquals(expectedContents, observedContents)
+    }
+
+    private fun assertJournalContents(file: File, size: Long, expected: ByteArray) {
+        val expectedContents = ByteArray(size.toInt())
+        expectedContents.fill(clearedByteValue)
+        expected.copyInto(expectedContents)
+        val journalPath = File(file.path + ".journal")
```
lemnik, line 193:
```suggestion
        val journalPath = File("${file.path}.journal")
```

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentTest.kt
```diff
@@ -0,0 +1,212 @@
...
+        assertFileContents(journalPath, expectedContents)
+    }
+
+    private fun assertSnapshotContents(file: File, expectedSnapshot: MutableMap<String, Any>) {
+        val snapshotPath = File(file.path + ".snapshot")
```
lemnik, line 198:
```suggestion
        val snapshotPath = File("${file.path}.snapshot")
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
lemnik: It might be worth also implementing `Closable` for clarity - and for the Kotlin `use` function

lemnik (approved):

PR #1317, opened 2021-07-13
Crash-time journal support

kstenerud:
## Goal

This PR implements the crash-time journal writing API for the journaling system

## Design

Everything in this part must be async-safe, so we can't use JSON libraries to implement it. Fortunately, the journal format only uses a subset of JSON.

## Testing

Added more unit tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
@@ -1,0 +1,460 @@
+#include "bugsnag_ndk.h"
+#include "utils/buffered_writer.h"
+#include "utils/number_to_string.h"
+#include "utils/string.h"
+#include <errno.h>
+#include <fcntl.h>
+#include <math.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+/* Crash-time journal writing functions. These must be async-safe and fast,
+ * which is why they use low-level open() and write() for everything, and make
+ * no calls to formatted output functions like sprintf.
+ *
+ * Journal writing functions end writing just after the opening quote whenever
+ * strings are involved. So for example, a function that leads up to writing a
+ * map key will end just after the '"', not before. The same goes for higher
+ * level string writing functions (not including write_string, write_raw_string,
+ * write_raw_bytes, and write_raw_literal).
+ *
+ * There are two versions of some functions to handle the requirement to close a
+ * quoted string, or to set up for a map.
+ */
+
+/* IMPORTANT: Keep this consistent with:
+ * - bugsnag_breadcrumb_type in
+ *       bugsnag-plugin-android-ndk/src/main/assets/include/event.h
+ * - BreadcrumbType in
+ *       bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbType.kt
+ */
+static const char *g_breadcrumb_types[] = {
+    "manual",     // BSG_CRUMB_MANUAL
+    "error",      // BSG_CRUMB_ERROR
+    "log",        // BSG_CRUMB_LOG
+    "navigation", // BSG_CRUMB_NAVIGATION
+    "process",    // BSG_CRUMB_PROCESS
+    "request",    // BSG_CRUMB_REQUEST
+    "state",      // BSG_CRUMB_STATE
+    "user",       // BSG_CRUMB_USER
+};
+static const int g_breadcrumb_types_count =
+    sizeof(g_breadcrumb_types) / sizeof(*g_breadcrumb_types);
+
+// The buffer size for the buffered writer
+#define BUFFER_SIZE (1024 * 128)
+
+// Implemented as defines to take advantage of compile-time size calculation
+// See write_raw_literal()
+#define PATH_METADATA_PREFIX "events.-1.metaData."
+#define PATH_NEXT_BREADCRUMB "events.-1.breadcrumbs."
+#define PATH_LAST_BREADCRUMB_METADATA_PREFIX                                   \
+  "events.-1.breadcrumbs.-1.metaData."
+#define PATH_USER "events.-1.user"
+#define KEY_ID "id"
+#define KEY_EMAIL "email"
+#define KEY_NAME "name"
+#define KEY_TIMESTAMP "timestamp"
+#define KEY_TYPE "type"
+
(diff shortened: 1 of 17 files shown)


bugsnag-plugin-android-ndk/src/test/cpp/main.c
```diff
@@ -8,24 +8,35 @@
...
+    if (temporary_folder == NULL) {
+        FAILm("Error retrieving temporary folder string");
+    }
+
     GREATEST_MAIN_BEGIN();
-    RUN_SUITE(string_utils);
-    RUN_SUITE(serialize_utils);
-    RUN_SUITE(breadcrumbs);
-    RUN_SUITE(event_mutators);
+// Uncomment these before merging
```
kstenerud, line 34: Reminder: Uncomment these before merging.

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
```diff
@@ -0,0 +1,587 @@
...
+    return write_literal("false");
+  }
+}
+
+static inline bool write_null() { return write_literal("null"); }
+
+// The following, including uint64_to_string(), int64_to_string(),
+// positive_double_to_string(), and double_to_string() were copied directly
+// from:
+// https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSJSONCodec.c#L90
```
kstenerud, line 121: This big chunk comes directly from bugsnag-cocoa.

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
```diff
@@ -0,0 +1,587 @@
...
+ */
+static size_t double_to_string(double value, char *dst, int max_sig_digits) {
+  if (value < 0) {
+    dst[0] = '-';
+    return positive_double_to_string(-value, dst + 1, max_sig_digits) + 1;
+  }
+  return positive_double_to_string(value, dst, max_sig_digits);
+}
+
+// ^^ End of copied code ^^
```
kstenerud, line 294: End of code copied from bugsnag-cocoa

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
```diff
@@ -0,0 +1,622 @@
...
+ * - BreadcrumbType in
+ *       bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbType.kt
+ */
+static const char *g_breadcrumb_types[] = {
+    "manual",     // BSG_CRUMB_MANUAL
+    "error",      // BSG_CRUMB_ERROR
+    "log",        // BSG_CRUMB_LOG
+    "navigation", // BSG_CRUMB_NAVIGATION
+    "process",    // BSG_CRUMB_PROCESS
+    "requsest",   // BSG_CRUMB_REQUEST
```
fractalwrench, line 36:
```suggestion
    "request",   // BSG_CRUMB_REQUEST
```

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
```diff
@@ -0,0 +1,622 @@
...
+ *
+ * There are two versions of some functions to handle the requirement to close a
+ * quoted string, or to set up for a map.
+ */
+
+/* IMPORTANT: Keep this consistent with:
+ * - bugsnag_breadcrumb_type in
+ *       bugsnag-plugin-android-ndk/src/main/assets/include/event.h
+ * - BreadcrumbType in
+ *       bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbType.kt
```
fractalwrench, line 28: This doesn't seem ideal - is there any way to enforce keeping these in sync?

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
fractalwrench: It might be worth extracting the functions to a separate file to make the separation obvious, and reduce the source file length

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
```diff
@@ -0,0 +1,622 @@
...
+
+// ^^ End of copied code ^^
+
+// Max significant digits + special chars for sign, exponent, etc.
+#define MAX_STRING_SIZE_FLOAT 22
+
+static inline bool write_double(double value) {
+  char buff[MAX_STRING_SIZE_FLOAT];
+  double_to_string(value, buff, MAX_SIGNIFICANT_DIGITS);
+  return write_raw_string(buff);
```
fractalwrench, line 346: What is the performance impact of calling `write()` each time one of these methods is called? A very long time ago the Cocoa notifier did a similar thing and introducing a buffer to avoid `write()` each time drastically improved performance. It'd be good to confirm that this isn't introducing similar issues: https://github.com/bugsnag/bugsnag-cocoa/pull/160#issuecomment-324703817

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
kstenerud: I can't think of any, short of some kind of multi-language code generator...

`event.h` already had synchronization requirements with `BreadcrumbType.kt` before. There is also other existing code with similar issues (any place we use enums that have JVM analogues).

fractalwrench: I guess we're stuck with this solution - at least the types of breadcrumbs should change very rarely, if at all.

kattrali: Yeah, the types have never changed.

bugsnag-plugin-android-ndk/src/main/jni/ctjournal.c
kstenerud: I've now replaced it with buffered_writer.c

fractalwrench (commented):
I'm happy with these changes - due to the changeset size it'd be nice to have a secondary review from either @lemnik or @kattrali to ensure that I've not missed anything.

kattrali (approved):

PR #1318, opened 2021-07-13
Capture process name in Event payload

fractalwrench:
## Goal

Captures the process name in the `Event` payload. This helps distinguish which process an error came from in multi-process apps. For API 28+ this uses `Application.processName()`, whereas for lower API levels reflection is used to access the same API using `ActivityThread`.

## Testing

Updated mazerunner tests to assert the process name is captured, and manually tested on API 16, 19, and 29 to confirm that each solution populates a valid value.


lemnik (approved):
LGTM

PR #1322, opened 2021-07-15
Ensure correct value always collected for activeScreen

fractalwrench:
## Goal

Ensures that the `activeScreen` always uses the value set in `SessionTracker`. This fixes a bug where the manual context (if set) was used instead, which was introduced in #1310.

## Testing

Added a test case that covers the issue.


lemnik (approved):
LGTM

PR #1324, opened 2021-07-16
Bump bugsnag dependencies in example app

fractalwrench:
## Goal

Bumps the bugsnag dependencies in the example app to newer versions as the current versions don't include newly released functionality.


lemnik (approved):
LGTM

PR #1325, opened 2021-07-19
Add thread safety to JournaledDocument.

kstenerud:
## Goal

Add thread-safety to the journal system.

A high-water system has also been added in order to support moving synchronization off to a background thread.

## Design

JournaledDocument now makes deep copies of the source document using concurrency-safe containers. All mutator methods are now synchronized.

## Testing

Added new tests for concurrency situations.


bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentTest.kt
```diff
@@ -137,6 +201,114 @@ class JournaledDocumentTest {
...
+            bufferSize - 10,
+            mutableMapOf("x" to 9)
+        )
+        document.addCommand(Journal.Command("a", 1))
+        document.addCommand(Journal.Command("b", 2))
+
+        assertJournalContents(
+            baseDocumentPath,
+            bufferSize,
+            (standardJournalInfoSerialized + "{\"a\":1}\u0000" + "{\"b\":2}\u0000").toByteArray(Charsets.UTF_8)
```
fractalwrench, line 248:
```suggestion
            "$standardJournalInfoSerialized{\"a\":1}\u0000{\"b\":2}\u0000".toByteArray(Charsets.UTF_8)
```

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
```diff
@@ -0,0 +1,145 @@
...
+import com.bugsnag.android.internal.JournaledDocument
+import com.bugsnag.android.internal.JsonHelper
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import java.io.File
+import java.lang.Thread
+
+class JournaledDocumentStressTest {
```
fractalwrench, line 12: This has added on ~7s to the unit test execution time, which is more than double its current speed. Can it be dialled down a bit so that it completes in <1s?

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
```diff
@@ -0,0 +1,145 @@
...
+            }
+            value = document["a$i"]
+            i = (i + 1) % maxI
+        }
+
+        // "use" value so that it doesn't get optimized away
+        print(value)
+
+        // Leave the OS time to flush the mem mapped file
+        Thread.sleep(100)
```
fractalwrench, line 58: Ideally this should be smarter - either there should be some polling to avoid waiting for 100ms each time, or there should be some way of detecting when the journal document has been modified and the entries can be read.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        while (modifyThread.state != Thread.State.TERMINATED) {
+            document.forEach {
+                // Just cycle through all entries
+            }
+            value = document["a$i"]
+            i = (i + 1) % maxI
+        }
+
+        // "use" value so that it doesn't get optimized away
+        print(value)
```
fractalwrench, line 55: Can an assertion be added to check each entry is valid? That would avoid the need for printing things to try and stop the compiler optimizing things away.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -27,20 +33,23 @@ class JournaledDocument
...
-    initialDocument: MutableMap<String, Any>
+    private val highWater: Long,
+    initialDocument: Map<String, Any>
 ) : Map<String, Any>, Closeable {
     private val journalPath = getJournalPath(baseDocumentPath)
     private val snapshotPath = getSnapshotPath(baseDocumentPath)
     private val newSnapshotPath = getNewSnapshotPath(baseDocumentPath)
 
-    private val document = initialDocument.toMutableMap()
+    private val document = convertDocumentToConcurrent(initialDocument)
```
fractalwrench, line 52: Is it possible to just expect callers to supply `ConcurrentHashMap`, or will the anticipated usage be passing in non-trivial data here? This would substantially reduce the complexity if this could be mandated as it'd allow for `convertDocumentToConcurrent` and associated functions to be removed.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -59,39 +83,57 @@ constructor(
...
-        journal.clear()
-        journalStream.clear()
-        journal.serialize(journalStream)
-        newSnapshotPath.renameTo(snapshotPath)
     }
 
     /**
      * Close the journal backing this document. All further modifications will throw exceptions.
      */
     override fun close() {
```
fractalwrench, line 128: This and other public functions could use some more error handling - for instance, if `snapshotInternal()` throws an exception then this would leave the class in an inconsistent state according to the contract of `Closeable`

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -59,39 +83,57 @@ constructor(
...
+        synchronized(journal) {
+            snapshotInternal()
+        }
+    }
+
+    /**
+     * Save a current snapshot if the journal has passed the high water mark.
+     * Use this in a BG thread to reduce the incidence of snapshotting on the critical path.
+     */
+    fun snapshotIfHighWater() {
```
fractalwrench, line 114: It'd be worth considering whether every function in this class needs to be public - this one seems like it could be `private`?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DocumentPathDirective.kt
```diff
@@ -58,7 +61,7 @@ internal interface DocumentPathDirective<C> {
      */
     open class ListIndexDirective(private val index: Int) : DocumentPathDirective<MutableList<in Any>> {
         override fun newContainer(): MutableList<in Any> {
-            return mutableListOf()
+            return CopyOnWriteArrayList()
```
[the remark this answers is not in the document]
kstenerud: This is the only way to be sure that `ConcurrentModificationException` never gets thrown. Alternatives would be to prohibit iteration over the document's contents at any level, or to include the caveat that anything iterating over the document in any way needs to also handle `ConcurrentModificationException`.

However, I figured that since there won't be more than 100 entries in a container at any level of the document ever, COW wouldn't be a big issue. Even if it were 1000 entries, that's only an 8k block of pointers to copy in two chunks for an insert or delete to that particular list/map in the document.

Also, not every journal entry is going to modify the same map or list. The most populous container we have is probably the stack trace, which is max 50 entries maybe, and more commonly 5-20. Next after that would be the breadcrumbs list, which we artificially limit.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: I wanted to ensure that there can be no modify-after-commit, and that requires a deep copy of the structure. I've had too many bad experiences in the past of phantom data changes because the same map got passed into two objects that didn't deep copy, and they then proceeded to stomp all over each others' data.

You can solve this with immutable containers, but this document is designed specifically to be modified (a lot), so I think it would be good to ensure that phantom modifications simply cannot occur by design.

We also won't be able to control the data being added by end users (if they add map metadata, we couldn't ensure the internal data's integrity should they modify the source map elsewhere), but this would be mitigated by deep copying.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: This function is designed to be called by a periodic housekeeping thread so that the actual act of snapshotting happens outside of the critical path as much as possible, so it needs to be public.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
kstenerud: I'm not sure how to make this smarter... Since it's a memory mapped file, we have no control over when it gets flushed to disk. I could use the modification time as a proxy, but then we run into race conditions where the flush happens before I sample the modification time, causing a flaky test. I can't just sample at the start because there could be any number of flushes during the course of the test and we can't be sure that the last modification was or was not flushed by the time the test verifies the file contents - it's OS controlled, which is why we're using it.

At this point in the test, it's making sure that an unclosed file contains what we expect it to, such that a crash before closing gives deterministic results (which we need or else we'll get corrupted data on relaunch). 100ms gives adequate time for the OS to flush, while not adding significant delays to the test.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentTest.kt
```diff
@@ -48,9 +56,15 @@ class JournaledDocumentTest {
     fun testEmptyDocumentNonEmptyJournal() {
         val baseDocumentPath = folder.newFile("mydocument")
```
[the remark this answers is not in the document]
kstenerud: I wasn't able to get this to work as a class member. It can't be initialized at object instantiation because `newFile()` would throw `java.lang.IllegalStateException: the temporary folder has not yet been created`.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
kstenerud: I've reduced the iterations by a factor of 10. Now the execution time is around 700ms.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DocumentPathDirective.kt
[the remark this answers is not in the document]
kstenerud: This is the only way to be sure that ConcurrentModificationException never gets thrown. Alternatives would be to prohibit iteration over the document's contents at any level, or to include the caveat that anything iterating over the document in any way needs to also handle ConcurrentModificationException.

However, I figured that since there won't be more than 100 entries in a container at any level of the document ever, COW wouldn't be a big issue. Even if it were 1000 entries, that's only an 8k block of pointers to copy in two chunks for an insert or delete to that particular list/map in the document.

Also, not every journal entry is going to modify the same map or list. The most populous container we have is probably the stack trace, which is max 50 entries maybe, and more commonly 5-20. Next after that would be the breadcrumbs list, which we artificially limit.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: I've added more checks and catches to ensure that every failure is checked, and `close()` completes its task before passing on an exception.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentTest.kt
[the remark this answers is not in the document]
fractalwrench: Could `File.createTempFile` do the job?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
fractalwrench: I don't have a good understanding of why the function needs to be public if `highWater` is a constructor parameter that controls when the function is called, and the function is only called internally. Is there other usage of this function anticipated than what's present at the moment?

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
```diff
@@ -0,0 +1,145 @@
...
+            bufferSize,
+            mutableMapOf()
+        )
+
+        val modifyThread = Thread {
+            for (i in 1..entryCount) {
+                document.addCommand("a$i", "x")
+            }
+        }
+        modifyThread.start()
```
lemnik, line 40:
```suggestion
        val modifyThread = thread {
            for (i in 1..entryCount) {
                document.addCommand("a$i", "x")
            }
        }
```

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
```diff
@@ -0,0 +1,145 @@
...
+            highWater,
+            mutableMapOf()
+        )
+
+        val modifyThread = Thread {
+            for (i in 1..entryCount) {
+                document.addCommand("a$i", "x")
+            }
+        }
+        modifyThread.start()
```
lemnik, line 91:
```suggestion
        val modifyThread = thread {
            for (i in 1..entryCount) {
                document.addCommand("a$i", "x")
            }
        }
```

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
```diff
@@ -0,0 +1,145 @@
...
+
+        Thread {
```
lemnik, line 93: Looks like this Thread isn't started? The Kotlin `thread` function starts by default.
```suggestion
        thread {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: This function is never called internally; it needs a controller object to call it.

The lines of responsibility are:
* `JournaledDocument` decides if it is in "high water" state (using the passed in constructor value) and snapshots if it is.
* `BugsnagMainJournal` decides when these "high water" checks occur.

`BugsnagMainJournal` will be triggering `document` (`JournaledDocument`) like this:

```kotlin
Thread {
    while(true) {
        Thread.sleep(highWaterPollingIntervalMS)
        document.snapshotIfHighWater()
    }
}
```

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentTest.kt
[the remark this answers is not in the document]
kstenerud: I started off using `File.createTempFile` but ran into permissions problems (it kept saying permission denied), which is why I switched to `TemporaryFolder`. `TemporaryFolder` also has the feature of auto-cleanup so that tests don't leave cruft behind even if the test fails or exceptions out.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
lemnik: Given that the snapshot & `addCommand` functions are both guarded by the `journal` monitor lock, I'm not clear on what a background thread adds here? Is it just so that snapshots can run concurrently with read operations?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
lemnik: Probably not strictly a problem: the mutable maps are leaked via delegation. So this is still technically a problem:
```kotlin
(journalDocument["key"] as MutableMap<String, Any?>)["something"] = true
```
I doubt it's worth addressing in this PR though.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DocumentPathDirective.kt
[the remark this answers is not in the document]
fractalwrench: I guess we can go with this for now and see later whether it impacts performance in real-world use.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
fractalwrench: Polling every ~5ms for a max timeout of ~2 seconds seems like it would do the trick? Waiting for 100ms wastes time unnecessarily but also might not be enough if I/O is slow - leading to flakes.

fractalwrench (approved):
Happy for this to be merged once all review comments are resolved

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: Yes, there's a big fat warning at the top of the class that says not to modify the document except via the journal commands. There's no way to actually enforce this invariant, unfortunately. But considering this class is not meant to be exposed for the user, we just need to be disciplined about it.

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledDocumentStressTest.kt
kstenerud: With the following modifications:

```kotlin
        while (modifyThread.state != Thread.State.TERMINATED) {
            document.forEach { _ ->
                // Just cycle through all entries
            }
            value = document["a$i"]
            i = (i + 1) % maxI
        }

        val baseline = document.snapshotPath.lastModified()

        for (i in 1..400) {
            if (document.snapshotPath.lastModified() != baseline) {
                break
            }
            Thread.sleep(5)
        }
```

Every test run duration was 2.6 seconds because the sync-to-disk had already happened before taking the baseline.

Disk I/O is to solid state storage, which even in the worst case will write the 500 byte block of data in far less than a millisecond. The 100ms pause was just to make absolutely sure that the OS has had so much time to schedule the write that it's impossible it didn't happen (it typically writes immediately, but can sometimes take a few milliseconds to do so on slower machines).

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: Yes, the locks protect against concurrent access (read and write) while the snapshot is going on.

The background thread is to push as many snapshot operations as possible onto that background thread so that they don't happen on the critical path. If the journal file overflows, it MUST be immediately flushed to disk to avoid data loss. So every time the high water code triggers on that separate thread, it mitigates an upcoming overflow, which means one less call to snapshot on the critical path.

lemnik (approved):
LGTM

PR #1326, opened 2021-07-21
Move example app AGP version to 3.6.4

fractalwrench:
## Goal

Fixes the example app build on CI by reverting to an older version of AGP.


lemnik (approved):
LGTM - Pending CI

PR #1328, opened 2021-07-22
Address pre-existing StrictMode violations

fractalwrench:
## Goal

Addresses some pre-existing StrictMode violations by moving affected code onto background threads where possible, and only calling `Future#get()` when the result is absolutely required.

## Changeset

- Altered various storage setup methods to avoid unnecessary calls (e.g. `DeviceIdStore` no longer calls `file.exists()` as `file.createNewFile()` is sufficient)
- Altered `ImmutableConfig` so that `persistenceDirectory` is a lazy property. This avoids calling `context#getCacheDir()` on the main thread
- Simplified the `Client` constructor by extracting the construction and configuration of objects into related `DependencyModules` which form a dependency graph. These make it simpler to run I/O on background threads and also lay the framework for greater dependency injection in future.
- Altered `device.freeDisk` calculation to run on IO thread
- Set a traffic stats tag on network requests
- Loaded plugins on a background thread

## Testing

Relied on existing test coverage. Confirmed the strictmode violations are resolved by enabling all penalties in the example app.


features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/StartupANRBehaviour.kt
```diff
@@ -20,23 +22,26 @@ fun Application.triggerStartupAnrIfRequired() {
         // we have to startup Bugsnag at this point
         Bugsnag.start(this)
 
-        thread {
-            // This is a dirty hack to work around the limitations of our current testing
-            // systems - where external key-events are pushed through our main thread (which we
-            // are pausing to test for ANRs)
+        // wait for Bugsnag's ANR handler to install first
+        Handler(Looper.getMainLooper()).post {
```
fractalwrench, line 26: Loading plugins has required an adjustment to the test scenario to wait until the first looper message has finished executing, as otherwise a startup ANR is missed. This is a bit of a trade off between always catching startup ANRs, and having a strictmode violation show up in the ANR plugin.

It might be possible to suppress the violation within `AnrPlugin`, or simply ignore it when creating a Bugsnag listener.

fractalwrench: In-person discussion decided that either the plugin should be responsible for loading specific parts on a bg thread, or the plugin should entirely load on the main thread and the violation can be ignored within the penaltyListener

lemnik (approved):
LGTM

PR #1329, opened 2021-07-22
Add main journal API

kstenerud:
## Goal

This is the top-level API to the journal system:

## Design

The journal at this layer is considered to be a read-only map that can only be modified via journal commands. There is an automatic housekeeping thread that snapshots the journal whenever it grows too big.

The top-level API is deliberately small and opinionated. Read access to the document is via the `Map` interface, modifications are via the `addCommand()` method, and there is also a `snapshot()` method for extenuating circumstances.

All opinionated configuration is set in BugsnagMainJournal's companion object, which we can tweak as needs arise. It cannot be changed by the end user.

This API will be initialized from `Client`, and will be accessed directly by the various systems that contribute data to the Bugsnag reports.

## Testing

Added top-level unit tests. The next phase will involve connecting the existing systems to the journal, which will affect their E2E tests.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
@@ -1,0 +1,171 @@
+package com.bugsnag.android
+
+import com.bugsnag.android.internal.DocumentPath
+import com.bugsnag.android.internal.JournaledDocument
+import java.io.File
+import java.io.IOException
+import java.lang.IllegalStateException
+import java.lang.Thread
+import kotlin.concurrent.thread
+
+/**
+ * The main journal contains the document that will be sent to the back end whenever an event occurs.
+ * All changes to the document must be made via journal commands to ensure that they are preserved
+ * in the event of a crash.
+ *
+ * The single instance to be used in Bugsnag is "instance", and startInstance() must be called before
+ * using it.
+ */
+internal class BugsnagMainJournal internal constructor(
+    private val logger: Logger,
+    baseDocumentPath: File,
+    initialDocument: Map<String, Any>
+) {
+    internal val journal = JournaledDocument(
+        baseDocumentPath,
+        journalType,
+        journalVersion,
+        mmapBufferSize,
+        highWater,
+        initialDocument
+    )
+
+    /**
+     * The document in its current state (with all current journal entries applied).
+     * Do NOT modify any of the document's contents directly; information will be lost if you do.
+     * Use addCommand() instead to modify the document.
+     */
+    val document: Map<String, Any> get() = journal
+
+    /**
+     * Add a journal command.
+     *
+     * @param documentPath The path describing where in the document to set the value
+     * @param value The value to set
+     * @see DocumentPath
+     */
+    fun addCommand(documentPath: String, value: Any?) {
+        if (documentPath.isEmpty()) {
+            logger.e("addCommand called with empty path (replace entire document not allowed)")
+            return
+        }
+        try {
+            return journal.addCommand(documentPath, value)
+        } catch (exc: IOException) {
+            logger.e("Could not add journal command", exc)
+        }
+    }
+
+    /**
+     * Save a current document snapshot and clear the journal.
+     * You'll rarely need to call this since the housekeeping thread already snapshots periodically.
+     */
+    fun snapshot() {
+        try {
+            journal.snapshot()
+        } catch (exc: IOException) {
+            logger.e("Could not write journal snapshot", exc)
+        }
+    }
+
+    private fun beginSnapshotHousekeepingThread() {
+        thread {
+            while (true) {
+                try {
+                    journal.snapshotIfHighWater()
+                } catch (exc: IOException) {
+                    logger.e("Could not write journal snapshot; exiting housekeeping thread...", exc)
+                    break
+                }
+                Thread.sleep(highWaterPollingIntervalMS)
+            }
+        }
+    }
+
+    companion object {
+        // Top-level key where version info will be saved.
+        internal const val versionInfoKey = "version-info"
+        internal const val journalTypeKey = "type"
+        internal const val journalVersionKey = "version"
+
+        // The "type" of the main journal. This will probably never change.
+        internal const val journalType = "Bugsnag Android"
+
+        // Version of this journal. This will change for migration purposes.
+        internal const val journalVersion = 1
+
+        // Size of the on-disk memory-mapped buffer in bytes.
+        private const val mmapBufferSize = 100000L
+
+        // If the memory-mapped buffer fills beyond this many bytes, it gets auto-snapshotted.
+        // Note: The size of mmapBufferSize and highWater have no effect on snapshot runtime cost.
+        private const val highWater = 50000L
+
+        // Polling interval in milliseconds for the high water check thread.
+        private const val highWaterPollingIntervalMS = 200L
+
+        /**
+         * The global instance of the main journal.
+         * You must call startInstance() before accessing this.
+         */
+        val instance: BugsnagMainJournal
+            get() = if (hasInitialized) {
+                mainJournal
+            } else {
+                throw IllegalStateException("Must call startInstance() before accessing instance")
+            }
+
+        /**
+         * Start the main journal at the specified base path.
+         * Any existing journal at this location will be overwritten.
+         *
+         * NOTE: This must be called once (and only once) before accessing instance.
+         *
+         * @param logger A logger
+         * @param baseDocumentPath Path to a filename that will be used as the basis for all journal and snapshot files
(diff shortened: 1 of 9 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
```diff
@@ -0,0 +1,175 @@
...
+            return docWithVersionInfo
+        }
+
+        /**
+         * Test interface to addCommand that removes all restrictions.
+         * This is required so that the unit tests can replace the entire document.
+         */
+        internal fun testintfAddCommand(documentPath: String, value: Any?) {
+            instance.journal.addCommand(documentPath, value)
+        }
```
lemnik, line 173: Given that all the variables here are `internal` is there any reason why these test-helper functions shouldn't be part of the Unit Test source set?

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
```diff
@@ -0,0 +1,175 @@
...
+                val journal = BugsnagMainJournal(logger, baseDocumentPath, initialDocument)
+                mainJournal = journal
+                hasInitialized = true
+                journal.beginSnapshotHousekeepingThread()
+                return previousContents
+            }
+        }
+
+        internal var hasInitialized = false
+        internal var mainJournal: BugsnagMainJournal? = null
```
lemnik, line 150: This looks like it might want to be `lateinit`:
```suggestion
        internal lateinit var mainJournal: BugsnagMainJournal
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ConcurrentContainers.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+import java.util.concurrent.CopyOnWriteArrayList
+
+/**
+ * Utility to convert containers to their concurrent counterparts.
+ */
+class ConcurrentContainers private constructor() {
+    companion object {
```
lemnik, line 11:
```suggestion
object ConcurrentContainers {
```

Alternatively - these might make nice extension functions? @fractalwrench ?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ConcurrentContainers.kt
```diff
@@ -0,0 +1,57 @@
...
+         */
+        internal fun asConcurrent(obj: Any?): Any? {
+            if (obj is Map<*, *>) {
+                return asConcurrent(obj)
+            }
+            if (obj is List<*>) {
+                return asConcurrent(obj)
+            }
+            return obj
+        }
```
lemnik, line 55:
```suggestion
        internal fun asConcurrent(obj: Any?): Any? = when(obj) {
            is Map<*, *> -> asConcurrent(obj)
            is List<*> -> asConcurrent(obj)
            else -> obj
        }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ConcurrentContainers.kt
```diff
@@ -0,0 +1,57 @@
...
+            @Suppress("UNCHECKED_CAST")
+            val concurrentMap = map as? ConcurrentMap<Any?, Any?> ?: ConcurrentHashMap(map)
+
+            map.forEach {
+                val entry = it.value
+                if (entry is Map<*, *>) {
+                    concurrentMap[it.key] = asConcurrent(entry)
+                } else if (entry is List<*>) {
+                    concurrentMap[it.key] = asConcurrent(entry)
+                }
```
lemnik, line 37:
```suggestion
                when(val entry = it.value) {
                    is Map<*, *> -> concurrentMap[it.key] = asConcurrent(entry)
                    is List<*> -> concurrentMap[it.key] = asConcurrent(entry)
                }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ConcurrentContainers.kt
```diff
@@ -0,0 +1,57 @@
...
+            @Suppress("UNCHECKED_CAST")
+            val concurrentList = list as? CopyOnWriteArrayList<Any?> ?: CopyOnWriteArrayList(list)
+
+            for (i in list.indices) {
+                val entry = list[i]
+                if (entry is Map<*, *>) {
+                    concurrentList[i] = asConcurrent(entry)
+                } else if (entry is List<*>) {
+                    concurrentList[i] = asConcurrent(entry)
+                }
```
lemnik, line 22:
```suggestion
                when(val entry = list[i])
                    is Map<*, *> -> concurrentList[i] = asConcurrent(entry)
                    is List<*> -> concurrentList[i] = asConcurrent(entry)
                }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
```diff
@@ -0,0 +1,175 @@
...
+         * @param baseDocumentPath Path to a filename that will be used as the basis for all journal and snapshot files
+         * @param initialDocument The initial contents of the document. This will be deep copied.
+         * @return The contents of the previous journal (if any)
+         */
+        fun startInstance(
+            logger: Logger,
+            baseDocumentPath: File,
+            initialDocument: Map<String, Any>
+        ): MutableMap<in String, out Any>? {
+            synchronized(hasInitialized) {
```
lemnik, line 130:
```suggestion
            synchronized(this) {
```

bugsnag-android-core/src/test/java/com/bugsnag/android/BugsnagMainJournalTest.kt
```diff
@@ -0,0 +1,133 @@
...
+    private val baseDocumentName = "mydocument"
+    private val baseDocumentPath: File by lazy {
+        if (!hasInitialized) {
+            hasInitialized = true
+            val basePath = File(folder.root, baseDocumentName)
+            basePath
+        } else {
+            baseDocumentPath
+        }
+    }
```
lemnik, line 25:
```suggestion
    private val baseDocumentPath by lazy { File(folder.root, "mydocument") }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * The main journal contains the document that will be sent to the back end whenever an event occurs.
+ * All changes to the document must be made via journal commands to ensure that they are preserved
+ * in the event of a crash.
+ *
+ * The single instance to be used in Bugsnag is "instance", and startInstance() must be called before
+ * using it.
+ */
+class BugsnagMainJournal internal constructor(
```
fractalwrench, line 19: Can this be made `internal` to hide it from our public API surface?

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
```diff
@@ -0,0 +1,175 @@
...
+        highWater,
+        initialDocument
+    )
+
+    /**
+     * The document in its current state (with all current journal entries applied).
+     * Do NOT modify any of the document's contents directly; information will be lost if you do.
+     * Use addCommand() instead to modify the document.
+     */
+    val document: Map<String, Any> get() = journal
```
fractalwrench, line 38: Perhaps this could be enforced with `Collections.unmodifiableMap()`?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ConcurrentContainers.kt
fractalwrench: Extension functions sounds good - all the symbols should also remain `internal`

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagMainJournal.kt
kstenerud: Doesn't `Map` already enforce that?

There are actually two levels of threat here:
- Modifying the map itself
- Modifying a map or list inside of the map

Unfortunately there's no way to enforce the second invariant because we need to be able to modify via the journal command system.

fractalwrench: Good point - I've got too much Java on the brain :)

fractalwrench (approved):
LGTM once the functions in `ConcurrentContainers` are converted into extension functions, and all symbols are made `internal`.

PR #1330, opened 2021-07-23
Add E2E test for strictmode VM policy violation detection

fractalwrench:
## Goal

Adds a new E2E test that triggers a StrictMode VmPolicy violation, specifically leaking a file URI. This currently only runs on Android <9, which [kills the app with a SIGKILL](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/StrictMode.java;l=2308?q=strictmode).

A later changeset will alter these tests to use `penaltyListener()` to call `Bugsnag.notify()` - for now the test simply asserts that no requests are sent.


lemnik (approved):
LGTM

PR #1331, opened 2021-07-26
Add Bugsnag listeners for StrictMode violation detection

fractalwrench:
## Goal

Adds `BugsnagThreadViolationListener` and `BugsnagVmViolationListener` to the public API. These classes implement StrictMode listeners and will call `Bugsnag.notify` if set as part of a [StrictMode policy](https://developer.android.com/reference/android/os/StrictMode).

By default, errors will be captured with a INFO severity level, a 'strictMode' severity reason, and an error message denoting whether it was a thread/vm policy violation. The classes supports chaining of listeners if a user desires this.

Note: this enforces that `Bugsnag.start()` must always be instantiated before setting the StrictMode listener. This was seen as preferable since otherwise a StrictMode violation might occur before bugsnag is initialized, and lead to `Bugsnag.getClient()` throwing an exception.

## Changeset

- Added `BugsnagThreadViolationListener` and `BugsnagVmViolationListener` to the public API
- Fixed bug in `updateSeverityInternal` which set the `severityReason` to `userCallbackSetSeverity`
- Added unit test coverage for new classes (including dealing with unexpected null inputs)
- Improved E2E test coverage for StrictMode scenarios so that on Android 9 they use the new APIs

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagThreadViolationListener.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagThreadViolationListener.java
@@ -1,0 +1,48 @@
+package com.bugsnag.android;
+
+import android.os.Build;
+import android.os.StrictMode.OnThreadViolationListener;
+import android.os.strictmode.Violation;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
+
+/**
+ * Sends an error report to Bugsnag for each StrictMode thread policy violation that occurs in
+ * your app.
+ * <p></p>
+ * You should use this class by instantiating Bugsnag in the normal way and then set the
+ * StrictMode policy with
+ * {@link android.os.StrictMode.ThreadPolicy.Builder#penaltyListener
+ * (Executor, OnThreadViolationListener)}.
+ * This functionality is only supported on API 28+.
+ */
+@RequiresApi(api = Build.VERSION_CODES.P)
+public class BugsnagThreadViolationListener implements OnThreadViolationListener {
+
+    private final Client client;
+    private final OnThreadViolationListener listener;
+
+    public BugsnagThreadViolationListener(@NonNull Client client) {
+        this(client, null);
+    }
+
+    public BugsnagThreadViolationListener(@NonNull Client client,
+                                          @Nullable OnThreadViolationListener listener) {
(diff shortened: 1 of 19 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagThreadViolationListener.java
```diff
@@ -0,0 +1,64 @@
...
+    public BugsnagThreadViolationListener(@NonNull Client client,
+                                          @Nullable OnThreadViolationListener listener) {
+        this.client = client;
+        this.listener = listener;
+    }
+
+    @SuppressWarnings("ConstantConditions")
+    @Override
+    public void onThreadViolation(@NonNull Violation violation) {
+        if (client == null || violation == null) { // bomb out early if supplied client was null
```
lemnik, line 43: Maybe this should still be delegating to any chained `listener`?

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagVmViolationListener.java
```diff
@@ -0,0 +1,64 @@
...
+    public BugsnagVmViolationListener(@NonNull Client client,
+                                      @Nullable OnVmViolationListener listener) {
+        this.client = client;
+        this.listener = listener;
+    }
+
+    @SuppressWarnings("ConstantConditions")
+    @Override
+    public void onVmViolation(@NonNull Violation violation) {
+        if (client == null || violation == null) { // bomb out early if supplied client was null
```
lemnik, line 43: Same question as above: this effectively swallows events for chained `listener`s

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagVmViolationListener.java
```diff
@@ -0,0 +1,64 @@
...
+            return;
+        }
+        client.notify(violation, new OnErrorCallback() {
```
lemnik, line 46: Could these two nearly-identical `OnErrorCallback` anonymous classes be refactored into a single package-protected implementation?

fractalwrench:
Updated implementation to always call the chained listener and to use the same `OnErrorCallback` implementation. It's worth noting that there's an outstanding question on the docs PR over what the constructor(s) should look like.

lemnik (approved):
LGTM

PR #1332, opened 2021-07-27
DocumentPath add/subtract support

kstenerud:
## Goal

Sometimes numeric document values need to be changed relative to their current value (incremented, decremented, added to, subtracted from) - for example counters. These operations must be supported as atomic operations, without requiring access to the original value.

## Design

Add support for an "addition" operator: If `+` is appended to the last path component, treat that last component (with the ending `+` character removed) as an instruction that the (assumed numeric) value is to be added to the (assumed numeric) value currently in the document. If there is no value at that place in the document, the document's "existing" value is assumed to be 0.

Note: The `+` character can be escaped, in which case it doesn't represent an addition operator.

Examples:

```
Original document:  {"foo": {"bar": 5}}
Path command:       {"foo.bar+": -100}
Resulting document: {"foo": {"bar": -95}}
```

```
Original document:  {}
Path command:       {"foo.bar+": -100}
Resulting document: {"foo": {"bar": -100}}
```

## Changeset

Added new DocumentPathDirectives to support addition, and updated the path parser to recognise `+`.

## Testing

Added more unit tests.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DocumentPath.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DocumentPath.kt
@@ -1,17 +1,20 @@
 package com.bugsnag.android.internal
 
+import java.lang.IllegalArgumentException
+
 /**
  * DocumentPath records a path into a hierarchical acyclic document, and can be used to modify
  * documents by navigating down the DAG and then applying a modification at that level.
  *
- * Paths are explained as follows in PLAT-6353:
- *
  * Paths follow a dotted notation similar to internet hostnames, such that each path component
- * is separated by a dot '.' If a literal dot is required, the path supports an escaping
- * mechanism using the backslash character:
+ * is separated by a dot '.' If a literal dot is required, use an escape sequence: A backslash
+ * character causes the parser to NOT treat the next character specially (regardless of whether
+ * it's a special character or not):
  *
  * \. = literal dot
  * \\ = literal backslash
+ * \+ = literal plus
+ * \z = literal z
  *
  * Path components are either integers or non-integers. Integers refer to the previous path
  * component as a list, providing the index to look up. Non-integers refer to the previous path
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DocumentPath.kt
```diff
@@ -199,28 +219,40 @@ class DocumentPath(path: String) {
...
-            for (i in path.length - 2 downTo 0) {
-                val ch = path[i]
-                isUnescapedDot = if (ch == ESCAPE_CHAR) {
-                    !isUnescapedDot
-                } else {
-                    break
+
+            if (path.length > 1 && path[path.lastIndex - 1] == ESCAPE_CHAR) {
+                return PathType.PATH_TYPE_NORMAL
+            }
```
kattrali, line 235: it looks like this case should be earlier, since it does not require `guessedPathType` and can be ruled out based solely on `path`.

bugsnag-android-core/src/test/java/com/bugsnag/android/DocumentPathTest.kt
```diff
@@ -14,7 +14,10 @@ class DocumentPathTest {
         val expected = mapOf<String, Any>("a" to 1)
 
         val observed = docpath.modifyDocument(document, value)
-        Assert.assertEquals(expected, observed)
+        Assert.assertEquals(
+            BugsnagTestUtils.normalized(expected),
+            BugsnagTestUtils.normalized(observed)
+        )
```
kattrali, line 20: If several assertions require the exact same setup, some variety of custom assertion seems worthwhile to keep things concise. Maybe something like `assertEqualNormalized()`?

kattrali (commented):
looks good overall, added a few questions inline

bugsnag-android-core/src/test/java/com/bugsnag/android/DocumentPathTest.kt
kstenerud: I've moved it to https://github.com/bugsnag/bugsnag-android/pull/1332/files#diff-972efa5cecbfa9d46ef7af453c6890918e814547ff95356ce929762014f4f83cR206

kattrali (approved):

PR #1333, opened 2021-07-29
7041 journal observer

kstenerud:
## Goal

PLAT-7041: Connect the main journal as a status event observer.

## Design

This is the first step in integrating the journal, and the lowest hanging fruit. There will need to be other PRs to connect other parts of the system to the journal at a later time.

## Testing

Added basic tests to verify that the journal gets updated from the standard observer events.


bugsnag-android-core/src/main/java/com/bugsnag/android/JournaledStateObserver.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    private fun handleInstallEvent(event: StateEvent.Install) {
+        addJournalCommand(JournalKeys.pathApiKey, event.apiKey)
+        addJournalCommand(JournalKeys.pathContext, client.context)
+
+        val appDataCollector = client.getAppDataCollector()
+        val appData = appDataCollector.generateAppWithState()
+        addJournalCommand(
+            JournalKeys.pathApp,
+            mapOf(
```
fractalwrench, line 118: I'm 90% sure that another function exists in the codebase that already coerces app/device into a map, it may be worth searching for it to reuse if possible

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournalKeys.kt
```diff
@@ -0,0 +1,59 @@
+package com.bugsnag.android.internal
+
+object JournalKeys {
```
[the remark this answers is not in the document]
kstenerud: The problem is that you have the same key values in different objects, and this will happen more as we add more information to the document.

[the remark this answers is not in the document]
lemnik: We could handle the duplicate-key issue by having the keys at multiple layers:

```object JournalKeys {
   internal const val totalMemory = "totalMemory"
   
   object Device {
      internal const val totalMemory = JournalKeys.totalMemory
   }
   
   object App {
      internal const val totalMemory = JournalKeys.totalMemory
   }
```

It might be a bit verbose, but the compiler will pack all of these down since they are all `const`.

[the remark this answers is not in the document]
kstenerud: The latest update adds `toJournalSection` methods to various data objects. I think this clears things up a fair bit since the keys are used in the objects themselves now.

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
```diff
@@ -56,6 +63,27 @@ internal class BugsnagMainJournal internal constructor(
...
+            if (command.first.isEmpty()) {
+                logger.e("addCommands called with empty path (replace entire document not allowed)")
+                return
+            }
+            try {
+                return journal.addCommands(*commands)
+            } catch (exc: IOException) {
+                logger.e("Could not add journal commands", exc)
+            }
+        }
```
lemnik, line 84:
```suggestion
        if(commands.any { (path, _) -> path.isEmpty() }) {
            logger.e("addCommands called with empty path (replace entire document not allowed)")
            return
        }
        
        try {
            return journal.addCommands(*commands)
        } catch (exc: IOException) {
            logger.e("Could not add journal commands", exc)
        }
```

bugsnag-android-core/src/test/java/com/bugsnag/android/BugsnagJournalTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            BugsnagTestUtils.normalized(expectedDocument),
+            BugsnagTestUtils.normalized(journal.document)
+        )
+    }
+
+    @Test
+    fun testManyJournalEntries() {
+        val journal = BugsnagJournal(NoopLogger, baseDocumentPath, mapOf())
+        journal.addCommand("one hundred", 100)
+        journal.addCommand("three\\.four\\.five\\.slash\\\\slash.-1", "x")
```
lemnik, line 81: Maybe these could be raw strings?
```suggestion
        journal.addCommand("""three\.four\.five\.slash\\slash.-1""", "x")
```

bugsnag-android-core/src/test/java/com/bugsnag/android/JournaledStateObserverTest.kt
```diff
@@ -0,0 +1,377 @@
...
+        return AppWithState(
+            BugsnagTestUtils.generateImmutableConfig(), null, null, null,
+            null, null, null, null, null, null
+        )
+    }
+
+    private fun journalContaining(document: Map<String, Any>): BugsnagJournal {
+        val journal = BugsnagJournal(NoopLogger, baseDocumentPath, document)
+        journal.snapshot()
+        return journal
```
lemnik, line 36:
```suggestion
        return BugsnagJournal(NoopLogger, baseDocumentPath, document).apply { snapshot() }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
```diff
@@ -82,12 +111,11 @@ internal class BugsnagMainJournal internal constructor(
         }
     }
 
-    companion object {
-        // Top-level key where version info will be saved.
-        internal const val versionInfoKey = "version-info"
-        internal const val journalTypeKey = "type"
-        internal const val journalVersionKey = "version"
+    override fun close() {
+        housekeepingThreadShouldKeepRunning = false
```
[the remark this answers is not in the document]
kstenerud: It will only sleep max 200ms, so it shouldn't be a problem.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
```diff
@@ -81,20 +82,40 @@ constructor(
...
-                command.serialize(journalStream)
-            }
-            document = command.apply(document)
-            journal.add(command)
+            requireOpen()
+            internalAddCommand(command)
+        }
+    }
+
+    fun addCommands(vararg commands: Journal.Command) {
```
fractalwrench, line 90: There are ~4 variations on `addCommand/addCommands` doing very similar things - could they be chained so that the important logic is only in one?

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
```diff
@@ -16,27 +17,33 @@ import kotlin.concurrent.thread
...
+    }
+
     /**
      * The document in its current state (with all current journal entries applied).
      * Do NOT modify any of the document's contents directly; information will be lost if you do.
      * Use addCommand() instead to modify the document.
      */
     val document: Map<String, Any> get() = journal
 
+    private var housekeepingThreadShouldKeepRunning = false
```
fractalwrench, line 45: Does this need to be marked as `@Volatile` if it could be set/read from multiple threads?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JsonHelper.kt
```diff
@@ -53,6 +54,8 @@ class JsonHelper private constructor() {
         fun deserialize(file: File): MutableMap<String, Any> {
             try {
                 FileInputStream(file).use { stream -> return deserialize(stream) }
+            } catch (ex: FileNotFoundException) {
```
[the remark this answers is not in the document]
kstenerud: I wanted it to show which filename had problems in the case of generic IO errors, but to specifically pass through file-not-found so that it can be reliably caught by a higher level caller (because in some cases it's not a fatal error if the file doesn't exist).

bugsnag-android-core/src/main/java/com/bugsnag/android/Device.kt
```diff
@@ -77,4 +79,19 @@ open class Device internal constructor(
         serializeFields(writer)
```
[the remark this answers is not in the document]
kstenerud: My next PR will add an interface `Journalable` that standardizes how to write to the journal, which will greatly simplify how this works across the 20 or so data objects that will eventually implement it. For now, I've only put in enough to make it work with the parts that are captured via the observers.

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
kstenerud: It's never read by another thread, and only changed when the journal is closed. The journal's housekeeping thread runs until this value becomes false (which only happens when close is called).

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/JournaledDocument.kt
kstenerud: They can't be chained because they run synchronized code at different levels to keep the amount of CPU churn down depending on how you're adding the commands. The common code has already been moved to `requireOpen()` and `internalAddCommand()`.

bugsnag-android-core/src/main/java/com/bugsnag/android/Device.kt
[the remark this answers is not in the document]
fractalwrench: As long as it's tracked by a ticket I'm happy!

PR #1334, opened 2021-08-02
Support zero-param constructor for StrictMode detection

fractalwrench:
## Goal

Adds a zero-parameter constructor for the StrictMode listeners. This avoids users having to call `Bugsnag.getClient()` unnecessarily.

The other constructors have been left as-is to allow for multiple `Client` instantiation and setting chained listeners.


lemnik (approved):
LGTM

lemnik (approved):

PR #1335, opened 2021-08-04
Add Journalable interface

kstenerud:
## Goal

Adds a `Journalable` interface. Implementations of this type can create a `Map<String, Any?>` which can be recorded as an entry in the journal. Most of the existing `Streamable` types have been converted to use this function where appropriate.

## Changeset

Implemented interface on App, AppWithState, Device, DeviceWithState, Metadata, Notifier, Stackframe, Thread, and User. The existing `Streamable` implementation calls `toJournalSection()` rather than writing a custom JSON serializer for each type.

## Testing

Relied on existing test coverage for JSON serializers to confirm that the values are populated correctly.


fractalwrench:
@lemnik I've now rebased these changes to the latest integration branch commit and implemented the Journalable interface across various Bugsnag types - so this will now be ready for review.

lemnik (approved):
LGTM

PR #1336, opened 2021-08-04
Bump gradle/AGP dependencies

fractalwrench:
## Goal

Bumps Gradle/AGP in the example app to the latest available versions without updating all the way to AGP 7.

The example app requires a clean assemble to workaround an apparent bug in AGP/Gradle which would throw a `ZipException` as some build cache exists.


lemnik (approved):
LGTM

PR #1337, opened 2021-08-04
Increase instrumentation test timeout

twometresteve:
## Goal.

Increase timeout for the instrumentations tests as they have been very variable recently (I've also raised a support ticket with BrowserStack).

## Changeset

Increased timeout and remove duplication of step from the pipeline files.

## Testing

Covered by CI.


fractalwrench (approved):
LGTM

PR #1342, opened 2021-08-13
fix(lowMemory): align the app.lowMemory field for JVM and NDK

lemnik:
## Goal
The `app.lowMemory` attribute should be reported in the same way for all errors.

## Design
Cache the most recent value sent to `ComponentCallbacks.onLowMemory` and `ComponentCallbacks2.onTrimMemory` in the `AppDataCollector` and report that value instead of looking-up the value at crash-time.

When we receive a trim memory event that indicates any state less than [COMPLETE](https://developer.android.com/reference/android/content/ComponentCallbacks2#TRIM_MEMORY_COMPLETE) we set the cached lowMemory state to `false`.

## Testing
Manual testing for both native and JVM crashes, and a new unit test to cover the mapping between memory-trim values and the `lowMemory` flag.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
@@ -340,6 +340,7 @@ public class Client implements MetadataAware, CallbackAware, UserAware {
                 }, new Function1<Boolean, Unit>() {
                     @Override
                     public Unit invoke(Boolean isLowMemory) {
+                        appDataCollector.setLowMemory(Boolean.TRUE.equals(isLowMemory));
                         clientObservable.postMemoryTrimEvent(isLowMemory);
                         return null;
                     }
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ClientComponentCallbacks.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ClientComponentCallbacks.kt
@@ -1,13 +1,13 @@
 package com.bugsnag.android
 
-import android.content.ComponentCallbacks
+import android.content.ComponentCallbacks2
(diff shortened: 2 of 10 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ClientComponentCallbacks.kt
```diff
@@ -18,6 +18,14 @@ internal class ClientComponentCallbacks(
         }
     }
 
+    override fun onTrimMemory(level: Int) {
+        if (level >= ComponentCallbacks2.TRIM_MEMORY_COMPLETE) {
+            callback(true)
+        } else {
+            callback(false)
+        }
```
kattrali, line 26:
```suggestion
        callback(level >= ComponentCallbacks2.TRIM_MEMORY_COMPLETE)
```
?

bugsnag-android-core/src/main/java/com/bugsnag/android/Client.java
```diff
@@ -340,6 +340,7 @@ public Unit invoke(String oldOrientation, String newOrientation) {
                 }, new Function1<Boolean, Unit>() {
                     @Override
                     public Unit invoke(Boolean isLowMemory) {
+                        appDataCollector.setLowMemory(isLowMemory == Boolean.TRUE);
```
kattrali, line 343: Should this be checking the value of the instance instead of equality with the static constant? I highly doubt this would end up with [a constructed `Boolean()` instance](https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#Boolean(boolean)) but still.

```suggestion
                        appDataCollector.setLowMemory(isLowMemory.booleanValue());
```

kattrali (approved):

PR #1353, opened 2021-09-01
Prepare project for AGP 7/Gradle 7

fractalwrench:
## Goal

Performs several pre-requisites before updating the project to use AGP 7/Gradle 7, the latest stable version.

## Changeset

- Bumped the test fixture and repo to use AGP 4.2.2
- Bumped the example app to AGP 7 and Gradle 7 as there is a clean upgrade path
- Removed unused bintray plugin and jcenter from the test fixture
- Altered dockerfile to copy settings.gradle before calling `./gradlew` (this results in an error on Gradle 7)
- Bumped gradle plugins to latest versions and suppressed any pre-existing static analysis warnings

## Testing

Manually ran example app and relied on CI checks.

PR #1354, opened 2021-09-01
Update project to build using Gradle/AGP 7

fractalwrench:
## Goal

Updates the repository to use Gradle 7, building on top of #1353.

## Changeset

- Updated Gradle wrapper to 7.2
- Bumped license plugin to 0.16.1 which supports Gradle 7
- Addressed most AGP 7 deprecations
- Bumped AGP to 7
- Set `abiFilters` on `android.defaultConfig.ndk` as suggested by the NDK setup docs
- Increased NDK revision from r16 to r17 to circumvent a [packaging bug](https://issuetracker.google.com/issues/198307779) in AGP 7. The [revision history](https://developer.android.com/ndk/downloads/revision_history) does not indicate that this will cause any breaking changes.

## Testing

Relied on existing CI coverage.


lemnik (approved):
LGTM

PR #1355, opened 2021-09-01
Unity/new session methods

rich-bugsnag:
For the unity notifier the android layer will now handle all session tracking. 

Therefore I needed to add StartSession, PauseSession and ResumeSession to the interface so that the Unity layer can interact with the current session.


fractalwrench (approved):

rich-bugsnag:
Please don't merge this yet, i want to wait until the feature is 100% working on unity just in case changes are needed

rich-bugsnag:
@fractalwrench This is ready now and the unity side is done. I'll leave it for you to merge and then before the next unity release we can work out the best place to point the Unity sub mod

fractalwrench (approved):

PR #1359, opened 2021-09-07
Setup OkHttp MockWebServer unit tests

fractalwrench:
## Goal

Creates a unit test that uses [MockWebServer](https://github.com/square/okhttp/tree/master/mockwebserver) to simulate network requests. This will be used to test the implementation of network breadcrumbs by confirming that `leaveBreadcrumb` is called on the `Client`.


lemnik (approved):
LGTM

PR #1360, opened 2021-09-08
Add OkHttp network breadcrumb scenario

fractalwrench:
## Goal

Adds an E2E scenario which makes an OkHttp call. When network breadcrumbs are fully implemented this test will provide a scaffold that allows for the functionality to be verified.

## Changeset

- Added a scenario which makes a network request to load google.com
- Setup test fixture to include OkHttp dependency
- Setup buildkite so that the r16 test fixture is built with OkHttp 3 (which has a minSdk of 9), and all other test fixtures are built with OkHttp (which has a minSdk of 21)


lemnik (approved):
LGTM

PR #1361, opened 2021-09-09
Capture breadcrumbs for OkHttp network requests

fractalwrench:


bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/BugsnagOkHttpPlugin.kt
```diff
@@ -1,14 +1,148 @@
...
+
+    /**
+     * Constructs a map of query parameters, redacting any sensitive values.
+     */
+    private fun Client.buildQueryParams(request: Request): Map<String, String?> {
+        val url = request.url
+        val params = mutableMapOf<String, String?>()
+
+        url.queryParameterNames.forEach { name ->
+            params[name] = url.queryParameter(name)
```
lemnik, line 109: We're only capturing a single value for each query parameter. Certainly the 99% case is covered, but I wonder if it's not worth having some logic to handle repeated parameters?

bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/BugsnagOkHttpPlugin.kt
```diff
@@ -1,14 +1,148 @@
...
+     * Sanitizes the URL by removing query params.
+     */
+    private fun sanitizeUrl(request: Request): String {
+        val url = request.url
+        val builder = url.newBuilder()
+
+        url.queryParameterNames.forEach { name ->
+            builder.removeAllQueryParameters(name)
+        }
+        return builder.build().toString()
```
lemnik, line 124:
```suggestion
        val builder = url.newBuilder().query(null).build().toString()
```

lemnik (approved):
LGTM With some minor comments

bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/BugsnagOkHttpPlugin.kt
fractalwrench: `query()` is deprecated in OkHttp 4 unfortunately - doing it the long-winded way maintains compatibility between 3 and 4 with the same plugin.

bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/BugsnagOkHttpPlugin.kt
fractalwrench: `queryParameterNames` is a `Set` so I think it shouldn't be possible to have duplicate keys?

fractalwrench: Updated so that a list is now added if more than one value is present.

PR #1363, opened 2021-09-13
Alter network breadcrumb metadata collection

fractalwrench:
## Goal

Alters the collection of network breadcrumb metadata so that the `status` and `responseContentLength` fields are only recorded when appropriate. Additionally this records an error breadcrumb when a request does not complete (e.g. due to a timeout).

## Changeset

- Respected the value of `config.enabledBreadcrumbTypes` by not capturing network breadcrumbs when `BreadcrumbType.REQUEST` is not present
- If a request is cancelled then an `"OkHttp call error"` breadcrumb is captured
- Set `requestContentLength/responseContentLength` to 0 by default
- Only record the HTTP status code and `responseContentLength` if the request did not error
- Mapped 1xx-3xx requests into success, 4xx-5xx requests into failure, and everything else into an error
- Suppressed unnecessary detekt warnings

## Testing

Updated unit tests to cover new functionality and tweaked existing assertions. Integration tests will be added in a separate changeset for ease of review.


bugsnag-plugin-android-okhttp/src/test/java/com/bugsnag/android/okhttp/NetworkBreadcrumbIntegrationTest.kt
```diff
@@ -1,63 +0,0 @@
-package com.bugsnag.android.okhttp
```
fractalwrench, line 1: This has been split into various integration tests which will be added in a separate changeset for ease of review

lemnik (approved):
LGTM

PR #1365, opened 2021-09-13
Add additional integration tests for network breadcrumbs

fractalwrench:
## Goal

Adds some additional integration tests for network breadcrumbs, which cover the following scenarios:

- A request served from a local cache
- Requests that timeout
- A request that is cancelled
- Requests whose body is not closed, which is recommended against in the OkHttp documentation as it leaks resources and prevents the request from finishing
- A redirected request


lemnik (approved):
LGTM

PR #1366, opened 2021-09-13
Increased default breadcrumb collection limit to 50

fractalwrench:
## Goal

Increases the default number of breadcrumbs collected to 50. This is required now that network breadcrumbs are anticipated to capture several more requests in a typical application, which would lead to increased breadcrumb churn at 25 items.

## Changeset

- Increased JVM and NDK limits to 50
- Added a migration to copy breadcrumbs correctly from the old `bugsnag_event` struct to the new one
- Altered setting of NDK user information, which was broken by increasing the breadcrumb array size in the struct

## Testing

Updated existing test assertions. Added extra test cases for the NDK struct migration, which cover the first index in the ring buffer starting at 0 and non-zero values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 * Update project to build using Gradle/AGP 7
   [#1354](https://github.com/bugsnag/bugsnag-android/pull/1354)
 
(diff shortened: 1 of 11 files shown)


lemnik (approved):
LGTM

PR #1367, opened 2021-09-13
Capture and report `Thread.state` for Android Runtime threads

lemnik:
## Goal
Capture and report the [Thread.state](https://developer.android.com/reference/java/lang/Thread.State) of all captured threads to help in diagnosing issues.

## Design
To separate our implementation from any future platform changes or errors, a new Bugsnag `Thread.State` enum was introduced, reflecting the possible thread-states currently available on Android. The new enum also includes an `UNKNOWN` state which is used in cases where we either cannot capture or map the platform state.

## Testing
Manual testing with the Example app and a debug server. Updated existing unit tests and end-to-end tests to include checks for the new Thread.State field.


bugsnag-android-core/src/test/java/com/bugsnag/android/ThreadSerializationTest.kt
```diff
@@ -98,6 +101,7 @@ internal class ThreadSerializationTest {
                 "main-one",
                 ThreadType.ANDROID,
                 false,
+                Thread.State.RUNNABLE,
```
fractalwrench, line 104: It'd be good to have a test case that serializes something other than `RUNNABLE`

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ThreadDeserializer.java
```diff
@@ -32,6 +32,7 @@ public Thread deserialize(Map<String, Object> map) {
                 MapUtils.<String>getOrThrow(map, "name"),
                 ThreadType.valueOf(type.toUpperCase(Locale.US)),
                 errorReportingThread,
+                Thread.State.byName(MapUtils.<String>getOrThrow(map, "state")),
```
fractalwrench, line 35: Will this require any updates to the React Native JavaScript layer for the `state` field to get passed down? It feels like it might work out-of-the-box but if not then this might result in threads being dropped from JS payloads. Just something to check when updating RN.

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadInternal.kt
```diff
@@ -18,6 +19,7 @@ class ThreadInternal internal constructor(
         writer.name("id").value(id)
         writer.name("name").value(name)
         writer.name("type").value(type.desc)
+        writer.name("state").value(state.name)
```
fractalwrench, line 22: Double-checking - will `getName()` be safe if ProGuard obfuscation is enabled?

fractalwrench (approved):

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadInternal.kt
lemnik: Progaurd doesn't obfuscate enum names by default - but that doesn't mean other tools won't. I'll sort out something more robust 👍

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ThreadDeserializer.java
lemnik: The ReactNative JS layer doesn't directly know about threads, however the typescript definitions in `common.d.ts` will need to be changed to reflect the `state` property.

fractalwrench (approved):

PR #1371, opened 2021-09-15
Introduce the `bsg_thread` structure in the NDK plugin

lemnik:
## Goal
Create and integrate a new structure to carry the NDK thread state information that we will report, not including the required data migration (which is in a separate task). A fixed-size array of `bsg_thread`s are included in `bugsnag_event` and are serialized into our standard JSON payload.

## Testing
A new unit test was added to test that the new model serializes to JSON correctly.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.h
@@ -24,6 +24,13 @@
  */
 #define BUGSNAG_DEFAULT_EX_TYPE "c"
 #endif
+#ifndef BUGSNAG_THREADS_MAX
+/**
+ * Maximum number of threads recorded for an event. Configures a default if not
+ * defined.
+ */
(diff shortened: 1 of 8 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -192,6 +205,9 @@ typedef struct {
   int crumb_first_index;
   bugsnag_breadcrumb breadcrumbs[BUGSNAG_CRUMBS_MAX];
 
+  int thread_count;
+  bsg_thread threads[BUGSNAG_THREADS_MAX];
```
kattrali, line 209: These fields can't be added without a migration, particularly since its in the middle of the structure instead of the end.

lemnik: Yup, migration is in a separate task. Will move the new data to the end of the struct to make that easier.

kattrali (approved):

PR #1373, opened 2021-09-16
Remove obsolete abiFilters check in publish script

fractalwrench:
## Goal

The NDK version and abiFilters used are now pinned via the Gradle DSL, whereas in the past they were subject to whatever NDK revision was installed. This check needs to be removed to allow publishing as it currently fails. I've confirmed that the 4 supported ABIs are included in the generated AAR.


lemnik (approved):
LGTM

PR #1375, opened 2021-09-16
Support integer values in buildUuid

fractalwrench:
## Goal

Supports integer values in the buildUuid field rather than assuming the Bundle value will always be a string.

## Testing

Added unit tests and confirmed in the dashboard that an event with an int buildUuid symbolicates.


lemnik (approved):
LGTM

PR #1376, opened 2021-09-16
migrate ndk reports for the new threads property

lemnik:
## Goal
Ensure that events stored by previous versions of the NDK plugin can be migrated to the new version, which includes space allocated for the `bsg_thread` structures.

## Design
As the new fields are at the end of the `bugsnag_event` structure - this migration is a simple copy operation (using `memcpy`) to allow for the expanded space where the threads will be stored. As we are using `calloc` no further action is required to zero the `thread_count`.

## Testing
Manually tested in the example app, and added migration unit tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.h
@@ -34,7 +34,7 @@
 /**
  * Version of the bugsnag_event struct. Serialized to report header.
  */
-#define BUGSNAG_EVENT_VERSION 6
+#define BUGSNAG_EVENT_VERSION 7
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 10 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -178,6 +185,12 @@ typedef struct {
   char url[64];
 } bsg_notifier;
 
+typedef struct {
+  pid_t id;
+  char name[16];
```
fractalwrench, line 190: Will this definitely be large enough to fit any thread name?

fractalwrench (approved):

fractalwrench:
One other comment - it'd be good to ensure this has an entry in the changelog

bugsnag-plugin-android-ndk/src/main/jni/event.h
lemnik: This mirrors the Kernel process name, as it'll report in `/proc/stat`. Happy to increase it, but it might be a bit of waste?

PR #1378, opened 2021-09-17
Use SystemClock.elapsedRealtime to track `app.durationInForeground`

fractalwrench:
## Goal

Uses `SystemClock.elapsedRealtime()` rather than `System.currentTimeMillis()` to track the `duration` and `durationInForeground` values. [SystemClock](https://developer.android.com/reference/android/os/SystemClock) is recommended for timing intervals as it is not affected by CPU power saving mode or the user altering the device time.

`System.currentTimeMillis()` is still used in the codebase elsewhere where accuracy is not as key - for example, adding a timestamp to the filename of persisted errors.

## Testing

Updated existing test coverage and manually tested to confirm that the values are correct.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
@@ -3,6 +3,8 @@ package com.bugsnag.android;
 import com.bugsnag.android.internal.DateUtils;
 import com.bugsnag.android.internal.ImmutableConfig;
 
+import android.os.SystemClock;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -320,11 +322,11 @@ DeliveryStatus deliverSessionPayload(Session payload) {
     }
 
     void onActivityStarted(String activityName) {
-        updateForegroundTracker(activityName, true, System.currentTimeMillis());
+        updateForegroundTracker(activityName, true, SystemClock.elapsedRealtime());
```
lemnik, line 325: This will also need a fix in `updateForegroundTracker` which tracks the session start as a `Date` using this value:
```java
startNewSession(new Date(nowMs), client.getUser(), true);
```

fractalwrench: Updated to use `new Date()` instead.

lemnik (approved):
LGTM

PR #1381, opened 2021-09-23
Remove flaky test case

fractalwrench:
## Goal

Removes a flaky test case that ensures a manually cancelled request leaves a network breadcrumb. This is effectively covered by the `socketTimeout` test case anyway because the call is cancelled by the OkHttpClient after a certain duration.


lemnik (approved):
LGTM Pending CI

PR #1382, opened 2021-09-23
Resolve warnings about field designator extension

fractalwrench:
## Goal

Resolves warnings in the NDK test code that have been introduced in the integration branch. Notably, the use of an old-style GNU field designator extension has been removed. The duplicate use of the `static` modifier and missing return types have also been addressed.


lemnik (approved):
LGTM

PR #1383, opened 2021-09-23
Make journal code internal and in separate package

fractalwrench:
## Goal

Moved the journalling code into a separate package named `com.bugsnag.android.internal.journal` where possible, as there are now a reasonable amount of classes. This also alters types to be `internal` where they do not need to be publicly visible.


lemnik (approved):
LGTM

PR #1384, opened 2021-09-23
Capture native thread states

lemnik:
## Goal
Include a list of application threads and their statuses in crashes produced by the NDK plugin. The data reported includes:
- id
- name
- state

for each thread running at the time the report is generated. This reporting can be enabled or disabled using the existing [`sendThreads`](https://docs.bugsnag.com/platforms/android/configuration-options/#sendthreads) option in `Configuration`.

## Design
Scan the `/proc/self/task` directory to retrieve the list of running tasks / threads using signal-safe functions. The status of these threads is read (where possible) into a pre-allocated buffer of `bsg_thread`s from [`/proc/self/task/{tid}/stat`](https://man7.org/linux/man-pages/man5/proc.5.html) and mapped to human-readable values where required.

## Testing
New unit tests and Mazerunner scenarios to test that the data is reported when expected, and not reported when the configuration forbids it. Manual end-to-end testing has also been done to ensure the data is ingested and rendered as expected.


examples/sdk-app-example/app/src/main/java/com/example/bugsnag/android/ExampleApplication.kt
```diff
@@ -26,7 +26,7 @@ class ExampleApplication : Application() {
         super.onCreate()
 
         val config = Configuration.load(this)
-        config.setUser("123456", "joebloggs@example.com", "Joe Bloggs")
+        config.setUser("123456", "joebloggs@example.com", "Joe Bzloggs")
```
fractalwrench, line 29: Just checking that this is intentional and not a stray key touch?

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -172,9 +170,7 @@ class NativeBridge : StateObserver {
                     arg.autoDetectNdkCrashes,
                     Build.VERSION.SDK_INT,
                     is32bit,
-                    makeSafe(arg.appVersion ?: ""),
-                    makeSafe(arg.buildUuid ?: ""),
-                    makeSafe(arg.releaseStage ?: "")
```
fractalwrench, line 177: Are these definitely not required? IIRC there is a ticket in the backlog that should address whether these should be populated by native events

fractalwrench (approved):

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
lemnik: They're not actually on the JNI signatures (in C) at the moment, so we definitely shouldn't be passing them down.

examples/sdk-app-example/app/src/main/java/com/example/bugsnag/android/ExampleApplication.kt
lemnik: that was definitely a stray key

fractalwrench (approved):

bugsnag-plugin-android-ndk/src/main/jni/utils/threads.c
```diff
@@ -0,0 +1,199 @@
...
+end:
+  // success if we hit the DONE marker
+  return state == PARSE_DONE;
+}
+
+/*
+ * Reads the /stat file fields we are looking for (TID, Name, State) into `dest`
+ * with a signal-safe approach.
+ */
+bool read_thread_state(bsg_thread *dest, const char *tid) {
```
kattrali, line 143: The helper functions in this file have fairly generic names, which could cause a conflict somewhere. They should be marked static so they aren't visible elsewhere.
```suggestion
static bool read_thread_state(bsg_thread *dest, const char *tid) {
```

kattrali (approved):
a few procedural comments but looks good!

PR #1385, opened 2021-09-24
Add StateEvent for when notifier info is updated

fractalwrench:
## Goal

Adds a `StateEvent` that is triggered when the `Notifier` information is changed (RN/Unity). This also serializes the `Notifier` object to the journal.

## Changeset

- Created `Client#setNotifier()` and removed `Client#getNotifier()`
- Added a `NotifierState` class which emits a `StateEvent` when the value changes
- Updated `BaseObservable` to have `emitObservableEvent` as an overridable function
- Created `BugsnagJournalEventMapper`, which loads the journal as a map and in future will generate an event

## Testing

Added unit test coverage and manually verified that the bugsnag journal is logged out in an example app.


lemnik (approved):
LGTM

PR #1388, opened 2021-09-28
Write extra data as journal entries

fractalwrench:
## Goal

This changeset ensures that journal entries are added whenever `metaData.app.networkAccess` and `metaData.app.memoryTrimLevel` change. Previously only the first value on initialization would be written to the journal, meaning any generated `Event` could be using an out-of-date value.

## Testing

Enhanced unit tests.


lemnik (approved):
LGTM

PR #1389, opened 2021-09-28
Alter react-native JSON serializers to use journal code

fractalwrench:
## Goal

Updates the JSON serializers in the React Native module to use the journal code instead, which reduces duplication. This also  fixes a couple of issues in the journal implementation that were caught by the RN unit tests:

- null values are no longer added to the map for `Stackframe`
- the cpuAbi is serialized as a list rather than an array
- the key for buildUuid now matches the one used in the RN json

## Testing

Relied on existing unit test coverage


lemnik (approved):
LGTM

PR #1391, opened 2021-09-28
Deserialize user from bugsnag journal

fractalwrench:
## Goal

Deserializes the `user` value from the Bugsnag journal entry. This has been achieved by altering the `User` class to [delegate its properties to a map](https://kotlinlang.org/docs/delegated-properties.html#storing-properties-in-a-map), and adding another constructor to `EventInternal` that provides default values.

## Changeset

- Added secondary constructor to `EventInternal` that sets default values
- Converted `User` to have its properties backed by a map. This will be rolled out to other classes on the `Event` interface where appropriate in future changesets
- Updated react native deserialization to use new constructor
- Loaded user from the bugsnag journal

## Testing

Added unit test coverage.


lemnik (approved):
LGTM

PR #1394, opened 2021-09-29
Deserialize metadata from bugsnag journal entries

fractalwrench:
## Goal

Deserializes metadata from any entries in the bugsnag journal and converts it into an `Event`.

## Testing

Added a unit test case.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/BugsnagJournalEventMapper.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/BugsnagJournalEventMapper.kt
@@ -43,6 +43,12 @@ internal class BugsnagJournalEventMapper(
         // populate user
         val userMap: Map<String, String?> = map.readJournalEntry("user")
         event.userImpl = User(userMap)
+
+        // populate metadata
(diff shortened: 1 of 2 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/BugsnagJournalEventMapper.kt
```diff
@@ -43,6 +43,12 @@ internal class BugsnagJournalEventMapper(
         // populate user
         val userMap: Map<String, String?> = map.readJournalEntry("user")
         event.userImpl = User(userMap)
+
+        // populate metadata
+        val metadataMap: Map<String, Map<String, Any?>> = map.readJournalEntry("metaData")
+        metadataMap.forEach { entry ->
+            event.addMetadata(entry.key, entry.value)
+        }
```
lemnik, line 51:
```suggestion
        val metadataMap: Map<String, Map<String, Any?>> = map.readJournalEntry("metaData")
        metadataMap.forEach { (key, value) ->
            event.addMetadata(key, value)
        }
```

lemnik (approved):
LGTM

PR #1395, opened 2021-09-29
Deserialize breadcrumbs from bugsnag journal entries

fractalwrench:
## Goal

Deserializes breadcrumbs from any entries in the bugsnag journal and adds them onto an `Event`. This converts `BreadcrumbInternal` to use a `Map` as a property delegate.

## Testing

Added unit tests to cover the functionality.


lemnik (approved):
LGTM

PR #1396, opened 2021-09-29
Deserialize context from bugsnag journal

fractalwrench:
## Goal

Deserializes the context from entries in the bugsnag journal into an event.

## Testing

Added unit tests to cover the case where context is present and when it isn't.


bugsnag-android-core/src/test/java/com/bugsnag/android/internal/journal/BugsnagJournalEventMapperTest.kt
```diff
@@ -40,6 +42,10 @@ class BugsnagJournalEventMapperTest {
                 "foo" to fooSection
             )
         )
+
+        journalMap = minimalJournalMap.plus(
+            mapOf("context" to "ExampleActivity")
+        )
```
lemnik, line 48: Would this not work here?

```suggestion
        journalMap = minimalJournalMap + ("context" to "ExampleActivity")
```

Leaning on:

```kotlin
operator fun <K, V> Map<out K, V>.plus(
    pair: Pair<K, V>
): Map<K, V>
```

lemnik (approved):
LGTM - With a trivial suggestion

PR #1398, opened 2021-09-30
Deserialize app from bugsnag journal entries

fractalwrench:
## Goal

Deserializes `AppWithState` to handle loading any journal entries added by Bugsnag. This converts the class so that its properties are backed by a Map delegate.

## Testing

Added unit test coverage.


bugsnag-android-core/src/main/java/com/bugsnag/android/App.kt
```diff
@@ -72,14 +97,10 @@ open class App internal constructor(
...
-        JournalKeys.keyId to id,
-        JournalKeys.keyReleaseStage to releaseStage,
-        JournalKeys.keyType to type,
-        JournalKeys.keyVersion to version,
-        JournalKeys.keyVersionCode to versionCode
-    )
+    override fun toJournalSection(): Map<String, Any?> {
+        val copy = map.toMutableMap()
+        copy["buildUUID"] = copy["buildUuid"]
+        copy.remove("buildUuid")
```
lemnik, line 103:
```suggestion
        copy["buildUUID"] = copy.remove("buildUuid")
```

lemnik (approved):
LGTM

PR #1400, opened 2021-09-30
Deserialize journal entries into device

fractalwrench:
## Goal

Deserializes the device from entries in the bugsnag journal into an event.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #1402, opened 2021-10-01
Deserialize session from bugsnag journal entry

fractalwrench:
## Goal

Deserializes the session written in the bugsnag journal if it is present, and adds it to the event.

## Testing

Added unit test coverage.


lemnik (approved):
LGTM

PR #1403, opened 2021-10-04
Deserialize projectPackages from bugsnag journal

fractalwrench:
## Goal

Deserializes projectPackages into an `Event` from entries stored in the journal.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #1456, opened 2021-10-25
Add complex metadata test

fractalwrench:
## Goal

Adds an E2E test which verifies that complex metadata values are serialized correctly. This has required some alterations/bug fixes to the JSON serializer.

## Changeset

Added an E2E test which has various metadata values. Notably the following have been included:

- A large string, to verify that there are no limits in the payload metadata
- A Date, as this was not being serialized correctly by dsl-json
- An unknown type, which has required updates to serialize a placeholder value rather than dsl-json throwing an exception
- Null values, which the `BugsnagJournalEventMapper` could not handle within a Metadata `Map` value
- A nested map

Unit test coverage has also been increased to cover this change.

### DIFF
--- a/features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXComplexMetadataScenario.kt
+++ b/features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXComplexMetadataScenario.kt
@@ -1,0 +1,55 @@
+package com.bugsnag.android.mazerunner.scenarios
+
+import android.content.Context
+import com.bugsnag.android.Bugsnag
+import com.bugsnag.android.Configuration
+import com.bugsnag.android.mazerunner.log
+import java.util.Date
+
+/**
+ * Sends a native crash to Bugsnag which has complicated metadata values.
+ */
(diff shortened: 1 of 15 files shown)


features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXComplexMetadataScenario.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    private external fun crash()
+
+    override fun startScenario() {
+        super.startScenario()
+        val longStr = "Have you tried turning it off and on again and off and on again and off" +
+            " and on again and off and on again and off and on again and off and on again" +
+            " and on again and off and on again and off and on again and off and on again?"
+        val map = mapOf(
+            "int_array" to intArrayOf(-5, 1000, 29),
```
lemnik, line 32:
```suggestion
            "int_array" to intArrayOf(-5, 1000, 29, Int.MAX_VALUE, Int.MIN_VALUE),
```

Assuming it doesn't do anything weird with Mazerunner?

lemnik (approved):
LGTM - with a suggestion of some more extreme values in the test

PR #1469, opened 2021-10-29
FeatureFlagState and FeatureFlags

lemnik:
## Goal
Implement `FeatureFlags` to encapsulate the containment of `FeatureFlag` objects, and `FeatureFlagState` to handle `StateEvent` notifications.

## Testing
New unit tests to assert the behaviour of `FeatureFlagState` and the serialization of `FeatureFlags`.


fractalwrench (approved):
LGTM

PR #1470, opened 2021-11-01
Fix duration and durationInForeground for native crashes

kstenerud:
## Goal

PLAT-7508

## Design

Use app launch, app "enter foreground" time, and event time to generate `duration` and `durationinForeground`. These timestamps are stored under a new "runtime" map under the top level of the journal:

```
  "runtime": {
      "timeLaunched": "2021-09-28T10:31:00.620Z",
      "timeEnteredForeground": "2021-09-28T10:31:05.620Z",
      "timeNow": "2021-09-28T10:32:00.620Z"
  }
```

The native layer now stores time values in the journal as hex-encoded strings based on unix time in milliseconds to avoid float conversion issues, and to remain async-safe.

## Testing

Updated tests to take into account the new information, and to check the duration calculations.


fractalwrench (approved):
LGTM

PR #1471, opened 2021-11-01
Remove unnecessary fields from NDK events

fractalwrench:
## Goal

The following fields are populated in unhandled NDK events when previously they weren't:

```
app.networkAccess
app.totalMemory
device.batteryLevel
device.charging
```

As these fields don't make sense in the context of an unhandled NDK error because they are not updated throughout the application's lifecycle, they should be omitted for now.


lemnik (approved):
LGTM

PR #1472, opened 2021-11-01
Move instrumentation tests to GitHub actions

twometresteve:
## Goal

TODO: Resolve failures in API levels 21/24.  It's not clear to me why these are failing, although I suspect they won't be hard to resolve.

Move the instrumentation tests to simulator-based GitHub Actions.

## Design

We've have various issues running Espresso tests on real devices and the time taken to deal with those has started to outweigh the benefits of running on real devices (given that we run e2e tests on them).

## Changeset

Remove Buildkite steps and add new GHA workflow.

## Testing

Covered by CI.


.buildkite/pipeline.yml
```diff
@@ -154,54 +154,6 @@ steps:
...
-      - docker-compose#v3.7.0:
-          run: android-ci
-    env:
-      SINGLE_RUNNER: "true"
-      INSTRUMENTATION_DEVICES: '["Google Pixel 2-8.0"]'
-      TEST_APK_LOCATION: 'bugsnag-plugin-android-ndk/build/outputs/apk/androidTest/debug/bugsnag-plugin-android-ndk-debug-androidTest.apk'
-    concurrency: 9
-    concurrency_group: 'browserstack-app'
-
-  - label: ':android: Performance benchmarks'
```
fractalwrench, line 189: This is my bad as I forgot to mention, but we'll probably want to keep running the performance benchmarks on BrowserStack in addition to an emulator, so that the results are realistic.

Although having said that, there's definitely an argument for running them on a full build only or omitting them entirely - as so far we've just tended to run the benchmarks ourselves on the conclusion of a big project.

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: instrumentation_tests
+
+on: [push, pull_request]
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
```
fractalwrench, line 6:
```suggestion
  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false -Dorg.gradle.parallel=true"
```

Making this a parallel build should speed things up at zero cost to us: https://docs.gradle.org/current/userguide/performance.html#parallel_execution

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: macos-latest
+
+    strategy:
+      fail-fast: true
+      matrix:
+        api-level:
+          - 21
+          - 24
+          - 26
+          - 29
```
fractalwrench, line 19: Let's just go with the latest and greatest API level - it'll speed up the feedback loop and the E2E tests already have sufficient coverage for us to know if something went wrong on a specific version.

fractalwrench (commented):
Looks nice. I think the API 21/24 failure is down to the benchmark tests not really being meant to run on an emulator - so we could either remove that step from CI or keep running it on browserstack.

Left a few other comments inline. The jobs also seem to be cancelled - have these passed originally?

.buildkite/pipeline.yml
twometresteve: I think I like the idea of just running benchmarking tests ourselves when we feel it's appropriate.  Having a human run and analyse performance test results allow us to be more subjective and explore areas in more detail if needed.  Having the tests run in CI could give a false sense of security whilst also using a limited resource.

fractalwrench: Sounds good to me 👍

.github/workflows/build.yml
twometresteve: Done - 29 appears to be the latest available with that action.

fractalwrench (approved):
LGTM

PR #1475, opened 2021-11-02
Use a fork of dsl-json

fractalwrench:
## Goal

Updates bugsnag-android to use a forked version of dsl-json. 

The primary concern against using the dsl-json supplied artefacts is that `DslJson` currently uses `java.awt.*` imports. This results in an Android lint warning that could break the build of end-users of bugsnag-android, necessitating this fork.

Forking the library will also allow us to customize the implementation as needed, and also remove code which isn't used by our project.

## Changeset

- Added dsl-json as a git submodule
- Added dsl-json submodule to the `bugsnag-android-core` sourcesets
- Updated existing code to use forked dsl-json package name
- Suppressed existing lint warnings in dsl-json with a baseline file

## Testing

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/JsonHelper.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/JsonHelper.kt
@@ -1,8 +1,8 @@
 package com.bugsnag.android.internal.journal
 
 import com.bugsnag.android.internal.DateUtils
-import com.dslplatform.json.DslJson
-import com.dslplatform.json.JsonWriter
+import com.bugsnag.dslplatform.json.DslJson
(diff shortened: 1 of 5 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/JsonHelper.kt
```diff
@@ -25,7 +25,12 @@ class JsonHelper private constructor() {
         private val dslJson = DslJson(settings)
 
         init {
-            dslJson.registerWriter(Date::class.java, ::serializeDate)
+            dslJson.registerWriter(Date::class.java) { writer: JsonWriter, value: Date? ->
```
fractalwrench, line 28: Inlining this prevents a lint warning about implicit SAM references.

lemnik (approved):
LGTM

PR #1476, opened 2021-11-02
Avoid ClassCastException in JSON serialization

fractalwrench:
## Goal

Avoids a class cast exception being thrown in a rare edge case by dsl JSON. If dsl-json is provided a list containing a primitive array and some other unrelated type, then it throws an exception.

This changeset updates bugsnag-android to use the fix in [this PR](https://github.com/bugsnag/dsl-json/pull/2). Additionally, it adds some unit test coverage to ensure that an exception is not thrown when serializing JSON with some unusual types.


lemnik (approved):
LGTM

PR #1482, opened 2021-11-04
[PLAT-7528] Remove requirement for concurrent containers in the journal

kstenerud:
## Goal

PLAT-7528

Concurrent containers were severely slowing performance for a feature we're not even using (concurrent journal access). We can deal with that issue if and when we actually need it.

## Testing

Manual performance testing to verify the improvement.


kstenerud:
Current implementation on next:

Screenshot 2021-11-03 at 14 08 34

New implemnentation:
Screenshot 2021-11-05 at 10 02 58

fractalwrench (approved):
LGTM pending CI

PR #1484, opened 2021-11-04
Increase resilience of NDK stackframe method capture

fractalwrench:
## Goal

Increases the resilience of the code used to instantiate a `StackTraceElement` object when calling `bugsnag_notify_env`. There were a couple of potential issues with the existing implementation:

- The return values of `filename` and `method` weren't checked for `NULL`, which could lead to exceptions being thrown in `bsg_safe_new_object`
- The return value of `calloc` was not checked for `NULL`, which could lead to a malformed string if `bsg_safe_new_string_utf` failed, resulting in a [SIGABRT](https://cs.android.com/android/platform/superproject/+/master:art/runtime/jni/check_jni.cc;l=1013?q=CheckNonHeapValue)

## Testing

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -90,9 +90,8 @@ void bsg_populate_notify_stacktrace(JNIEnv *env, bugsnag_stackframe *stacktrace,
     // populate method
     jstring method = NULL;
     if (strlen(frame.method) == 0) {
-      size_t frame_len = sizeof(char) * 32;
       char frame_address[32];
-      snprintf(frame_address, frame_len, "0x%lx",
+      snprintf(frame_address, sizeof(frame_address), "0x%lx",
                (unsigned long)frame.frame_address);
       method = bsg_safe_new_string_utf(env, frame_address);
     } else {
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -81,15 +81,25 @@ void bsg_populate_notify_stacktrace(JNIEnv *env, bugsnag_stackframe *stacktrace,
       goto exit;
     }
 
+    // populate filename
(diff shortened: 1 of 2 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -81,16 +81,30 @@ void bsg_populate_notify_stacktrace(JNIEnv *env, bugsnag_stackframe *stacktrace,
...
+
+    // populate method
+    jstring method = NULL;
     if (strlen(frame.method) == 0) {
-      char *frame_address = calloc(1, sizeof(char) * 32);
-      sprintf(frame_address, "0x%lx", (unsigned long)frame.frame_address);
-      method = bsg_safe_new_string_utf(env, frame_address);
-      free(frame_address);
+      size_t frame_len = sizeof(char) * 32;
+      char *frame_address = calloc(1, frame_len);
```
kstenerud, line 94: Since this is only 32 bytes, why not just allocate it on the stack?

fractalwrench: Good idea - I've tweaked the code to do this.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -81,16 +81,26 @@ void bsg_populate_notify_stacktrace(JNIEnv *env, bugsnag_stackframe *stacktrace,
...
+
+    // populate method
+    jstring method = NULL;
     if (strlen(frame.method) == 0) {
-      char *frame_address = calloc(1, sizeof(char) * 32);
-      sprintf(frame_address, "0x%lx", (unsigned long)frame.frame_address);
+      size_t frame_len = sizeof(char) * 32;
+      char frame_address[32];
+      snprintf(frame_address, frame_len, "0x%lx",
+               (unsigned long)frame.frame_address);
```
kstenerud, line 96:
```suggestion
      char frame_address[32];
      snprintf(frame_address, sizeof(frame_address), "0x%lx",
               (unsigned long)frame.frame_address);
```

kstenerud (approved):

PR #1485, opened 2021-11-05
Pipeline improvements

twometresteve:
## Goal

Makes a few changes to simplify the repo file structure and reduce the amount of noise on a full CI build.

## Changeset

- Flatten the two batches of files under `features/full_tests`, using the Cumber `--exclude` option in the pipeline to separate them instead.
- Android 9 tests now duplicated in quick and full builds
- `features/smoke_tests` included in full build steps, avoiding the need to include the quick pipeline (reducing the number of steps to look at and removing a number of Appium session start times from the equation).

## Testing

Covered by separate "quick" and "full" CI runs, performed before raising this PR.


.buildkite/pipeline.quick.yml
```diff
@@ -85,9 +85,9 @@ steps:
     concurrency_group: 'browserstack-app'
     concurrency_method: eager
 
-  - label: ':android: Android 9 NDK r21 smoke tests'
-    key: 'android-9-smoke'
-    depends_on: "fixture-r21"
+  - label: ':android: Android 9 NDK r21 end-to-end tests - batch 1'
```
fractalwrench, line 88: This can go in another PR/ticket as it's a future optimization, but is it worth discussing what device the quick step should run on?

IIRC Android 9 was picked because it was most popular at the time, but Android 11 is the most commonly used device right now - so perhaps it'd make sense to run by default on Android 11 (or 12). There's probably also a case for looking at Appium performance over a couple of CI runs, and leaning towards the fastest option.

.buildkite/pipeline.quick.yml
```diff
@@ -140,10 +143,13 @@ steps:
...
+          - "features/smoke_tests"
+          - "features/full_tests"
+          - "--exclude=features/full_tests/[^o-z].*.feature"
           - "--app=/app/build/release/fixture-r21.apk"
           - "--farm=bs"
           - "--device=ANDROID_9_0"
+          - "--fail-fast"
     concurrency: 9
     concurrency_group: 'browserstack-app'
     concurrency_method: eager
```
fractalwrench, line 155: This could also be addressed in a separate changeset/later, but it's worth mentioning that commits targeting an integration branch only run the steps defined in `pipeline.yml`. That means it's only possible to run the smoke tests.

I think it'd be really useful to have a gated step that runs all the scenarios like is done here. That would prevent devs from adding `[full ci]` and causing CI congestion.

fractalwrench (approved):
LGTM - I added a couple of comments inline but they're more future suggestions.

One other query - it looks like the CI run doesn't perform the full E2E tests on Android 4. I'm assuming that's intentional and there is some technical blocker preventing us from running tests on that API level.

twometresteve:
> LGTM - I added a couple of comments inline but they're more future suggestions.
> 
> One other query - it looks like the CI run doesn't perform the full E2E tests on Android 4. I'm assuming that's intentional and there is some technical blocker preventing us from running tests on that API level.

Yes, the Android 4 device we were using on BrowserStack was extremely slow and unreliable.  You may remember we agreed to remove it and do manual testing as and when we deemed it necessary.  This could be another good reason for getting the e2e tests running properly on local devices and I have raised internal ticket PLAT-7576 specifically for Android 4.

.buildkite/pipeline.quick.yml
twometresteve: That should be easy enough to achieve by moving a line in `pipeline_trigger.sh`:
```
elif [[ "$BUILDKITE_MESSAGE" == *"[gated-full ci]"* ||
  "$BUILDKITE_BRANCH" == "next" ]]; then
  echo "Running pre-release build"
  buildkite-agent pipeline upload .buildkite/pipeline.quick.yml
  buildkite-agent pipeline upload .buildkite/block.full.yml
elif [[ "$BUILDKITE_MESSAGE" == *"[quick ci]"* ||
  "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" == "next" ||
  "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" == "integration/"* ]]; then
  echo "Running quick build"
  buildkite-agent pipeline upload .buildkite/block.quick.yml
else
```

Moving the line for `"integration/"*` into the `gated-full` section _should_ be ok, I think.  But you wouldn't want a full build due to the impact on draft PRs for integration branches (a fresh build will occur after every merge, swamping our resources).

.buildkite/pipeline.quick.yml
twometresteve: No objection to moving on from 9, I just don't think it's been explicitly considered until now.  Theissue we might have is availability when 11/12 are [tier 3/4 devices](https://www.browserstack.com/device-tiers) when Android 9 is tier 1.  So although the devices might be quicker, we could spend longer waiting for them (and potentially timing out).

.buildkite/pipeline.quick.yml
fractalwrench: Ah - I hadn't realized that adding `[quick ci]` to a commit message was an available option. That seems like it'll achieve what is necessary in the short-term

.buildkite/pipeline.quick.yml
fractalwrench: Got it. Sticking with tier 1 sounds the best option for reliability for now in that case 👍

PR #1487, opened 2021-11-05
[Plat-7497] int encoding

kstenerud:
## Goal

PLAT-7497

Encode all integer values as hex encoded strings in the native layer because we can't guarantee no rounding errors with floating point from an async-safe environment.

## Testing

Re-ran all unit tests, verified that the web UI displays as expected.


tomlongridge:
Closing as we're pausing journalling work for now.

PR #1489, opened 2021-11-05
Avoid formatting date to improve addBreadcrumb performance

fractalwrench:
## Goal

Delays the date formatting for breadcrumbs to the point where the event JSON payload is generated, rather than at the point of breadcrumb collection. The journal now stores the breadcrumb timestamp as a `Long`.

Other timestamps are still stored as strings - these can be altered at a future date as they don't impact performance as much as breadcrumb timestamps.

## Testing

When adding 100 breadcrumbs, the previous implementation spent ~100ms on `toIso8601()` when profiled with a full Java method trace:
baseline

Avoiding date formatting has improved performance by ~13%:

changes


lemnik (approved):
LGTM

PR #1490, opened 2021-11-05
Temporarily skip failing e2e test steps

twometresteve:
## Goal

Comment out checks on failing fields pending further investigation, as well as some other CI fixes and improvements.

I've also corrected the inclusion of scenarios for a couple of build steps and made a change to `pipeline_trigger.sh` so that future PRs into integration branches behave like PRs into `next` (which, in particular, means a complete set of e2e tests will get run).

TODO: An instrumentation test failure still needs to be resolved/removed.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -357,7 +357,6 @@ steps:
         pull: android-maze-runner
         run: android-maze-runner
         command:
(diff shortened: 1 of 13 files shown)


fractalwrench (approved):
LGTM pending resolution of the failing NDK instrumentation tests

PR #1492, opened 2021-11-08
Allow onSend callbacks to run

fractalwrench:
## Goal

Fixes some bugs in how JVM exceptions were deserialized from disk when running an `OnSend` callback. This is down to `BugsnagJournalEventMapper` converting both an `Event` JSON and journal entries into an `Event`.

## Changeset

- Removed unnecessarily verbose logging around the `BugsnagJournal` which may have confused end-users
- Only loaded the duration from the 'runtime' journal entry if it is present, rather than throwing an exception
- Handling a long/string timestamp for breadcrumbs

This has been tested by uncommenting the E2E scenario and existing unit tests.


lemnik (approved):
LGTM

PR #1493, opened 2021-11-09
[PLAT-7581] Use mutable maps for anything that can be put in the journal

kstenerud:
## Goal

The journal requires mutable containers and auto-converts anything non-conformant. Make sure anything we pass in is already mutable.

## Testing

Re-ran all automated tests.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ClientInternal.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ClientInternal.kt
@@ -316,7 +316,7 @@ internal class ClientInternal constructor(
     private fun onConnectivityChange(hasConnection: Boolean, networkState: String?) {
         leaveAutoBreadcrumb(
             "Connectivity changed", BreadcrumbType.STATE,
-            mapOf(
+            mutableMapOf(
                 "hasConnection" to hasConnection,
                 "networkState" to networkState
             )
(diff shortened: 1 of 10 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ClientInternal.kt
```diff
@@ -316,7 +316,7 @@ internal class ClientInternal constructor(
     private fun onConnectivityChange(hasConnection: Boolean, networkState: String?) {
         leaveAutoBreadcrumb(
             "Connectivity changed", BreadcrumbType.STATE,
-            mapOf(
+            mutableMapOf(
```
fractalwrench, line 319: I can't recall what the conclusion of the discussion we had with @lemnik earlier was - but I think that using `HashMap` would be marginally less of a footprint than `mutableMapOf` due to its use of `LinkedHashMap` behind the scenes.

fractalwrench (approved):
LGTM - pending resolution of whether we should use `mutableMapOf` or `hashMapOf`

bugsnag-android-core/src/main/java/com/bugsnag/android/ClientInternal.kt
kstenerud: Yes, I wanted to start out with just mutableMapOf and then we could work from that baseline to determine how much (if any) HashMap would gain us. That way if the gains are negligible, we can remain idiomatic.

PR #1494, opened 2021-11-09
[Plat-7581] Remove mutable map conversions from journal

kstenerud:
## Goal

Since we are able to control what goes into the journal, a defensive mutable copy is not actually required, and not converting could offer speedups.

## Testing

Re-ran all automated tests.


fractalwrench (approved):
LGTM

PR #1495, opened 2021-11-09
Optimize Command.serialize()

fractalwrench:
## Goal

Optimizes the `serialize()` command to avoid the unnecessary instantiation of a `Pair`, and switches to using the cheaper `HashMap` object.

The rest of the method is dominated by dsl-json which doesn't have any low-hanging fruit for optimization.

## Testing

Verified against a baseline and confirmed that the time taken to instantiate and populate maps reduced from ~55ms to ~25ms.

command_serialize_baseline

command_serialize_improvement

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/Journal.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/Journal.kt
@@ -97,7 +97,7 @@ class Journal(val type: String, val version: Int) {
          */
         @Throws(IOException::class)
         fun serialize(out: OutputStream) {
-            JsonHelper.serialize(mutableMapOf(path to value), out)
+            JsonHelper.serialize(mapOf(path to value), out)
             out.write(0)
         }
 


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/Journal.kt
```diff
@@ -97,7 +97,10 @@ class Journal(val type: String, val version: Int) {
          */
         @Throws(IOException::class)
         fun serialize(out: OutputStream) {
-            JsonHelper.serialize(mutableMapOf(path to value), out)
+            // optimization - avoid unnecessary instantiation of Pair() on a hot path
```
kstenerud, line 100: I was actually touching the same line for a different PR. Can you try comparing it with just changing to `JsonHelper.serialize(mapOf(path to value), out)`? Since it's just being serialized, it doesn't need to be mutable. When I did this, it pretty much disappeared from the flame graph in my tests.

fractalwrench: `mapOf(path to value)` is marginally faster - I'll update the PR to use that.

Screenshot 2021-11-09 at 15 13 57

kstenerud (approved):

PR #1496, opened 2021-11-09
Memoize document path directives

fractalwrench:
## Goal

Creating a `Journal.Command` object initializes a list of `DocumentPathDirective`. This was relatively expensive to calculate and memoization reduces the amount of time required by an order of magnitude. 

## Testing

Verified that the amount of time dedicated to creating path directives when setting the context drops by an order of magnitude, from ~200ms to ~22ms:

command_init_baseline
command_init_changes

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/DocumentPath.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/DocumentPath.kt
@@ -146,6 +146,10 @@ class DocumentPath(path: String) {
         private const val PATH_SEPARATOR = '.'
         private const val ADD_OPERATOR = '+'
 
+        // optimization: memoize the results for the path directive
+        @JvmField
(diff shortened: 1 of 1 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/journal/DocumentPath.kt
```diff
@@ -184,6 +188,15 @@ class DocumentPath(path: String) {
          * The path cannot have empty components (e.g. "a..b" is invalid).
          */
         internal fun toPathDirectives(path: String): List<DocumentPathDirective<Any>> {
+            var result = memo[path]
+            if (result == null) {
+                result = toPathDirectivesImpl(path)
+                memo[path] = result
+            }
+            return result
```
lemnik, line 196:
```suggestion
            return memory.getOrPut(path) { toPathDirectivesImpl(path) }
```
?

PR #1497, opened 2021-11-09
[Plat-7528] improve performance 3

kstenerud:
## Design

Allocate a 10-entry array list for document path directives to avoid the overhead of adding to a linked list.


fractalwrench (approved):
LGTM once the style violations are addressed.

It'd also be good to document why 10 was chosen. Is that the optimal number?

lemnik (approved):
LGTM

PR #1498, opened 2021-11-09
Plat 7528 remove breadcrumbs legacy path

kstenerud:
## Goal

Breadcrumbs now use the journal, so we don't need to run the legacy path anymore.


fractalwrench (approved):
LGTM pending CI

PR #1499, opened 2021-11-10
feat(feature flags): added FeatureFlags to `Event`

lemnik:
## Goal
Expose `FeatureFlags` on the `Event` object so that they can be altered by crash-time callbacks before being uploaded. The `FeatureFlags` are copied from the client state when the `Event` is created, and can then be altered on a per-event basis before being sent.

## Testing
A new unit test covers the `FeatureFlags` deep-copy behaviour and per-event changes in the same way as we test per-event `Metadata`


fractalwrench (approved):
LGTM

PR #1500, opened 2021-11-10
[Plat-7528] improve journal performance

kstenerud:
## Goal

`unspecialMapPath` is a guard against special characters in a path component, but this is the exception, not the norm. Optimize for the happy path.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
@@ -194,9 +194,19 @@ internal class BugsnagJournal @JvmOverloads internal constructor(
          * - All occurrences of special characters (+ . \)
          */
         fun unspecialMapPath(pathComponent: String): String {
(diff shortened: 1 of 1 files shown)


fractalwrench (approved):
LGTM pending CI

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagJournal.kt
```diff
@@ -194,9 +194,24 @@ internal class BugsnagJournal @JvmOverloads internal constructor(
...
+                val index = pathComponent.toIntOrNull()
+                if (index != null) {
+                    return "\\" + pathComponent
+                }
+            }
+
+            if (!(
+                pathComponent.contains('.') ||
+                    pathComponent.contains('+') ||
+                    pathComponent.contains('\\')
```
lemnik, line 211:
```suggestion
                pathComponent.indexOfFirst { it == '.' || it == '+' || it == '\\' } != -1
```
then we only have a single string scan?

lemnik (approved):

PR #1503, opened 2021-11-10
[PLAT-7584] Remove unneeded NDK calls

kstenerud:
## Goal

Now that the journal is connected to everything, many of the NDK calls we made before are no longer needed, and just slow things down.


tomlongridge:
Closing as we're pausing journalling work for now.

PR #1504, opened 2021-11-10
test(feature flags): mazerunner feature to test feature flags

lemnik:
## Goal
Mazerunner test feature covering the new feature flags logic. This covers feature flags added globally (using `Bugsnag.addFeatureFlag`) and directly to an `Event` via an `OnErrorCallback`.

Tests include checking for:
* A feature flag with a name only
* A feature flags with names and variants
* Adding a feature flag with a null name (should be ignored and not sent)

### DIFF
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/FeatureFlagScenario.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/FeatureFlagScenario.kt
@@ -1,0 +1,47 @@
+package com.bugsnag.android.mazerunner.scenarios
+
+import android.content.Context
+import com.bugsnag.android.Bugsnag
+import com.bugsnag.android.Configuration
+import com.bugsnag.android.FeatureFlag
+
+class FeatureFlagScenario(
+    config: Configuration,
+    context: Context,
+    eventMetadata: String
+) : Scenario(config, context, eventMetadata) {
+    override fun startScenario() {
+        super.startScenario()
+
+        Bugsnag.addFeatureFlag("demo_mode")
+
+        Bugsnag.addFeatureFlags(
+            listOf(
+                FeatureFlag("should_not_be_reported_1"),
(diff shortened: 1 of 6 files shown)


features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/FeatureFlagScenario.kt
```diff
@@ -0,0 +1,31 @@
+package com.bugsnag.android.mazerunner.scenarios
+
+import android.content.Context
+import com.bugsnag.android.Bugsnag
+import com.bugsnag.android.Configuration
+
+class FeatureFlagScenario(
```
tomlongridge, line 7: This scenario doesn't cover adding multiple flags or clearing them - shouldn't it cover the whole API surface (hopefully within the one scenario)?

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/FeatureFlagScenario.kt
```diff
@@ -0,0 +1,31 @@
...
+
+        if (eventMetadata?.contains("callback") == true) {
+            Bugsnag.addOnError { error ->
+                error.addFeatureFlag("sample_group", "a")
+                return@addOnError true
+            }
+        }
+
+        if (eventMetadata?.contains("unhandled") == true) {
+            throw RuntimeException("FeatureFlagScenario unhandled")
```
fractalwrench, line 26: It'd be good to add a test to ensure that the functionality works for unhandled NDK errors too.

lemnik: The native functionality was added in its own scenario (batch_1 / batch_2). I'll rebase and merge them together.

fractalwrench (approved):
LGTM

PR #1505, opened 2021-11-11
Plat 7599/onsend port

lemnik:
## Goal
Rework the `OnSend` implementation and `dsl-json` submodule to target `next` instead of the journal integration. Still missing from this PR is any ability to parse the stored Events into the existing model.

## Testing
This merge is not expected to pass tests and contains incomplete code.


lemnik:
The remaining Lint error on this build is expected. The `TODO` it references is where we need to parse the event JSON files back into an `Event` object

bugsnag-android-core/src/main/java/com/bugsnag/android/CallbackState.kt
```diff
@@ -1,11 +1,12 @@
...
+import java.util.concurrent.CopyOnWriteArrayList
 
 internal data class CallbackState(
-    val onErrorTasks: MutableCollection<OnErrorCallback> = ConcurrentLinkedQueue<OnErrorCallback>(),
-    val onBreadcrumbTasks: MutableCollection<OnBreadcrumbCallback> = ConcurrentLinkedQueue<OnBreadcrumbCallback>(),
-    val onSessionTasks: MutableCollection<OnSessionCallback> = ConcurrentLinkedQueue()
+    val onErrorTasks: MutableCollection<OnErrorCallback> = CopyOnWriteArrayList(),
+    val onBreadcrumbTasks: MutableCollection<OnBreadcrumbCallback> = CopyOnWriteArrayList(),
+    val onSessionTasks: MutableCollection<OnSessionCallback> = CopyOnWriteArrayList(),
+    val onSendTasks: MutableCollection<OnSendCallback> = CopyOnWriteArrayList()
```
kattrali, line 9: why is this better as CoW lists?

lemnik: CoW lists will typically have slightly better space / time performance for this use-case (haven't directly benchmarked for our use case). Specifically: we tend to add / remove from the list very few times, but iterate a very large numbers of times.

Happy to change it back if that's desirable.

kattrali (approved):

PR #1507, opened 2021-11-12
Fix journal increment in the crashtime journal

kstenerud:
## Goal

The crashtime journal needs to support increment operations for an upcoming PR.

## Testing

Added new unit and integration tests.


lemnik (approved):
LGTM

PR #1510, opened 2021-11-15
Parse Stacktrace for OnSend Callbacks

lemnik:
## Goal
Parse the stackframes and stacktraces from stored Events back into the Event model so that they can be processed by registered `OnSendCallback`s.

## Testing
Updated `StackframeSerializationTest` to also test deserialization using the existing fixtures.


bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -89,6 +89,31 @@ class Stackframe : JsonStream.Streamable {
...
+            method = json["method"] as? String
+            file = json["file"] as? String
+            lineNumber = json["lineNumber"] as? Number
+            inProject = json["inProject"] as? Boolean
+            columnNumber = json["columnNumber"] as? Number
+            @Suppress("UNCHECKED_CAST")
+            code = json["code"] as? Map<String, String?>
+            type = (json["type"] as? String)?.let { ErrorType.fromDescriptor(it) }
+        }
+    }
```
kattrali, line 115: Presumably, up until now, no callbacks ever exposed the `NativeStackframe` structure, so no one had to consider the implications of having it be a private and inaccessible part of the Stackframe class. Does this implementation imply no one would be able to introspect or modify the stacktrace for native frames? If so, it might be easier to just make `Stackframe` and `NativeStackframe` into a single class.

lemnik: 🤔  That is definitely an issue worth considering, although given that we only have a `Stackframe` or `NativeStackframe` it's also worth considering lifting them onto the same level under a `sealed class`

kattrali: As far as I could tell, the `NativeStackframe` class is only used in a single place, via JNI, and adds a few optional fields. If they are all in a single `Stackframe` class, then the existing use can be easily adapted and then all fields are available in callbacks.

kattrali (approved):

PR #1514, opened 2021-11-16
Parse DeviceWithState and AppWithState for OnSendCallbacks

lemnik:
## Goal
Parse the `AppWithState` and `DeviceWithState` data from stored events into our in-memory model so that they can be processed by registered `OnSendCallback`s.

## Testing
Added deserialization tests to `AppDataCollectorSerializationTest` and `DeviceDataCollectorSerializationTest` using the existing test fixtures.


fractalwrench (approved):

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataCollectorSerializationTest.kt
```diff
@@ -31,7 +31,7 @@ internal class AppDataCollectorSerializationTest {
             val memoryTrimState = MemoryTrimState()
 
             // populate summary fields
-            config.appType = "React Native"
+            config.appType = "ReactNative"
```
kattrali, line 34: Is this related?

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataCollectorSerializationTest.kt
```diff
@@ -74,8 +74,17 @@ internal class AppDataCollectorSerializationTest {
...
     @Test
     fun testJsonSerialisation() {
         verifyJsonMatches(testCase.first, testCase.second)
     }
+
+    @Test
+    fun testJsonDeserialization() {
+        verifyJsonParser(testCase.first, testCase.second) {
+            eventMapper.convertAppWithState(it)
+        }
```
kattrali, line 88: What does this test actually do? I'm sure I'm missing context here from various helpers, and its nice that its concise, but it seems unclear what the `first` and `second` and `it` are to determine if this covers everything in the changeset.

lemnik: The helper works by

1. parsing the json fixture with the `convertAppWithState` function
2. turning the in-memory model back into json using the existing `toStream(writer: JsonStream)` function
3. comparing the output of the in-memory model with the test fixture (they should be equal)

It could definitely be a bit clearer. I'll add some named parameters to help.

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataCollectorSerializationTest.kt
lemnik: This will be fixed once the integration branch is based on the latest from `next`. This is our "all whitespace in json is stripped" friend.

bugsnag-android-core/src/test/java/com/bugsnag/android/AppDataCollectorSerializationTest.kt
kattrali: Ah, thanks for that, makes sense!

kattrali (approved):

PR #1517, opened 2021-11-16
Parse Event for OnSend Callbacks

lemnik:
## Goal
Parse `Event` objects from stored events, so that they can be processed by registered `OnSendCallbacks`.

## Testing
Added a deserialization test to EventSerializationTest


bugsnag-android-core/src/main/java/com/bugsnag/android/MarshalledEventSource.kt
```diff
@@ -29,6 +33,11 @@ internal class MarshalledEventSource(private val eventFile: File) : () -> Event
 
     @Suppress("StopShip")
```
kattrali, line 34: This can go now! :smile:

kattrali (approved):
looks good

PR #1523, opened 2021-11-24
Mazerunner test for the OnSend callback

lemnik:
## Goal
Add Mazerunner coverage for the OnSend Callback to target next.

## Testing
Copied the OnSend Mazerunner scenario from the Journal branch.


fractalwrench (approved):
LGTM

PR #1526, opened 2021-11-24
`redactedKeys` now correctly apply to metadata on Breadcrumbs

lemnik:
## Goal
Configured `redactedKeys` should be applied to metadata within Breadcrumbs, and not just metadata associated with Events.

## Design
Allowed the `JsonStream` class to wrap another `JsonStream` so that the `redactedKeys` can be enforced in `Event.toStream`. This naturally enforces the `redactedKeys` for the entire sub-tree.

## Testing
Adjusted existing unit tests.

### DIFF
--- a/bugsnag-android-core/src/test/java/com/bugsnag/android/EventRedactionTest.kt
+++ b/bugsnag-android-core/src/test/java/com/bugsnag/android/EventRedactionTest.kt
@@ -13,17 +13,35 @@ internal class EventRedactionTest {
     fun testEventRedaction() {
         val event = Event(
             null,
+            generateImmutableConfig()
+                .copy(redactedKeys = setOf("password", "changeme")),
+            SeverityReason.newInstance(SeverityReason.REASON_HANDLED_EXCEPTION),
+            NoopLogger
+        )
+
+        testEventRedaction(event, "event_redaction.json")
+    }
+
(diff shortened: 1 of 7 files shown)


bugsnag-android-core/src/test/java/com/bugsnag/android/EventRedactionTest.kt
```diff
@@ -13,7 +13,8 @@ internal class EventRedactionTest {
     fun testEventRedaction() {
         val event = Event(
             null,
-            generateImmutableConfig(),
+            generateImmutableConfig()
+                .copy(redactedKeys = setOf("password", "changeme")),
```
fractalwrench, line 17: Is there still a test case in the codebase which asserts that the default value of `redactedKeys` equals "password"? If not, then it'd probably be worth adding one.

fractalwrench (approved):
LGTM

bugsnag-android-core/src/test/java/com/bugsnag/android/EventRedactionTest.kt
lemnik: There are in other tests, but they're testing `JsonStream` and such. I'll add another test to `EventRedactionTest` for the default "password" key.

PR #1527, opened 2021-11-24
Small fixes for the onSend implementation

kattrali:
## Goal

Resolve a few issues encountered during testing, can be split up into different changesets if needed.

## Changeset

* Make the API publicly accessible. While there are many things that can be done via JNI, dynamically constructing a class and bytecode for a method implementation for `onSend()` is not one that dalvik makes particularly ... easy. :sweat_smile: 
* Run callbacks on handled events. The existing implementation only worked for unhandled events (though including C/C++ crashes). At some point it would be advisable to revisit the delivery code paths because persisted and immediately sent events differ so wildly that this logic is necessary twice.
* Support altering API key in unhandled events - the key was passed into the event marshaller and not re-evaluated after the callbacks were run, so any alterations were ignored.
* Remove the Client add/remove functions - we don't need them internally, and they may be confusing since events begin being sent during initialization. There are no equivalent to these on similar platforms either. I realized while trying to devise a test where there's a crash but a callback was added after the fact that maybe these functions shouldn't exist.

## Testing

* manually tested on handled and unhandled events, including C/C++ crashes
* added new scenario for handled events
* added assertions for altering API key in callbacks


lemnik (approved):
LGTM

fractalwrench (changes requested):
This is the right way to make these changes - but I'm not convinced that there's been enough discussion around the API changes (although I may have missed it in one of the many slack channels).

The initial plan was to release `onSend` as a non-public API on Android so that we could gain feedback about its use on Unreal/Unity and then decide whether to make it public. Making this public now means we loses the ability to alter the interface - so probably needs flagging up to product and others.

If we end up removing `onSend` from `Client` then it'd also be necessary to remove it from `Bugsnag` as well. I understand the intent but again think it'd be worth discussing with the product team. There are other callbacks such as `OnBreadcrumb` where adding to the client is discouraged in favour of `Configuration`, but they remain included as an escape hatch for users who can't/won't edit their `Configuration` object.

kattrali:
> If we end up removing onSend from Client then it'd also be necessary to remove it from Bugsnag as well.

There's no onSend on Bugsnag to begin with

gareththackeray (approved):

fractalwrench (approved):
LGTM

PR #1528, opened 2021-11-25
Leave error breadcrumbs before attempting event delivery

lemnik:
## Goal
Ensure the at Error Breadcrumbs are created and stored as soon as an `Event` is populated, rather than after the `Event` has been delivered.

## Changeset
Moved the `leaveErrorBreadcrumb` from `DeliveryDelegate` into the `Client` so that it can be called directly after the `Event` has been populated.

## Testing
Added new unit tests for the `Client` facade, and a new Mazerunner scenario that create an handled error directly followed by an unhandled exception. The unhandled exception should include an Error breadcrumb with details of the handled error.


fractalwrench (approved):
LGTM pending CI + a rebase

kattrali (approved):

PR #1529, opened 2021-11-25
Update onsend-next to next and align tests

lemnik:
## Goal
Update `onsend-next` to `next` and align the serialization tests with the new JSON comparison which retains quoted whitespace when comparing.


fractalwrench (approved):
LGTM

PR #1536, opened 2021-11-29
Bump compileSdkVersion to apiLevel 31

fractalwrench:
## Goal

Bumps the [compileSdkVersion](https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd) to apiLevel 31 from 30. This ensures that our SDK is compiled against the latest version of Android while retaining compatibility for previously supported versions.

## Changeset

- Bumped compileSdkVersion to 31
- Bumped targetSdkVersion where possible (bugsnag-benchmarks is the only place still targetting 30, as the microbenchmark library does not support this yet)
- Avoided closing the system dialogs in the mazerunner fixture as this is [no longer allowed in Android 12](https://developer.android.com/about/versions/12/behavior-changes-all#close-system-dialogs)

## Testing

Relied on existing test coverage and ran the example app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+* Bump compileSdkVersion to apiLevel 31
+  [#1536](https://github.com/bugsnag/bugsnag-android/pull/1536)
+
 ## 5.16.0 (2021-11-29)
 
-## Bug fixes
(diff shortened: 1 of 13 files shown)


lemnik (approved):
LGTM - Pending a Rebase. The CHANGELOG entry might need to be moved post v5.16.0

PR #1538, opened 2021-12-01
Flush in memory session first

fractalwrench:
## Goal

The in-memory session should be sent first in precedence over sessions which are already persisted. This will reduce the chance of losing the session completely if a signal is raised as it reduces the time spent waiting for sessions to be sent.

## Testing


lemnik (approved):
LGTM

PR #1539, opened 2021-12-01
[PLAT-7680] Use signal stack unwinding mechanism when unwinding an ANR

kstenerud:
## Goal

The ANR plugin was being initialized with a stack unwinder function that used `bsg_global_env->unwind_style` instead of the `bsg_global_env->signal_unwind_style` that signal handlers require (for safety in an async-safe environment).

The ANR handler is triggered from a signal handler, so this PR updates it to use the proper signal-safe unwind style.

## Testing

Reran all tests and also tested manually.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -43,6 +43,17 @@ bsg_unwinder bsg_configured_unwind_style() {
   return BSG_CUSTOM_UNWIND;
 }
 
+/**
+ * Get the configured unwind style for async-safe environments such as signal
+ * handlers.
+ */
+bsg_unwinder bsg_configured_signal_unwind_style() {
+  if (bsg_global_env != NULL)
+    return bsg_global_env->signal_unwind_style;
+
+  return BSG_CUSTOM_UNWIND;
+}
+
 void bugsnag_add_on_error(bsg_on_error on_error) {
   if (bsg_global_env != NULL) {
     bsg_global_env->on_error = on_error;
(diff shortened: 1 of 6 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -698,14 +705,20 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_updateMetadata(
   bsg_release_env_write_lock();
 }
 
-ssize_t
-bsg_unwind_stack_default(bugsnag_stackframe stacktrace[BUGSNAG_FRAMES_MAX],
-                         siginfo_t *info, void *user_context) __asyncsafe;
+// Unwind the stack using the default unwind style.
```
tomlongridge, line 708: Presumably this comment needs updating?

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -43,6 +43,13 @@ bsg_unwinder bsg_configured_unwind_style() {
   return BSG_CUSTOM_UNWIND;
 }
 
+bsg_unwinder bsg_configured_signal_unwind_style() {
```
tomlongridge, line 46: Some comments on this for future readers would be appreciated.

fractalwrench (commented):
This change seems reasonable but it'd be good to understand more about how it's been manually tested. Has that involved checking the NDK stackframes in the ANR? Will there be any difference in the stackframes that are currently reported for ANRs on 32 + 64 bit devices?

kstenerud:
I've run manual tests in 32 and 64-bit mode with native and Java ANRs and didn't see any differences from before, but this would be expected since both stack walking styles are already long established in the codebase for signal handling. The only difference we would expect is that it doesn't crash on certain 32-bit devices in certain circumstances (that we haven't been able to repro).

fractalwrench:
One more query, it looks like the code could still hit the problematic stack unwinder if a 64-bit device is using 32-bit libraries. Unless I've misinterpreted things that is: https://github.com/bugsnag/bugsnag-android/blob/defbdc4e2ed242439e553be517a2a941aee19883/bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder_libunwind.c#L99

kstenerud:
It depends on what unwind style is being used. The entry point used by both the signal handler and the ANR handler is https://github.com/bugsnag/bugsnag-android/blob/defbdc4e2ed242439e553be517a2a941aee19883/bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.c#L63

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -36,13 +36,28 @@ void bsg_release_env_write_lock(void) {
   pthread_mutex_unlock(&bsg_global_env_write_mutex);
 }
 
+/**
+ * Get the configured unwind style for non-async-safe environments.
+ * DO NOT USE THIS IN A SIGNAL HANDLER!
+ */
 bsg_unwinder bsg_configured_unwind_style() {
```
kattrali, line 43: This is only used from `bugsnag.c` to unwind for `notify()`, so maybe this function should move to be internal to that file? Otherwise, the comment would make more sense in the header doc for this function rather than the implementation.

kattrali (approved):
highlighted one concern for future travelers, otherwise lgtm.

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
kstenerud: It makes use of the static `bsg_global_env`, which would have to be exposed if the function were placed in another file. This seemed like the lesser evil.

PR #1540, opened 2021-12-01
Avoid unnecessary network connectivity change breadcrumb

fractalwrench:
## Goal

Android's `ConnectivityManager` always invokes a callback with the current network state when `registerDefaultNetworkCallback` is called. This led to a network connectivity change breadcrumb being logged on Bugsnag initialization for API >= 24, which wasn't strictly a network change as such.

This alters the changeset to ignore the initial change on API 24 and to only log a breadcrumb for subsequent network changes.

## Testing

Added unit test coverage and manually confirmed that a breadcrumb is no longer left on Bugsnag initialization as a matter of course, but that genuine network changes afterwards are recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 * Flush in-memory sessions first
   [#1538](https://github.com/bugsnag/bugsnag-android/pull/1538)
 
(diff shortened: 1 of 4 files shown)


lemnik (approved):
LGTM

PR #1543, opened 2021-12-03
Clear the stack-trace memory to avoid UTF-8 encoding problems

lemnik:
## Goal
Ensure that `bugsnag_notify_env` does not fail due to uninitialised memory in the captured stack trace.

## Testing
Added a new function to the native mazerunner scenarios which will "dirty" the stack to emulate the behaviour of a longer running application. The function simply populates a 1Mb temporary array with data and returns.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -132,6 +132,7 @@ void bugsnag_notify_env(JNIEnv *env, const char *name, const char *message,
   jbyteArray jmessage = NULL;
 
   bugsnag_stackframe stacktrace[BUGSNAG_FRAMES_MAX];
+  memset(stacktrace, 0, sizeof(stacktrace));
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM

bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -132,6 +132,7 @@ void bugsnag_notify_env(JNIEnv *env, const char *name, const char *message,
   jbyteArray jmessage = NULL;
 
   bugsnag_stackframe stacktrace[BUGSNAG_FRAMES_MAX];
+  memset(stacktrace, 0, sizeof(bugsnag_stackframe) * BUGSNAG_FRAMES_MAX);
```
kattrali, line 135: Its an array, and declared immediately beforehand (instead of being passed as a parameter to this function, so you can grab the size directly:
```suggestion
  memset(stacktrace, 0, sizeof(stacktrace));
```

PR #1545, opened 2021-12-08
refactor(ndk): separate event reading funcs into own file

kattrali:
## Goal

The first of a few PRs, reducing the scope of `serializer.c`. It currently has four jobs:

* Writing the event format to disk
* Reading the event format from disk
* Serializing an event to JSON
* Being the entry point for any other code that needs one of those three things.

## Design

This change will start movement towards making the serializer just the interface to groups of functionality, the end result of the API being:

```c
bool bsg_serialize_event_to_file(bsg_environment *env) __asyncsafe;

bool bsg_serialize_last_run_info_to_file(bsg_environment *env) __asyncsafe;

bugsnag_event *bsg_deserialize_event_from_file(char *filepath);

char *bsg_serialize_event_to_json_string(bugsnag_event *event);
```

And the file structure looking something like:

```
utils/
  serializer
  serializer/
    event_reader (reads disk storage format to event)
    event_writer (writes event to disk format)
    json_writer  (writes event to json buffer)
```

This will put all of the async-signal-safe functions into the writer and hide a few details that are unneeded outside of the serialization code, to avoid confusion.

## Changeset

Direct move of the event reading code from serializer to a new event_reader file. Intentionally avoided further divisions to preserve change history in each subset (viewable with `git blame -C path/to/event_reader.c`), though removing the dependence of migrations on the public interface should be an eventual goal.

## Testing

Existing tests should cover this change, as its just a move.


lemnik (approved):
LGTM

fractalwrench (approved):
LGTM. It may be worth considering targeting `next` to reduce the chance of conflicts, depending on when the feature flags work is likely to ship

kattrali:
Ah - I targeted this branch as the landing place for the migration changes, did a quick scan and there's no other NDK-related stuff going in before the holiday code freeze :eyes:

PR #1547, opened 2021-12-08
refactor(ndk): move event writing funcs into own file

kattrali:
## Goal
Similar to the creation of serializer/event_reader.c, this moves all async-signal-safe functions for writing an event into their own file.

The serializer.c file is now a frontend to event reading and writing, while organizing the helper functions as their own units. A future changeset will do something comparable with the json string writer functions.

## Design

Moved `buffered_writer.*` and event writing functions into the serializer utilities since its only needed for writing events.

More info in #1545 

## Changeset

```
 bugsnag-plugin-android-ndk/src/main/CMakeLists.txt                               |  3
 bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c                       | 94 -
 bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.h                       |  7 
 bugsnag-plugin-android-ndk/src/main/jni/utils/{ => serializer}/buffered_writer.c |  4 
 bugsnag-plugin-android-ndk/src/main/jni/utils/{ => serializer}/buffered_writer.h |  0
 bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_writer.c          | 84 +
 bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_writer.h          |  9
 7 files changed, 105 insertions(+), 96 deletions(-)
```

## Testing

Existing tests should maintain coverage, as this is almost entirely a move.

PR #1549, opened 2021-12-09
Move NDK JSON serialization logic into own files

kattrali:
## Goal

Similar to the struct reading/writing functions (#1546, #1547), this separates the JSON writing from other bits, hiding things that aren't needed externally (like the parson library). This is the last stop before fixing things. 🚂 🛠️ 

## Changeset

The bulk of it is the move here:
```
 bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c                               | 418 -
 bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.h                               |  40 -
 bugsnag-plugin-android-ndk/src/main/jni/utils/{serializer.c => serializer/json_writer.c} |  26 +
 bugsnag-plugin-android-ndk/src/main/jni/utils/{serializer.h => serializer/json_writer.h} |  29 +
```

Added a commit to move the logger function into its own file, to avoid deep linking to `bugsnag_ndk.h` from serializer functions. Bonus: Do we want to turn off logging in prod builds? Or some subset? That would be pretty easy to add later by changing the macro based on whether `NDEBUG` ("not debug", a CMake-ism) is set, or add a few more (`LOGI`, `LOGW`, `LOGE`?) as needed.

## Testing

[Existing tests should cover this change.](https://tenor.com/bDWSw.gif)


fractalwrench (approved):
LGTM

PR #1550, opened 2021-12-09
Fix feature flags in NDK

lemnik:
## Goal
Feature Flags added in Java / Kotlin code need to be included as an array in unhandled NDK events. This requires that the `FeatureFlagState` be observed by the NDK plugin, and that the `featureFlags` be correctly encoded as a JSON array (and not as an object).

## Testing
Fixed existing Mazerunner tests and upgraded Mazerunner to ensure coverage. This PR will also go through a full CI run to ensure all of the tests are exercised.

### DIFF
--- a/features/full_tests/native_feature_flags.feature
+++ b/features/full_tests/native_feature_flags.feature
@@ -1,22 +1,26 @@
-#Feature: Synchronizing feature flags to the native layer
-#
-#  Scenario: Feature flags are synchronized to the native layer
-#    When I run "CXXFeatureFlagNativeCrashScenario"
-#    And I configure Bugsnag for "CXXFeatureFlagNativeCrashScenario"
-#    And I wait to receive an error
-#    And the error payload contains a completed unhandled native report
-#    And the exception "errorClass" equals "SIGILL"
-#    And event 0 contains the feature flag "demo_mode" with no variant
-#    And event 0 contains the feature flag "sample_group" with variant "a"
-#    And event 0 does not contain the feature flag "should_not_be_reported_1"
-#    And event 0 does not contain the feature flag "should_not_be_reported_2"
-#    And event 0 does not contain the feature flag "should_not_be_reported_3"
-#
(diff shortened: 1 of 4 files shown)


features/full_tests/native_feature_flags.feature
```diff
@@ -1,11 +1,10 @@
 Feature: Synchronizing feature flags to the native layer
 
   Scenario: Feature flags are synchronized to the native layer
-    When I run "CXXFeatureFlagNativeCrashScenario"
+    When I run "CXXFeatureFlagNativeCrashScenario" and relaunch the app
     And I configure Bugsnag for "CXXFeatureFlagNativeCrashScenario"
     And I wait to receive an error
     And the error payload contains a completed unhandled native report
-    And the exception "errorClass" equals "SIGILL"
```
kattrali, line 8: Does this change have some effect on error class...?

features/full_tests/native_feature_flags.feature
```diff
@@ -1,11 +1,10 @@
 Feature: Synchronizing feature flags to the native layer
 
   Scenario: Feature flags are synchronized to the native layer
```
fractalwrench, line 3: It'd be good to get an understanding of why this test passed previously and make any changes based on that. As it stands it looks like the test should have been failing previously.

kattrali: As written, I don't see how it could have ever ran successfully - the app would not have relaunched.

features/full_tests/native_feature_flags.feature
lemnik: The actual error name depends on Android version, so I've replaced this with:
```
And the exception "errorClass" equals one of:
          | SIGILL |
          | SIGTRAP |
```
Which is inline with the other similar scenarios

kattrali (approved):

fractalwrench (approved):
LGTM pending a rebase and confirmation that the new mazerunner tests pass on CI

PR #1551, opened 2021-12-10
NDK event migration v4 tests and fixes

kattrali:
## Goal

_Don't panic, this changeset is 90% tests_

1. Fix fix omissions (context) and mis-sized memcpys (breadcrumbs, etc) in v4 migration 
2. Add new testing strategy that should be simpler to use going forward
3. Fix feature flags being an object (stolen from #1550, but with impact on this test)

## Design

Introduced a new testing style so that assertions are more visible to the developer while running the tests, and so the event generator code and expected event output can be compared side-by-side easily. Left the existing tests as-is since they are more unit-y and these new tests verify event writing and json serialization together.

## Testing

Added new tests for migration v4, with more to follow for later migrations in subsequent changeset.


bugsnag-plugin-android-ndk/detekt-baseline.xml
```diff
@@ -3,6 +3,7 @@
   <ManuallySuppressedIssues/>
   <CurrentIssues>
     <ID>ComplexMethod:NativeBridge.kt$NativeBridge$override fun onStateChange(event: StateEvent)</ID>
+    <ID>LongMethod:EventMigrationV4Tests.kt$EventMigrationV4Tests$@Test fun testMigrateEventToLatest()</ID>
```
kattrali, line 6: This seemed like a sensible long method, since it verifies several things and pulling the objects out of the test scope didn't make sense. Though there will be another "issue" for every test in this format. 🤷

bugsnag-plugin-android-ndk/src/androidTest/java/com/bugsnag/android/ndk/migrations/EventMigrationV4Tests.kt
```diff
@@ -0,0 +1,182 @@
...
+        val infoFile = createTempFile()
+
+        val info = migratePayloadInfo(infoFile.absolutePath)
+
+        assertEquals(
+            mapOf(
+                "apiKey" to "5d1e5fbd39a74caa1200142706a90b20",
+                "notifierName" to "Test Library",
+                "notifierURL" to "https://example.com/test-lib",
+                "notifierVersion" to "2.0.11"
```
kattrali, line 21: This info is persisted but currently wholly unused - this key _should_ be used to send the event, but is not afaict.

bugsnag-plugin-android-ndk/src/androidTest/java/com/bugsnag/android/ndk/migrations/EventMigrationTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.content.Context
+import androidx.test.platform.app.InstrumentationRegistry
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.junit.Before
+import java.io.File
+
+open class EventMigrationTest {
+
+    private lateinit var context: Context
+    private var objectMapper = ObjectMapper()
```
lemnik, line 12: This doesn't appear to be re-assigned anywhere, so maybe it can be a `val`?
```suggestion
    private val objectMapper = ObjectMapper()
```

bugsnag-plugin-android-ndk/build.gradle
```diff
@@ -10,6 +10,9 @@ apply plugin: "com.android.library"
 
 dependencies {
     api(project(":bugsnag-android-core"))
+    // newest version that didn't require fiddling with sourceCompatibility -
+    // didn't use kotlin module to avoid runtime version conflicts
+    androidTestImplementation("com.fasterxml.jackson.core:jackson-databind:2.12.5")
```
fractalwrench, line 15: It feels preferable to use dsl-json or GSON rather than bringing another JSON serializer into the project, even if it's just for test code

fractalwrench (approved):
LGTM pending one review comment

bugsnag-plugin-android-ndk/build.gradle
fractalwrench: e.g. 

```
byte[] input;
DslJson<Object> dslJson = new DslJson<Object>();
Map<String, Object> map = dslJson.deserialize(Map.class, input, input.length);
```

lemnik (approved):
LGTM

bugsnag-plugin-android-ndk/build.gradle
kattrali: I attempted to pull in GSON (since we only have the streaming bits vendored in), but its giving me some JSON validity warnings for reasons that don't seem worth chasing for tests. dsl-json does not (yet?) exist in the repo this juncture, but I can fix it after the next merge or somthing to use `JsonHelper`.

fractalwrench: Yeah - it'd be worth merging in `next` if the integration branch is behind, as that'd allow this to use `JsonHelper`.

kattrali: Its a non-trivial merge (with a few conflicts in `Event` classes), so it might be worth holding until I get the tests in. I've got tests for all of the subsequent migrations and JSON serialization ready to go.

PR #1553, opened 2021-12-13
Merge next into feature flags integration branch

fractalwrench:
## Goal

Merges the latest from `next` into the feature flags integration branch. Relied on existing unit + mazerunner test coverage.


CHANGELOG.md
```diff
@@ -1,19 +1,44 @@
...
+
+* Clear native stacktrace memory in `bugsnag_notify_env` before attempting to unwind the stack
+  [#1543](https://github.com/bugsnag/bugsnag-android/pull/1543)
+
+## 5.16.0 (2021-11-29)
 
 ### Bug fixes
 
 * Increase resilience of NDK stackframe method capture
   [#1484](https://github.com/bugsnag/bugsnag-android/pull/1484)
```
kattrali, line 27: I think this changelog section should be in 5.15.0 - looks like the sections merged incorrectly.

kattrali (commented):
Saw just one thing, it may mean there's other weirdness happening in the changeset, but ideally its just the change log because tests would catch anything else.

CHANGELOG.md
fractalwrench: These seem to match up with the release notes on GH for 5.16.0, unless I've got the wrong end of the stick? https://github.com/bugsnag/bugsnag-android/releases/tag/v5.16.0

kattrali: weird - it must be that a previous merge from next into the feature flag integration branch went awry, as [it has it in 5.15.0](https://github.com/bugsnag/bugsnag-android/blob/integration/feature-flags/CHANGELOG.md#5150-2021-11-04)

PR #1554, opened 2021-12-13
Fix rare incorrect serialization of JSON fields in NDK errors

fractalwrench:
## Goal

Fixes the serialization of certain JSON numbers by serializing them into a string instead. This alters the Parson serializer so that if a number fits into an int without losing any precision, it will always be serialized as an int rather than a float.

For now I've only altered fields which are likely to hold large values, and have altered the serialization so that they are _always_ serialized as a string. This approach could be altered once the other NDK struct changes have landed if necessary.

## Testing

Relied on existing test coverage, and manually tested against the dashboard to confirm that values are displayed correctly.


bugsnag-plugin-android-ndk/src/main/jni/deps/parson/parson.c
```diff
@@ -998,6 +1000,20 @@ static int json_serialize_to_buffer_r(const JSON_Value *value, char *buf, int le
             }
             written_total += written;
             return written_total;
+        case JSONLong:
+            l_value = json_value_get_long(value);
+            if (buf != NULL) {
+                num_buf = buf;
+            }
+            written = sprintf(num_buf, "\"%lld\"", (long long) l_value);
```
kattrali, line 1008: Provided you import [`<inttypes.h>`](https://www.cplusplus.com/reference/cinttypes/), you can do:

```suggestion
            written = sprintf(num_buf, "\"%" PRId64 "\"", l_value);
```

and you'll get the right number type on each arch.

Regarding stringification, [RFC 7159](https://www.rfc-editor.org/rfc/rfc7159#section-6) states that JSON number values should be recognized correctly within the range +/- 2^53 (exclusive):

> Note that when such software is used, numbers that are integers and
   are in the range [-(2\**53)+1, (2\**53)-1] are interoperable in the
   sense that implementations will agree exactly on their numeric
   values.

So if you want a smaller change footprint (e.g. `versionCode` will still be a number for people until they get to 2 ^ 1023), you can check if a value is inside of the range and only stringify if not:

```c
if (l_value =< MAX_NUM || l_value >= MIN_NUM) { // not the best constant names, but you know
  written = sprintf(num_buf, "%" PRId64, l_value);
} else {
  written = sprintf(num_buf, "\"%" PRId64 "\"", l_value);
}
```

features/full_tests/native_event_tracking.feature
```diff
@@ -5,7 +5,7 @@ Feature: Synchronizing app/device metadata in the native layer
         And I wait to receive an error
         Then the error payload contains a completed handled native report
         And the event "app.inForeground" is true
-        And the event "app.duration" is greater than 0
+        And the event "app.duration" is not null
```
kattrali, line 8: Rather than just asserting _something_ is here, I'd add a new assertion to check that the stringified value is sensible. i.e.
```cucumber
And the event "app.duration" is a stringified number greater than 0
```

and check that the value is something like `/^[1-9][0-9]+$/` and `value.to_i > 0`

tomlongridge:
Closing for now whilst we monitor effect of #1578

PR #1555, opened 2021-12-13
Reduce duplication of test code

fractalwrench:
## Goal

Avoids duplication of test code in the unit tests by creating a shared sourceSet for `BugsnagTestUtils`.


lemnik (approved):
LGTM

PR #1558, opened 2021-12-14
Improve the memory and performance handling delivery response status codes

lemnik:
## Goal
Improve the memory use and performance of `DefaultDelivery.getDeliveryStatus` by avoiding the allocation and searching of a `List<Int>`.

## Changeset
Replaced the `IntRange` and `filter` with an equivalent method `isUnrecoverableStatusCode` which performs a simple set of comparisons.

## Testing
Additional unit tests added to cover more of the unrecoverable errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM

PR #1560, opened 2021-12-16
Sync feature flags from Android -> ReactNative

lemnik:
## Goal
Deliver FeatureFlag events that occur in Android to the ReactNative bridge and JavaScript code. This ensures that the two layers remain in-sync and that events modified in JavaScript contain the expected feature flags.

JavaScript implementation is here: https://github.com/bugsnag/bugsnag-js/pull/1618

## Testing
New unit test to check the expected event format of the events


fractalwrench (approved):
LGTM

PR #1561, opened 2021-12-17
fix(ndk): keep representable values as numbers not strings

kattrali:
## Goal

Reduce change footprint of #1554, and add some confidence with additional testing for the new integer value logic

Fixes an issue where `json_value_set_long()` had incorrect parameter type

## Changeset

* Serializes integer values over a threshold as strings, keeping everything else as-is. This keeps some consistency between what will be reported on each layer for the various values, and longer term the behavior of the different serializers can be synchronized.

## Testing

* Added new ParsonTest, which validates values sent over the wire, for ease of inspection. The provided cases fail without using the new `_long()` value type


bugsnag-plugin-android-ndk/src/test/cpp/ParsonTest.cpp
```diff
@@ -0,0 +1,27 @@
...
+extern "C" {
+#endif
+
+JNIEXPORT jstring JNICALL
+Java_com_bugsnag_android_ndk_ParsonTest_longToJsonArray(JNIEnv *env,
+                                                        jobject _this,
+                                                        jlong value) {
+  JSON_Value *ary_value = json_value_init_array();
+  JSON_Array *ary = json_value_get_array(ary_value);
+  json_array_append_long(ary, value);
```
kattrali, line 15: Use `json_array_append_number()` here to see the test fail

bugsnag-plugin-android-ndk/src/main/jni/deps/parson/parson.h
```diff
@@ -145,7 +145,7 @@ int json_object_dothas_value_of_type(const JSON_Object *object, const char *name
 JSON_Status json_object_set_value(JSON_Object *object, const char *name, JSON_Value *value);
 JSON_Status json_object_set_string(JSON_Object *object, const char *name, const char *string);
 JSON_Status json_object_set_number(JSON_Object *object, const char *name, double number);
-JSON_Status json_object_set_long(JSON_Object *object, const char *name, long number);
+JSON_Status json_object_set_long(JSON_Object *object, const char *name, int64_t number);
```
kattrali, line 148: Might be worth separately renaming all of these function to use `_integer()` instead of long, since that is the underlying type now, and it took a bit to spot that the type no longer matched

fractalwrench: This is a good suggestion 👍

fractalwrench (approved):
LGTM - pending rename of the function as suggested

PR #1566, opened 2021-12-21
Harden ndk layer with use of const

fractalwrench:
## Goal

Updates the NDK layer to use `const` in a few places where it could be using the keyword.

## Testing


bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -163,8 +163,8 @@ bool bugsnag_event_get_metadata_bool(void *event_ptr, const char *section,
   return false;
 }
 
-void bugsnag_event_start_session(bugsnag_event *event, char *session_id,
-                                 char *started_at, int handled_count,
+void bugsnag_event_start_session(bugsnag_event *event, const char *session_id,
+                                 const char *started_at, int handled_count,
```
kstenerud, line 167: Can we const `handled_count` and `unhandled_count` as well?

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer.c
```diff
@@ -15,7 +15,8 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
-bool bsg_event_write(bsg_report_header *header, bugsnag_event *event, int fd);
+bool bsg_event_write(const bsg_report_header *header,
+                     const bugsnag_event *event, int fd);
```
kstenerud, line 19: Can we const fd as well?

kstenerud (approved):

PR #1568, opened 2021-12-21
Avoid dependencies on the global namespace

twometresteve:
## Goal

Refactor test assertions to avoid dependencies on items in the global namespace and the brittleness that can bring.

## Design

Note that this is a temporary change to stop the next Maze Runner release from breaking the build (is is not inherently a breaking change in itself).  The new class added in this PR will be removed and `Maze.check` used instead.

## Testing

Covered by CI and I have also successfully run the e2e tests against the Maze Runner release that's currently waiting in the wings.

### DIFF
--- a/features/steps/android_steps.rb
+++ b/features/steps/android_steps.rb
@@ -1,3 +1,30 @@
+class Assertions
+  class << self
+    include Test::Unit::Assertions
+
+    def true(test, message = nil)
+      assert_true(test, message)
+    end
+
+    def block(message = 'block failed', &block)
+      assert_block(message, &block)
+    end
+
+    def not_nil(test, message = nil)
+      assert_not_nil(test, message)
+    end
+
+    def equal(expected, act, message = nil)
+      assert_equal(expected, act, message)
+    end
+
+    def include(collection, object, message = nil)
+      assert_include(collection, object, message)
+    end
+    alias includes include
+  end
+end
+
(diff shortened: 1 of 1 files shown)


imjoehaines (commented):
Doesn't https://github.com/bugsnag/maze-runner/pull/331 make this unnecessary?

twometresteve:
> Doesn't [bugsnag/maze-runner#331](https://github.com/bugsnag/maze-runner/pull/331) make this unnecessary?

Indeed it does and I want to go straight to using that.  But because the Android pipeline uses the `latest-v6-cli` Docker image for running the tests, any other branch would pick up the change as soon as the release is made, breaking the build.  I could have had the fixing PR all lined up and waiting, but with the time CI takes at times it could still have introduced an unwelcome dependency between PRs.

imjoehaines:
> &hellip; because the Android pipeline uses the `latest-v6-cli` Docker image for running the tests, any other branch would pick up the change as soon as the release is made, breaking the build

I still don't get it 😅 — wouldn't https://github.com/bugsnag/maze-runner/pull/331 be included in `latest-v6-cli` when a new release is made and therefore nothing would break?

imjoehaines (approved):
I see — somehow Minitest is being required in the CI docker image, which causes it to overwrite the new functions from https://github.com/bugsnag/maze-runner/pull/331

PR #1569, opened 2021-12-21
Use Maze.check

twometresteve:
## Goal

Use the new `Maze.check` mechanism for making test assertions, avoiding potential problems from using functions in the global namespace.

## Testing

Covered by a `[quick ci]` run.


fractalwrench (approved):
LGTM

PR #1572, opened 2022-01-04
[PLAT-6090] Harmonize all string copying in the NDK layer

kstenerud:
## Goal

We're currently using two `strlen` implementations and three `strcpy` implementations. Pick one proper implementation of each and use it exclusively.

## Testing

Re-ran all tests, and tested manually.


lemnik (approved):
LGTM

PR #1574, opened 2022-01-05
Improve JNI safety

kstenerud:
## Goal

Some Java objects were still leaking, or being removed at the wrong time, and the caches for the JVM handles were getting reloaded on every call (and leaking).

* All class and method handles have been consolidated to jni_cache.c
* Class handles are refreshed on each JNI call because they get deleted when the call returns.
* All JVM handles are now properly disposed of in sub-functions or loops (for top-level JNI functions there's no need since the handles will get reaped as soon as the function returns)
* Functions that weren't being used outside of their files have been marked static.

## Testing

Reran e2e tests, and tested manually.


lemnik (approved):
LGTM

PR #1578, opened 2022-01-06
fix(ndk): support storing large ints for memory and time values 

kattrali:
## Goal

Added additional changes to event format v8 to expand the size of version code, total memory, and duration values to fit 64-bit integers.

Technically any stored values can now be as large as ~52 bits when delivered as that is how much can fit in a number within the current JSON format (though #1554 mitigates this).

## Changeset

* Fetched values as longs where needed instead of doubles
* Updated storage type of total memory and friends to be `int64_t`
* Added migration for existing app and device contents of events

## Testing

* Updated existing assertions to use big big numbers

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_reader.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_reader.c
@@ -55,6 +55,9 @@ void migrate_breadcrumb_v1(bugsnag_report_v2 *report_v2,
                            bugsnag_report_v3 *event);
 void migrate_breadcrumb_v2(bugsnag_report_v5 *report_v5, bugsnag_event *event);
 
+void migrate_device_v2(bsg_device_info *output, bsg_device_info_v2 *input);
+void migrate_app_v3(bsg_app_info *output, bsg_app_info_v3 *input);
+
 void bsg_read_feature_flags(int fd, bsg_feature_flag **out_feature_flags,
                             size_t *out_feature_flag_count);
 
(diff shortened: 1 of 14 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_reader.c
```diff
@@ -376,6 +425,42 @@ bugsnag_event *bsg_map_v2_to_report(bugsnag_report_v2 *report_v2) {
   return bsg_map_v3_to_report(event);
 }
 
+static void add_metadata_string(bugsnag_metadata *meta, char *section,
+                                char *name, char *value) {
+  if (meta->value_count < BUGSNAG_METADATA_MAX) {
+    bsg_metadata_value *item = &meta->values[meta->value_count];
+    strcpy(item->section, section);
```
fractalwrench, line 432:
```suggestion
    bsg_strncpy(item->section, section, sizeof(item->section));
```

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_reader.c
```diff
@@ -402,22 +487,39 @@ void migrate_device_v1(bugsnag_report_v2 *report_v2, bugsnag_report_v3 *event) {
...
+                      report_v2->device.screen_resolution);
+}
+
+void migrate_device_v2(bsg_device_info *output, bsg_device_info_v2 *input) {
+  output->api_level = input->api_level;
+  output->cpu_abi_count = input->cpu_abi_count;
+  memcpy(&output->cpu_abi, input->cpu_abi, sizeof(input->cpu_abi));
+  memcpy(&output->orientation, input->orientation, sizeof(input->orientation));
+  output->time = input->time;
+  memcpy(&output->id, input->id, sizeof(input->id));
```
fractalwrench, line 513: Would it be preferable to use `bsg_strncpy` here? I can't see anything obvious going wrong with using `memset` but it'd be more in keeping with how the rest of the migrations work IIRC

fractalwrench (approved):
LGTM pending resolution of comments

bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_reader.c
kattrali: Its functionally equivalent -  the benefit of using `memcpy` was not needing to switch functions when copying array types other than `char` (like `cpu_abi`).

PR #1584, opened 2022-01-19
Propagate discarded exceptions to previously registered handlers

lemnik:
## Goal
Fix edge cases where crashing the app in a disabled `releaseStage` in the `Application.onCreate` method caused the app to hang instead of terminating. This appears to be caused by an interaction between the Android Runtime and it's startup behaviour, and the fact that we were not forwarding the uncaught exception to the default `UncaughtExceptionHandler` leaving the application partly running.

## Design
By forwarding uncaught exceptions before discarding them, we allow the default handler to do it's work and the app cleans up and terminates as expected.

## Changeset
Uncaught exceptions that will be discarded by Bugsnag are first forwarded to any previously registered `UncaughtExceptionHandler`.

## Testing
Manually tested the edge case as described in the ticket, and added a unit test for the new `ExceptionHandler` behaviour.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/ExceptionHandler.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/ExceptionHandler.java
@@ -1,6 +1,7 @@
 package com.bugsnag.android;
 
 import android.os.StrictMode;
+
 import androidx.annotation.NonNull;
 
 import java.lang.Thread;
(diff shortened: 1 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ExceptionHandler.java
```diff
@@ -66,6 +68,10 @@ public void uncaughtException(@NonNull Thread thread, @NonNull Throwable throwab
                     metadata, severityReason, null);
         }
 
+        forwardToOriginalHandler(thread, throwable);
```
kstenerud, line 71: I think this leaves us open to a death loop if anything in `uncaughtException` throws an exception. What about a try/finally (with forwarding code in finally) wrapping an inner try/catch that discards anything that goes wrong in our handler code?

kstenerud (changes requested):

bugsnag-android-core/src/main/java/com/bugsnag/android/ExceptionHandler.java
lemnik: We *might* be, although the docs on `uncaughtException` clearly state:
> Any exception thrown by this method will be ignored by the Java Virtual Machine.

https://developer.android.com/reference/java/lang/Thread.UncaughtExceptionHandler#uncaughtException(java.lang.Thread,%20java.lang.Throwable)

kstenerud: Maybe it's just because I'm getting older, but I never trust API promises :P

Another benefit would be a single point where it calls `forwardToOriginalHandler`, no matter how many return points we have.

kstenerud (approved):

PR #1585, opened 2022-01-21
Don't attempt to share the `bsg_jni_cache` between threads

lemnik:
## Goal
Fix `SIGABRT` crashes caused by multiple threads reading and writing to the global `bsg_jni_cache` simultaneously.

## Design
Replace the global JNI cache with a single cache for each JNI function call. A stack-allocated cache is populated on-entry to the NDK layer and a pointer to the cache is then passed into each function that requires it. This avoids any allocation errors, and ensures that JNI objects don't accidentally cross thread-boundaries.

## Testing
Manual testing with an app known to cause this issue. Due to the race-condition nature of this crash, no additional automated tests can be added at this time.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -118,7 +118,9 @@ void bugsnag_notify_env(JNIEnv *env, const char *name, const char *message,
   jbyteArray jname = NULL;
   jbyteArray jmessage = NULL;
 
-  if (!bsg_jni_cache_refresh(env)) {
+  bsg_jni_cache jni_cache;
+
+  if (!bsg_jni_cache_init(env, &jni_cache)) {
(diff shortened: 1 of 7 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -118,7 +118,9 @@ void bugsnag_notify_env(JNIEnv *env, const char *name, const char *message,
   jbyteArray jname = NULL;
   jbyteArray jmessage = NULL;
 
-  if (!bsg_jni_cache_refresh(env)) {
+  bsg_jni_cache jni_cache;
+
+  if (!bsg_jni_cache_refresh(env, &jni_cache)) {
```
kstenerud, line 123: This name doesn't really describe it properly anymore. It's not refreshing the cache, but rather filling it completely. Can we change it?

kstenerud (approved):

PR #1589, opened 2022-01-27
Parse FeatureFlags when OnSendCallbacks are configured

lemnik:
## Goal
Fix a bug where configuring an `OnSendCallback` would potentially cause feature flags to be dropped from reports. This was caused by `BugsnagEventMapper` not parsing the feature flags list back into the in-memory model, causing them to be dropped from any `Event` that was stored on disk (although only if an `OnSendCallback` was configured).

## Testing
Additional unit tests, and new Mazerunner tests to cover this scenario.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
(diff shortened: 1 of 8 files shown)


nickdowell (approved):
LGTM 👍

PR #1590, opened 2022-01-28
[PLAT-7808] Attach thread unconditionally when reporting ANR

kstenerud:
## Goal

Calling GetEnv to check if the env is currently attached to a thread is unnecessary in the ANR reporting code because it will never be attached to a thread (we're calling from a C thread). Also, it causes problems on certain Android versions.

## Testing

Tested manually, and ran unit and e2e tests.

### DIFF
--- a/bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
+++ b/bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
@@ -174,22 +174,10 @@ static void notify_anr_detected() {
     return;
   }
 
-  bool should_detach = false;
   JNIEnv *env;
-  int result = (*bsg_jvm)->GetEnv(bsg_jvm, (void **)&env, JNI_VERSION_1_4);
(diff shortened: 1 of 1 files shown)


bugsnag-plugin-android-anr/src/main/jni/anr_handler.c
```diff
@@ -236,10 +224,7 @@ static void notify_anr_detected() {
     check_and_clear_exc(env);
   }
 
-  if (should_detach) {
-    (*bsg_jvm)->DetachCurrentThread(
-        bsg_jvm); // detach to restore initial condition
-  }
+  (*bsg_jvm)->DetachCurrentThread(bsg_jvm);
```
lemnik, line 227: Could we not run into issues where current thread was already attached, and our `AttachCurrentThread` was a no-op?

lemnik (approved):
LGTM

PR #1591, opened 2022-01-31
v5.19.2

lemnik:
### Bug fixes

* Fixed an issue where feature-flags were not always sent if an OnSendCallback was configured
  [#1589](https://github.com/bugsnag/bugsnag-android/pull/1589)


nickdowell (approved):
Verified and confirmed that this fixes the feature flags issue for Unreal Engine 👍

PR #1592, opened 2022-01-31
Fix React Native events that override the apiKey

lemnik:
## Goal
Allow React Native applications to override the `Event.apiKey` in their callbacks.

## Changeset
If the `apiKey` field is present and not empty, override it in the `Event` before delivery to the Android layer

## Testing
Additional unit tests added, and manually tested with the example React Native app.

### DIFF
--- a/bugsnag-plugin-react-native/src/test/java/com/bugsnag/android/EventDeserializerTest.kt
+++ b/bugsnag-plugin-react-native/src/test/java/com/bugsnag/android/EventDeserializerTest.kt
@@ -90,6 +90,7 @@ class EventDeserializerTest {
         assertEquals("app-id", event.app.id)
         assertEquals("device-id", event.device.id)
         assertEquals("123", event.getMetadata("custom", "id"))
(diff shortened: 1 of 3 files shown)


bugsnag-plugin-react-native/src/test/java/com/bugsnag/android/EventDeserializerTest.kt
```diff
@@ -115,4 +115,30 @@ class EventDeserializerTest {
...
+            "metadata" to metadataMap(),
+            "app" to mapOf("id" to "app-id"),
+            "device" to mapOf(
+                "id" to "device-id",
+                "runtimeVersions" to hashMapOf<String, Any>()
+            )
+        )
+
+        val event = EventDeserializer(client, emptyList()).deserialize(map)
+        assertEquals("abc123", event.apiKey)
```
imjoehaines, line 142: Seems worth adding an assertion when there is no API key in the map (presumably it's `null` in that case?)

imjoehaines (approved):

PR #1596, opened 2022-02-02
[PLAT-7915] Init JNI cache once, and don't store JNIEnv

kstenerud:
## Goal

JNIEnv is only valid for the current thread, so we shouldn't be permanently storing it as a global. Store JavaVM instead.

Complete the JNI cache fix by making the cache once again global, ensuring that all object references have been first made global.

## Testing

Re-ran all automated tests, did manual testing using the example app.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
@@ -9,62 +9,72 @@
 #include "utils/stack_unwinder.h"
 #include "utils/string.h"
 #include <jni.h>
+#include <pthread.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
-static JNIEnv *bsg_global_jni_env = NULL;
-
-void bugsnag_start(JNIEnv *env) { bsg_global_jni_env = env; }
-
-void bugsnag_notify(const char *name, const char *message,
-                    bugsnag_severity severity) {
-  if (bsg_global_jni_env != NULL) {
-    bugsnag_notify_env(bsg_global_jni_env, name, message, severity);
-  } else {
-    BUGSNAG_LOG("Cannot bugsnag_notify before initializing with bugsnag_start");
+void bugsnag_start(JNIEnv *env) {
+  if (!bsg_jni_cache_init(env)) {
+    BUGSNAG_LOG("Could not init JNI jni_cache.");
   }
 }
 
-void bugsnag_set_user(const char *id, const char *email, const char *name) {
-  if (bsg_global_jni_env != NULL) {
-    bugsnag_set_user_env(bsg_global_jni_env, id, email, name);
-  } else {
-    BUGSNAG_LOG(
-        "Cannot bugsnag_set_user before initializing with bugsnag_start");
+void bugsnag_notify(const char *name, const char *message,
+                    bugsnag_severity severity) {
+  JNIEnv *env = bsg_jni_cache_get_env();
+  if (env == NULL) {
+    return;
   }
+
+  bugsnag_notify_env(env, name, message, severity);
+}
+
+void bugsnag_set_user(const char *id, const char *email, const char *name) {
+  JNIEnv *env = bsg_jni_cache_get_env();
+  if (env == NULL) {
+    return;
+  }
+
+  bugsnag_set_user_env(env, id, email, name);
 }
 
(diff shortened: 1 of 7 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag.c
```diff
@@ -13,58 +13,125 @@
...
+  JNIEnv *env = NULL;
+  jint state = (*bsg_jni_cache->jvm)
+                   ->GetEnv(bsg_jni_cache->jvm, (void **)&env, JNI_VERSION_1_6);
+  switch (state) {
+  case JNI_EDETACHED:
+    if ((*bsg_jni_cache->jvm)
+            ->AttachCurrentThread(bsg_jni_cache->jvm, &env, NULL) != JNI_OK) {
+      BUGSNAG_LOG("Failed to %s: Could not attach to JVM thread", purpose);
+      return NULL;
+    }
```
lemnik, line 37: Is it worth using a pthread destructor to ensure we detach the thread on termination?
https://github.com/bugsnag/bugsnag-plugin-android-cxxapi/blob/master/src/main/cpp/utils/JavaLoader.cpp#L34

kstenerud: I think that code is to handle the normal case where a thread you're calling from will eventually terminate?
This particular thread will never terminate until forcibly shut down on app termination, so there'd be no JVM to detach at that point, which might cause a crash (I think? These are pretty strange waters).

kattrali: Its for either case. this isn't _that_ weird. it makes sense to tidy up and detach in the case that the thread somehow terminates so it doesn't take the whole app with it, even for the case that we revisit this later to adjust how/when the thread is started and ends.

kstenerud: My worry is the possibility that JVM cleanup happens before thread reaping when the app terminates. In that case, there could be a danger of the env pointer or related JVM pointers being invalid by the time the thread cleanup code runs.

If we don't make it clean up the env, it would just leak memory, wouldn't it?

kattrali: The strategy in question is used in unreal engine and a few other places, its not dangerous.

lemnik: Failing to detach a thread that terminates will cause an app-level crash.

kstenerud: Hang on... I'm thinking of the wrong PR. This code is part of a two-function api. Anyone calling `acquire_env` must call `release_env`, which handles detaching. It's used by the non-env functions `bugsnag_notify`, `bugsnag_set_user`, and `bugsnag_leave_breadcrumb`, which all call the release function.

lemnik: These would still cause problems if the thread was externally terminated for any reason (thinking `pthread_cancel` or similar).

bugsnag-plugin-android-ndk/src/main/jni/safejni.c
```diff
@@ -1,12 +1,32 @@
...
+#include "utils/stack_unwinder.h"
 #include <stdbool.h>
+#include <string.h>
 #include <utils/string.h>
 
+#include <android/log.h>
+#ifndef BUGSNAG_LOG
+#define BUGSNAG_LOG(fmt, ...)                                                  \
+  __android_log_print(ANDROID_LOG_WARN, "BugsnagNDK", fmt, ##__VA_ARGS__)
+#endif
```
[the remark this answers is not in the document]
kstenerud: Oh right. Leftovers from ANR side...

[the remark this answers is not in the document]
kattrali: Unsure what happened in the commit, but this change wasn't resolved?

lemnik (approved):
LGTM

bugsnag-plugin-android-ndk/src/main/jni/safejni.c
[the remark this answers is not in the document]
kstenerud: Weird... It's pushed now!

kattrali (approved):

PR #1597, opened 2022-02-03
Fix local e2e test runs

twometresteve:
## Goal

Fixes a problem running Maze Runner locally.  The same issue doesn't occur on CI.

## Testing

Basic CI covers any regression, tested locally that it fixes the problem.


lemnik (approved):
LGTM

PR #1605, opened 2022-02-23
Update NDK unwinding logic

kattrali:
## Goal

Provide a backport of libunwindstack from android 12.0.0r31 compatible with API level 14. Consolidates to use a single unwinding implementation across everything.

This change is targeting an integration branch as there are additional required improvements, but this changeset has become large enough already. This is 1 of (probably) 3.

## Changeset

:warning: Don't panic - the vast majority of the changeset is removing now-unused files. The easiest way to understand the changes is to view each commit separately.

* Updated the unwinder. This required a few changes to the internals, but there are now two paths to unwind, one intended for a crashing context.
* Bumped the NDK revision used to build to r23 - an update from r17 was necessary for c++17 language features, and the guidance specifies that newer is better when you can't match the target app's version exactly.

## Testing

* Added a few new unit tests for the unwinder, which run on x86 / x64 / arm32 / arm64
* Tested manually that the stack contents are correct, mapping files are applied, and events group across architectures on arm32 and arm64 running on android 4.4, 5, 9, 11, and 12.


lemnik (approved):
LGTM!

PR #1606, opened 2022-02-23
Avoid exporting lib ndk internals

kattrali:
## Goal

Only expose the minimum surface required to make the library work, via a linker [version script](https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_25.html). They should be:

* `Java_*` for JNI functions
* `bugsnag_*` for the native API
* `__cxa*`, `__gxx*` and friends for C++ exception handling

## Changeset

* Add linker script to build process
* Remove exception property reporting dependent on arbitrary type info being present

## Testing

Tested manually that exception handling still worked as expected, though this change should be covered by the existing tests and addition of a specific test for C++ exception handling.


lemnik (approved):
LGTM

PR #1607, opened 2022-02-25
[PLAT-8098] Add a configuration option to limit the number of threads captured

kstenerud:
## Goal

Implement ROAD-1407: Provide a configuration option to limit the number of threads that get reported on an event.

## Design

See ED

## Testing

Manual testing, new unit tests, new e2e tests.


bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -29,49 +28,55 @@ internal class ThreadState @Suppress("LongParameterList") @JvmOverloads construc
...
-            stackTraces[currentThread] = currentThread.stackTrace
-        }
-        if (exc != null && isUnhandled) { // unhandled errors use the exception trace for thread traces
-            stackTraces[currentThread] = exc.stackTrace
-        }
+        val currentThread = java.lang.Thread.currentThread()
+        val threadCount = java.lang.Thread.activeCount()
+        val wantedThreadCount = kotlin.math.min(threadCount, maxThreadCount)
+        val threads: Array<java.lang.Thread?> = arrayOfNulls(wantedThreadCount)
+        val recordedThreadCount = java.lang.Thread.enumerate(threads)
```
lemnik, line 47: `Thread.enumerate` will only report threads in our current thread-group, while `getAllStackTraces` will list all threads in this apps runtime.

To get all of the process threads, we still need to find the system `ThreadGroup`:
```kotlin
private fun rootThreadGroup(): ThreadGroup {
    var group = Thread.currentThread().threadGroup
    
    while(group.parent != null) {
        group = group.parent
    }
    
    return group
}
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -29,49 +28,55 @@ internal class ThreadState @Suppress("LongParameterList") @JvmOverloads construc
...
-            }.toMutableList()
+            } else {
+                thread.stackTrace
+            }, projectPackages, logger)
+
+            return Thread(thread.id, thread.name, ThreadType.ANDROID, isErrorThread, Thread.State.forThread(thread), stackTrace, logger)
+        }
+
+        val list = ArrayList<Thread>(recordedThreadCount)
+        for (i in 0..recordedThreadCount-1) {
```
lemnik, line 67:
```suggestion
        for (i in 0 until recordedThreadCount) {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -29,49 +28,55 @@ internal class ThreadState @Suppress("LongParameterList") @JvmOverloads construc
...
     ): MutableList<Thread> {
-        // API 24/25 don't record the currentThread, add it in manually
-        // https://issuetracker.google.com/issues/64122757
-        if (!stackTraces.containsKey(currentThread)) {
-            stackTraces[currentThread] = currentThread.stackTrace
-        }
-        if (exc != null && isUnhandled) { // unhandled errors use the exception trace for thread traces
-            stackTraces[currentThread] = exc.stackTrace
-        }
+        val currentThread = java.lang.Thread.currentThread()
```
lemnik, line 43: We could clean these up nicely by using:
```kotlin
import java.lang.Thread as JavaThread
```

and then referring to them as `JavaThread` (or maybe `JThread`?) instead of the fully qualified thread name

bugsnag-android-core/src/androidTest/java/com/bugsnag/android/ThreadStateTest.kt
```diff
@@ -92,35 +109,38 @@ class ThreadStateTest {
...
 
-        // find the stack trace for the current thread that was passed as a parameter
-        val expectedTrace = allStackTraces.filter {
-            it.key.id == currentThread.id
-        }.map { it.value }.first()
+        // find the stack trace for the current thread that was passed as a parameter.
+        // Drop the top 3 stack elements because we're capturing the trace from a different location.
+        val expectedTrace = allThreads.filter {
+            it.id == currentThread.id
+        }.first().stackTrace.drop(3)
```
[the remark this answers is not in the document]
kstenerud: Yup, worked fine so added.

PR #1610, opened 2022-03-04
Add expectations for the symbolication results of ndk stacktraces

kattrali:
:warning: [Don't panic](https://tenor.com/KHmF.gif) - this is a big PR, but it should be possible to follow in each commit.

## Changeset

* Add assertions to existing unwinder unit tests to assert that the addresses matched to known functions are "reasonable" within a threshold. Avoided asserting specific values to not bog down the test with architecture-specific cases, but it should be a good "canary" for modifications going terribly wrong going forward
* Refactors existing native crash test cases to assert that the addresses returned from stack traces correspond to the correct file and line information in fixture files, in both the smoke test and full test runs. Addresses are validated using `addr2line` against the unstripped version of the fixture libraries under test.
* Added a few new test cases for C++ exceptions involving triggering `std::unexpected()`
* Updates the CI configuration to upload the unstripped libraries to make the above tests work on CI (@twometresteve this is the part you may care about - its similar to how the UE tests on iOS work)
* Added a three-liner for preventing simultaneous unwinds from an ANR and an app crash at the same time - technically this should be in its own changeset .. but I felt sneaky since this changeset was getting a full CI run.

## Testing

* Ran the full suite against the changes since it makes alterations to the CI configuration
* Tested locally on arm32 and arm64 to ensure the tests are consistent


kstenerud (approved):
Looks great! Just a couple of minor nits.

twometresteve (commented):
No comments from me.

PR #1612, opened 2022-03-04
ci: disable android 5 builds

kattrali:
This tests fail for spurious reasons so frequently that their results are not useful, because of either inherent OS instability or the device farm quality.


twometresteve (approved):
I'm happy with removing Android 5 for now.  There's a different approach we can take to avoid the Appium errors that cause so many of the failures, so we can look at re-adding them in future.

PR #1614, opened 2022-03-04
build: remove unused ABI from size reports

kattrali:
The comment added to PRs included a line for armeabi, which is no longer shipped as a supported ABI


lemnik (approved):
LGTM

PR #1616, opened 2022-03-04
build: auto-bump the version of bugsnag-android in example deps

kattrali:
Update the example app to use the latest version of bugsnag-android during the release process


lemnik (approved):
LGTM

PR #1618, opened 2022-03-08
Account for null termination in char array buuid

Reflejo:
This change allows to use for example sha256 as build uuids

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.h
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.h
@@ -51,7 +51,7 @@ typedef struct {
   char version[32];
   char active_screen[64];
   int64_t version_code;
-  char build_uuid[64];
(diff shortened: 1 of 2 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.h
```diff
@@ -51,7 +51,7 @@ typedef struct {
   char version[32];
   char active_screen[64];
   int64_t version_code;
-  char build_uuid[64];
+  char build_uuid[65];
```
Reflejo, line 54: Do I need to modify migrations? (ie https://github.com/bugsnag/bugsnag-android/blob/next/bugsnag-plugin-android-ndk%2Fsrc%2Fmain%2Fjni%2Futils%2Fserializer%2Fmigrate.h#L83, ...)

PR #1621, opened 2022-03-11
Add Bugsnag.isStarted

slack-jallen:
## Goal

Add Bugsnag.isStarted so that it is possible to check whether it is safe to call the other methods. Currently to use Bugsnag safely you either have to keep track of this yourself or wrap calls in try/catch.

## Design

Add a static method

## Changeset

Bugsnag.java

## Testing

Added unit tests


ZacSweers (approved):

luke-belton:
hey @slack-jallen - thanks for the PR!

Would you be able to share a little more info on your use case here please?

Why do you need the `isStarted` method? We’d usually recommend starting Bugsnag immediately at app launch, so is there a reason why `Bugsnag.start` might be delayed in your app?

Are you trying to store things (e.g. breadcrumbs, metadata) potentially before Bugsnag has started?

slack-jallen:
> hey @slack-jallen - thanks for the PR!
> 
> Would you be able to share a little more info on your use case here please?
> 
> Why do you need the `isStarted` method? We’d usually recommend starting Bugsnag immediately at app launch, so is there a reason why `Bugsnag.start` might be delayed in your app?
> 
> Are you trying to store things (e.g. breadcrumbs, metadata) potentially before Bugsnag has started?

It's not so much that it is delayed, it is that it might not be started at all depending on the configuration. It would be nice to know if Bugsnag API's are going to throw "IllegalStateException" before calling them.

luke-belton:
Hi @slack-jallen - thanks for the info. We'd usually recommend starting Bugsnag in all cases and using the [`enabledReleaseStages`](https://docs.bugsnag.com/platforms/android/configuration-options/#enabledreleasestages) config option to disable reporting in certain environments. Is there a reason why this isn't appropriate here?

Feel free to write in to support@bugsnag.com if you'd prefer to answer this away from a public forum!

slack-jallen:
> Hi @slack-jallen - thanks for the info. We'd usually recommend starting Bugsnag in all cases and using the [`enabledReleaseStages`](https://docs.bugsnag.com/platforms/android/configuration-options/#enabledreleasestages) config option to disable reporting in certain environments. Is there a reason why this isn't appropriate here?

Thanks for the info. The app is using `enabledReleaseStages`, but on top of that the app also has a separate config that determines whether to call start at all. I guess you want to guide people to always call Bugsnag.start()? I guess if that's your guideline we'll have to inject this config decision ourselves.

ZacSweers:
There are three cases that come to mind:
- early stage app initialization where crash reporting isn't necessarily enabled
- sample/playground apps that use common nonfatal reporting of some sort but aren't registered bugsnag apps themselves
- instrumentation/e2e tests that run the full app but without crash reporting

I don't think that enabling bugsnag everywhere is a reasonable solution, especially the latter two in a data-privacy sensitive environment

tomlongridge:
Thanks @slack-jallen – I've included your changes on #1640 and we'll include in the next release.

xander-jones:
Thanks again for the contribution @slack-jallen, this has been released in https://github.com/bugsnag/bugsnag-android/releases/tag/v5.22.0

slack-jallen:
> Thanks again for the contribution @slack-jallen, this has been released in https://github.com/bugsnag/bugsnag-android/releases/tag/v5.22.0

Awesome, thanks for taking it forward!

PR #1623, opened 2022-03-14
ci: remove android 5 builds

kattrali:
these tests still consistently fail, either due to poor test or device farm quality (unsure how I missed this one in #1612)


twometresteve (approved):
It's easy to miss things like this with the number of pipeline files that we have - I'll being look to reduce them to be more in line with bugsnag-unity, as that two step approach seems to have worked well.

PR #1624, opened 2022-03-14
Rework the background task service shutdown procedure to give more headroom against the 5s ANR timeout

kstenerud:
## Goal

In a worst case scenario, `BackgroundTaskService.shutdown()` could wait for 4.5 seconds to complete a shutdown, which is dangerously close to the 5s limit for an ANR if this happens on the main thread.

Rework so that all the shutdowns happen at the same time.

## Testing

Rerun all tests. Unfortunately this manifest as a flake in CI so there's no definitive test for it.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/BackgroundTaskService.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/BackgroundTaskService.kt
@@ -158,18 +158,13 @@ internal class BackgroundTaskService(
         internalReportExecutor.shutdownNow()
         defaultExecutor.shutdownNow()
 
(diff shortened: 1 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/BackgroundTaskService.kt
```diff
@@ -164,12 +164,10 @@ internal class BackgroundTaskService(
         // shutdown the IO executor last.
```
lemnik, line 164: I think the above comment can be safely removed.

lemnik (commented):
LGTM - Pending a comment and a CHANGELOG entry

lemnik (approved):
LGTM

PR #1639, opened 2022-03-28
Ensure handling_crash flag is atomic

kstenerud:
## Goal

The flag to mark that a native crash is being handled was not atomic, leaving room for a race condition should two threads trigger a native crash at the same time.

## Design

Use atomics to ensure that no race condition is possible.

## Testing

Reran all tests.


lemnik (approved):
LGTM

PR #1640, opened 2022-03-29
[PLAT-8191] Add Bugsnag.isStarted

tomlongridge:
## Goal

From #1621:

> Add Bugsnag.isStarted so that it is possible to check whether it is safe to call the other methods. Currently to use Bugsnag safely you either have to keep track of this yourself or wrap calls in try/catch.

Raised internally for CI and merge purposes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,17 @@
 
 ## TBD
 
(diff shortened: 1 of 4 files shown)


lemnik (approved):
LGTM

PR #1648, opened 2022-04-08
Remove unnecessary toList() call that was slowing down initialization

kstenerud:
## Goal

`NativeBridge.is32Bit()` had an unnecessary `toList()` call that for some reason was adding 25ms to the startup time. Removing it brings the time back down to sub-millisecond.

## Testing

Manual testing, reran all tests.


lemnik (approved):
LGTM - Pending CI

PR #1649, opened 2022-04-08
Resolve cppcheck errors

twometresteve:
## Goal

Resolve errors found by `cppcheck` locally, also ensuring that CI fails for errors in the future.

## Changeset

I've made the changes that seem appropriate, I don't think they should have any functional bearing.

## Testing

Covered by a full CI run.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -159,8 +159,7 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_install(
   // copy event path to env struct
   const char *event_path = bsg_safe_get_string_utf_chars(env, _event_path);
   if (event_path == NULL) {
-    free(bugsnag_env);
-    return;
+    goto error;
   }
   sprintf(bugsnag_env->next_event_path, "%s", event_path);
   bsg_safe_release_string_utf_chars(env, _event_path, event_path);
@@ -169,8 +168,7 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_install(
   const char *last_run_info_path =
       bsg_safe_get_string_utf_chars(env, _last_run_info_path);
   if (last_run_info_path == NULL) {
-    free(bugsnag_env);
-    return;
+    goto error;
   }
   bsg_strncpy(bugsnag_env->last_run_info_path, last_run_info_path,
               sizeof(bugsnag_env->last_run_info_path));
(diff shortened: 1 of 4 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -159,6 +159,7 @@ JNIEXPORT void JNICALL Java_com_bugsnag_android_ndk_NativeBridge_install(
   // copy event path to env struct
   const char *event_path = bsg_safe_get_string_utf_chars(env, _event_path);
   if (event_path == NULL) {
+    free(bugsnag_env);
```
kstenerud, line 162: For cleanup code in C, it's generally safer to use a "poor man's finally" approach using goto. That way, any future checked code that can fail doesn't require you to hunt for and match up every freeable resource and every exit point in the function, and all resource-freeing code exists in one place.

```c
{
    resource *resource_a = NULL;
    resource *resource_b = NULL;
    object *some_object = NULL;

    resource_a = get_resource_a();
    if (resource_a == NULL) {
        goto exit;
    }

    if (!validate_resource(resource_a)) {
        goto exit;
    }

    resource_b = get_resource_b(resource_a);
    if (resource_b == NULL) {
        goto exit;
    }

    if (!validate_resource(resource_b)) {
        goto exit;
    }

    some_object = copy_object_from_resource(resource_b);

exit:
    free(resource_a);
    free(resource_b);
    return some_object;
}
```

bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -119,6 +119,8 @@ bsg_metadata_value bugsnag_get_metadata_value(void *event_ptr,
   }
   bsg_metadata_value data;
   data.type = BSG_METADATA_NONE_VALUE;
+  data.bool_value = false;
+  data.double_value = 0.0;
```
kstenerud, line 123:
```c
  bsg_metadata_value data = {
          .type = BSG_METADATA_NONE_VALUE
  };
```

kstenerud (changes requested):

bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
twometresteve: Refactored to `free` when the allocated struct is not assigned to `bsg_global_env`.

bugsnag-plugin-android-ndk/src/main/jni/event.c
twometresteve: Changed as suggested.

kstenerud (approved):

PR #1651, opened 2022-04-12
Pull e2e test commands from Maze Runner

twometresteve:
## Goal

Start a thread when the Test Fixture launches to poll Maze Runner for the next command action to perform.

## Design

Android likes to throw up system dialogs for crashy apps, which have needed to be closed so that Appium actions aren't blocked.  However, error dialogs aren't always shown and so we've been spending a lot of time trying to dismiss dialogs that are not present (around 4s each time).  With this change the runtime for each full set of e2e tests is reduced by around 10 minutes.

## Changeset

`run_command` button removed, being replaced by a polling thread.
Removal of error dialog closing calls in most situations (still needed for touch events so that ANR scenarios work reliably).

## Testing

Covered by a full CI run.

### DIFF
--- a/features/fixtures/mazerunner/app/detekt-baseline.xml
+++ b/features/fixtures/mazerunner/app/detekt-baseline.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" ?>
 <SmellBaseline>
   <ManuallySuppressedIssues></ManuallySuppressedIssues>
-  <CurrentIssues></CurrentIssues>
+  <CurrentIssues>
+    <ID>MagicNumber:MainActivity.kt$MainActivity$1000</ID>
+  </CurrentIssues>
 </SmellBaseline>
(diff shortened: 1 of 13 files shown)


twometresteve:
> Making sure I understand correctly, because we loop with 1 sec intervals we can provide the command after the app has launched, meaning any pre-processing (such as setting the test endpoints to a different URL) will still be effective?

If there are no commands queued up then the test fixture will happy poll away and do nothing.  So yes, a step could be written to change the endpoints first.

Having said that, it's probably time to overhaul the test fixture UI - the endpoint config could be pushed into a command themselves and we could also revisit the feedback that we get on the screen (when watching the recording we now have no clue where in the test run we're looking at).  I can take that as an immediate follow-on task.

twometresteve:
Dropping back to a draft PR to rework a little for ANR scenarios.

twometresteve:
Branch is broken, restarting with a clean one.

PR #1655, opened 2022-04-14
Allow max threads to be configured via the manifest

kstenerud:
## Goal

The new maxReportedThreads config option was not settable via the manifest. This PR adds that capability.

## Testing

New tests to verify manifest settings.


lemnik (approved):
LGTM - Might need a Changelog entry as well?

PR #1658, opened 2022-04-14
Minor CI improvements

twometresteve:
## Goal

Combine the coding standard check steps, making for a faster run.  Also try to balance the two batches of e2e tests so they take roughly the same amount of time.

## Testing

Combined steps covered by a standard CI run and I verified that the `--continue` did result in all checks being performed if one fails using a temporary commit.  

For balancing the batches, I ran a full build and whilst runtimes can always vary, the batches were pretty balanced - within a minute of each other.


joshedney (approved):
LGTM!

PR #1659, opened 2022-04-19
Relaunch test fixture based on app_state

twometresteve:
## Goal

Use a more sophisticated approach to relaunching the test fixture after a crash, rather than a simple sleep.

## Design

If a slow device takes longer than the 5s currently allowed to write reports to file, flakes can occur.  Conversely, the app normally closes down far faster and so 5s on every crash is a waste of time.

## Changeset

Uses Appium's `app_state` to try and determine when the app has stopped running.  This identified some non-crashy scenarios that needed changing to avoid using the relaunch step, although there are still some scenarios for which the state still doesn't end in `not_running`.  While this is investigated further the step will simply log a warning and carry one regardless (just as it did with a simple sleep).

## Testing

Covered by a `[full ci]` run.

### DIFF
--- a/features/full_tests/auto_notify.feature
+++ b/features/full_tests/auto_notify.feature
@@ -21,9 +21,13 @@ Feature: Switching automatic error detection on/off for Unity
 
   @skip_android_8_1
   Scenario: ANR not captured with autoDetectAnrs=false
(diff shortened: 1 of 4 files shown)


joshedney (approved):
LGTM!

PR #1663, opened 2022-04-26
Use step ‘Bugsnag confirms it has no errors to send’ where feasible

twometresteve:
## Goal

Avoid some idle waiting time in e2e tests by checking for Bugsnag log messages rather than waiting for a set period of time for nothing to happen.

## Design

Follows the existing pattern for using the `Bugsnag confirms it has no errors to send` step, it just hasn't been employed uniformally.

## Changeset

Feature files updated to use the step where possible.  Scenario code updated to override `getInterceptedLogMessages`, which is needed so it knows what log messages to look out for.

## Testing

Covered by a full CI run.


joshedney (approved):
LGTM

PR #1665, opened 2022-04-27
Reduce change of CXXDelayedErrorScenario flaking

twometresteve:
## Goal

Reduce change of CXXDelayedErrorScenario flaking.

## Design

Any test scenario that depends on timely execution is inherently flaky, but the 250ms allowed here seems a little optimistic  - for it to allow enough time for the first error to be sent within the launch period.

## Changeset

Increased launch duration and added logging in case it continues to flake.

## Testing

CI shows the scenario still run as intended - only time will tell if the flake is really fixed.


joshedney (approved):
LGTM!

PR #1669, opened 2022-04-28
Update test DiscardOldEventsScenario to check for Bugsnag log

twometresteve:
## Goal

Enhances the checks performed in `DiscardOldEventsScenario`, whilst also saving a relaunch of the app.

## Design

I've built slightly on the existing mechanism that allows us to capture and report specific log messages to Maze Runner.

## Changeset

Updates the log capture mechanism to allow logs that match a regex (rather than just a string) to be sent to Maze Runner.  The test scenario is then reworked to be more specific about the requests that are being received and to check for the log message (saving a relaunch of the app).

## Testing

Covered by CI and successfully run several times.


joshedney (approved):
LGTM!

PR #1671, opened 2022-04-29
Fix NDK stack frame symbolication

lemnik:
# Goal
Fix NDK stack traces when libraries are dynamically linked after Bugsnag has been started.

# Design
When a stack frame symbol cannot be found during unwinding, fallback to `dladdr` to determine its symbol.

We don't attempt full reloads of the cache, as the code generated by ART should not change while the process is active so the only code we should fallback to `dladdr` for is linked in with either `dlopen` or `System.loadLibrary`.

## Testing
Modified an existing end-to-end test to cover this behaviour.


kstenerud (approved):

Makefile
```diff
@@ -32,16 +32,18 @@ test-fixtures:
...
-		cp $$LIB_CXX/obj/armeabi-v7a/libcxx-scenarios.so build/libcxx-scenarios-arm32.so && \
-		cp $$LIB_CXX_BUGSNAG/obj/x86_64/libcxx-scenarios-bugsnag.so build/libcxx-scenarios-bugsnag-x86_64.so && \
-		cp $$LIB_CXX_BUGSNAG/obj/x86/libcxx-scenarios-bugsnag.so build/libcxx-scenarios-bugsnag-x86.so && \
-		cp $$LIB_CXX_BUGSNAG/obj/arm64-v8a/libcxx-scenarios-bugsnag.so build/libcxx-scenarios-bugsnag-arm64.so && \
-		cp $$LIB_CXX_BUGSNAG/obj/armeabi-v7a/libcxx-scenarios-bugsnag.so build/libcxx-scenarios-bugsnag-arm32.so
+	LIB_CXX_BUGSNAG_64=$$(dirname `ls -dtr features/fixtures/mazerunner/cxx-scenarios-bugsnag/build/intermediates/cxx/RelWithDebInfo/*/obj/x86_64 | head -n 1`) && \
+	LIB_CXX_BUGSNAG_32=$$(dirname `ls -dtr features/fixtures/mazerunner/cxx-scenarios-bugsnag/build/intermediates/cxx/RelWithDebInfo/*/obj/armeabi-v7a | head -n 1`) && \
+	LIB_CXX_64=$$(dirname `ls -dtr features/fixtures/mazerunner/cxx-scenarios/build/intermediates/cxx/RelWithDebInfo/*/obj/x86_64 | head -n 1`) && \
+	LIB_CXX_32=$$(dirname `ls -dtr features/fixtures/mazerunner/cxx-scenarios/build/intermediates/cxx/RelWithDebInfo/*/obj/armeabi-v7a | head -n 1`) && \
```
twometresteve, line 38: The `-r` option to `ls` needs removing on each of these`.

lemnik: I actually plan on removing this file from this PR once https://github.com/bugsnag/bugsnag-android/pull/1670 is merged down. Hence the Draft status

kattrali (approved):

PR #1674, opened 2022-05-04
Build test fixtures on Macs

twometresteve:
## Goal

Reduce build times by using our Mac servers rather than Docker images (which often need costly pulls and/or rebuilds).  This change also removes a modest amount of noise from the Buildkite pipeline.

## Design

The core of this change is a refactor of the `Makefile`, which aligns local and CI builds.  

## Changeset

- Makefile refactor
- Buildkite pipeline updates

The publishing step step that occurs during a release still depends on the Docker images.  Further work on our infrastructure is needed to facilitate the move (specific signing keys are used) and I wanted to keep the size of this PR down.

## Testing

Covered by a full CI run.


Cawllec (approved):
Looks really good, that's a lot of time saved per run!

kstenerud (approved):
Tested all builds locally and an E2E run. Alles gut.

PR #1676, opened 2022-05-05
Add short delay to reduce probability of CXXDelayedCrash flake

Cawllec:
## Goal

Currently the `CXXDelayedCrashScenario` flakes, and is presumed to be the result of a race condition between the background trigger in the scenario and the test crash.

This introduces a 1s delay between the background trigger and crash to ensure the foreground state is properly changed.

Required #1674 to be merged before merging


features/fixtures/mazerunner/cxx-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/CXXDelayedCrashScenario.kt
```diff
@@ -26,7 +27,9 @@ class CXXDelayedCrashScenario(
 
     override fun onActivityStopped(activity: Activity) {
         super.onActivityStopped(activity)
-        log("App sent to background, triggering crash.")
-        activate(405)
+        log("App sent to background, triggering crash in 1 seconds.")
+        Handler().postDelayed({
+            activate(405)
+        }, 1000)
```
twometresteve, line 33: This makes logical sense to me, but it feels like there's something else going on here.  I ran it in a loop 22 times and it failed 6 times, but increasing the delay to 3s made it worse.

Perhaps we should be scrutinising the internals of the notifier more and adding some logging (as I think you were suggesting IRL)?

lemnik: Maybe we need some more logging? In say `onActivityResumed` as well (or possibly even *all* of the activity lifecycle methods)?

PR #1677, opened 2022-05-05
Build r16 test fixture with AGP 4.2.0

twometresteve:
## Goal

Ensure the NDK r16 test fixture can run on all architectures by building it with AGP 4.2.0.

## Design

After some investigation it was found that AGP 7 doesn't populate all subfolders of `lib` in the built APK with all `.so` files, causing linking errors in the scenarios that require the missing library.  Building with AGP 4 for r16 fixes this.

## Changeset

- New Gradle build property to control the use of AGP 4.
- Ported the script that copies built libraries into place to Ruby, allowing for greater complexity.

I've stopped short of adding the Samsung Galaxy S7 back into CI as it seems to be generally less stable than the existing devices.

## Testing

Covered by a full CI run and I've verified locally that scenarios such as `MultiProcessUnhandledCXXErrorScenario` (which dynamically load the C++ scenarios ;library) run on the Samsung.


Makefile
```diff
@@ -36,38 +36,39 @@ fixture-r16: notifier
 	# Build the r16 test fixture
 	@./gradlew -PTEST_FIXTURE_NDK_VERSION=16.1.4479499 \
                -PTEST_FIXTURE_NAME=fixture-r16.apk \
+               -PUSE_LEGACY_AGP=true \
                -PUSE_LEGACY_OKHTTP=true \
 	           -p=features/fixtures/mazerunner assembleRelease -x check
-	@scripts/copy-build-files.sh release fixture-r16
+	@ruby scripts/copy-build-files.rb release r16
```
kstenerud, line 42: carlton3

kstenerud (approved):

PR #1679, opened 2022-05-06
Fixed concurrency bug that could be triggered via the React Native plugin

kstenerud:
## Goal

`DeviceDataCollector.generateDevice` makes a clone of its internal `runtimeVersions` map, which can also be modified by calling `addRuntimeVersionInfo`. This can cause a concurrency exception in some cases.

## Design

Switched to a copy-on-write approach to sidestep the issue.

## Testing

New unit test to verify that the new code is concurrency-safe.


lemnik:
I'm not convinced that we should be opening up our classes just to satisfy Mockito limitations.

We should rather look at:
* The `kotlin-allopen` gradle plugin
* Upgrading Mockito to a version that supports mocking final Android classes (this will have a knock-on effect to our min Android version support)
* Avoiding the need for mocks in this test (I doubt this is realistic)

For now I'd also be willing to drop the unit test if it proves too invasive.

kstenerud:
Alright, I'm leaving the test code there (commented out) because it's a pain to write. Once we can figure out a better solution we can enable it again.

lemnik (approved):
LGTM

PR #1682, opened 2022-05-09
Allow retry for specific failure in MultiThreadedStartupScenario

twometresteve:
## Goal

Avoid test flake in `MultiThreadedStartupScenario`.

## Design

`MultiThreadedStartupScenario` is designed to verify that calling Bugsnag methods while it is being started on another thread does not raise an exception.  The scenario can occasionally flake if the OS end up starting the `leaveBreadcrumb` thread first (which we cannot control completely, even by setting thread priorities).  If that happens that Bugsnag raises an exception stating that `Bugsnag.start` must be called first.

## Changeset

This change allow a retry in the specific case that the exception is raised because the threads started in reverse order.

## Testing

Covered by CI and I saw the retry take affect once during development (it's not that easy to reproduce).

### DIFF
--- a/features/full_tests/threaded_startup.feature
+++ b/features/full_tests/threaded_startup.feature
@@ -6,4 +6,5 @@ Feature: Switching automatic error detection on/off for Unity
   Scenario: Starting Bugsnag & calling it on separate threads
     When I run "MultiThreadedStartupScenario" and relaunch the crashed app
     And I configure Bugsnag for "MultiThreadedStartupScenario"
-    Then I should receive no error
+    Then I wait to receive an error
+    And the error is correct for "MultiThreadedStartupScenario" or I allow a retry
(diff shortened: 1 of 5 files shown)


features/full_tests/threaded_startup.feature
```diff
@@ -6,4 +6,5 @@ Feature: Switching automatic error detection on/off for Unity
   Scenario: Starting Bugsnag & calling it on separate threads
     When I run "MultiThreadedStartupScenario" and relaunch the crashed app
     And I configure Bugsnag for "MultiThreadedStartupScenario"
-    Then I should receive no error
+    Then I wait to receive an error
+    And the error is correct for "MultiThreadedStartupScenario" or I allow a retry
```
Cawllec, line 10: Is this the best approach here? If it receives a request it shouldn't that indicates an actual issue.

PR #1687, opened 2022-05-12
Use correct version of the notifier during the building of the example app

joshedney:
## Goal

Use the correct notifier version during the building of the example app.

## Changeset

Add command to `Makefile` to build the notifier with the correct version and build the example app.
Bump the version of the plugin used to the current version.

## Testing

Version bumped locally and notifier/example app has been built
Covered by [full ci]


Makefile
```diff
@@ -27,10 +27,11 @@ MINIMAL_FIXTURE ?= false
 TEST_FIXTURE_NDK_VERSION ?= 16.1.4479499
 TEST_FIXTURE_NAME ?= fixture
 USE_LEGACY_OKHTTP ?= false
+NOTIFIER_VERSION ?= 9.9.9
```
lemnik, line 30: Not 100% sold on this name, since this version number is only used for test fixtures? By my current reading of it, you could also break the test fixtures by changing it's value (as they all expect 9.9.9)?

lemnik (approved):
LGTM

PR #1688, opened 2022-05-16
Improve CI stability

twometresteve:
## Goal

Attempts to improve CI stability.

## Changeset

- Use Google Nexus 6 in preference, as a higher tier device on BrowserStack
- Move the starting of the thread the polls for test Command to `onResume`, in the hope that the app is "more ready"
- USe `Maze.check.equal` rather than `assert_equal`

## Testing

Ran a full batch of e2e tests on Android 6 twenty times with no failures due to how the polling started (some other flakes do still exist).


joshedney (approved):
LGTM! Pending Tests

Cawllec (approved):

PR #1690, opened 2022-05-18
Only update session tracking after Activities have changed state

lemnik:
## Goal
On SDK 29 and higher use the `onActivityPostStarted` and `onActivityPostStopped` to update the session tracking, ensuring the `Activity` has finished changing state before we fetch new session state attributes.

## Design
When the foreground activities are started and stopped we capture flags such as context and `inForeground` and send them to observing layers (such as ReactNative and NDK). By capturing them after the `onStart`/`onStop` methods have returned we ensure that these attributes are properly up-to-date.

## Testing
TBD


Cawllec (approved):
LGTM!  I'm quickly testing another test change for a very similar test-flake, so I'd hold off from merging just yet.

PR #1700, opened 2022-06-10
Avoid phantom values in NDK Metadata

lemnik:
## Goal
Fix a bug when overwriting of clearing metadata, the NDK did not always remove the value as expected leading to phantom metadata values in reports. For example:

```
addMetadata("section", "overwrite", "value1")
addMetadata("section", "overwrite", "value2")
clearMetadata("section", "overwrite")
```

Would typically result in the Event having `"value1"` in its reported metadata, instead of no value as expected.

## Design
Introduce a simple compacting function which will remove any duplicates and gaps in the metadata array and compact it.

This new function is used when either:
* an entry or entries are removed from the metadata
* the metadata array is full and a new value is being added

By retaining the insertion order of the metadata we introduce minimal additional overhead on the most common path: adding metadata. The correct values are retained as the most recent values are always the last in the array (effectively acting as a primitive journal).

### Designs considered but not used

* A linear scan to check for existing values in the `add` functions - this would have introduce unacceptable overheads
* A full hash-map style implementation - requires a significant rewrite of the entire metadata model

## Testing
Modified the existing end-to-end test which deletes metadata before a native crash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
+* Fixed Bugsnag interactions with the Google ANR handler on newer versions of Android
+  [#1699](https://github.com/bugsnag/bugsnag-android/pull/1699)
+* Overwriting & clearing event metadata in the NDK plugin will no longer leave phantom values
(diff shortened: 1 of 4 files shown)


kstenerud (approved):
Ugly as hell O^2 algo, but for so few entries it's the sane choice :P

PR #1701, opened 2022-06-14
Add telemetry controls for internal reports.

kstenerud:
## Goal

See ROAD-1488-ED


lemnik:
I think this PR is also missing a CHANGELOG entry?

PR #1702, opened 2022-06-15
Make telemetry an EnumSet so that it can be modified

kstenerud:
This allows users to do:

```kotlin
config.telemetry.remove(Telemetry.INTERNAL_ERRORS)
```
instead of having to make a copy first (which is ugly in Java).


lemnik (approved):
LGTM

PR #1707, opened 2022-06-20
Implement a new OPAQUE metadata type

lemnik:
## Goal
Introduce a new `OPAQUE` metadata value type to allow complex metadata values to be stored by the NDK layer for later delivery. This PR introduces the type enum, fields to store the values within, and migrations for existing event files. It does not include any of the encoding or decoding logic.

## Changeset

* Add `BSG_METADATA_OPAQUE_VALUE` as a possible metadata value type
* Add `opaque_value` and `opaque_value_size` to the `bsg_metadata_value` struct for storage
* Bump the size limit of metadata value `section` and `name` to 64 chars
* Migrate all existing report formats to a new version (9) with the new structure

## Testing
New test for the latest binary event format, and updated all existing migration tests to cover the change to metadata and breadcrumbs (which also contain metadata).


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/event_reader.c
```diff
@@ -205,41 +250,50 @@ bugsnag_event *bsg_report_v8_read(int fd) {
 
   // read the feature flags, if possible
   bsg_read_feature_flags(fd, &event->feature_flags, &event->feature_flag_count);
+  // TODO: read the opaque data, if possible
```
kstenerud, line 253: Is this TODO supposed to be here?

lemnik: Yes - it'll be completed in another ticket

kstenerud (approved):

PR #1708, opened 2022-06-21
Encode opaque metadata

lemnik:
## Goal
Detect metadata values that cannot be represented directly in the C layer and encode them as `OPAQUE`. This includes breadcrumbs which use a `OpaqueValue` marker object in the `Map` delivered to the C layer for storage.

## Testing
Introduced a new end-to-end test that add complex metadata via the Java layer. This does not fully function yet as the data cannot be retrieved in the error reports (upcoming functionality). The test does however ensure that the new functionality introduces no other failures when adding the metadata.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/OpaqueValue.kt
+++ b/bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/OpaqueValue.kt
@@ -1,0 +1,45 @@
+package com.bugsnag.android.ndk
+
+import com.bugsnag.android.JsonStream
+import java.io.StringWriter
+
+/**
+ * Marker class for values that are `BSG_METADATA_OPAQUE_VALUE` in the C layer
+ */
+internal class OpaqueValue(val json: String) {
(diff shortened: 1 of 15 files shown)


bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/OpaqueValue.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.bugsnag.android.JsonStream
+import java.io.StringWriter
+
+/**
+ * Marker class for values that are `BSG_METADATA_OPAQUE_VALUE` in the C layer
+ */
+internal class OpaqueValue(val json: String) {
+    companion object {
+        private const val MAX_NDK_STRING_LENGTH = 64
+        private const val US_ASCII_MAX_CODEPOINT = 128
```
kstenerud, line 12: ASCII max codepoint is actually 127, but you also wouldn't want codepoint 127 in a string because it's the DEL control code.

kstenerud (approved):

PR #1713, opened 2022-06-27
Encode opaque metadata into JSON events to be delivered

lemnik:
## Goal
Encode the opaque JSON values into the Parson JSON model so that they are delivered with the event.

## Changeset

- When not in a signal handler: `free` any opaque values that cannot be referenced
  - clear metadata
  - clear metadata section
  - clear breadcrumbs
  - add breadcrumb (when overriding an existing breadcrumb)
- end-to-end tests covering opaque metadata in events and breadcrumbs
  - clearing opaque metadata in a C `on_error` callback to ensure we don't deadlock within a signal handler
- small fixes for bugs picked up in end-to-end testing
- new Mazerunner steps to cover metadata within breadcrumbs
  - searching breadcrumbs by name instead of index makes the tests more robust

## Testing
Modified the existing `CXXSignalSmokeScenario` to include a large amount of opaque metadata at both the event level, and within breadcrumbs. Some of this data is cleared out as part of the C crash handler to ensure out opaque metadata logic remains async safe.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/event.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/event.c
@@ -4,6 +4,9 @@
 #include <utils/logger.h>
 #include <utils/memory.h>
 
+static void bsg_clear_metadata_value(bsg_metadata_value *value);
+static void bsg_free_opaque_metadata(bugsnag_metadata *metadata);
+
 /**
  * Compact the given metadata array by removing all duplicate entries and NONE
  * values. This will retain the order of the values. After this function
@@ -32,15 +35,9 @@ static bool bsg_metadata_compact(bugsnag_metadata *const metadata) {
       if (strcmp(value->section, section) == 0 &&
           strcmp(value->name, name) == 0) {
 
-        // ensure we free any OPAQUE values we're going to remove
-        if (value->type == BSG_METADATA_OPAQUE_VALUE) {
-          bsg_free(value->opaque_value);
-          value->opaque_value_size = 0;
-        }
-
         // this will be picked up by our primaryIndex search on a future
         // iteration
-        value->type = BSG_METADATA_NONE_VALUE;
+        bsg_clear_metadata_value(value);
       }
     }
   }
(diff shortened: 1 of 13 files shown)


bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -215,12 +233,23 @@ void bugsnag_event_clear_metadata(void *event_ptr, const char *section,
...
 void bugsnag_event_clear_metadata_section(void *event_ptr,
                                           const char *section) {
+  size_t clearedCount = 0;
   bugsnag_event *event = (bugsnag_event *)event_ptr;
   for (int i = 0; i < event->metadata.value_count; ++i) {
     if (strcmp(event->metadata.values[i].section, section) == 0) {
+      if (event->metadata.values[i].type == BSG_METADATA_OPAQUE_VALUE) {
+        bsg_free(event->metadata.values[i].opaque_value);
+        event->metadata.values[i].opaque_value_size = 0;
+      }
```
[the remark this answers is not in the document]
lemnik: Agreed. I've added a `bsg_free_metadata_value` which is safe to call on any `metadata_value` and encapsulates the `type` check and memory release for a single value.

bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -104,6 +104,7 @@ static bool bsg_event_clear_metadata(bugsnag_metadata *metadata,
       if (value->type == BSG_METADATA_OPAQUE_VALUE) {
         bsg_free(value->opaque_value);
         value->opaque_value_size = 0;
+        value->opaque_value = NULL;
```
kstenerud, line 107: `bsg_free_metadata_value()`

bugsnag-plugin-android-ndk/src/main/jni/event.c
```diff
@@ -142,6 +138,26 @@ static int allocate_metadata_index(bugsnag_metadata *metadata,
   return index;
 }
 
+void bsg_free_metadata_value(bsg_metadata_value *value) {
+  if (value->type == BSG_METADATA_OPAQUE_VALUE &&
+      value->opaque_value_size > 0) {
+    bsg_free(value->opaque_value);
+
+    value->opaque_value = NULL;
+    value->opaque_value_size = 0;
```
kstenerud, line 147: Quick question: If an opaque value has been cleared, is its type not now `BSG_METADATA_NONE_VALUE`?

kstenerud (approved):

bugsnag-plugin-android-ndk/src/main/jni/event.c
lemnik: That is left up to the caller, since it's not required for the functions that clear the entire metadata or all breadcrumbs.

Cawllec (approved):
Looks good to me from a test perspective

PR #1714, opened 2022-06-27
[PLAT-8301] Store addresses in string form when writing to JSON

kstenerud:


bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -128,9 +128,9 @@ class Stackframe : JsonStream.Streamable {
 
         writer.name("columnNumber").value(columnNumber)
 
-        frameAddress?.let { writer.name("frameAddress").value(it) }
-        symbolAddress?.let { writer.name("symbolAddress").value(it) }
-        loadAddress?.let { writer.name("loadAddress").value(it) }
+        frameAddress?.let { writer.name("frameAddress").value("0x%x".format(it)) }
+        symbolAddress?.let { writer.name("symbolAddress").value("0x%x".format(it)) }
+        loadAddress?.let { writer.name("loadAddress").value("0x%x".format(it)) }
```
lemnik, line 133: This should probably use a specialised formatter to ensure that the outputted values are always output in an unsigned format.

bugsnag-android-core/src/main/java/com/bugsnag/android/BugsnagEventMapper.kt
```diff
@@ -193,15 +193,15 @@ internal class BugsnagEventMapper(
...
 
         (frame["symbolAddress"] as? String)?.let {
-            copy["symbolAddress"] = java.lang.Long.decode(it)
+            copy["symbolAddress"] = it
         }
 
         (frame["loadAddress"] as? String)?.let {
-            copy["loadAddress"] = java.lang.Long.decode(it)
+            copy["loadAddress"] = it
         }
```
lemnik, line 205: Maybe these should rather handle backwards compatibility with old reports that might still be in storage:

```
(frame["frameAddress"] as? Number)?.let {
    copy["frameAddress"] = "0x${it.toString(16)}"
}
```

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeStackframe.kt
```diff
@@ -59,9 +59,9 @@ class NativeStackframe internal constructor(
         writer.name("method").value(method)
         writer.name("file").value(file)
         writer.name("lineNumber").value(lineNumber)
-        writer.name("frameAddress").value(frameAddress)
-        writer.name("symbolAddress").value(symbolAddress)
-        writer.name("loadAddress").value(loadAddress)
+        frameAddress?.let { writer.name("frameAddress").value("0x%x".format(it)) }
+        symbolAddress?.let { writer.name("symbolAddress").value("0x%x".format(it)) }
+        loadAddress?.let { writer.name("loadAddress").value("0x%x".format(it)) }
```
lemnik, line 64: This will need to handle cases where the value came from an unsigned high-value long (where the most-significant bit is set).

bugsnag-android-core/src/main/java/com/bugsnag/android/Stackframe.kt
```diff
@@ -106,10 +106,10 @@ class Stackframe : JsonStream.Streamable {
         lineNumber = json["lineNumber"] as? Number
         inProject = json["inProject"] as? Boolean
         columnNumber = json["columnNumber"] as? Number
-        frameAddress = (json["frameAddress"] as? Number)?.toLong()
-        symbolAddress = (json["symbolAddress"] as? Number)?.toLong()
-        loadAddress = (json["loadAddress"] as? Number)?.toLong()
-        codeIdentifier = (json["codeIdentifier"] as? String)
+        frameAddress = (json["frameAddress"] as? String)?.let { java.lang.Long.decode(it) }
+        symbolAddress = (json["symbolAddress"] as? String)?.let { java.lang.Long.decode(it) }
+        loadAddress = (json["loadAddress"] as? String)?.let { java.lang.Long.decode(it) }
```
lemnik, line 111: This will need to handle cases where the String value represents an unsigned high-value long (where the most-significant bit is set). `Long.decode` will throw a `NumberFormatException` in these cases.

kstenerud:
I've moved all of the conversion code into JsonHelper.

lemnik (approved):
LGTM

PR #1720, opened 2022-07-13
[PLAT-8704] Allow feature flags to be accessed from the event object during callbacks

kstenerud:
## Goal

Allow feature flags to be accessed from the event object during OnSend callbacks

## Testing

Updated e2e tests to check accessing the feature flags from OnSend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Enhancements
 
-* Feature flags can now be accessed in the onSend callbacks.
+* Feature flags can now be accessed in the onSend and onError callbacks.
   [#1720](https://github.com/bugsnag/bugsnag-android/pull/1720)
 
 ## 5.24.0 (2022-06-30)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+* Feature flags can now be accessed in the onSend callbacks.
+  [#1720](https://github.com/bugsnag/bugsnag-android/pull/1720)
+
 ## 5.24.0 (2022-06-30)
 
 ### Enhancements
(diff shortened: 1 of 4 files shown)


features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/OnSendCallbackScenario.kt
```diff
@@ -14,6 +15,8 @@ internal class OnSendCallbackScenario(
     init {
         config.addOnSend(
             OnSendCallback { event ->
+                event.clearFeatureFlag("deleteMe")
+                event.addFeatureFlag(event.featureFlags[0].name, "b")
```
lemnik, line 19: Do we maybe want to use `getFeatureFlags()` as well here? Something like:

```suggestion
                event.clearFeatureFlag("deleteMe")
                if (event.getFeatureFlags().single() != FeatureFlag("fromStartup", "a")) {
                    return false
                }
                event.addFeatureFlag(event.featureFlags[0].name, "b")
```

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+* Feature flags can now be accessed in the onSend callbacks.
```
lemnik, line 7:
```suggestion
* Feature flags can now be accessed in the onSend and onError callbacks.
```

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/OnSendCallbackScenario.kt
kstenerud: That's in essence what it's already doing. `event.featureFlags[0].name` calls `getFeatureFlags()`, fetches index 0, and gets its name. It then adds that feature flag's name with a new variant. If any of that fails, the test will fail.

lemnik (approved):
LGTM

PR #1725, opened 2022-08-03
Run ANR smoke test first on Android 4

twometresteve:
## Goal

Attempt to avoid flakes in the ANR scenario on Android 4 by running it first.

## Design

One theory of why the ANR flakes only on Android 4 is that previous crashy scenarios put Android into an "unstable" stable, causing it to behave differently.  Whilst this PR may not fix the problem, it at least takes that factor out of the equation.

## Changeset

Feature files are now prefixed with a number to make the intention of running them in order reasonably clear.

## Testing

Covered by a basic CI run.


lemnik (approved):
LGTM

PR #1726, opened 2022-08-04
Added `codeIdentifier` to the NDK event model

lemnik:
## Goal
Introduce the `codeIdentifier` stack frame field into the NDK model, so that we can start reporting `Build ID`.

## Changeset
- Added the `code_identifier` to the stack frame model
- Event model is now version 10
- New migration tests for version 10

## Testing
New unit tests to ensure the migration code works. Manual testing of normal event reporting, and migrations.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/json_writer.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/json_writer.c
@@ -275,6 +275,15 @@ void bsg_serialize_stackframe(bugsnag_stackframe *stackframe, bool is_pc,
     json_object_set_string(frame, "method", (*stackframe).method);
   }
 
+  if (*stackframe->code_identifier != 0) {
+    char code_identifier[sizeof(stackframe->code_identifier) + 1];
+    code_identifier[sizeof(stackframe->code_identifier)] =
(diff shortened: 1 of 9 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/serializer/json_writer.c
```diff
@@ -275,6 +275,15 @@ void bsg_serialize_stackframe(bugsnag_stackframe *stackframe, bool is_pc,
     json_object_set_string(frame, "method", (*stackframe).method);
   }
 
+  if (*stackframe->code_identifier != 0) {
+    char code_identifier[sizeof(stackframe->code_identifier) + 1];
+    code_identifier[sizeof(stackframe->code_identifier)] =
+        0; // force zero terminator
```
kattrali, line 281: What is the expected length for these? Are they consistent?

lemnik: They're not strictly consistent and depend on which hash option is being used: https://android.googlesource.com/toolchain/llvm-project/+/refs/heads/master/lld/ELF/SyntheticSections.cpp#278

kattrali (approved):

PR #1727, opened 2022-08-08
Report "Build ID" as codeIdentifier in NDK stacktraces

lemnik:
## Goal
When reporting native stack traces, we need to include the "Build ID" as `codeIdentifier` for each stack frame (when it's available) so that we can later retrieve the correct symbol data.

## Changeset

- Add the "Build ID" to the stack-frames in the NDK event model during a crash
- Use `NativeStackframe` instead of `StackTraceElement` to relay `bugsnag_notify` events to the JVM layer

## Testing
New unit tests and additional checks on top of existing end-to-end testing

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
@@ -3,14 +3,18 @@ package com.bugsnag.android;
 import com.bugsnag.android.internal.ImmutableConfig;
 
 import android.annotation.SuppressLint;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import java.io.File;
 import java.nio.charset.Charset;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -36,6 +40,26 @@ public class NativeInterface {
         } else {
             return Bugsnag.getClient();
         }
+    }
+
+    /**
+     * Create an empty Event for a "handled exception" report. The returned Event will have
+     * no Error objects, metadata, breadcrumbs, or feature flags. It's indented that the caller
+     * will populate the Error and then pass the Event object to
+     * {@link Client#populateAndNotifyAndroidEvent(Event, OnErrorCallback)}.
+     */
+    private static Event createEmptyEvent() {
+        Client client = getClient();
+
+        return new Event(
+                new EventInternal(
+                        (Throwable) null,
+                        client.getConfig(),
+                        SeverityReason.newInstance(SeverityReason.REASON_HANDLED_EXCEPTION),
+                        client.getMetadataState().getMetadata().copy()
+                ),
+                client.getLogger()
+        );
     }
 
     /**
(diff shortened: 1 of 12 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -38,6 +42,36 @@ private static Client getClient() {
         }
     }
 
+    /**
+     * Create an empty Event for a native "handled exception" report. The returned Event will have
+     * no Error objects, metadata, breadcrumbs, or feature flags. It's indended that the caller
```
kattrali, line 47:
```suggestion
     * no Error objects, metadata, breadcrumbs, or feature flags. It's intended that the caller
```

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -38,6 +42,36 @@ private static Client getClient() {
...
+     * will populate the Error and then pass the Event object to
+     * {@link Client#populateAndNotifyAndroidEvent(Event, OnErrorCallback)}.
+     */
+    private static Event createNativeEvent() {
```
kattrali, line 51: What about this makes it inherently a native event? Does `createEmptyEvent()` make sense here (or perhaps as a convenience initializer on Event)?

bugsnag-android-core/src/test/java/com/bugsnag/android/NativeInterfaceApiTest.kt
```diff
@@ -220,11 +235,22 @@ internal class NativeInterfaceApiTest {
...
+            Severity.ERROR,
+            arrayOf<StackTraceElement>()
+        )
         verify(client, times(1)).notify(any(), any())
     }
 
+    @Test
+    fun notifyNativeTraceCall() {
+        NativeInterface.notify("SIGPIPE", "SIGSEGV 11", Severity.ERROR, arrayOf<NativeStackframe>())
+        verify(client, times(1)).populateAndNotifyAndroidEvent(any(), any())
```
kattrali, line 251: perhaps there's something more specific to be verified here, like to ensure the event was created with the correct error class / message?

bugsnag-plugin-android-ndk/src/main/jni/utils/string.c
```diff
@@ -24,3 +24,20 @@ void bsg_strncpy(char *dst, const char *src, size_t dst_size) {
...
+      (max_chars > byte_count * 2) ? byte_count : (max_chars - 1) / 2;
+
+  char *cursor = (char *)src;
+  char *outCursor = dst;
+  for (size_t i = 0; i < byte_copy_count; ++i) {
+    *outCursor++ = hex[(*cursor >> 4) & 0xF];
+    *outCursor++ = hex[(*cursor++) & 0xF];
+  }
+
+  *outCursor = 0;
```
kattrali, line 42: What is this set again for?

lemnik: null-terminator, maybe `\0` would be clearer?

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
lemnik: Actually I can use an existing constructor and get the same effect.

PR #1738, opened 2022-09-06
Encode line numbers as hex strings instead of numbers

lemnik:
## Goal
Avoid potential overflow issues when storing addresses in the `lineNumber` property by encoding them as hex strings. When native crashes occur we store the relative PC in the `lineNumber` property, which may be larger than a JSON number can hold.

## Design
Encode all `lineNumber` values as hex strings instead of raw JSON numbers. These are automatically decoded again on the server-side and still present as decimal numbers in the dashboard.

## Testing
Altered all of the existing test fixtures that check line-numbers to use the hex-encoded string syntax

### DIFF
--- a/bugsnag-plugin-android-ndk/src/androidTest/java/com/bugsnag/android/ndk/migrations/EventMigrationV10Tests.kt
+++ b/bugsnag-plugin-android-ndk/src/androidTest/java/com/bugsnag/android/ndk/migrations/EventMigrationV10Tests.kt
@@ -136,7 +136,7 @@ class EventMigrationV10Tests : EventMigrationTest() {
                     "stacktrace" to listOf(
                         mapOf(
                             "frameAddress" to "0xfffffffe",
-                            "lineNumber" to 4194967233L,
+                            "lineNumber" to "0xfa0a1ec1",
                             "loadAddress" to "0x242023",
                             "symbolAddress" to "0x308",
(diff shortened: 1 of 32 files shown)


bugsnag-plugin-android-ndk/src/androidTest/java/com/bugsnag/android/ndk/migrations/EventMigrationV10Tests.kt
```diff
@@ -136,7 +136,7 @@ class EventMigrationV10Tests : EventMigrationTest() {
                     "stacktrace" to listOf(
                         mapOf(
                             "frameAddress" to "0xfffffffe",
-                            "lineNumber" to 4194967233L,
+                            "lineNumber" to "0xfa0a1ec1",
```
kstenerud, line 139: Should these tests be changed? The existing legacy data on disk will be integers, not hex encoded strings...

lemnik: The binary files are only converted to JSON as part of delivery, so the legacy events in these tests would go through the new JSON encoding (resulting in legacy data having hex encoded strings, and not integers).

kstenerud (approved):

PR #1740, opened 2022-09-15
Skip the startup ANR end to end tests on Android 13

lemnik:
## Goal
Stop the `startup_anr` scenario from failing on Android 13.

## Changelog
Skip the startup ANR tests on Android 13, since we no longer have access to the `input` command from within the app process. As a result we can't force an ANR from within the `Application`.

## Testing
End-to-end tests pass again.


features/full_tests/startup_anr.feature
```diff
@@ -7,6 +7,9 @@ Feature: onCreate ANR
 # and then terminate the "misbehaving" app with a KILL (9) signal (almost like the ANR code doesn't
 # fire at all). Since we can't cover a KILL signal in a test, we skip Android 10.
   @skip_android_10
+# On Android 13 we no longer have permission to inject BACK button events, which are used to trigger
+# the ANR - so the test is not valid there, as such we skip all Android versions 10+
```
twometresteve, line 11: > as such we skip all Android versions 10+

Doesn't seem correct to me

Gemfile
```diff
@@ -4,7 +4,7 @@ source "https://rubygems.org"
 #gem 'bugsnag-maze-runner', path: '../maze-runner'
 
 # Or a specific release:
-gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.15.0'
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.26.0'
```
twometresteve, line 7: If you bump to 6.26.1 then you won't see ugly errors coming out of Maze Runner when running locally.

Issue #1741, opened 2022-09-19
Help needed contributing


guuilp:
Hey everyone!

I'm trying to make a contribution to the `bugsnag-android` project, but I'm having a hard time running the unit tests locally.

Tried to first build the project via command-line (`./gradlew clean assembleRelease`) but I'm having an NDK error, [here's the build scan](https://scans.gradle.com/s/n3qbmvgcvyjxc/activate).

I'm not used to the NDK setup, so I'd appreciate if you could help me! 🙇

guuilp:
Tried to first build the project via command-line (`./gradlew clean assembleRelease`) and received this:

```
C/C++: /Users/guilherme.pereira/Projects/forks/bugsnag-android/bugsnag-plugin-android-ndk/CMakeLists.txt release|armeabi-v7a : CMake Warning at /Users/guilherme.pereira/Library/Android/sdk/ndk/23.1.7779620/build/cmake/android-legacy.toolchain.cmake:416 (message):
  An old version of CMake is being used that cannot automatically detect
  compiler attributes.  Compiler identification is being bypassed.  Some
  values may be wrong or missing.  Update to CMake 3.19 or newer to use
  CMake's built-in compiler identification.
```

luke-belton:
Hi @guuilp - are you still having trouble with this? Perhaps it's no longer an issue given the resolution of https://github.com/bugsnag/bugsnag-android/pull/1742. Let me know if you want us to look into this further 👍

PR #1742, opened 2022-09-19
OkHttp Plugin - Allow filtering sensitive query params

guuilp:
## Goal

Allow filtering sensitive query params. Some of our requests contains sensitive information, such as latitude and longitude and we don't want Bugsnag to store that information.

## Design

I thought about receiving a function as constructor parameter, but that could harm the library behavior, so I opted to a more simple approach which is receiving a list os Strings and filtering the terms inside the Plugin. I think this approach makes it clear the intention of the new constructor parameter.

## Changeset

- BugsnagOkHttpPlugin
  - Receiving a new constructor parameter `sensitiveQueryParamList` to allow clients of bugsnag to tell the SDK which query param should be filtered.
  - Updated the `buildQueryParams` function to not add the query params that are in the `sensitiveQueryParamList` list.
## Testing

- BugsnagOkHttpPluginTest
  - Updated the `loadedClientCallEnd` test to guarantee the new logic.

I tried to run the tests locally, but I had a hard time configuring the CMake. Help registered here: https://github.com/bugsnag/bugsnag-android/issues/1741

### DIFF
--- a/bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/BugsnagOkHttpPlugin.kt
+++ b/bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/BugsnagOkHttpPlugin.kt
@@ -25,6 +25,7 @@ import java.util.concurrent.ConcurrentHashMap
  * https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/#the-response-body-must-be-closed
  */
 class BugsnagOkHttpPlugin @JvmOverloads constructor(
+    val sensitiveQueryParamList: List<String> = emptyList(),
     internal val timeProvider: () -> Long = { System.currentTimeMillis() }
 ) : Plugin, EventListener() {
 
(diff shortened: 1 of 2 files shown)


luke-belton:
Hi @guuilp - thanks for raising this! This is actually exactly what the [`redactedKeys`](https://docs.bugsnag.com/platforms/android/configuration-options/#redactedkeys) config option is for! For your use case something like this should work:
```
val config = Configuration.load(this)
config.redactedKeys = setOf("latitude", "longitude")
Bugsnag.start(this, config)
```

I'm going to close this out, but feel free to respond if this doesn't answer your question!

guuilp:
Oh, didn't know about the redact keys things. It seems like what I really want indeed, thanks!

guuilp:
@luke-belton I just tested the redacted keys and it didn't work as expected.

```
        val config = Configuration(debugConfig.bugsnagKey)
        config.addPlugin(bugsnagOkHttpPlugin)
        config.redactedKeys = setOf("latitude", "longitude", "lat", "lon")
        Bugsnag.start(application, config)
```


image

PR #1744, opened 2022-09-22
[PLAT-8915] Allow more breadcrumbs

kstenerud:
## Goal

- Allow more breadcrumbs and increase default maximum
- Add max metadata string value config
- Trim breadcrumbs when the Bugsnag payload is too big
- Keep internal metrics of trimmed metadata and breadcrumbs (when metrics are enabled)

Also fixed incorrectly placed usage metrics in the event JSON.

## Testing

Added e2e tests


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/StringUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    fun trimNullableStringValuesTo(maxStringLength: Int, map: MutableMap<String, Any?>): Pair<Int, Int> {
+        var charCount = 0
+        @Suppress("UNCHECKED_CAST")
+        val toTrim = map.filterValues { it is String && it.length > maxStringLength } as Map<String, String>
+        toTrim.forEach {
+            charCount += it.value.length - maxStringLength
+            map[it.key] = stringTrimmedTo(maxStringLength, it.value)
+        }
```
lemnik, line 20: This might be better done using the `MutableMap.Entry` objects, which allow the value to be changed without needing additional hashmap lookups?

bugsnag-android-core/src/main/java/com/bugsnag/android/EventInternal.kt
```diff
@@ -235,6 +248,48 @@ internal class EventInternal : FeatureFlagAware, JsonStream.Streamable, Metadata
...
+            stringAndCharCounts = breadcrumb.impl.trimMetadataStringsTo(maxLength)
+            stringCount += stringAndCharCounts.first
+            charCount += stringAndCharCounts.second
+        }
+        return Pair(stringCount, charCount)
+    }
+
+    internal fun serializeJsonPayload(streamable: JsonStream.Streamable): ByteArray {
+        return ByteArrayOutputStream().use { baos ->
+            JsonStream(PrintWriter(baos).buffered()).use(streamable::toStream)
```
lemnik, line 268:
```suggestion
            JsonStream(PrintWriter(baos)).use(streamable::toStream)
```

bugsnag-android-core/src/main/java/com/bugsnag/android/DefaultDelivery.kt
```diff
@@ -39,18 +44,53 @@ internal class DefaultDelivery(
...
     ): DeliveryStatus {
+        TrafficStats.setThreadStatsTag(1)
+        if (connectivity != null && !connectivity.hasNetworkConnection()) {
+            return DeliveryStatus.UNDELIVERED
+        }
 
+        var json = serializeJsonPayload(eventPayload)
+        if (json.size > maxPayloadSize) {
+            val breadcrumbAndBytesRemovedCounts =
+                eventPayload.event?.impl?.trimBreadcrumbsBy(json.size - maxPayloadSize)
```
lemnik, line 58: `eventPayload.event` will be `null` for unhandled errors, we'll need to load these from `eventFile`

bugsnag-android-core/src/main/java/com/bugsnag/android/BreadcrumbInternal.kt
```diff
@@ -32,4 +33,13 @@ internal class BreadcrumbInternal internal constructor(
         writer.value(metadata, true)
         writer.endObject()
     }
+
+    internal fun trimMetadataStringsTo(maxStringLength: Int): Pair<Int, Int> {
```
lemnik, line 37: This sort of `Pair` is used in a few places, is it possibly worth adding a new `internal data class` for these metrics so that it's clearer? Something like:
```kotlin
internal data class TrimMetrics(
   val itemsTrimmed: Int, // breadcrumbs, strings, whatever
   val dataTrimmed: Int // chars, bytes, whatever
)
```

features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/scenarios/BreadcrumbTooBigScenario.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    init {
+        config.maxStringValueLength = 2000000
+    }
+
+    override fun startScenario() {
+        super.startScenario()
+
+        Bugsnag.leaveBreadcrumb("test" + "*".repeat(1100000))
+        Bugsnag.notify(generateException())
```
lemnik, line 21: It would be good if at least one of these scenarios failed with an unhandled error, just to ensure that path functions as expected.

features/full_tests/trimming.feature
```diff
@@ -0,0 +1,77 @@
...
+    And the event "metaData.custom.foo" equals "abcdefghijklmnopqrstuvwxyzabcdefghijkl***<2> CHARS TRUNCATED***"
+    And the breadcrumb named "test" has "metaData.a" equal to "abcdefghijklmnopqrstuvwxyzabcdefghijkl***<2> CHARS TRUNCATED***"
+    And the event "usage.system.stringsTruncated" equals 2
+    And the event "usage.system.stringCharsTruncated" equals 4
+
+  # This scenario may flake if the payload structure changes significantly because it relies upon overflowing
+  # the max payload size by a certain amount. If it fails after a structural change, verify
+  # maze_output/failed/Payload_is_too_big_by_3_breadcrumbs/errors.log to make sure it's behaving as expected,
+  # and then modify "When I configure the app to run in the "100" state" to a number that generates 3 breadcrumbs
+  # worth of data too much.
```
twometresteve, line 56: I was going to say please don't add scenarios that are known to be flaky, but having read it more I think you really mean "start to fail consistently" rather than flake here?

twometresteve: ...although having said that isn't it likely to flake due to things like network and other breadcrumbs being slight vary in nature?  Can we turn off any breadcrumb types like to be sure that we get exactly the right number of breadcrumbs?

PR #1745, opened 2022-09-28
Small cleanups to Client constructor

kstenerud:
## Goal

Clean up the Client constructor.

## Design

Move non-constructor-y code to a separate `start()` function, which must be called before using a client.

## Testing

Reran all tests to make sure they still pass.


lemnik:
LGTM

PR #1746, opened 2022-09-30
Fix placement of usage statistics so that they are all placed under "usage" in the payload

kstenerud:
## Goal

The usage statistics were being put in the wrong place in the payload.

This PR moves the stats to their proper place under "usage".

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/InternalMetricsImpl.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/InternalMetricsImpl.kt
@@ -7,16 +7,21 @@ class InternalMetricsImpl : InternalMetrics {
     private val callbackCounts = hashMapOf<String, Int>()
 
     override fun toJsonableMap(): Map<String, Any> {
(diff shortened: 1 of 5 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/InternalMetricsImpl.kt
```diff
@@ -7,16 +7,21 @@ class InternalMetricsImpl : InternalMetrics {
...
-        return mapOf(
-            "config" to configDifferences,
-            "callbacks" to allCallbacks()
-        )
+        val callbacks = allCallbacks()
+
+        return listOfNotNull(
+            if (configDifferences.isNotEmpty()) "config" to configDifferences else null,
+            if (callbacks.isNotEmpty()) "callbacks" to callbacks else null,
+        ).toMap()
```
lemnik, line 15: I feel like if we're going to keep using this pattern, we should probably create our own helpers for it. Not for this PR but something to consider.

lemnik (approved):
LGTM

PR #1747, opened 2022-09-30
Make Scenario a little more error friendly

kstenerud:
## Goal

Get rid of the generic catch-all "is it a valid JVM class" error in Scenario and make it a little more intelligent when it encounters problems instantiating and running scenarios.


lemnik (approved):
LGTM

PR #1749, opened 2022-09-30
Attempt delivery on crash if configured

lemnik:
## Goal
Attempt to deliver crashes when they happen if configured.

## Design
The `DeliveryDelegate.cacheEvent` flag for `attemptSent` now attempts immediate delivery of just that single event

## Testing
A new test scenario was introduced to test that a crash can be delivered without a fixture restart.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
@@ -7,13 +7,14 @@ import com.bugsnag.android.internal.ImmutableConfig;
 import androidx.annotation.NonNull;
 import androidx.annotation.VisibleForTesting;
 
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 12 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/DeliveryDelegate.java
```diff
@@ -55,7 +56,11 @@ void deliver(@NonNull Event event) {
             String severityReasonType = event.getImpl().getSeverityReasonType();
             boolean promiseRejection = REASON_PROMISE_REJECTION.equals(severityReasonType);
             boolean anr = event.getImpl().isAnr(event);
-            cacheEvent(event, anr || promiseRejection);
+            if (immutableConfig.getAttemptDeliveryOnCrash()) {
+                cacheAndSendSynchronously(event);
```
kstenerud, line 60: `DeliveryDelegate.deliver` gets called for all kinds of higher level errors, including ANRs and handled errors (via `Client.populateAndNotifyAndroidEvent`). I don't think we should be blocking the thread for ANRs and non-fatal events.

lemnik: I agree on ANRs and such. I'll modify the order of these `if` checks to only act on unhandled errors that were not previously `attemptSend`.

kstenerud (approved):

PR #1756, opened 2022-10-05
chore: update danger to v9 and git gem to avoid vulnerability

tomlongridge:
## Goal

Avoid reported vulnerability (CVE-2022-25648) in Danger reporting of test fixture.

## Changeset

The latest Danger gem still relies on `git` <1.11, so forced the update with a dependency from our Gemfile.

## Testing

Looking for successful run of the Danger pipeline.


lemnik (approved):
LGTM - Pending an update and another CI run

PR #1760, opened 2022-10-07
Plat 8915 trim metadata

kstenerud:
## Goal

PLAT-9019: Trim metadata when the payload is too big

## Testing

New e2e tests for this


lemnik (approved):
LGTM Pending CI

PR #1769, opened 2022-10-18
Replaced "first significant stack frames" step with "any significant frames"

lemnik:
## Goal
Allow end to end tests to be more tolerant of platform changes when looking at native symbolicated stack traces. Some versions of Android cause `libunwind` to include signal handler frames at the top of the stack-trace causing a strict "first significant frames" check to fail.

## Design
Modified the "first significant stack frames" step to allow the list of expected frames to appear anywhere in the symbolicated stack-trace, as long as the frames remain sequential.

## Testing
Enabled all of the tests that were failing on Android 7 / 7.1 and verified they all run again.

### DIFF
--- a/features/full_tests/native_crash_handling.feature
+++ b/features/full_tests/native_crash_handling.feature
@@ -3,8 +3,6 @@ Feature: Native crash reporting
   Background:
     Given I clear all persistent data
 
-  # TODO: Pending PLAT-9011
-  @skip_android_7
   Scenario: Dereference a null pointer
     When I run "CXXDereferenceNullScenario" and relaunch the crashed app
     And I configure Bugsnag for "CXXDereferenceNullScenario"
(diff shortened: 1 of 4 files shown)


features/full_tests/native_crash_handling.feature
```diff
@@ -19,7 +17,7 @@ Feature: Native crash reporting
     And the event "device.totalMemory" is not null
     And the error payload field "events.0.device.cpuAbi" is a non-empty array
     And the error payload field "events.0.metaData.app.memoryLimit" is greater than 0
-    And the first significant stack frames match:
+    And any significant stack frames match:
```
twometresteve, line 20: nit: does "significant" actually mean anything now?  i.e. is it really just any frame?

lemnik: Ignore this PR - I just realised there is a better way to solve this 😆

PR #1773, opened 2022-10-19
e2e test improvements

twometresteve:
## Goal

Fix a couple of test flakes and make the fixture output less noisy in some cases.

## Changeset

- Safely get JSON element values in the test fixture to avoid ugly stack traces (if Maze Runner were to change behaviour, for example).
- Trigger scenario actions on the main thread to make times more consistent and avoid flakes in 2 scenarios
- Stop the test fixture polling for commands as soon as a positive action is received, to avoid noisy output in the device log.

## Testing

Covered by a basic CI run, and I've rerun the flaky scenarios 100 times locally to verify they are fixed (except in the case of  
 CXXDelayedCrashScenario which needs more work, but is no worse than it was).

### DIFF
--- a/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -91,6 +91,11 @@ class MainActivity : Activity() {
         }
     }
 
+    // As per JSONObject.getString but returns and empty string rather than throwing if not present
+    private fun getStringSafely(jsonObject: JSONObject, key: String): String {
+        return if (jsonObject.has(key)) jsonObject.getString(key) else ""
+    }
+
     // Starts a thread to poll for Maze Runner actions to perform
     private fun startCommandRunner() {
         // Get the next maze runner command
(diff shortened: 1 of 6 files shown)


features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -111,17 +116,22 @@ class MainActivity : Activity() {
...
-                    val action = command.getString("action")
-                    val scenarioName = command.getString("scenario_name")
-                    val scenarioMode = command.getString("scenario_mode")
-                    val sessionsUrl = command.getString("sessions_endpoint")
-                    val notifyUrl = command.getString("notify_endpoint")
+                    val action = getStringSafely(command, "action")
+                    val scenarioName = getStringSafely(command, "scenario_name")
+                    val scenarioMode = getStringSafely(command, "scenario_mode")
+                    val sessionsUrl = getStringSafely(command, "sessions_endpoint")
+                    val notifyUrl = getStringSafely(command, "notify_endpoint")
```
lemnik, line 123:
```suggestion
                    val action = command.optString("action")
                    val scenarioName = command.optString("scenario_name")
                    val scenarioMode = command.optString("scenario_mode")
                    val sessionsUrl = command.optString("sessions_endpoint")
                    val notifyUrl = command.optString("notify_endpoint")
```

features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -91,6 +91,11 @@ class MainActivity : Activity() {
         }
     }
 
+    // As per JSONObject.getString but returns and empty string rather than throwing if not present
+    private fun getStringSafely(jsonObject: JSONObject, key: String): String {
+        return if (jsonObject.has(key)) jsonObject.getString(key) else ""
+    }
```
lemnik, line 97: This function is effectively what `JSONObject.optString` does. Included a suggestion below.

lemnik (approved):
LGMT - with a suggestion for more easily safe-getting the JSON strings

PR #1776, opened 2022-11-01
Update e2e tests to use Maze Runner v7

twometresteve:
## Goal

Update the e2e tests to use Maze Runner v7.

## Changeset

- Use v7 Gem/Docker images (legacy driver still needed for Android 4)
- Explicitly set Appium version 
- Handle “noop” action from Maze Runner when no commands are queued
- Explicitly set orientation to portrait at start of scenarios that check it
- Use w3c driver methods

## Testing

Covered by a full CI run (conducted prior to the last commit that reinstated `--fail-fast`).


lemnik (approved):
LGTM

PR #1781, opened 2022-11-10
Avoid race conditions in refreshSymbolTable

lemnik:
## Goal
Avoid race conditions between the refreshing of the crash-time stack unwinder in the NDK module, and the `refreshSymbolTable` functions.

## Changelog

- `bsg_unwinder_refresh` now completely rebuilds the crash-time stack unwinder and only updates the pointer once the new unwinder is fully initialised.
- `bsg_unwind_crash_stack` localises its reference to the crash-time unwinder upon entry to keep the reference stable.
- `unwinding_crash_stack` is used to ensure that the unwinder used by `bsg_unwind_crash_stack` is not destroyed by a concurrent `bsg_unwinder_refresh`

## Testing
A new mazerunner scenario was introduced that crashes the app while repeatedly rebuilding the unwinder. This scenario will break (most of the time) if the refresh interferes with the stack unwind. Under manual testing this break happens for almost every test, but given the race-condition nature of this we can expect some fakes (but only on regression).

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.cpp
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.cpp
@@ -16,7 +16,7 @@ static unwindstack::Unwinder *crash_time_unwinder;
 // soft lock for using the crash time unwinder - if active, return without
 // attempting to unwind. This isn't a "real" lock to avoid deadlocking in the
 // event of a crash while handling an ANR or the reverse.
-static bool unwinding_crash_stack;
+static std::atomic_bool unwinding_crash_stack = ATOMIC_VAR_INIT(false);
 
 // Thread-safe, reusable unwinder - uses thread-specific memory caches
 static unwindstack::LocalUnwinder *current_time_unwinder;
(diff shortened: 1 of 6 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.cpp
```diff
@@ -16,7 +16,7 @@ static unwindstack::Unwinder *crash_time_unwinder;
 // soft lock for using the crash time unwinder - if active, return without
 // attempting to unwind. This isn't a "real" lock to avoid deadlocking in the
 // event of a crash while handling an ANR or the reverse.
-static bool unwinding_crash_stack;
+static volatile bool unwinding_crash_stack;
```
kstenerud, line 19: Can we use an atomic type here to make absolutely sure? C compilers have a tendency to play fast and loose with volatile.

kstenerud (approved):

Marudhas (approved):

Marudhas (approved):

Marudhas (approved):

Marudhas (approved):

Marudhas (approved):

Marudhas (approved):

Marudhas (approved):

Marudhas (approved):

PR #1783, opened 2022-11-15
Refresh symbols in background

lemnik:
## Goal
Avoid app stutters when refreshing the native symbol tables.

## Design
When the app startup is marked as complete, use the `BackgroundTaskService` to invoke `refreshSymbolTable` instead of directly invoking it on the main thread.

## Testing
Relied on existing tests.


bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
```diff
@@ -164,12 +174,13 @@ class NativeBridge : StateObserver {
     }
 
     private fun deliverPendingReports() {
-        lock.lock()
         val filenameRegex = """.*\.crash$""".toRegex()
+        lock.lock()
```
kstenerud, line 178: Can we do `withLock` here as well?

kstenerud (approved):

bugsnag-plugin-android-ndk/src/main/java/com/bugsnag/android/ndk/NativeBridge.kt
lemnik: We could, but it doesn't improve things much since we also have a `catch` in this block. So we would end up with a `withLock` block, and a `try {} catch {}` block as well.

PR #1787, opened 2022-11-18
Sort received requests to avoid test flakes

twometresteve:
## Goal

Avoids a flake in `ErrorBreadcrumbsScenario` where the errors can be received out of order.

## Changeset

Sorts the requests received in that scenario and refactors the existing scenarios that sort requests to use the step available in Maze runner.

## Testing

Covered by CI (once the device farm starts working again).


lemnik (approved):
LGTM

PR #1788, opened 2022-11-18
Temporary soft fail of tests on Android 6 and 7

twometresteve:
## Goal

A repeat of #976 .

Allows tests on Android 6 and 7 to soft fail whilst the secure tunnel outage at BrowserStack is resolved.

Once the outage is resolved and the jobs start running again we should remove the soft fails.

## Testing

Covered by CI.


lemnik (approved):
LGTM

PR #1789, opened 2022-12-06
Remove soft fails for Android 6 and 7 e2e tests

twometresteve:
## Goal

Remove soft fails for Android 6 and 7 e2e tests.

## Design

A persistent problem with the BrowserStack secure tunnel meant that these tests were consistently failing for a number of days.  The problem is now confirmed as fixed and so these soft_fails can now be removed.

## Testing

Covered by CI.


joshedney (approved):
Good to me it looks

PR #1790, opened 2022-12-19
Reconcile Cucumber steps

twometresteve:
## Goal

Fixes the build after `the event has {int} breadcrumbs` was added to Maze Runner, causing a Cucumber ambiguous step error.

Also takes the opportunity to remove `the event has a {string} breadcrumb with the message {string}` in favour of `the event has a {string} breadcrumb name {string}` (no "the"), also already defined in Maze Runner.

## Testing

Covered by a basic CI run.


imjoehaines (approved):
LGTM

There's also a duplicate step just below this: "the exception stacktrace matches the thread stacktrace". The second definition is wrong as it uses non-namespaced Maze stuff

twometresteve:
> 

And of course neither are used (Cucumber would complain about an ambiguous definition if they were).  I've removed them both.

Cawllec (approved):

PR #1792, opened 2022-12-22
Allow a single retry on flaky scenarios

twometresteve:
## Goal

Allow a retry for some e2e scenarios in specific situations.

## Design

These scenarios are known to be flaky and a ticket exists to investigate further.  This PR uses a feature of Maze Runner to allow a single retry in a specific circumstance - in this case if `"app.inForeground"` is `true`.  If any other difference occurs or the request is not received, then the scenario will fail.

## Testing

Covered by CI and I ran a scenario multiple times locally to verify what happens on a retry.


rich-bugsnag (approved):
NOICE 👍

Marudhas (approved):

Marudhas (approved):

PR #1794, opened 2023-01-04
add addMetadata overload to allow Unity to pass metadata as a hashmap

rich-bugsnag:
add addMetadata overload to nativeInterface.java to allow Unity to pass metadata as a hashmap

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
@@ -266,6 +266,13 @@ public class NativeInterface {
     }
 
     /**
+     * Add metadata to subsequent exception reports with a Hashmap
+     */
+    public static void addMetadata(@NonNull final String tab, HashMap<String, Object> metadata) {
+        getClient().addMetadata(tab, metadata);
+    }
+
+    /**
      * Return the client report release stage
      */
     @Nullable


bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -265,6 +265,13 @@ public static void addMetadata(@NonNull final String tab,
         getClient().addMetadata(tab, key, value);
     }
 
+    /**
+     * Add metadata to subsequent exception reports with a Hashmap
+     */
+    public static void addMetadata(@NonNull final String tab, HashMap<String, Object> metadata) {
```
lemnik, line 271:
```suggestion
    public static void addMetadata(@NonNull final String tab, Map<String, ?> metadata) {
```

lemnik (approved):
LGTM - Pending CI

bugsnag-android-core/src/main/java/com/bugsnag/android/NativeInterface.java
```diff
@@ -265,6 +265,13 @@ public static void addMetadata(@NonNull final String tab,
         getClient().addMetadata(tab, key, value);
     }
 
+    /**
+     * Add metadata to subsequent exception reports with a Hashmap
+     */
+    public static void addMetadata(@NonNull final String tab, @NonNull final Map<String, ?> metadata) {
```
lemnik, line 271: Checkstyle isn't happy with how long this line is. 
```suggestion
    public static void addMetadata(@NonNull final String tab,
                                   @NonNull final Map<String, ?> metadata) {
```

PR #1796, opened 2023-01-20
Read mazerunner address from config file

yousif-bugsnag:
## Goal

The test fixture is currently hard-coded to reach Maze Runner on [bs-local.com:9339](http://bs-local.com:9339/).  We’re moving to BitBar, so that address will need to change - but also we’re going to use public IP address that we’ll need to determine dynamically.

Maze Runner is being updated to write its address to a “Fixture config” (json) file and push this to the app using Appium.  

This PR updates the fixture app to poll/watch for the arrival of the file and read the Maze Runner address before it continues with what it normally does (poll the command endpoint for stuff to do).

The file is called `fixture_config.json` and will be located in `/sdcard/Android/data/<app package>/files` .  Contents is a simple JSON object:
```
{
  "maze_address": "bs-local.com:9339"
}
```

If the config file is not found within 5 seconds the fixture app will revert back to the current address.

### DIFF
--- a/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -125,7 +125,7 @@ class MainActivity : Activity() {
 
     // As per JSONObject.getString but returns and empty string rather than throwing if not present
     private fun getStringSafely(jsonObject: JSONObject?, key: String): String {
-        return if (jsonObject!!.has(key)) jsonObject.getString(key) else ""
+        return jsonObject?.optString(key) ?: ""
     }
 
     // Starts a thread to poll for Maze Runner actions to perform
--- a/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -20,6 +20,8 @@ import java.net.URL
 import kotlin.concurrent.thread
 import kotlin.math.max
 
+const val CONFIG_FILE_TIMEOUT = 5000
+
 class MainActivity : Activity() {
 
     private val mainHandler = Handler(Looper.getMainLooper())
@@ -29,6 +31,7 @@ class MainActivity : Activity() {
 
     var scenario: Scenario? = null
     var polling = false
+    var mazeAddress: String? = null
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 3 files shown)


features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -73,6 +76,33 @@ class MainActivity : Activity() {
...
 
+    private fun setMazeRunnerAddress() {
+        val context = MazerunnerApp.applicationContext()
+        val externalFilesDir = context.getExternalFilesDir(null)
+        val configFile = File(externalFilesDir, "fixture_config.json")
+        log("Attempting to read Maze Runner address from config file ${configFile.path}")
+
+        // Poll for the fixture config file
+        val pollStart = System.currentTimeMillis()
+        while (System.currentTimeMillis() - pollStart < CONFIG_FILE_TIMEOUT) {
```
lemnik, line 87: nit: assigning an "end time" avoids the arithmetic in the polling check

```suggestion
        val pollEnd = System.currentTimeMillis() + CONFIG_FILE_TIMEOUT
        while (System.currentTimeMillis() < pollEnd) {
```

features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -73,6 +76,33 @@ class MainActivity : Activity() {
...
+            if (configFile.exists()) {
+                val fileContents = configFile.inputStream().use { it.reader().readText() }
```
lemnik, line 89: Wrapping an `InputStream` in a `Reader` negates the need for `use` here, as `InputStreamReader.close()` will close the underlying `InputStream`.

```suggestion
                val fileContents = configFile.inputStream().reader().readText()
```

features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -73,6 +76,33 @@ class MainActivity : Activity() {
...
+                val fixtureConfig = JSONObject(fileContents)
```
lemnik, line 90: While very unlikely, this could fail to parse the JSON file if it is still being written or is otherwise corrupt. In that case we would want to ignore the error and try again.

```suggestion
                val fixtureConfig = runCatching { JSONObject(fileContents) }.getOrNull()
```

features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
yousif-bugsnag: Not sure I follow - would I then need to close the steam explicitly? Could this be simplified even further?
```suggestion
                val fileContents = configFile.readText()
```

lemnik: `use` is shorthand for "close this when I'm done with it" - I assumed that is what you meant there.

Yes, you could 👍  I forgot they added that to `File` as well 😆

lemnik (approved):
LGTM

PR #1797, opened 2023-01-26
Add conditional execution of e2e test on BitBar

twometresteve:
## Goal

Lay the tracks for regularly running e2e tests on BitBar.

## Design

Work to make the e2e tests and their fixture app agnostic of device farm is ongoing.  This change makes it easy to run tests on BitBar by either putting `[bitbar]` in a commit message or by setting the `BITBAR` pipeline variable.  By default tests will only be run on BrowserStack and the extra steps are in a separate pipeline yaml file to avoid polluting the main one too much.

## Changeset

- New pipeline steps added in a separate file
- Buildkite plugin version bumped
- New Docker Compose service for running on BitBar
- Maze Runner command responses now set conditionally depending on the environmental context.

## Testing

That running on BrowserStack is unaffected is verified by standard CI.  I've run some informal tests as part of development to see the new steps running on BitBar.  They are not expected to pass as yet.


docker-compose.yml
```diff
@@ -13,9 +13,7 @@ services:
     environment:
       BROWSER_STACK_USERNAME:
       BROWSER_STACK_ACCESS_KEY:
-      SAUCE_LABS_USERNAME:
```
twometresteve, line 16: We removed support for Sauce Labs in Maze Runner some time ago and do not have an account with them in any case.

docker-compose.yml
```diff
@@ -48,14 +46,39 @@ services:
...
-      SAUCE_LABS_ACCESS_KEY:
       TEST_FIXTURE_SYMBOL_DIR:
     volumes:
       - ./build:/app/build
       - ./features/:/app/features/
       - ./maze_output:/app/maze_output
 
+  android-maze-runner-bitbar:
+    image: 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:tms-remove-tms-cli
```
twometresteve, line 56: Ideally I should change this to a proper release before we can merge this.

yousif-bugsnag (approved):

PR #1802, opened 2023-02-07
Check ConnectivityManager for network status

lemnik:
## Goal
Make MazeRunner slightly less network-chatty by checking with the `ConnectivityManager` for the current internet status instead of trying to ping external services.

## Testing
Relied on existing tests.


twometresteve (approved):
LGTM

PR #1803, opened 2023-02-07
Relax checks on device orientation

twometresteve:
## Goal

Relax checks on device orientation to avoid unnecessary flakes.

## Design

The device orientation in the test environment can be quite unreliable, as discussed here:
https://www.browserstack.com/docs/app-automate/appium/troubleshooting/app-orientation-issues

Rather rather battle against those factors, it seems more pragmatic to simply allow either portrait or landscape to be present in the field.

## Testing

Covered by a basic CI run.


lemnik (approved):
LGTM

PR #1807, opened 2023-02-10
Update manifest config scenario to be device farm agnostic

yousif-bugsnag:
## Goal

The Android test fixture’s manifest has references to `bs-local.com`.  We still want to test that the notifier can read all values from the manifest, but we need to make the scenario agnostic of device farm.

## Design

`LoadConfigurationFromManifestScenario` now records the endpoint values from the manifest as metadata in the notify call and switches to using the endpoint configuration that the scenario is initialised with.

Also updated `ConfigureStartupAnrScenario` as this was also relying on the hardcoded endpoints in the manifest. This scenario now sets the notify and sessions endpoint to be used on restart via `SharedPreferences`

## Testing

Covered by a CI run


twometresteve (approved):
LGTM and I've built the fixture locally to verify that both of the affected scenarios now run on bitbar.

PR #1808, opened 2023-02-13
Add IntRange annotations to applicable configuration so that Android Studio can warn users

hannah-smartbear:
Added an @IntRange annotation to any configuration options that had hard limits on integer ranges, so that Android Studio can warn users when they try to set a configuration option to a number outside the range of options without them having to read the code, documentation or experience a runtime warning.
Tested this by building the library with my changes locally and checking if the example app gave me a warning in android studio when I set a configuration option to be outside the range.


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
```
lemnik, line 5:
```suggestion
### Enhancements
```

lemnik (approved):
LGTM

PR #1812, opened 2023-02-26
Enable ANR test scenarios on BitBar

twometresteve:
## Goal

Enable ANR test scenarios on BitBar.

## Design

We discovered that the ANR scenarios were not working on BitBar due to the developer options.  They have now been updated and the ANR scenarios pass.

## Changeset

As part of the investigation, I changed the scenarios to go back to tapping the screen rather than pressing the back button.  I don't think it was strictly necessary, but have kept the change in having retested it like that.  I don't think there's really anything between tapping and using the back button anyhow. 

I've also tagged all the ANR scenarios with `@anr` for future convenience.

## Testing

Retested against BitBar on CI before raising this PR.

### DIFF
--- a/features/full_tests/auto_notify.feature
+++ b/features/full_tests/auto_notify.feature
@@ -20,6 +20,7 @@ Feature: Switching automatic error detection on/off for Unity
     Then Bugsnag confirms it has no errors to send
 
   @skip_android_8_1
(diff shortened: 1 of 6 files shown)


joshedney (approved):
LGMT!

PR #1816, opened 2023-03-10
Move Android 10 tests to BitBar

twometresteve:
## Goal

Move Android 10 tests to BitBar.

## Design

Both the smoke and full tests for Android 10 now run on BitBar.  I've moved the pipeline files back around so that BrowserStack/BitBar sit alongside each other rather than in a BS specific pipeline file.  The BB only file remains for the time being to make development a little easier.

## Changeset

I've also fixed a bug in the clearPersistentData routine so that native and session directories are cleared out.

## Testing

Covered by a full CI run.


lemnik (approved):
LGTM

PR #1819, opened 2023-03-15
Correct and tidy docker-compose.yml

twometresteve:
## Goal

Fixes the build by changing to the correct Maze Runner docker image.

## Changeset

I've also tidied up the compose file a little to avoid repetition and remove an unused service.

## Testing

Covered by a basic CI run (which covers both BB and BS).


lemnik (approved):
LGTM

PR #1822, opened 2023-03-23
Move Android 13 tests to BitBar

twometresteve:
## Goal

Move Android 13 tests to BitBar.

## Testing

Achieved 3 successful full builds while developing this PR.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -133,6 +133,60 @@ steps:
     concurrency_group: 'bitbar-app'
     concurrency_method: eager
 
+  - label: ':android: Android 13 NDK r21 end-to-end tests - batch 1'
+    depends_on: "fixture-r21"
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.full.yml
```diff
@@ -29,7 +29,7 @@ steps:
       queue: macos-12-arm
     command: './gradlew clean assembleRelease check --scan'
 
-  - label: ':android: Minimal fixture end-to-end tests'
+  - label: ':browserstack: Minimal fixture end-to-end tests'
```
twometresteve, line 32: The intention with these is so that we can see at a glance on the Buildkite results page which device farm a step runs on.  BrowserStack already have a custom emojicon, BitBar is on its way (so will just display as ":bitbar:" in the short term.

lemnik (approved):

PR #1826, opened 2023-03-27
Move Android 9 & 11 tests to BitBar

Cawllec:
## Goal

Move Android 9 & 11 tests to BitBar

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -29,7 +29,7 @@ steps:
       queue: macos-12-arm
     command: './gradlew clean assembleRelease check --scan'
 
-  - label: ':browserstack: Minimal fixture end-to-end tests'
+  - label: ':bitbar: Minimal fixture end-to-end tests'
     depends_on: "fixture-minimal"
     timeout_in_minutes: 30
     plugins:
@@ -52,7 +52,7 @@ steps:
     concurrency_group: 'bitbar-app'
     concurrency_method: eager
 
-  - label: ':browserstack: Debug fixture smoke tests'
+  - label: ':bitbar: Debug fixture smoke tests'
     depends_on: "fixture-debug"
     timeout_in_minutes: 30
     plugins:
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.full.yml
```diff
@@ -60,26 +62,28 @@ steps:
...
-    concurrency_group: 'browserstack-app'
+    concurrency: 25
+    concurrency_group: 'bitbar-app'
     concurrency_method: eager
 
   #
   # BitBar steps
   #
-  - label: ':bitbar: Android 10 NDK r21 end-to-end tests - batch 1'
+  - label: ':android: Android 9 NDK r21 end-to-end tests - batch 1'
```
twometresteve, line 86: I had put `:browserstack:` and `:bitbar:` in deliberately to make it easy to see at a glance on the Buildkite UI where tests have been run.  I've got a ticket in with Marketing to get `:bitbar:` added to the list of custom Buildkite emojis.

I think that merge conflicts probably caused this to happen - note to myself to ensure that we branch from existing branches if appropriate to avoid the situation.

twometresteve:
I think that conflicts have occurred in a merge/rebase and not been resolved correctly.  The diff that GitHub is presenting is impossible to follow, but checking out the branch locally I can see that in `pipeline.full.yml`:
- The `# BrowserStack steps` header appears twice and the first has BitBar steps below it
- BitBar steps are listed in the order Android 9,13,10,11
- Steps for Android 10 are still running on BrowserStack and should be removed

Also, the smoke tests for Android 9 (on BB) are missing from `pipeline.yml`.

twometresteve (changes requested):

twometresteve:
> I've fixed the headers and ordering - it looks like we already had coverage for Android 10 on both bitbar and browserstack, which I assumed was intentional. If not, I'll remove it. 

Ah, I see now - not intentional (I see now I missed the removal on the previous PR). If you could remove them from BS that would be great.

> We also didn't previously have Android 9 smoke tests on any platforms, but I'm happy to add those in if we want the extra coverage without the full test runs.

Oh, I hadn't realised that.  Perhaps the view was that the minimal/debug tests in the full pipeline was sufficient.  Let's see what @lemnik thinks about adding "standard" smoke tests for Android 9 to the basic pipeline, I'll go with his view.

PR #1829, opened 2023-04-12
Implement deterministic Build ID generator

lemnik:
## Goal
Use the `.dex` file signatures to generate a stable `buildUuid` if none is specified in either `Configuration` or `AndroidManifest.xml` metadata.

## Design
The new `DexBuildIdGenerator` will attempt to extract the 20 byte signature from the `classes.dex` file in the app `.apk` file. This is done without unpacking the `.dex` files using `java.util.zip.ZipFile`. Each `classes.dex` file signature is then XOR'ed together in numeric order (`classes.dex`, `classes2.dex`, `classes3.dex`, `classes(N).dex`) to create a Build ID that is stable for the bytecode being used in the application (similar to the `build_id` notes we extract from `.so` files in the NDK plugin).

## Changeset
This PR does *not* include using the new ID generator as a fallback to the existing configured ID (this will be in a separate PR to keep the tests easier to review).

## Testing
A new set of unit tests which assert expected values from a known `.dex` and `.apk` file resource.


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/DexBuildIdGenerator.kt
```diff
@@ -0,0 +1,108 @@
...
+import java.io.File
+import java.util.zip.ZipEntry
+import java.util.zip.ZipFile
+import kotlin.experimental.xor
+
+internal object DexBuildIdGenerator {
+    private const val MAGIC_NUMBER_BYTE_COUNT = 8
+    private const val CHECKSUM_BYTE_COUNT = 4
+    private const val SIGNATURE_START_BYTE = MAGIC_NUMBER_BYTE_COUNT + CHECKSUM_BYTE_COUNT
+    private const val SIGNATURE_BYTE_COUNT = 20
```
kstenerud, line 14:
```dogma
header     = magic & checksum & signature & ...;
magic      = "dex\[a]" & '0'~'9'{3} & "\[0]";
checksum   = uint(32,~);
signature  = uint(8,~){20};
```
😉

kstenerud (approved):

PR #1832, opened 2023-04-18
NDK and ANR detection warning messages

SmartbearYing:
## Goal

When NDK and ANR plugins are excluded, there will not warning messages in the log when application is initialised. 

## Changeset

Added condition to make sure ANR and NDR crash warning message shows correctly.

## Testing

Manual test


lemnik (approved):
LGTM - Pending Lints

PR #1834, opened 2023-04-21
Do not run any ANR scenarios on Samsungs

twometresteve:
## Goal

No not run any ANR scenarios on Samsungs.

## Design

Follow on to #1831, not all ANR scenarios were considered.

## Testing

Covered by CI.


lemnik (approved):
LGTM

PR #1843, opened 2023-05-10
Add Binary compatibility validator

SmartbearYing:
## Goal
Add binary compatibility validator to ensure backwards compatibility of the public API surface.

## Design
Added the [Kotlin compatibility validator](https://github.com/Kotlin/binary-compatibility-validator) plugin to the build process, and a new step in BuildKite that will fail if the API surface changes.

## Testing
Manually tested by introducing incompatible changes and checking for expected failure.


lemnik (approved):
LGTM

PR #1847, opened 2023-05-12
fix(minimalapp): updated gem versions for CVE-2022-46648

tomlongridge:
## Goal

Fix some reported CVEs in a gem of the test fixture. These are not production code, but fixing to squash the warnings.

## Changeset

Updated Gemfile.lock and manually required `git` < 1.13.0

## Testing

Used by danger, so relying on the reporting on this PR, which appears to work.


lemnik (approved):
LGTM

PR #1851, opened 2023-05-16
Avoid secondary crashes from background notify calls during a crash

lemnik:
## Goal
Avoid secondary crashes when multiple threads trigger unhandled exceptions simultaneously while a crash report is being built.

## Design
Instead of shutting down the `BackgroundTaskService` when handling a crash (in `notifyUnhandledException`) we now wait for any pending tasks to complete, which avoids secondary crashes due to `RejectedExecutionException`s being thrown.

## Testing
Due to the inherent race-conditions required to trigger these failures, this has been tested using a unit test where we can directly trigger the internals and assert that no failure takes place.


kstenerud (approved):

lemnik:
This approach will cause new threads to be allocated on executors (specifically the `internalReportExecutor`, and `defaultExecutor`) that have currently have no work. This is not an acceptable outcome for 99% of cases. More work will need to be done on this to ensure that new threads are not allocated just to await a single event.

PR #1852, opened 2023-05-19
Log notifier startup times from mazerunner

lemnik:
## Goal
Start gathering metrics for the notifier startup time using our MazeRunner fixtures.

## Design
Benchmarking the notifier startup time using normal tools is almost impossible and largely meaningless, as the `Bugsnag.start` method only has an effect once in an application run and creating thousands of `Client` objects has a cumulative performance hit that would not be realistic in user applications.

In order to gather meaningful startup times this PR instruments all of the calls to `Bugsnag.start` in the Mazerunner fixture, and reports them to the new `/metrics` endpoint (which is exported as `metrics.csv` at the end of the tests). This allows us to gather real startup times from physical devices running in a release configuration.

## Testing
Manually tested.

### DIFF
--- a/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/MazerunnerHttpClient.kt
+++ b/features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/MazerunnerHttpClient.kt
@@ -1,11 +1,9 @@
 package com.bugsnag.android.mazerunner
 
 import android.util.JsonWriter
-import java.io.ByteArrayOutputStream
 import java.io.StringWriter
 import java.net.HttpURLConnection
 import java.net.URL
-import java.security.MessageDigest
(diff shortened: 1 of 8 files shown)


features/fixtures/mazerunner/jvm-scenarios/src/main/java/com/bugsnag/android/mazerunner/MazerunnerHttpClient.kt
```diff
@@ -36,6 +42,37 @@ class MazerunnerHttpClient(private val logEndpoint: URL) {
...
+                    json.name(name).value(value)
+                }
+
+                json.endObject()
+            }
+
+            bytes.toByteArray()
+        }
+
+        // Mazerunner expects all of these requests to also have a Bugsnag-Integrity header
```
twometresteve, line 65: Not true as of 7.30.2

twometresteve (approved):

PR #1856, opened 2023-06-15
Support regexes for redacted keys and discard classes in android

SmartbearYing:
## Goal
Offer a more flexible way to specify discard classes and redacted keys in order to reduce the boiler-plate for customers with large numbers of either.

## Changeset
Redacted keys and discard classes are now a `Set<Pattern>` instead of `Set<String>` offering a more flexible API.

## Testing
Updated the existing tests to use the new API


bugsnag-android-core/src/main/java/com/bugsnag/android/ObjectJsonStreamer.kt
```diff
@@ -12,7 +13,7 @@ internal class ObjectJsonStreamer {
         internal const val OBJECT_PLACEHOLDER = "[OBJECT]"
     }
 
-    var redactedKeys = setOf("password")
+    var redactedKeys = setOf(Pattern.compile(".*password.*"))
```
lemnik, line 16: It's probably worth making this case insensitive, so that we also cover "Password", "passWord", etc.

```suggestion
    var redactedKeys = setOf(Pattern.compile(".*password.*", Pattern.CASE_INSENSITIVE))
```

lemnik (approved):
LGTM

PR #1859, opened 2023-06-28
Replace thread type to error type

SmartbearYing:
## Goal

Replace thread type with error type

## Changeset

Replaced thread type with error type and removed thread type

## Testing

Updated the existing tests to use only error type


bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorType.kt
```diff
@@ -5,6 +5,11 @@ package com.bugsnag.android
  */
 enum class ErrorType(internal val desc: String) {
 
+    /**
+     * A error captured from Android's JVM layer
+     */
+    EMPTY(""),
```
lemnik, line 11: I think we can rather make this a more useful / error related idea, even if the content and use remains the same.

```suggestion
     * An error with an unknown type or source
     */
    UNKNOWN(""),
```

lemnik (approved):
LGTM

PR #1861, opened 2023-07-04
Prevent device data collector may causing a rare app crash while calculating free memory

SmartbearYing:
## Goal

Prevent DeviceDataCollector crashing the app when the ActivityManager cannot be reached.

## Changeset

Added try and catch for throwable exception.

## Testing

Additional unit tests


CHANGELOG.md
```diff
@@ -4,9 +4,12 @@
 
 ### Bug fixes
 
-* Prevent rare app crash while migrating old `SharedPreferences` data from older versions of `bugsnag-android`
+* Prevent rare app crash while migrating old `SharedPreferences` data from older versions of `bugsnag-android`.
```
lemnik, line 7: nit: I don't think we need to change this CHANGELOG line ;)

```suggestion
* Prevent rare app crash while migrating old `SharedPreferences` data from older versions of `bugsnag-android`
```

CHANGELOG.md
```diff
@@ -4,9 +4,12 @@
...
   [#1860](https://github.com/bugsnag/bugsnag-android/pull/1860)
 
+* Prevent device data collector may causing a rare app crash while calculating free memory.
```
lemnik:
```suggestion
* Prevent free memory calculation from potentially crashing the app when `ActivityManager` cannot be reached
```

lemnik (approved):
LGTM

PR #1862, opened 2023-07-10
Thread id is more compatible

SmartbearYing:
## Goal

Thread id can handle both string and number, it will be more compatible to server and system setup.

## Changeset

Changed thread id type to string.

## Testing

Updated the existing tests


bugsnag-android-core/src/main/java/com/bugsnag/android/Thread.java
```diff
@@ -40,14 +40,15 @@ private void logNull(String property) {
     /**
      * Sets the unique ID of the thread (from {@link java.lang.Thread})
      */
-    public void setId(long id) {
+    public void setId(@NonNull String id) {
         impl.setId(id);
     }
```
lemnik, line 45: this hides a `NullPointerException` from the Kotlin side, so we need a `null` check:

```suggestion
    public void setId(@NonNull String id) {
        if (id != null) {
            impl.setId(id);
        } else {
            logNull("id");
        }
    }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -109,13 +109,13 @@ internal class ThreadState @Suppress("LongParameterList") constructor(
             // API 24/25 don't record the currentThread, so add it in manually
             // https://issuetracker.google.com/issues/64122757
             // currentThread may also have been removed if its ID occurred after maxThreadCount
-            keepThreads.take(Math.max(maxThreadCount - 1, 0)).plus(currentThread).sortedBy { it.id }
+            keepThreads.take((maxThreadCount - 1).coerceAtLeast(0)).plus(currentThread).sortedBy { it.id }
```
lemnik, line 112: nit: an `import kotlin.math.max` allows:

```suggestion
            keepThreads.take(max(maxThreadCount - 1, 0)).plus(currentThread).sortedBy { it.id }
```

lemnik (approved):
LGTM

PR #1863, opened 2023-07-11
Setting persist user default value to true

SmartbearYing:
## Goal

Reset persist user default value to match with other platforms.

## Changeset

Changed persist user default value to true

## Testing

Updated the existing tests


lemnik (approved):
LGTM

PR #1865, opened 2023-07-12
Stable Build ID Generator

lemnik:
## Goal
Generate a stable Build ID value if one is not provided in the `AndroidManifest.xml`

## Design
When a Build ID cannot be found in the `AndroidManifest.xml` the startup process will inspect the `classes.dex` files and extract the checksums in order to generate a Build ID that can be tied to the bytecode. This allows us to identify specific bytecode deployments against source maps without needing to inject logic into the build process.

## Testing
New unit tests introduced.


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 * Added `Session.apiKey` so that it can be changed in an `OnSessionCallback`
   [#1855](https://github.com/bugsnag/bugsnag-android/pull/1855)
+* Generate a Build ID value if one is not provided in the AndroidManifest.xml
+  []()
```
tomlongridge, line 10:
```suggestion
  [#1865](https://github.com/bugsnag/bugsnag-android/pull/1865)
```

tomlongridge (commented):
LGTM - will hold off approving for another set of eyes.

PR #1866, opened 2023-07-13
Support regexes for redacted keys and discard classes in android

SmartbearYing:
## Goal
Offer a more flexible way to specify discard classes and redacted keys in order to reduce the boiler-plate for customers with large numbers of either.

## Changeset
Redacted keys and discard classes are now a `Set<Pattern>` instead of `Set<String>` offering a more flexible API.

## Testing
Updated the existing tests to use the new API


lemnik (approved):
LGTM

PR #1868, opened 2023-07-17
Apikey validation check after bugsnag start

SmartbearYing:
## Goal

Prevent exception occurs before bugsnag start.

## Changeset

Apikey validation check will be carried on after bugsnag start in ImmutableConfig when sanitising the configuration

## Testing

Updated the existing tests


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -181,6 +182,34 @@ internal fun convertToImmutableConfig(
     )
 }
 
+private fun validateApiKey(value: String) {
+    if (isInvalidApiKey(value)) {
+        w(
```
lemnik, line 187: This should log to the configured `Logger` rather than always directing to the `DebugLogger`

```suggestion
        logger.w(
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -181,6 +182,34 @@ internal fun convertToImmutableConfig(
...
+        throw IllegalArgumentException("No Bugsnag API Key set")
+    }
+    if (apiKey.length != VALID_API_KEY_LEN) {
+        return true
+    }
+    // check whether each character is hexadecimal (either a digit or a-f).
+    // this avoids using a regex to improve startup performance.
+    for (k in 0 until VALID_API_KEY_LEN) {
+        val chr = apiKey[k]
+        if (!Character.isDigit(chr) && (chr < 'a' || chr > 'f')) {
```
lemnik, line 206: nit:

```suggestion
        if (!chr.isDigit() && chr !in 'a'..'f') {
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -193,6 +222,8 @@ internal fun sanitiseConfiguration(
         packageManager.getApplicationInfo(packageName, PackageManager.GET_META_DATA)
     }.getOrNull()
 
+    validateApiKey(configuration.apiKey, configuration.logger!!)
```
SmartbearYing, line 225: @lemnik I am not sure about this part. Since the apiKey does need the validation check so the configuration.logger can not be null. But this might throw an exception.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -181,6 +182,34 @@ internal fun convertToImmutableConfig(
     )
 }
 
+private fun validateApiKey(value: String, logger: Logger?) {
+    if (isInvalidApiKey(value)) {
+        logger?.w(
+            "Invalid configuration. " +
+                "apiKey should be a 32-character hexademical string, got " + value
```
lemnik, line 189: I think we should be using `DebugLogger` to avoid hiding the log message when a customer or no-op `Logger` is being used.

```suggestion
private fun validateApiKey(value: String) {
    if (isInvalidApiKey(value)) {
        DebugLogger.w(
            "Invalid configuration. apiKey should be a 32-character hexademical string, got $value"
```

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -208,9 +237,13 @@ internal fun sanitiseConfiguration(
...
         if (loggingEnabled) {
             configuration.logger = DebugLogger
+            validateApiKey(configuration.apiKey, configuration.logger as DebugLogger)
         } else {
             configuration.logger = NoopLogger
+            validateApiKey(configuration.apiKey, configuration.logger as NoopLogger)
         }
+    } else {
+        validateApiKey(configuration.apiKey, configuration.logger)
     }
```
lemnik, line 247: I think that the `validateApiKey` can be moved out of the `if` statement

lemnik (changes requested):

bugsnag-android-core/dsl-json
lemnik, line 1: I don't think that this was included intentionally?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -181,6 +181,33 @@ internal fun convertToImmutableConfig(
...
+    }
+    // check whether each character is hexadecimal (either a digit or a-f).
+    // this avoids using a regex to improve startup performance.
+    for (k in 0 until VALID_API_KEY_LEN) {
+        val chr = apiKey[k]
+        if (!chr.isDigit() && chr !in 'a'..'f') {
+            return true
+        }
+    }
+    return false
```
lemnik, line 208: Can't believe I didn't see this last time around 🙈 

```suggestion
    return !apiKey.all { it.isDigit() || it in 'a'..'f' }
```

lemnik (changes requested):

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/ImmutableConfig.kt
```diff
@@ -211,6 +232,8 @@ internal fun sanitiseConfiguration(
         } else {
             configuration.logger = NoopLogger
         }
+    } else {
+        validateApiKey(configuration.apiKey)
     }
```
lemnik, line 237: This shouldn't be an in `else` since as things stand, we are only validating the API-Key when we don't configure the logger

lemnik (changes requested):

lemnik (approved):
LGTM

PR #1873, opened 2023-07-25
Revert Example appcompile & target SDK

lemnik:
## Goal
Avoid version conflicts on CI.

## Changelog
Reverted the Example app to using SDK 31 instead of 34


joshedney (approved):
LGTM

PR #1888, opened 2023-08-08
CI improvements

twometresteve:
## Goal

Improve some aspects of CI:
- Bump Maze Runner to v8
- Add standardised logging for specific events
- Only upload each test fixture to BitBar once

## Design

Logging the string "bugsnag info" at INFO level looks like redundant information in an isolated context - but it means that I can go to any repo and just look for the same string for things that I care about.

## Testing

Covered by a full CI run.


lemnik (approved):
LGTM

PR #1893, opened 2023-08-23
Correlate ExitInfo with Event by session ID

SmartbearYing:
## Goal

Match the android exit information with the event in a onSend callback

## Design

Once we match the android exit information with the event, the plug in will allow user to get more details about the android exit reasons.

## Changeset

Over write onSend callback to match android exit information with captured event content

## Testing

Introduced new tests


bugsnag-plugin-android-exitinfo/detekt-baseline.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>MagicNumber:ExitInfoCallback.kt$ExitInfoCallback$100</ID>
```
lemnik, line 5: This shouldn't be required anymore

bugsnag-plugin-android-exitinfo/src/main/java/com/bugsnag/android/ExitInfoCallback.kt
```diff
@@ -0,0 +1,30 @@
...
+            if (event.session?.id?.let { exitInfo.processStateSummary?.contentEquals(it.toByteArray()) } == true) {
+                eventEnhancer(event, exitInfo)
+                return@onSend true
+            }
+        }
+        return true
+    }
+
+    internal companion object {
+        const val MAXIMUM_NUMBER = 100
```
lemnik, line 28: A more descriptive name would be good:

```suggestion
        const val MAX_EXIT_INFO = 100
```

bugsnag-plugin-android-exitinfo/src/main/java/com/bugsnag/android/ExitInfoCallback.kt
```diff
@@ -0,0 +1,30 @@
...
+    @RequiresApi(Build.VERSION_CODES.R)
+    override fun onSend(event: Event): Boolean {
+        val am = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
+        val allExitInfo = am.getHistoricalProcessExitReasons(context.packageName, 0, MAXIMUM_NUMBER)
+        allExitInfo.forEach { exitInfo ->
+            if (event.session?.id?.let { exitInfo.processStateSummary?.contentEquals(it.toByteArray()) } == true) {
+                eventEnhancer(event, exitInfo)
+                return@onSend true
+            }
+        }
```
lemnik, line 23: I think this might be better expressed as a `find`?

```suggestion
        val sessionIdBytes = event.session?.id?.toByteArray() ?: return true
        val exitInfo = allExitInfo.find { it.processStateSummary?.contentEquals(sessionIdBytes) == true }
           ?: return true
        
        eventEnhancer(event, exitInfo)
```

bugsnag-plugin-android-exitinfo/src/test/java/com/bugsnag/android/ExitInfoCallbackTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+@RunWith(MockitoJUnitRunner::class)
+internal class ExitInfoCallbackTest {
+
+    private lateinit var exitInfoCallback: ExitInfoCallback
+
+    @Mock
+    private lateinit var context: Context
+
+    private val eventEnhancer: (Event, ApplicationExitInfo) -> Unit = { _, _ -> }
```
lemnik, line 24: Using a mocked `eventEnhancer` would allow the tests to `verify` that it was called when the session IDs match up

bugsnag-plugin-android-exitinfo/detekt-baseline.xml
lemnik, line 1: Given that this is effectively empty, I think it's better if we remove it.

lemnik (approved):
LGTM

PR #1895, opened 2023-09-01
Remove event on-disk migration code

kstenerud:
## Goal

Remove the NDK event file migration code.

## Testing

Reran all tests (the migration tests have been replaced with a single store/load test)


kstenerud:
Reopened as https://github.com/bugsnag/bugsnag-android/pull/1898

PR #1896, opened 2023-09-04
Integrate Protocol Buffers to parse tombstone files

SmartbearYing:
## Goal

Integrate Protocol Buffers to parse tombstone files

## Changeset

Integrate the Protocol Buffers compiler plugin and `javalite` dependency to allow the parsing of tombstone files in the exitinfo plugin

## Testing

Introduced new unit tests


bugsnag-plugin-android-exitinfo/build.gradle
```diff
@@ -10,4 +11,30 @@ dependencies {
...
+        }
+    }
+}
+
+android {
+    compileOptions {
+        sourceCompatibility 1.8
+        targetCompatibility 1.8
+    }
+}
```
lemnik, line 38: This should rather be changed in build plugin

bugsnag-plugin-android-exitinfo/proguard-rules.pro
lemnik, line 1: This rule should be in the consumer proguard rules so that it's applied to consumer applications. It should also be made more specific so that it only affects classes that the exitinfo plugin generates.

bugsnag-plugin-android-exitinfo/src/main/proto/tombstone.proto
```diff
@@ -0,0 +1,170 @@
+//
+// Protobuf definition for Android tombstones.
+//
+// An app can get hold of these for any `REASON_CRASH_NATIVE` instance of
+// `android.app.ApplicationExitInfo`.
+//
+// https://developer.android.com/reference/android/app/ApplicationExitInfo#getTraceInputStream()
+//
+syntax = "proto3";
+option java_package = "com.android.server.os";
```
lemnik, line 10: I think it might be worth renaming this package to avoid collisions:
```suggestion
option java_package = "com.bugsnag.android.repackaged.server.os";
```

lemnik (approved):
LGTM

PR #1897, opened 2023-09-05
Merge threads from trace files

lemnik:
## Goal
Merge threads parsed from trace files into the associated `Event`s

## Testing
New unit test for the event enhancer


SmartbearYing (approved):
LGTM

PR #1901, opened 2023-09-13
Enhance events with tombstone and trace files in the BugsnagExitinfo plugin

SmartbearYing:
## Goal
Merge tombstone & trace data into BugSnag `Event`s when available.

## Changeset
Introduced `BugsnagExitInfoPlugin` to correlate `ApplicationExitInfo` data with BugSnag `Event`s before they are sent.

## Testing
Introduce new unit test and manual test on example app

### DIFF
--- a/bugsnag-plugin-android-exitinfo/src/main/java/com/bugsnag/android/BugsnagExitInfoPlugin.kt
+++ b/bugsnag-plugin-android-exitinfo/src/main/java/com/bugsnag/android/BugsnagExitInfoPlugin.kt
@@ -6,8 +6,24 @@ import android.os.Build
 import androidx.annotation.RequiresApi
 
 @RequiresApi(Build.VERSION_CODES.R)
-class BugsnagExitInfoPlugin(
-    private val context: Context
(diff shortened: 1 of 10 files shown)


bugsnag-plugin-android-exitinfo/src/main/java/com/bugsnag/android/BugsnagExitInfoPlugin.kt
```diff
@@ -0,0 +1,53 @@
+package com.bugsnag.android
+
+import android.app.ActivityManager
+import android.content.Context
+import android.os.Build
+import androidx.annotation.RequiresApi
+
+@RequiresApi(Build.VERSION_CODES.R)
+class BugsnagExitInfoPlugin(
```
lemnik, line 9: To keep this easy to use from Java, we should add `@JvmOverloads` to the constructor
```suggestion
class BugsnagExitInfoPlugin @JvmOverloads constructor(
```

lemnik (approved):
LGTM

PR #1902, opened 2023-09-14
Pre on send callback mechanism

SmartbearYing:
## Goal

Internal API for for first-party OnSend callbacks.

## Changeset

Move onSend to package protected and prior to the callbacks

## Testing

Introduce new unit test and manual test on example app


lemnik (approved):
LGTM

PR #1908, opened 2023-09-28
Remove Warning log when persisted user file not loaded

SmartbearYing:
## Goal

Warning log would not show when persisted user file is not loaded

## Changeset

Add persisted user file exist checks and config persist user check in when loading user

## Testing

Manual test and refactor existing tests

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/UserStore.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/UserStore.kt
@@ -3,7 +3,6 @@ package com.bugsnag.android
 import com.bugsnag.android.internal.ImmutableConfig
 import com.bugsnag.android.internal.StateObserver
 import java.io.File
-import java.io.IOException
 import java.util.concurrent.atomic.AtomicReference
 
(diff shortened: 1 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/UserStore.kt
```diff
@@ -87,13 +81,19 @@ internal class UserStore @JvmOverloads constructor(
...
+            synchronizedStreamableStore.file.canRead() &&
+            synchronizedStreamableStore.file.length() > 0L &&
+            persist
+        ) {
+            try {
                 synchronizedStreamableStore.load(User.Companion::fromReader)
-            } catch (exc: Exception) {
-                logger.w("Failed to load user info", exc)
+            } catch (e: Exception) {
+                logger.w("Failed to load user info")
```
lemnik, line 92:
```suggestion
            } catch (exc: Exception) {
                logger.w("Failed to load user info", exc)
```

lemnik (approved):
LGTM

PR #1911, opened 2023-10-03
Session payload versions

SmartbearYing:
## Bug

V3 session payload is delivered as V1 session payload

## Changeset

V3 session payload will be processed as V2 session payload and be delivered as V3 session payload

## Testing

Existing unit tests

PR #1913, opened 2023-10-11
v5.31.2

SmartbearYing:
### Bug fixes

* V3 session payload will be processed and delivered as non legacy payload.
  [#1911](https://github.com/bugsnag/bugsnag-android/pull/1911)


lemnik (approved):
LGTM

PR #1915, opened 2023-10-11
New foreground detector

lemnik:
## Goal
Replace the use of `ActivityManager.importance` with `Activity` based foreground tracking.

## Design
Replaced the `ForegroundDetector` with a new `internal object` that implements foreground tracking using `ActivityLifecycleCallbacks`. This allows us to determine app foreground status without waiting for IPC calls and also aligns the "inForeground" field with [bugsnag-android-performance](https://github.com/bugsnag/bugsnag-android-performance) and [androidx.lifecycle.ProcessLifecycleOwner](https://developer.android.com/reference/androidx/lifecycle/ProcessLifecycleOwner).

## Testing
Relied on existing tests all passing


tomlongridge (commented):
LGTM

SmartbearYing (approved):
LGTM

PR #1916, opened 2023-10-12
Move foreground logic out of SessionTracker

lemnik:
## Goal
Move the foreground/background tracking logic from `SessionTracker` to the new `ForegroundDetector` singleton.

## Changes
Moved the `lastExitedForegroundMs` and `lastEnteredForegroundMs` timestamps out of `SessionTracker` and into `ForegroundDetector`.

This PR also reworks the "background" trigger slightly to avoid the 700ms timeout when the last `Activity` stops (where possible). Now we use [isChangingConfigurations](https://developer.android.com/reference/android/app/Activity#isChangingConfigurations()) when an `Activity` stops to detect whether we expect the `Activity` to (almost) immediately restart. If this is not the case we send the "inBackground" event immediately.

## Testing
Modified existing tests.


SmartbearYing (approved):
LGTM

PR #1927, opened 2023-11-06
v5.31.3

SmartbearYing:
### Bug fixes

* Handle cases in NDK when `GetStringUTFChars` fails for callback names
  [#1918]


lemnik (approved):
LGTM

PR #1932, opened 2023-11-09
Clear files between e2e test scenarios

twometresteve:
## Goal

Clear additional files between e2e test scenarios that, if left, can cause scenarios to bleed into each other.

## Design

A forthcoming release of Maze Runner will remove use of the deprecated driver `reset` method.  This means that test fixtures and scenarios will need to take responsibility for 

## Testing

Covered by CI - and I've also run the affected test against an in-development version of Maze Runner that doesn't call `reset`.


lemnik (approved):
LGTM

PR #1935, opened 2023-11-14
Time limitation for thread collection

SmartbearYing:
## Goal

Users is able to set a maximum time to avoid the timeout while collecting threads

## Changeset

Adding a maximum traces time variable in config that allow setting a time limitation while capturing threads.

## Testing

Introduce new unit test and existing unit tests


bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -109,16 +113,24 @@ internal class ThreadState @Suppress("LongParameterList") constructor(
...
         // Note: We must ensure that currentThread is always present in the final list regardless.
+
+        val reportThreads = mutableListOf<Thread>()
+        val timeout = SystemClock.elapsedRealtime() + maxTime
         val keepThreads = allThreads.sortedBy { it.id }.take(maxThreadCount)
+        keepThreads.forEach {
+            if (SystemClock.elapsedRealtime() < timeout) {
+                reportThreads.add(toBugsnagThread(it))
+            }
+        }
```
lemnik, line 124: this might be better expressed as a `for` loop with a `break`?

```suggestion
        for(thread in keepThreads) {
            if (SystemClock.elapsedRealtime() >= timeout) {
                break
            }
            reportThreads.add(toBugsnagThread(it))
        }
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -78,6 +81,7 @@ internal class ThreadState @Suppress("LongParameterList") constructor(
         exc: Throwable?,
         isUnhandled: Boolean,
         maxThreadCount: Int,
+        maxTime: Long,
```
lemnik, line 84: naming: this could be clearer

```suggestion
        stacktraceCollectionTimeout: Long,
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ThreadState.kt
```diff
@@ -109,16 +115,41 @@ internal class ThreadState @Suppress("LongParameterList") constructor(
...
+        val currentThreadIndex = sortedThreads.binarySearch(0, min(maxThreadCount, sortedThreads.size)) {
+            it.id.compareTo(currentThread.id)
+        }
+
+        // API 24/25 don't record the currentThread, so add it in manually
+        // https://issuetracker.google.com/issues/64122757
+        // currentThread may also have been removed if its ID occurred after maxThreadCount
+        val keepThreads = sortedThreads.take(
+            if (currentThreadIndex >= 0) maxThreadCount else max(maxThreadCount - 1, 0)
+        )
```
lemnik, line 129: Given the change in logic, I think we should enhance the comment to explain the `maxThreadCount - 1`:

```suggestion
        // API 24/25 don't record the currentThread, so add it in manually
        // https://issuetracker.google.com/issues/64122757
        // currentThread may also have been removed if its ID occurred after maxThreadCount
        // as such we may need to leave a space in new list for currentThread
        val keepThreads = sortedThreads.take(
            if (currentThreadIndex >= 0) maxThreadCount else max(maxThreadCount - 1, 0)
        )
```

bugsnag-android-core/src/main/java/com/bugsnag/android/Configuration.java
lemnik, line 1: There is automatic formatting which has inserted far to much into this diff (all of the `<p>` elements).

lemnik (approved):
LGTM

PR #1940, opened 2023-11-15
Feature flags are modified in-place.

SmartbearYing:
## Goal

Feature flags position will be in the place when it is added.

## Changeset

Feature flags would not be added to the end of the collection when it is modified.

## Testing
Exiting unit tests


lemnik (approved):
LGTM

PR #1941, opened 2023-11-16
Corrections to fix v6 publishing

lemnik:
## Goal
Fix the publishing step by removing the legacy `bugsnag-android-ndk` module (which was previously replaced by `bugsnag-plugin-android-ndk`).

Also removed the flaky "StrictMode Activity leak violation" scenario as it flakes on Android 7 more than it passes, and does not assert that the notifier is working as expected. Instead it was only running on Android 7 & 8, and asserted that the StrictMode violation was *not* being caught by the notifier (as the notifier is incapable of reporting that violation on those platform versions).

The scenario flaked as it has a hard-dependency on the device being able to open a "leaked" file by Intent. When no third-party app is installed to action the leaking intent the scenario flaked. Handling this case within the scenario would exactly mimic the *expected* behavior of the scenario (negating the test completely).

## Testing
Relied on existing testing.

### DIFF
--- a/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ConfigSerializer.java
+++ b/bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ConfigSerializer.java
@@ -16,7 +16,6 @@ class ConfigSerializer implements MapSerializer<ImmutableConfig> {
         map.put("autoDetectErrors", config.getAutoDetectErrors());
         map.put("autoTrackSessions", config.getAutoTrackSessions());
         map.put("sendThreads", config.getSendThreads().toString());
-        map.put("discardClasses", config.getDiscardClasses());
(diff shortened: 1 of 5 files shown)


tomlongridge (approved):

bugsnag-plugin-react-native/src/main/java/com/bugsnag/android/ConfigSerializer.java
```diff
@@ -2,11 +2,14 @@
...
 import com.bugsnag.android.internal.ImmutableConfig;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
```
tomlongridge, line 12: I imagine these new imports aren't required?

PR #1942, opened 2023-11-16
Update workflow to use auth provided by bearer token

Cawllec:
## Goal

Fixes auth by using explicit variable references rather than environment variables, and uses a given bearer token instead of the default.


twometresteve (approved):
Seems fine to me if we know it works?

PR #1951, opened 2023-12-05
Convert Event to Kotlin and remove EventInternal

SmartbearYing:
## Goal

Convert Event.java to kotlin class and integrate EventInternal to Event

## Testing
Existing tests


lemnik:
The current effect on the public API surface means this PR currently introduces more problems than it really solves. Closing until this can be addressed properly.

PR #1952, opened 2023-12-05
Time limitation for thread collection (V5 backport)

SmartbearYing:
## Goal

Users is able to set a maximum time to avoid the timeout while collecting threads

## Changeset

Adding a maximum traces time variable in config that allow setting a time limitation while capturing threads.

## Testing

Introduce new unit test and existing unit tests


lemnik (approved):
LGTM

PR #1958, opened 2023-12-15
Handle NaN & Infinity values in serialization

lemnik:
## Goal
Avoid creating invalid / incomplete event payloads when metadata includes non-finite values (typically via ReactNative).

## Design
The non-finite double values (`NaN`, `-Infinity` and `+Infinity`) are now omitted (along with any associated keys) instead of throwing an `IllegalArgumentException`

## Testing
Adjusted the existing unit tests.


tomlongridge (approved):

SmartbearYing (approved):
LGTM

PR #1962, opened 2024-01-15
Update proguard rules for React Native plugin

yousif-bugsnag:
## Goal

React Native 0.73 has updated AGP to 8.x, which now runs R8 in "full mode" by default. This causes RN projects with Bugsnag to crash, as the current proguard rule for `BugsnagReactNativePlugin` (`-keepnames`) is not sufficient to prevent the class from being obfuscated.

## Design

Updated the proguard rule to use `-keep` instead of `-keepnames`

## Testing

Tested manually on a React Native project. Test fixtures for `@bugsnag/react-native` will also be updated to enable Proguard once this fix is released.


lemnik (approved):
LGTM

PR #1966, opened 2024-01-26
Modified-utf8 decoder

lemnik:
## Goal
Implement a Modified-UTF8 decoder for `ByteBuffer`s so that strings stored by the C layer can be directly decoded by the Kotlin layer.

## Design
Followed the Java modified-utf8 decode with some small modifications:

- null-bytes (zero) are considered an early-exit to comply with C-string expectations
- the returned strings do not contain trailing null-bytes
- invalid character sequences cause a UTF replacement character to be used instead of throwing an exception

## Testing
New unit test with various example strings and failure modes


kstenerud (approved):

SmartbearYing (approved):
LGTM

PR #1969, opened 2024-01-31
Flush command to mazerunner fixture

SmartbearYing:
## Goal

Added command to mazerunner fixture for flushing all the stored event files.

## Changeset

Added a `Flush` class to mazerunner package for accessing eventStore flushAsync method

### DIFF
--- a/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -11,6 +11,7 @@ import android.os.Looper
 import android.view.Window
 import android.widget.Button
 import android.widget.EditText
+import com.bugsnag.android.BugsnagInternals
 import com.bugsnag.android.mazerunner.scenarios.Scenario
 import org.json.JSONObject
 import java.io.File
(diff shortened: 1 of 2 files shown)


features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/Flush.java
```diff
@@ -0,0 +1,8 @@
+package com.bugsnag.android;
+
+public class Flush {
+    public void flush() {
```
lemnik, line 4: This class should be named more clearly, and the `flush` method can be `static` so that we don't need to create object instances to call `flush()`

```suggestion
public class BugsnagInternals {
    private BugsnagInternals() {}
    
    public static void flush() {
```

features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -177,6 +178,9 @@ class MainActivity : Activity() {
                                 runScenario(scenarioName, scenarioMode, sessionsUrl, notifyUrl)
                             }
                             "clear_persistent_data" -> clearPersistentData()
+                            "flush" -> {
+                                Flush().flush()
+                            }
```
lemnik, line 183:
```suggestion
                            "flush" -> BugsnagInternals.flush()
```

PR #1972, opened 2024-02-05
Added BugsnagExitInfoPlugin  in to minimal test fixture 

SmartbearYing:
Added BugsnagExitInfoPlugin to  MINIMAL_FIXTURE so in minimal test fixture can capture it


lemnik (approved):
LGTM

PR #1980, opened 2024-02-14
Add missing test expectations

twometresteve:
## Goal

Adds two checks that were missing from a test scenario.

## Design

The checks are essential to the test, which is designed to verify that the error report context reflects the Activity that is changed to.

## Testing

Covered by a basic CI run.


SmartbearYing (approved):
LGTM

PR #1982, opened 2024-02-16
Decode user info to event

SmartbearYing:
## Goal

Decoder for bsg_user_info to event 

## Changeset

Moving decoding user info process to NativeEventDecoder in Kotlin

## Testing

Introduced new tests


lemnik (approved):
LGTM

PR #1983, opened 2024-02-16
Update e2e tests to Maze Runner v9

twometresteve:
## Goal

Upgrade CI tests to use Maze Runner v9.

## Changeset

- Pinned to Appium v1, as a number of complexities and the need for workarounds had become evident with v2.
- Discovered that the persistence folders that need to be cleared have changed.
- Found a feature file that wasn't clearing persistent data (now needed before every scenario).

## Testing

Covered by a full CI run (run immediately before creating this PR).


lemnik (approved):
LGTM

PR #1988, opened 2024-02-28
Avoid possible crashes in the ANR handler where the local reference table is exhausted

lemnik:
## Goal
Avoid possible crashes in the ANR handler where the local reference table is exhausted.

## Design
Replaced the use of `DeleteLocalRef` with `PushLocalFrame`/`PopLocalFrame` to simplify the handling of the local references.

## Testing
Manually tested and relied on existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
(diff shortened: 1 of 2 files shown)


kstenerud (approved):
Me likes!!!

YYChen01988 (approved):
LGTM

PR #1989, opened 2024-03-05
Adding generateAnonymousId config option

hannah-smartbear:
CHANGES

Added generateAnonymousId config option, which, when set to false:
- Does not use a previously persisted device ID if there is one
- Does not generate or store a new device ID
- However, it also does not wipe any device ID that has been previously stored. I’m not sure if this is something I should add?

I implemented this by returning null early in loadDeviceId function in DeviceIdStore.kt if generateAnonymousId is set to false. This seemed the easiest way to do this with the least amount of files changed.

Where UserStore.kt tries to use the deviceId as the user ID, deviceId should be null anyway if generateAnonymousId config option is false, due to my changes to loadDeviceId. However, there’s one scenario where there the current behaviour is perhaps not ideal:
- Let’s say a customer previously had default config options for persistUser and generateAnonymousId (i.e. true for both), and they are not setting user info themselves so both device ID and user ID are set to the device ID value
- The customer then decides they don’t want that Device ID value reported on events, so looks at the docs and concludes that they need to set generateAnonymousId to false.
- However, since they still have persistUser = true, the device ID will actually still appear on their events in the user tab, but we won’t necessarily know not to add this to events since we don’t know that the user ID stored on the device is actually the device ID and that we shouldn’t use it
So I’m not sure if there’s anything we can do to prevent this happening, or if we should just include a note in the docs to make this clear? We couldn’t just not return the persisted user if generateAnonymousId is false, since the persisted user ID may be something set by the customer and not the device ID.

TESTING

- I tested this locally by publishing my changes to Maven local and setting the generateAnonymousId config option in my example app and triggering events
- I added a unit test to DeviceIdStoreTest.kt to test that a previously persisted device ID is not loaded when generateAnonymousId is set to false
- I was also thinking that I would like to add a test to check that no new device ID is generated or persisted when generateAnonymousId is set to false, but wasn’t too sure how to do this

### DIFF
--- a/bugsnag-plugin-android-ndk/src/test/cpp/test_serializer.c
+++ b/bugsnag-plugin-android-ndk/src/test/cpp/test_serializer.c
@@ -50,7 +50,7 @@ void loadDeviceTestCase(bugsnag_event *event) {
     strcpy(device->orientation, "portrait");
 
     struct tm time = { 0, 0, 0, 1, 12, 128 };
-    device->time = mktime(&time);
+    device->time = timegm(&time);
     strcpy(device->id, "f5gh7");
     device->jailbroken = true;
     strcpy(device->locale, "En");


bugsnag-android-core/src/androidTest/java/com/bugsnag/android/DeviceIdStoreTest.kt
```diff
@@ -237,4 +252,26 @@ internal class DeviceIdStoreTest {
...
         assertEquals(prefDeviceId, storeDeviceId)
     }
+
+    /**
+     * If generateAnonymousId is false, no device ID is returned (even if one is saved)
+     */
+    @Test
+    fun loadWithoutGenerateAnonymousId() {
+        file.writeText("{\"id\": \"${ids[0]}\"}")
+        fileInternal.writeText("{\"id\": \"${ids[1]}\"}")
```
lemnik, line 262: these look like a good candidate for raw string literals:

```suggestion
        file.writeText("""{"id": "${ids[0]}"}""")
        fileInternal.writeText("""{"id": "${ids[1]}"}""")
```

lemnik (approved):
LGTM - With a small suggestion

PR #1993, opened 2024-03-14
Rescue error from Appium press operation during ANRs

twometresteve:
## Goal

Prevent test runs from failing due to timeouts with tap/press operations during an ANR.  The press seems to work anyway.

## Changeset

I've also changed `tap` to `press` as the Appium docs seemed to suggest that it works in more cases.

## Testing

Covered by CI.


lemnik (approved):
LGTM

PR #2005, opened 2024-04-05
Don't copy FeatureFlags on the crashing thread

lemnik:
## Goal
Avoid making defensive copies of the feature flags on the notifying thread (unless they are altered in an `OnErrorCallback`).

## Design
Replaced the `LinkedHashMap` with a Copy-On-Write data structure. To avoid multiple levels of locking, and the need for iterators this is a simple encapsulated array which is copied each time a change is required.

The new implementation is slower when adding & removing feature flags (since each addition or removal is now O(n) rather than O(1)), but since there are relatively few feature flags added per run (<1000) this should be a reasonable trade-off as now the feature flags do not need *any* synchronization or defensive copying when an error report is required.

## Testing
Additional tests for the new implementation and existing tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
+* FeatureFlags are now a copy-on-write structure, and so don't need to be defensive copied on a crashing thread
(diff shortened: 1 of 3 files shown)


kstenerud (approved):

YYChen01988 (approved):
LGTM

PR #2006, opened 2024-04-08
Allow early session starts in auto mode

YYChen01988:
## Goal
Allow `Bugsnag.startSession` to be called with automatic session tracking, and not have the first manual session be over written by the first automatic session.

## Design
The first automatically started session will be discarded if there is already a manually started session. All subsequent sessions behave as previous.

## Testing
Manual testing and end to end test


bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -36,6 +36,7 @@ class SessionTracker extends BaseObservable implements ForegroundDetector.OnActi
     private volatile Session currentSession = null;
     final BackgroundTaskService backgroundTaskService;
     final Logger logger;
+    private boolean shouldSuppressFirstAutoSession = true;
```
lemnik, line 39: Given that this *only* happens when the first session is manual, maybe this field should be named something more like: `firstAutoSessionSuppressed` and rather start as `false`?

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionTracker.java
```diff
@@ -92,12 +93,27 @@ Session startNewSession(@NonNull Date date, @Nullable User user,
...
+            return true;
+        } else {
+            if (autoCaptured && currentSession != null) {
+                if (!currentSession.isAutoCaptured() && shouldSuppressFirstAutoSession) {
+                    shouldSuppressFirstAutoSession = false;
+                    return true;
+                }
+            }
+            return false;
+        }
```
lemnik, line 113: To avoid possible `NullPointerException`s I think that the `currentSession` should rather be captured here. It's also possible to simplify this `if` statement:

```suggestion
        } else {
            Session existingSession = currentSession;
            if (autoCaptured &&
                    existingSession != null &&
                    !existingSession.isAutoCaptured() &&
                    shouldSuppressFirstAutoSession) {
                    
                shouldSuppressFirstAutoSession = false;
                return true;
            }
        }
        
        return false;
```

features/smoke_tests/03_sessions.feature
```diff
@@ -112,3 +112,6 @@ Feature: Session functionality smoke tests
     And the event "user.email" equals "ABC.CBA.CA"
     And the event "user.name" equals "ManualSessionSmokeScenario"
 
+  Scenario: Start session in auto mode
+    When I run "StartSessionAutoModeScenario"
+    And I wait to receive a session
```
lemnik, line 117: Its worth adding another check to ensure *exactly one* session is received.

lemnik (approved):
LGTM

PR #2008, opened 2024-04-09
Avoid compiling regexes in ObjectJsonStreamer

lemnik:
## Goal
Reduce the overhead of constructing `ObjectJsonStreamer` instances (and therefore leaving complex breadcrumbs) by sharing the default `redactedKeys` (which is immutable).

## Design
An `ObjectJsonStreamer` is allocated internally each time a breadcrumb metadata item needs to be JSON encoded early (because it is too complex for the NDK layer to handle directly). Each of these was causing the default `redactedKeys` pattern to be compiled. Since the default `redactedKeys` set is immutable, it can be safely shared instead.

## Testing
Relied on existing tests.


YYChen01988 (approved):
LGTM

PR #2013, opened 2024-04-12
Reduce use of Atomics in Session

YYChen01988:
## Goal

Reduce use of Atomics in Session, so the value is more accurate syncing between multiple threads. 

## Changeset

Change autoCaptured type from AtomicBoolean to volatile boolean since it does not use specific atomic method.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
@@ -28,7 +28,7 @@ public final class Session implements JsonStream.Streamable, UserAware {
     private App app;
     private Device device;
 
-    private final AtomicBoolean autoCaptured = new AtomicBoolean(false);
+    private volatile boolean autoCaptured = false;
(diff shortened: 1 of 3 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/Session.java
```diff
@@ -198,16 +198,20 @@ Session incrementUnhandledAndCopy() {
...
     }
 
-    AtomicBoolean isTracked() {
-        return tracked;
+    boolean markTracked() {
+        return tracked.compareAndSet(false, true);
+    }
+
+    boolean markPaused() {
+        return isPaused.compareAndSet(true, false);
```
lemnik, line 206: Given that this are now encapsulated, can we change the visibility of the `isPaused` to `private` instead of package-protected?

YYChen01988: `isPaused` in `Session` can not be private since it is used and reset in `SessionTracker`

PR #2018, opened 2024-04-16
Don't use strings for `BreadcrumbType` in NDK

lemnik:
## Goal
Small reduction in overhead when leaving breadcrumbs: use integers to represent the `BreadcrumbType` instead of strings.

## Design
While the string values for `BreadcrumbType` on the Java/Kotlin code were constants, they required a call to `GetStringUTFChars` and `strcmp` to evaluate which enum value to use.

The new implementation uses an exhaustive `when` which means:
- the transfer values are not sensitive to the ordering of the `enum` values (as `ordinal` would be)
- any adding or removing of `enum` values will cause compiler failures

## Testing
Relied on existing tests.


YYChen01988 (approved):
LGTM

kstenerud (approved):

PR #2020, opened 2024-04-23
NDK should obey Configuration.maxBreadcrumbs

lemnik:
## Goal
NDK should obey `Configuration.maxBreadcrumbs`

## Design
Change the `bugsnag_event.breadcrumbs` to be dynamically allocated instead of a fixed-size block. This blob is written out to the event file after the core event, but before the feature-flags.

## Changes

- Bumped the event version to 14
- Added `bugsnag_event.max_crumb_count` to track the size of the `breadcrumbs` buffer
- Changed `bugsnag_event.breadcrumbs` to be `bugsnag_breadcrumb*` instead of `bugsnag_breadcrumb[50]

## Testing
Manual testing, and a new end-to-end test added


DariaKunoichi (approved):
LGTM

PR #2021, opened 2024-04-24
Don't redactKeys in OpaqueValues

lemnik:
## Goal
Avoid redacting keys in the OpaqueValue class. The encoding used does not obey the configured set of redacted keys, and this behavior leads to some metadata being redacted in NDK crashes, while some is not.

## Testing
Relied on existing tests.


YYChen01988 (approved):
LGTM

PR #2023, opened 2024-04-25
Add the correlation property to the Event model

lemnik:
## Goal
Add the correlation property to the Event model so that trace id / span id can be sent as part of an Event.

## Testing
Updated existing unit test


tomlongridge (approved):
LGTM

PR #2026, opened 2024-05-08
Remove Failed delivery internal reports

YYChen01988:
## Goal

Replaced failed delivery internal reports with more effective error information 

## Changeset

Removed the error report when a file is not (de)serialized correctly and replaced it with a logger error message. 

## Testing

Existing tests


lemnik (approved):
LGTM

PR #2030, opened 2024-05-16
Update AGP and related versions for mazerunner

YYChen01988:
## Goal

The MazeRunner test fixtures can be built with AGP 8.3.2 to ensure we remain compatible.

## Changeset

Update AGP, Gradle and okhttp version

## Testing

Existing end to end tests


lemnik (approved):
LGTM

PR #2032, opened 2024-05-17
Add Android 14 tests

twometresteve:
## Goal

Add end-to-end tests for Android 14.

## Changeset

An ANR test has also also been amended, unrelated to the addition of Android 14.  It had been causing Appium issues with Android 8 for some time.

Scenario `Unhandled JVM error` has also been skipped pending further investigation - it's flaked occasionally for a while, but now seems to fail consistently on Android 14.

## Testing

Covered by a full CI run.


lemnik (approved):
LGTM

PR #2033, opened 2024-05-23
Fixed the logic of discarding session

YYChen01988:
## Goal

Discard the first automatic session when sessions are mixed with manual session

## Changeset

Changing the logic for discarding session method

## Testing

unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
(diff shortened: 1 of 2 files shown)


bugsnag-android-core/detekt-baseline.xml
lemnik, line 1: This change doesn't look like it relates to this PR.

PR #2034, opened 2024-05-24
Reinstate instrumentation tests using MacOS test boxes as a platform

Cawllec:
## Goal

We've had to remove our instrumentation tests from github actions for a variety of reasons.  This change adds a script which should add Android Virtual Devices and SDKs to our MacOS test boxes as required, using them to run the instrumentation tests.  The version of android we're testing against is determined by the API_LEVEL environment variable, which should allow us to run against several different API levels if we see the need to.

## Testing

It's successfully completed several test runs, and a [negative test run](https://buildkite.com/bugsnag/bugsnag-android/builds/11382#018faa41-5ed4-4b38-83d7-7f79fafa9f8b/62-1093) to ensure the script failed successfully.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -91,7 +91,7 @@ steps:
     plugins:
       artifacts#v1.9.0:
         upload:
-            - "bugsnag-*/build/reports/androidTests/connected/**/*"
+            - "bugsnag-*/build/reports/androidTests/connected/**/*.html"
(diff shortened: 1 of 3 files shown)


lemnik:
It's probably worth using the `compressed` option in the BK artifact plugin so that we just get a single zip file for the test reports.

lemnik (approved):
LGTM

PR #2035, opened 2024-05-30
NDK metadata iterator

lemnik:
## Goal
Reduce the overhead of copying Metadata into the NDK layer. This PR includes some minor fixes to existing tests.

## Design
Directly iterate over the `Map.entrySet` instead of first copying all of the keys to a temporary `ArrayList`. Since the most common `Map` implementations maintain their data in `Map.Entry` objects, this has significantly lower overheads and complexity than the previous pattern.

## Testing
Relied on existing tests.


Cawllec (approved):

Cawllec:
I've rebased and re-tested using the instrumentation tests.  They now pass with this change.

kstenerud (approved):

Issue #107, opened 2016-05-02
Why are consumer ProGuard rules included?


JakeWharton:
As far as I can tell none of them are needed. This library has no reflection so both obfuscation and stripping will behave correctly without them.

Currently it disables both obfuscation and stripping without a clear reason as to why.

JakeWharton:
From the history it looks like these rules were once used to ProGuard the release jar and erroneously were converted to consumer rules?

Unless someone sees something that I don't, these should be removed to allow obfuscation and unused code stripping.

JakeWharton:
Having manually removed the embedded proguard rules, everything still works fine and allowed 77 methods to be removed as well as obfuscation to occur.

I'm sending a PR.

kattrali:
They were used at one point, but that has long since passed. Merging.

PR #2049, opened 2024-06-28
Proguard rules for ErrorType & Telemetry.values()

### LINKED ISSUE #2048: Telemetry can be obfuscated but is looked up reflectively via EnumSet
### Describe the bug
Bugsnag can fail to initialize when loading on startup in an obfuscated/minified app because `Telemetry` is loaded into an `EnumSet` her: 

This results in a runtime crash like so: 

```
Caused by: java.lang.RuntimeException: java.lang.NoSuchMethodException: com.bugsnag.android.b1.values []
                              	at java.lang.Enum.enumValues(Enum.java:315)
                              	at java.lang.Enum.-$$Nest$smenumValues(Unknown Source:0)
                              	at java.lang.Enum$1.create(Enum.java:322)
                              	at java.lang.Enum$1.create(Enum.java:320)
                              	at libcore.util.BasicLruCache.get(BasicLruCache.java:63)
                              	at java.lang.Enum.getSharedConstants(Enum.java:334)
                              	at java.lang.Class.getEnumConstantsShared(Class.java:4036)
                              	at java.util.EnumSet.getUniverse(EnumSet.java:410)
                              	at java.util.EnumSet.noneOf(EnumSet.java:112)
                              	at java.util.EnumSet.of(EnumSet.java:238)
                              	at com.bugsnag.android.s.<init>(Unknown Source:128)
                              	at A.M.<init>(SourceFile:11)
                              	at com.bugsnag.android.l.h(Unknown Source:4)
                              	at dev.zacsweers.fieldspottr.MainActivity.onCreate(Unknown Source:26)
```

Worse still, because this happens during initialization, it will never be reported!

### Environment
* Android version: N/A
* Bugsnag version: 6.6.0
* Emulator or physical device: N/A

The solution would either to be to bundle consumer proguard rules that keep this enum, annotate it with `@Keep`, or consider moving off of `EnumSet` usage.

lemnik:
## Goal
Don't crash apps on startup where the `enum.values()` functions are not included in the `keep` list by default. Fixes #2048 

## Design
Included specific Proguard consumer rules for `Telemetry.values()` and `ErrorType.values()`, ensuring that these members are not removed or obfuscated.

## Testing
Manually tested with app using fully customized Proguard rules.

### DIFF
--- a/bugsnag-android-core/proguard-rules.pro
+++ b/bugsnag-android-core/proguard-rules.pro
@@ -5,3 +5,9 @@
 -keep class com.bugsnag.android.BreadcrumbState { *; }
 -keep class com.bugsnag.android.BreadcrumbType { *; }
 -keep class com.bugsnag.android.Severity { *; }
+-keepclassmembers enum com.bugsnag.android.Telemetry {
+    public static com.bugsnag.android.Telemetry[] values();
+ }
+-keepclassmembers enum com.bugsnag.android.ErrorType {
+    public static com.bugsnag.android.Telemetry[] values();
+ }


tomlongridge (approved):
LGTM - with a changelog

AlexanderGH:
@lemnik this PR mentions "Proguard consumer rules" but the rules modified in this PR are not consumer rules AFAICT. Consumer rules were disabled in https://github.com/bugsnag/bugsnag-android/issues/107

I also was unable to find any documentation that says the referenced rules should be copy pasted manually and kept up-to-date with the various artifacts.

Do you have a reference to how these rules (and those of the other sub-projects) are supposed to be included in apps?

lemnik:
Hi @AlexanderGH 

The file might not be named as "consumer" but these are our configured consumer Proguard rules, as set here:
https://github.com/bugsnag/bugsnag-android/blob/next/buildSrc/src/main/kotlin/com/bugsnag/android/BugsnagBuildPlugin.kt#L206

AlexanderGH:
Thanks for the explanation! My confusion was a result of Android Studio not showing the full set of files in the .aar. But downloading it and opening the aar directly does indeed show that the proguard rules are there. 👍

PR #2054, opened 2024-07-03
Correctly copy `build.gradle.kts` instead of the old `build.gradle` during publishing

lemnik:
## Goal
Correct the publishing step by copying the correct `build.gradle.kts` file.


YYChen01988 (approved):
LGTM

PR #2056, opened 2024-07-04
Add other downstream repos for automatic updates

Cawllec:
## Goal

Adds bugsnags flutter, JS (RN), and unreal repos to the bumpsnag workflow, to update them when a new version of the notifier is released.


lemnik (approved):
🥳

PR #2058, opened 2024-07-09
Handle cases when deserializing threads with no state

lemnik:
## Goal
Handle extremely rare edge cases when deserializing reports (for `OnSendCallback`) where a `thread.state` property was not included.

## Changeset
Replace the missing `state` with an empty string.

## Testing
Introduced a new suite of unit tests to cover this edge case, and cases such as numeric & string thread IDs and extremely long thread IDs


tomlongridge (approved):

YYChen01988 (approved):
LGTM

PR #2065, opened 2024-08-06
[PLAT-12353] Split out ANR tests to BS

joshedney:
## Goal

Split out the Android ANR tests to BrowserStack for stability

## Testing

Covered by CI


.buildkite/pipeline.yml
```diff
@@ -131,6 +136,36 @@ steps:
...
+        run: maze-runner
+        service-ports: true
+        command:
+          - "features/smoke_tests/01_anr.feature"
+          - "--app=@build/bs-fixture-r19-url.txt"
+          - "--app-activity=com.bugsnag.android.mazerunner.MainActivity"
+          - "--app-package=com.bugsnag.android.mazerunner"
+          - "--appium-version=1.22.0"
+          - "--farm=bs"
+          - "--device=ANDROID_7"
```
twometresteve, line 161: An annoying thing to realise now is that all the android 7 and 8 devices we have for BS in Maze Runner are Samsungs and so the scenarios will all be skipped.

Maybe we don't actually need to skip Samsung?  Could be worth checking that.  Or see if BS have any other options we can use?

We can skip the Android 7 and 8 steps in this PR if you like and raise a separate ticket to reinstate them somehow.

joshedney: Looking at BS it doesn't look like they have any none Samsung devices for Android 7/8. I'll remove the skip and see what issues we get.

twometresteve (approved):

PR #2066, opened 2024-08-07
Restored the expected timestamp format to NDK breadcrumbs

lemnik:
## Goal
Stop NDK breadcrumbs from being dropped due to invalid timestamps.

## Testing
Manually tested that the correct timetstamp is used. Also added to existing unit tests to assert the timestamp is as expected.


YYChen01988 (approved):
LGTM

PR #2069, opened 2024-08-12
Rework startup threading

lemnik:
## Goal
Improve the threading during startup and reduce the overheads.

## Design
Replaced the various layers of startup tasks with a dedicated structure. The `RunnableProvider` class replaces the previously used `Future`, `SafeFuture` and `Callable` / `Runnable` tasks. The `RunnableProvider` safely computes values on background threads, and blocks only when required while also avoiding running the work (such as IO) on the main thread.

## Testing
Updated tests to handle the new structures, and relied on existing tests along with manual testing with `StrictMode` using `detectAll()`.


bugsnag-android-core/src/main/java/com/bugsnag/android/StorageModule.kt
```diff
@@ -2,56 +2,72 @@ package com.bugsnag.android
...
-    private fun lastRunInfo(): LastRunInfo? {
+    val lastRunInfo = lastRunInfoStore.map { lastRunInfoStore ->
         val info = lastRunInfoStore.load()
         val currentRunInfo = LastRunInfo(0, crashed = false, crashedDuringLaunch = false)
         lastRunInfoStore.persist(currentRunInfo)
-        return info
+        return@map info
+    }
+
+    fun loadUser(initialUser: User): Provider<UserState> = provider {
```
tomlongridge, line 68: This appears to be doubly-wrapped in providers - what's the reason for UserStore being handled differently?

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionStore.kt
```diff
@@ -13,14 +12,14 @@ import java.util.Date
...
+    maxPersistedSessions: Int,
+    private val apiKey: String,
     logger: Logger,
     delegate: Delegate?
 ) : FileStore(
-    File(
-        config.persistenceDirectory.value, "bugsnag/sessions"
-    ),
-    config.maxPersistedSessions,
+    File(bugsnagDir, "sessions"),
```
tomlongridge, line 21: Are we consistent in when we add a `bugsnag` subdir under the persistence directory?

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/dag/Provider.kt
```diff
@@ -0,0 +1,114 @@
...
+    override fun get(): E {
+        return getOr { throw value as Throwable }
+    }
+
+    private inline fun getOr(failureHandler: () -> E): E {
+        while (true) {
+            when (state.get()) {
+                TASK_STATE_RUNNING -> awaitResult()
+                TASK_STATE_PENDING -> {
+                    if (isMainThread()) {
```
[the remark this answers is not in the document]
lemnik: No. All of the `Provider` work is always carried out on background threads, so they never wait for the main thread.

bugsnag-android-core/src/main/java/com/bugsnag/android/StorageModule.kt
tomlongridge: Discussed IRL - other fields in client will probably go this way in future changes.

bugsnag-android-core/src/main/java/com/bugsnag/android/SessionStore.kt
tomlongridge: I see now - these are the uses where we go through `migrateLegacyFiles`; the others are using the config value directly.

bugsnag-android-core/src/main/java/com/bugsnag/android/internal/dag/DependencyModule.kt
```diff
@@ -1,3 +1,26 @@
...
+    @JvmField
+    val taskType: TaskType = TaskType.DEFAULT
+) : DependencyModule {
+    inline fun <R> provider(crossinline provider: () -> R): RunnableProvider<R> {
+        return bgTaskService.provider(taskType, provider)
+    }
+
+    internal inline fun <E, R> Provider<E>.map(crossinline mapping: (E) -> R): RunnableProvider<R> {
+        val task = object : RunnableProvider<R>() {
+            override fun invoke(): R = mapping(this@map.get())
```
YYChen01988, line 20: Maybe nice to have a note describing what the invoke do and also this class's task?

YYChen01988 (changes requested):

tomlongridge (approved):

YYChen01988 (approved):
LGTM, pending CI

PR #2070, opened 2024-08-16
AnrLinkError and NdkLinkError report

YYChen01988:
## Goal

Improve AnrLinkError and NdkLinkError report so when loading our native libraries.

## Changeset

Error reporting added to metadata when loading library

## Testing

Existing unit tests


bugsnag-plugin-android-anr/src/main/java/com/bugsnag/android/AnrPlugin.kt
tomlongridge, line 1: Needs a changelog entry, along the lines of

> Add original error class and message to metadata for link errors loading BugSnag libraries

tomlongridge (approved):

PR #2074, opened 2024-09-17
Simplifying discardOldestFileIfNeeded function

hannah-smartbear:
## Goal

The function had a fairly high overhead, making several copies of the list of files. The goal was to reduce this.

## Changeset

- Return earlier if the number of files is less than the maxStoreCount (previously this wasn't checked until a second copy of the list had already been made)
- Sort the first file list by last modified timestamp, instead of copying into a list and then sorting

## Testing

- Tests already written for this function are still applicable
- Tested this locally

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.kt
@@ -7,8 +7,6 @@ import java.io.FileNotFoundException
 import java.io.FileOutputStream
 import java.io.OutputStreamWriter
 import java.io.Writer
-import java.util.Collections
-import java.util.Comparator
 import java.util.concurrent.ConcurrentSkipListSet
 import java.util.concurrent.locks.Lock
 import java.util.concurrent.locks.ReentrantLock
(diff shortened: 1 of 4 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.kt
```diff
@@ -115,23 +112,21 @@ internal abstract class FileStore(
...
+                    return
+                } else if (!queuedFiles.contains(file)) {
+                    logger.w(
+                        "Discarding oldest error as stored error limit reached: '" +
+                            file.path + '\''
+                    )
+                    deleteStoredFiles(setOf(file))
+                    discardedCount++
                 }
             }
```
lemnik, line 131: I suspect we can further simplify this logic using a Sequence:
```suggestion
            deleteStoredFiles(
                listFiles
                    .asSequence()
                    // only discard files that are not currently in the queue
                    .filter { it !in queuedFiles }
                    // discard the oldest files first
                    .sortedBy { it.lastModified() }
                    // take the 'numberToDiscard' oldest files off the top of the sequence
                    .take(numberToDiscard)
                    // log each file that will be discarded
                    .onEach {
                        logger.w(
                            "Discarding oldest error as stored error limit reached: '${it.path}'"
                        )
                    }
                    .toSet()
            )
```

lemnik (approved):
LGTM

PR #2078, opened 2024-09-25
Add OkHttpDelivery class to allow customers to use their OkHttpClient for delivery

hannah-smartbear:
## Goal

Create a OkHttpDelivery class as part of the bugsnag-plugin-android-okhttp plugin to allow customers to use their OkHttpClient for delivery instead of using our default delivery

## Changeset

- Created OkHttpDelivery class that extends the Delivery interface, that uses similar logic to the DefaultDelivery class for preparing the payload for delivery, logging, etc. but using the customer's OkHttpClient instead if one is provided
- Moved DefaultDelivery's serializePayload function out of the DefaultDelivery class so it could be used by the OkHttpDelivery class too, and made JsonHelper object public
- Customer would use this as follows: `config.useOkHttpDelivery(myExistingHttpClient)`

## Testing
- Tested locally
- Since this class extends Delivery, it looks like a lot of the testing (e.g. testing the headers are correct) are already covered?


bugsnag-android-core/src/main/java/com/bugsnag/android/TruncateEvent.kt
lemnik, line 1: This seems like a file that should be in the `com.bugsnag.android.internal` packge

bugsnag-plugin-android-okhttp/src/main/java/com/bugsnag/android/okhttp/OkHttpDelivery.kt
```diff
@@ -0,0 +1,100 @@
...
+    companion object {
+        @SuppressLint("NotConstructor")
+        @JvmSynthetic
+        fun Configuration.OkHttpDelivery(okHttpClient: OkHttpClient): OkHttpDelivery {
+            return OkHttpDelivery(okHttpClient, apiKey, maxStringValueLength, logger!!)
+        }
+
+        fun Configuration.useOkHttpDelivery(okHttpClient: OkHttpClient) {
+            delivery = OkHttpDelivery(okHttpClient)
+        }
```
lemnik, line 39: These should be outside of the `OkHttpDelivery` class (ie: in the global namespace).

PR #2080, opened 2024-09-27
Startup timeout fixes

lemnik:
## Goal
Use a reasonable timeout for startup crash delivery in cases when `Bugsnag.start` is called long after the app start.

## Changeset
If the startup crash synchronous timeout would be `<=0` we revert to the old behavior and block for up-to a full 2 seconds.

## Testing
Relied on the existing tests being more reliable.


YYChen01988 (approved):
LGTM

PR #2088, opened 2024-10-07
Move Build UUID retrieval / calculation into a provider

lemnik:
## Goal
Ensure calculating the Build UUID is done on a background worker when it's derived from the `.dex` file signatures.

## Changeset
The Build UUID has internally been changed from a `String?` to a `Provider<String?>?`:

- When there is not a Build UUID and it will not be calculated, the `Provider` is null
- When the ID is being calculated it will be a non-null `Provider`
- If the Build UUID cannot be calculated, the `Provider` will return null

## Testing
Updated the existing tests to handle the new pattern.


kstenerud (approved):

YYChen01988 (approved):
LGTM

PR #2089, opened 2024-10-08
Startup improvements

lemnik:
## Goal
Improve the startup threading behavior of the SDK to reduce the chances of deadlocks & ANRs.

## Design
Previously the SDK startup included several calls which moved work to a background thread, and then used `Future.get()` to block until the work was complete. This was done to keep I/O operations (and similar) off the main thread, but introduced the cross-thread overheads without any direct benefit (all the cost of a thread, without actually doing anything in parallel).

This PR introduces a new model based around a `Provider` interface, which is similar to `FutureTask` in concept. The primary implementation is `RunnableProvider` which has a very primitive work-stealing ability to ensure that `Providers` do not block each other (avoiding possible deadlocks where tasks are blocked waiting for tasks on the same queue) unless required. That is: a `Provider` will typically try and "do" another providers work (unless that provider has already done, or is busy doing said work).

As far as possible the new model does not block until the data from a `Provider` is actually required by a non-provider (such as the NDK module, or a crash handler). This should allow better use of the threads available during startup.

## Testing
Modified the existing unit tests to handle the changes to the models.


YYChen01988 (approved):
LGTM, pending changelog entrance

PR #2092, opened 2024-10-23
Improved EventPayload for Delivery

lemnik:
## Goal
Make it easier to build new `Delivery` implementations without requiring the use of internal classes.

## Design
Updated the `EventPayload` class to have the trimming logic and integrity-header generation.

## Changeset
`EventPayload` and `Session` both now include a `toByteArray()` function that will encode them in JSON suitable for direct delivery over HTTP.

`EventPayload` also includes the calculations for the `Bugsnag-Integrity` header, and the payload trimming logic that was previously in `DefaultDelivery`

A new `OkHttpDelivery` implementation of `Delivery` has been added to the `bugsnag-plugin-android-okhttp` module, allowing for the reuse of externally configured `OkHttpClient`s to be used for delivery of sessions and events.

## Testing
New end to end scenario


bugsnag-android-core/src/main/java/com/bugsnag/android/EventPayload.kt
```diff
@@ -13,25 +15,85 @@ import java.io.IOException
...
+        val localEvent = event
+        if (localEvent != null) {
+            return localEvent
         }
+
+        val eventSource = MarshalledEventSource(eventFile!!, apiKey ?: config.apiKey, logger)
+        val decodedEvent = eventSource()
+
+        // cache the decoded Event object
+        event = decodedEvent
```
kstenerud, line 56: Maybe call this `cachedEvent`? Or is it already an exposed API?

kstenerud (approved):

bugsnag-android-core/src/main/java/com/bugsnag/android/EventPayload.kt
lemnik: Unfortunately it's already exposed API. It's also sometimes the "original" event (when we don't have an event file)

YYChen01988 (approved):
LGTM

PR #2100, opened 2024-11-08
In-memory ExitInfoPluginStore

lemnik:
## Goal
Simplify the use of the `ExitInfoPluginStore` by retaining the data in-memory instead of constantly reloading from disk.

## Testing
Modified existing tests to match the new fields & behaviour


YYChen01988 (approved):
LGTM

PR #2107, opened 2024-11-18
Replaced the buildSrc plugin with discrete build files

lemnik:
## Goal
Break up the build plugin moving the majority of the build logic back into the `build.gradle.kts` files.

The build plugin within `buildSrc` created a Gradle classpath structure and plugin application order that has blocked various upgrades (AGP and Kotlin mostly). By splitting up the build plugin and moving to simpler utility functions, we can upgrade components and then migrate more of the build logic back into utilities when/if we can.

## Design
The common version details have been rearranged within the `Versions` object, which now includes a `ModuleWithVersion` utility class. This serves a similar purpose to a Gradle Version catalog, but can easily be used from within the build sources (`buildSrc`) and therefore used within utility functions (such as `addCommonModuleDependencies()` and `loadDefaultPlugins()`).

## Testing
The existing build steps all still work without change. Manually published to Sonatype staging repository.


tomlongridge (approved):
LGTM - largely going on looking for inconsistencies and letting the build speak for itself. Seems like a good approach.

YYChen01988 (approved):
LGTM

PR #2112, opened 2024-11-29
Pipeline tidyup

twometresteve:
## Goal

General tidy-up of the Buildkite pipeline to:
- Ensure all steps have an adequate but not excessive timeout
- Agent queues are explicitly stated in each file (even if a default is used)
- Generic `macos` queue is used instead of specfic macOS versions where possible

## Testing

Covered by a full CI run.

### DIFF
--- a/.buildkite/block.step.yml
+++ b/.buildkite/block.step.yml
@@ -1,7 +1,12 @@
+agents:
+  queue: 'macos'
+
 steps:
-  - block: 'Trigger full build'
+  - label:
+    block: 'Trigger full build'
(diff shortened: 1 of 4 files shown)


.buildkite/block.step.yml
```diff
@@ -1,7 +1,12 @@
+agents:
+  queue: 'macos'
```
joshedney, line 2: You'll also need to update the queue within buildkite if you haven't done so already

joshedney (approved):
LGTM

Cawllec (approved):

PR #2127, opened 2025-01-20
Update makefile version checking

clr182:
## Goal

make bump does no verification of the version syntax, allowing us to accidentally bump to version strings that Gradle cannot order / match and potentially breaking our recommended com.bugsnag:bugsnag-android:5.+ syntax.

## Changeset

Added regex version checking for major.minor.patch on make bump.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,23 +67,13 @@ example-app:
 	@cd ./examples/sdk-app-example/ && ./gradlew clean assembleRelease
 
 bump:
-ifneq ($(shell git diff --staged),)
-	@git diff --staged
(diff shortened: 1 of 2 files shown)


lemnik:
I think ideally this PR should use a script based on something like: https://github.com/bugsnag/bugsnag-android-performance/blob/next/scripts/bump-version.sh

clr182:
> I think ideally this PR should use a script based on something like: https://github.com/bugsnag/bugsnag-android-performance/blob/next/scripts/bump-version.sh

lemnik (approved):
LGTM

PR #2170, opened 2025-03-26
LibraryLoader will retry System.loadLibrary exactly once

lemnik:
## Goal
Work around possible `System.loadLibrary` race conditions by retrying native library loads exactly once if they fail.

## Testing
Relied on existing tests

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/LibraryLoader.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/LibraryLoader.java
@@ -39,8 +39,14 @@ class LibraryLoader {
             try {
                 System.loadLibrary(name);
                 loaded = true;
-            } catch (UnsatisfiedLinkError error) {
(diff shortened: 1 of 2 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/LibraryLoader.java
```diff
@@ -39,8 +39,13 @@ void loadLibInternal(String name, Client client, OnErrorCallback callback) {
             try {
                 System.loadLibrary(name);
                 loaded = true;
-            } catch (UnsatisfiedLinkError error) {
-                client.notify(error, callback);
+            } catch (UnsatisfiedLinkError ignored) {
+                try {
```
tomlongridge, line 43: I feel like a comment justifying this might be reasonable :)

tomlongridge (approved):

YYChen01988 (approved):
LGTM

PR #2177, opened 2025-04-15
Release/v6.13.0

YYChen01988:
### Enhancements

* Attempt a second `System.loadLibrary` for native modules in case of library loading race-conditions
  [#2170](https://github.com/bugsnag/bugsnag-android/pull/2170)


lemnik (approved):
LGTM

PR #2181, opened 2025-05-01
Increase AVD timeout to 120s to cover increased startup time

joshedney:
## Goal

The instrumentation tests fail because the AVD does not boot in 60 seconds. I've increased the timeout to 120 seconds, enough time for the AVD to boot.

## Testing

Covered by CI


lemnik (approved):
LGTM

PR #2185, opened 2025-05-19
Isolate CI step that used AVD

joshedney:
## Goal

Due to issues with running multiple AVD instances on the same box, its best if we move that step to the isolated queue to prevent the steps from erroring when multiple tests are running on the same box.

## Testing

Covered by CI


lemnik (approved):
LGTM

PR #2189, opened 2025-05-26
Change to discardOldestFileIfNeeded sorting

clr182:
## Goal
Fix rare `"Comparison method violates its general contract!"` crashes in `discardOldestFileIfNeeded` when the file timestamps change during the `sort()`.

## Design
Capture all of the `lastModified()` timestamps into a stable list before sorting.

## Testing
Relied on existing tests

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.kt
@@ -112,34 +112,32 @@ internal abstract class FileStore(
 
     fun discardOldestFileIfNeeded() {
         // Limit number of saved payloads to prevent disk space issues
-        if (isStorageDirValid(storageDir)) {
-            val listFiles = storageDir.listFiles() ?: return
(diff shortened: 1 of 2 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/FileStore.kt
```diff
@@ -114,11 +114,21 @@ internal abstract class FileStore(
         if (isStorageDirValid(storageDir)) {
             val listFiles = storageDir.listFiles() ?: return
             if (listFiles.size < maxStoreCount) return
-            val sortedListFiles = listFiles.sortedBy { it.lastModified() }
+
+            // Store lastModified to ensure it doesn't change between sorting
```
lemnik, line 118:
```suggestion
            // Store lastModified to ensure it doesn't change during sort
```

clr182:
Hey @lemnik 
wondering if you've had a chance to consider this further? FYI as far as I am aware it hasn't been reported again by any other user.

PR #2190, opened 2025-05-26
Set default endpoints based on API key

richardelms:
## Goal

Set default endpoints based on API key. If no endpoint is configured, payloads should be sent to `*.insighthub.smartbear.com` if the API key starts with `00000`.

## Changeset

- changed ConfigInternal.kt to return the correct defaults based on api key if no custom endpoints have been configured.
- added comparison methods to EndpointConfiguration class to ensure unit tests compare the values of the endpoints inside rather than the objects. 

## Testing

Added unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Enhancements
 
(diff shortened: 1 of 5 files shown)


lemnik (approved):
LGTM

PR #2196, opened 2025-06-04
Fix the publishing repository

lemnik:
## Goal
Fix the `release` Gradle script to point to the correct repository


joshedney (approved):
LGTM!

PR #2202, opened 2025-06-20
Fix RootDetection background ANR

lemnik:
## Goal
Fix background ANRs that occur on startup when waiting for the `RootDetection` sub-process to complete.

## Design
Added a timed `Process.waitFor` to the `RootDetection` to limit the amount of time the sub-process can take before we continue with startup.

Also generally moved more of the startup process waits to later to allow more to complete in parallel.

## Testing
Manually tested blocking the sub-process, and relied on existing tests for the remaining changes.


YYChen01988 (approved):
LGTM

PR #2203, opened 2025-06-24
Properly scope the build-id capturing in the stack_unwinder

lemnik:
## Goal
Improve the capturing of the `.so` build-id when unwinding native stacks.

## Testing
Relied on existing tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.cpp
+++ b/bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.cpp
@@ -102,26 +102,28 @@ static void populate_code_identifier(const unwindstack::FrameData &frame,
   }
 
   auto shared_build_id = elf_fields->build_id_.load();
-  std::string_view build_id;
   if (shared_build_id == nullptr) {
     auto elf = elf_fields->elf_.get();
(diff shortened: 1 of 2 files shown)


bugsnag-plugin-android-ndk/src/main/jni/utils/stack_unwinder.cpp
```diff
@@ -102,26 +102,28 @@ static void populate_code_identifier(const unwindstack::FrameData &frame,
...
+    }
 
-  if (build_id.empty()) {
-    return;
+    // MapInfo.GetPrintableBuildID is *not* async-safe so we need our own
+    // safe hex encoder to copy BuildID into code_identifier.
+    bsg_hex_encode(dst_frame.code_identifier, build_id.data(),
+                   build_id.length(), sizeof(dst_frame.code_identifier));
+  } else if (!shared_build_id->empty()) {
+    std::string build_id = *shared_build_id;
```
DariaKunoichi, line 121: why are you shadowing that variable?

lemnik: I've added a `static_cast` to clarify what's going on here

PR #2205, opened 2025-06-27
Use JNI_OnLoad to link the NDK plugin

lemnik:
## Goal
Avoid lazy-linking the JNI functions in `bugsnag-plugin-android-ndk`.

## Design
This PR introduces a small Gradle plugin to `buildSrc` that generates the JNI natives lists as header files which are `#include`d into `bugsnag_ndk.c`. The plugin means that the Kotlin `external`s and the native code are guaranteed to be in-sync and no new functions can be introduced without their implementation.

Using `JNI_OnLoad` and `RegisterNatives` also helps improve reliability because the entire module will either load, or not load. So we cannot suffer from spurious `LinkError`s when unlinked function cannot be called due to a failed loading of some kind.

## Testing
Manually tested failure modes, and relied on existing tests.

### DIFF
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -933,6 +933,9 @@ static void JNI_NativeBridge_setInternalMetricsEnabled(JNIEnv *env,
   bsg_set_internal_metrics_enabled(enabled);
 }
 
+// These headers are included here to ensure that the JNI methods are
+// declared (above) as static, avoiding the need to forward declare them
+
 #include "NativeBridge_JNI.h"
 #include "NdkPlugin_JNI.h"
 
--- a/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
+++ b/bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
@@ -84,8 +84,7 @@ void bsg_finish_handling_crash() {
   bsg_global_env->handling_crash = false;
 }
 
-JNIEXPORT void JNICALL Java_com_bugsnag_android_NdkPlugin_enableCrashReporting(
-    JNIEnv *env, jobject _this) {
+static void JNI_NdkPlugin_enableCrashReporting(JNIEnv *env, jobject _this) {
   if (bsg_global_env == NULL) {
     BUGSNAG_LOG(
         "Attempted to enable crash reporting without first calling install()");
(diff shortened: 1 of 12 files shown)


bugsnag-plugin-android-ndk/src/main/jni/bugsnag_ndk.c
```diff
@@ -962,12 +927,55 @@ Java_com_bugsnag_android_ndk_NativeBridge_notifyRemoveCallback(
...
-    JNIEnv *env, jobject thiz, jboolean enabled) {
+static void JNI_NativeBridge_setInternalMetricsEnabled(JNIEnv *env,
+                                                       jobject thiz,
+                                                       jboolean enabled) {
   bsg_set_internal_metrics_enabled(enabled);
 }
 
+#include "NativeBridge_JNI.h"
+#include "NdkPlugin_JNI.h"
+
```
tomlongridge, line 938: Should we include these earlier, with the others?

buildSrc/build.gradle.kts
```diff
@@ -4,8 +4,19 @@ plugins {
 
 repositories {
     mavenCentral()
+    google()
+}
+
+gradlePlugin {
+    plugins {
+        register("bugsnag-ndk-table-plugin") {
```
tomlongridge, line 12: Is this a more descriptive name?

```suggestion
        register("bugsnag-jni-link-table-plugin") {
```

buildSrc/src/main/kotlin/com/bugsnag/android/gradle/JNILinkTablePlugin.kt
```diff
@@ -0,0 +1,67 @@
...
+
+import com.android.build.api.instrumentation.InstrumentationScope
+import com.android.build.api.variant.AndroidComponentsExtension
+import com.android.build.api.variant.Variant
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.file.Directory
+import org.gradle.api.provider.Provider
+
+class JNILinkTablePlugin : Plugin<Project> {
```
tomlongridge, line 11: It would be nice to see a bit of code docs here to explain the plugin and rationale.

tomlongridge (commented):
Negligible comments/queries, but also can you include a changelog entry? Even though it's internal, I think it worth calling out so we spot it if we notice any change in the link errors.

buildSrc/build.gradle.kts
lemnik: Yes, I think that's a much nicer name.

tomlongridge (approved):

tomlongridge (approved):

PR #2221, opened 2025-07-23
Reduce the level of warnings in the ExitInfo plugin

lemnik:
## Goal
Avoid emitting potentially confusing warnings during normal `exit-info` plugin operations


YYChen01988 (dismissed):
LGTM

tomlongridge (approved):
Not sure it justifies a minor release on it's own, but we'll argue that out on release.

PR #2235, opened 2025-08-01
ExitInfo Plugin: Avoid synthesizing duplicate ANRs when in fallback modes

lemnik:
## Goal
Improve the robustness of the `exitinfo` error matching and handling.

## Design
The plugin now maintains 2 state objects (with the same data structure):
- the previously serialized state (typically stored during the last run of the app)
- the current state of the plugin (updated and stored while running)

The states now include a timestamp which is used as a cut-off point when attempting to match events. This is set to the startup time of the plugin, so any `ApplicationExitInfo` reports *after* the most `previousState` will represent exits that occurred either for the previous run of the app (or some other processes that crashed on startup before the plugin was initialized).

When synthesizing ANR reports, the PID of the report must match the `previousState.pid`. This means that we will only ever synthesize a single ANR report at a time, and only after we know that our ANR plugin didn't catch it (for example: on a background ANR).

## Testing
New unit tests added with synthetic condition sets to check matching and edge-cases.


YYChen01988 (dismissed):
LGTM

YYChen01988:
Should we have a CHANGELOG entry?

PR #2237, opened 2025-08-05
Do not report `lowMemory` or `memoryTrimLevel` when not available

lemnik:
## Goal
Stop reporting the `app.lowMemory` and `app.memoryTrimLevel` attributes when they are not reported by the OS. Newer versions of Android have deprecated these callbacks, and removing the attributes (unless the callbacks have been triggered) reduces the amount of metadata and noise.

## Testing
Updated existing unit tests and end-to-end tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Enhancements
+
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## 6.16.0 (2025-07-31)
```
YYChen01988, line 3: TBD?

PR #2244, opened 2025-08-13
Fix reported loadAddress for system libs in NDK errors

lemnik:
## Goal
Correct the `loadAddress` based on the mapped offset so that the `frameAddress - loadAddress == relPC`.

## Changeset
When a valid map offset is reported by the unwinder, subtract it from the reported `loadAddress` so that the reported address is the expected value. The only observed offset maps have been system libraries (such as `libart.so`) but there are likely others that may have had the wrong `loadAddress` reported.

In order to retain `frameAddress` being the raw PC while `lineNumber` remains the reported relative PC (`rel_pc`) we need to adjust the `loadAddress` so that `frameAddress - loadAddress == lineNumber` (or rather `pc - load_addr == rel_pc`). The `rel_pc` is typically calculated as:
```c
pc - map_info->start() + load_bias_ + map_info->elf_offset();
```
where `elf_offset` is `elf_offset == offset - elf_start_offset`.

However: the `load_bias` for a JITsym frame is `7fffffffffffffff` which ruins these calculations, so when calculating the `loadAddress` we check the total offset applied won't cause an underflow.

## Testing
- Added a frameAddress / lineNumber validation check to the Mazerunner smoke tests.
- Manually checked against various crash types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 * Handle edge-cases to reduce the changes of duplicate ANR reporting from `bugsnag-plugin-android-exitinfo` when `reportUnmatchedANR = true && disableProcessStateSummaryOverride = true`
   [#2235](https://github.com/bugsnag/bugsnag-android/pull/2235)
+* Correct the reporting of the loadAddress for native system libraries and JIT frames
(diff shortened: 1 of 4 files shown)


yousif-bugsnag (approved):
LGTM

YYChen01988 (approved):
LGTM

PR #2273, opened 2025-09-04
Error grouping

YYChen01988:
## Goal

- Introduce`Bugsnag.groupingDiscriminator` and `Event.groupingDiscriminator` as an additional way to differentiate (and split) groups of events


tomlongridge (approved):
Changelog to be added on release.

PR #2275, opened 2025-09-04
Correct detection of out of project frames in e2e tests

twometresteve:
## Goal

Resolve a common flake caused by incorrect detection of out of project frames.

## Design

If a `!` is present in the stackframe file, `symbolicate` was only passing everything after it to `is_out_of_project?`.  But that function expects to receive the whole path to detect out of project frames correctly.  This was causing the `Stack buffer overflow` scenario to flake.  Sometimes the frame is provided as:
```
/apex/com.android.runtime/lib64/bionic/libc.so
```
and sometimes (the flaky case) it is:
```
/apex/com.android.runtime/lib64/bionic/libc.so!libc.so
```

## Changeset

I've also fixed a few style errors highlighted by my IDE.

## Testing

Covered by a full CI run.

### DIFF
--- a/features/steps/symbol_steps.rb
+++ b/features/steps/symbol_steps.rb
@@ -66,7 +66,7 @@ Then("the exception {string} demangles to {string}") do |keypath, expected_value
                   "expected '#{actual_value}' to demangle to '#{expected_value}' but was '#{demangled_value}'")
 end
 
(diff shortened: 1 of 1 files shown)


lemnik (approved):
LGTM

PR #2291, opened 2025-09-18
ExitInfo: add the `abortMessage` to applicable events

lemnik:
## Goal
When reporting a crash due to a failed assertion we want to report the assertion message. This is reported as the `abortMessage` in the tombstone file.

## Testing
Updated the unit tests & manually tested.


YYChen01988 (approved):
LGTM

PR #2296, opened 2025-09-23
Remote Config Retrieval & Storage

lemnik:
## Goal
Download, store and retrieve the `RemoteConfig` objects.

## Design
Added the `RemoteConfigStore` to store & retrieve the persistent `RemoteConfig` json files in the config directory.

The `RemoteConfigRequest` is a `Callable<RemoteConfig?>` that will attempt to retrieve new configuration, but will also return the existing `RemoteConfig` with a new expiry if instructed to by the `Cache-Control` header.

Finally, the `RemoteConfigState` encapsulates loading the `RemoteConfig` either from local storage or requesting updated config from the server (along with a function to schedule a download if required).

## Changeset

- Replaced the `RemoteConfig.fromParser` with `RemoteConfig.fromMap` which better aligns with the new `JsonCollectionParser`

## Testing
New unit tests added


YYChen01988 (approved):
LGTM

PR #2299, opened 2025-09-25
SDK Side Event Discarding

lemnik:
## Goal
Apply `DiscardRule`s to events before they are delivered, discarding matched events instead of delivering them.

## Design
Introduced a new `DeliveryPipeline` class that encapsulates the final stages of `Event` processing before `Delivery` gets called. This handles both in-memory and previously persisted events (in the form of `EventPayload`).

Both the `DeliveryDelegate` and `EventStore` have been altered to send events via the `DeliveryPipeline` and handle the response (including the possibility of the events being discarded). The `DeliveryPipeline` will attempt to retrieve the `RemoteConfig` and apply any applicable `DiscardRule`s *after* running any registered `OnSendCallback`s, ensuring the payload is in the same form as it would be server-side.

## Changeset

* `EventPayload.apiKey` now encapsulates much of the logic to decide which api-key is applicable for a given `Event`
* A `RemoteConfigState` is now created within each `Client` and serves to download and store the applicable remote configuration

## Testing
Manually tested the discard functionality by injecting an appropriate config file into the example app.

Several of the unit tests needed some basic updates to support the new `DeliveryPipeline` class


YYChen01988 (approved):
LGTM

YYChen01988 (approved):

PR #2301, opened 2025-09-29
Simple path implementation for JSON data extraction

lemnik:
## Goal
Enable the next phase of our remote SDK configuration with a simple data extractor.

## Design
The new `JsonCollectionPath` class implements the parsing of the path into nodes, and encapsulates the data extraction.

Calling `JsonCollectionPath.fromString` splits the string on the dot (`'.'`) character and then builds a chain of `PathNode` objects based on each unit of the path:

- `'*'` becomes a `WildcardNode` and matches all elements in an array, and all values in an object
- Positive integers become `PositiveIndex` objects and attempt to match the numbered element of an array. If asked to match against an object (`Map`) the index is treated as a property name
- Negative integers become `NegativeIndex` objects and attempt to match an element from the end of an array (`-1` is the last element, `-2` is the second-from-last, and so-on). If asked to match against an object (`Map`) the index is treated as a property name (verbatim)
- All other strings are treated as a property name for a JSON object (`Map`) and will not match anything else.

## Testing
New unit tests introduced.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/json/JsonCollectionPath.kt
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/internal/json/JsonCollectionPath.kt
@@ -1,0 +1,130 @@
+package com.bugsnag.android.internal.json
+
+/**
+ * A simple path implementation similar to json path, but much simpler. The notation is strictly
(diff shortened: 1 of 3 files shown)


yousif-bugsnag (approved):
LGTM

PR #2318, opened 2025-10-24
HASH Discard Rule

lemnik:
## Goal
Introduce the `Hash` Discard rule to the `RemoteConfig` capabilities.

## Changes

- added the `Hash` discard rule to the `RemoteConfig` class
- `Hash` objects do not parse their `DiscardRule` content until they are first used
- Changed the `All` and `AllHandled` discard rules to be objects instead of classes

## Testing
Added new unit tests for the `Hash` parsing and serialization. Also tested manually.


bugsnag-android-core/src/main/java/com/bugsnag/android/internal/remoteconfig/RemoteConfigState.kt
```diff
@@ -112,7 +112,7 @@ internal class RemoteConfigState(
     }
 
     internal companion object {
-        const val REFRESH_BUFFER_MS = 2 * 60 * 60 * 1000L // 2 hours
+        const val REFRESH_BUFFER_MS = 0L
```
DariaKunoichi: why is it reset to 0?

PR #2322, opened 2025-10-30
Use Maze Runner dedicated idempotent commands

twometresteve:
## Goal

Moves the e2e test fixture to the idempotent commands pattern.

## Design

The thrust of the change should be familiar from the bugsnag-android-performance fixture that already uses idempotent commands.  The new `/idem-command` endpoint that only supports idempotent commands is used as the functionality will eventually be removed from the `/command` endpoint.

## Changeset

New `PersistentData` class created to factor code out of the `MainActivity` and avoid static analysis failures (we could do more here in time).

Whilst implementing the change and, in particular, running the tests several times with additional logging I made some additional changes to reduce test flakes:
- `CONFIG_FILE_TIMEOUT` increased as sometimes the file was not found in time.  Further resilience may be needed in this area in future.
- `onCreate` now records whether the activity is being recreated and prevents a new command runner from being started if that's the case.  Thanks to extra logging I found that a second thread was being created, causing the same command to be run more than once and tests to fail.  The existing `polling` flag was not sufficient as a new object must be instantiated causing it to be false (in hindsight I could have tried making it static).

## Testing

Covered by a full CI run.

### DIFF
--- a/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
+++ b/features/fixtures/mazerunner/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
@@ -20,7 +20,7 @@ import java.net.URL
 import kotlin.concurrent.thread
 import kotlin.math.max
 
-const val CONFIG_FILE_TIMEOUT = 5000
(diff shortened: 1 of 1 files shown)


lemnik (approved):
LGTM with a tiny naming suggestion

PR #2324, opened 2025-10-30
Do not close launch window in flaky e2e test

twometresteve:
## Goal

Attempts to resolve a test flake by keeping the "app is launching" window open indefinitely.

## Design

We're not sure if this is definitely the cause of the flakes, but in theory the `lastRunInfo` file, which is written when the launch window ends, could have been getting corrupted by the app crashing (as per the scenario) while it was being written.  That would explain what we were seeing - no lastRunInfo included in the metadata for the next error, but if nothing else this change rules out a theory for the cause of the flake.

## Testing

Ran the test locally 10 times with no flakes, but only time will tell if it\s a complete fix.


lemnik (approved):
LGTM

PR #2328, opened 2025-11-10
implement ErrorOptions and CaptureOptions

YYChen01988:
## Goal

Introduced new capability to control which data fields are captured when reporting via notify function.

## Changeset

Added `ErrorOptions` to notify function in both `Client` and `Bugsnag`

## Testing

Updated existing tests

### DIFF
--- a/features/full_tests/error_capture_options.feature
+++ b/features/full_tests/error_capture_options.feature
@@ -9,7 +9,7 @@ Feature: ErrorCaptureOptions
     Then I wait to receive an error
     And the error is valid for the error reporting API version "4.0" for the "Android Bugsnag Notifier" notifier
 
-        # Stacktrace validation
+    # Stacktrace validation
     And the error payload field "events.0.exceptions.0.stacktrace" is a non-empty array
     And the event "exceptions.0.stacktrace.0.method" ends with "com.bugsnag.android.mazerunner.scenarios.Scenario.generateException"
     And the exception "stacktrace.0.file" equals "SourceFile"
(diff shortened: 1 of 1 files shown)


bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorOptions.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Controls which custom metadata tabs are included.
+     * - null: all metadata tabs captured
+     * - empty set: only app and device tabs captured
+     * - set of names: app, device, and specified tabs captured
+     *
+     * Note: app and device tabs are always captured.
+     */
+    var metadata: Set<String>? = null
```
lemnik:
```suggestion
    var metadata: MutableSet<String>? = null
```

bugsnag-android-core/src/main/java/com/bugsnag/android/ErrorOptions.kt
```diff
@@ -0,0 +1,82 @@
...
+                this.metadata = metadata
+            }
+        }
+
+        /**
+         * Return CaptureOptions that will not capture any optional fields.
+         */
+        @JvmStatic
+        fun captureNothing(): CaptureOptions {
+            return captureOnly(0)
```
lemnik:
```suggestion
            return captureOnly(0, emptySet())
```

yousif-bugsnag (approved):
LGTM

PR #2332, opened 2025-11-20
AppHang Plugin

lemnik:
## Goal
Allow ANR like reporting with a configurable plugin

## Design
The new `bugsnag-plugin-android-apphang` allows heartbeat based monitoring of a specific `Looper` (the `main` thread by default), and will report the target thread stacktrace as an error if it appears blocked for a configurable amount of time (3 seconds by default).

## Testing
Manually tested, added new e2e test and unit tests.


yousif-bugsnag (approved):
LGTM

PR #2337, opened 2025-11-25
Added Request/Response as optional properties on an Event

lemnik:
## Goal
Add model classes and properties for `Request` and `Response` to the `Event` model.

## Design
In order to report richer HTTP related errors we're adding optional first-party entities to the `Event` class.

## Testing
Added unit tests


bugsnag-android-core/src/main/java/com/bugsnag/android/Event.java
```diff
@@ -509,7 +509,7 @@ public boolean isUnhandled() {
     /**
      * Whether the event was a crash (i.e. unhandled) or handled error in which the system
      * continued running.
-     *
+     * <p>
```
gingerbenw:
```suggestion
     *
```

gingerbenw (approved):

PR #2345, opened 2025-12-05
Add patch version to device metadata

YYChen01988:
## Changeset

Including security patch version to device metadata

## Testing
updated existing unit tests and end to end tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * Added the OS security patch version to the reported device metadata
   [#2345](https://github.com/bugsnag/bugsnag-android/pull/2345)
+* Added `ErrorCaptureOptions.CAPTURE_ALL` field to more easily change only the captured metadata when specifying `ErrorOptions`
+  [#2358](https://github.com/bugsnag/bugsnag-android/pull/2358)
 
 ### Bug fixes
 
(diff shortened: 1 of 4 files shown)


features/smoke_tests/02_handled.feature
lemnik: I think we should check for these on all our smoke tests

bugsnag-plugin-android-ndk/src/main/jni/metadata.c
```diff
@@ -239,6 +239,12 @@ static void populate_device_metadata(JNIEnv *env, bugsnag_event *event,
                         sizeof(screen_resolution));
   bugsnag_event_add_metadata_string(event, "device", "screenResolution",
                                     screen_resolution);
+
+  char security_patch[64];
+  copy_map_value_string(env, data, "securityPatch", security_patch,
+                        sizeof(security_patch));
+  bugsnag_event_add_metadata_string(event, "device", "securityPatch",
+                                    security_patch);
```
lemnik:
```suggestion
  char security_patch[64] = {0};
  copy_map_value_string(env, data, "securityPatch", security_patch,
                        sizeof(security_patch));
  if (security_patch[0] != 0) {
    bugsnag_event_add_metadata_string(event, "device", "securityPatch",
                                      security_patch);
  }
```

PR #2351, opened 2025-12-15
Implement ErrorOptions for fatal notify

YYChen01988:
## Goal

Introduced `isFatal` error option to event

## Design

Event should only be delivered the last fatal error when app start again from the fatal crash

## Changeset

Added the `isFatal` to `errorOptions` and change the event to severity reason and delivery only one event when app has fatal crash

## Testing

Introduced new end to end tests


features/smoke_tests/04_unhandled.feature
```diff
@@ -313,3 +313,15 @@ Feature: Unhandled smoke tests
     And the event "threads.0.name" is not null
     And the event "threads.0.type" equals "c"
     And the thread with name "roid.mazerunner" contains the error reporting flag
+
+  Scenario: Unhandled exceptions with captureNone options
```
lemnik: I think this can rather be moved to a non-smoke test.

PR #2352, opened 2025-12-15
Add User info to synthesized ANR events

YYChen01988:
## Goal
Included User information to synthesized ANRs generated by the `exitinfo` plugin.

## Changeset
When the `exitinfo` plugin generates a synthesized ANR (one reported in `ApplicationExitInfo` but not captured by the `bugsnag-plugin-android-anr` module), the current (at time the event is synthesized) `User` is now added to the `Event`.

## Testing
Manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## TBD
+
+### Bug fixes
+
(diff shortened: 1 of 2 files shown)


lemnik (approved):
LGTM

PR #2363, opened 2026-01-06
Replace the heartbeat lock in AppHang Plugin

lemnik:
## Goal
Remove the heartbeat lock in the AppHang plugin to avoid possibly locking the main thread when the monitor holds the lock at the same time.

## Design
Replaced the lock & condition signalling with the `park`/`unpark` primitives. These only allow a single "block" / "unblock" behaviour, but the "parked" thread suits our use-case significantly better. This should reduce the overhead on the main thread.

## Changes

- Replaced the lock & condition with `park`/`unpark`
- Replaced the `elapsedRealtime` clock with `uptimeMillis` which doesn't tick during deep sleeps, avoiding deep sleep triggering false-positive AppHangs
- Moved the heartbeat re-posting to the monitor thread so that the heartbeat, improving the throughput on the main thread

```mermaid
sequenceDiagram
    activate monitor
    monitor->>main: heartbeat
    deactivate monitor
    activate main
    main->>monitor: unpark
    deactivate main
    activate monitor
    monitor->>main: heartbeat
    deactivate monitor
    activate main
    main->>monitor: unpark
    deactivate main
    activate monitor
```

## Testing
Relied on existing tests.

### DIFF
--- a/bugsnag-plugin-android-apphang/src/main/java/com/bugsnag/android/internal/LooperMonitorThread.kt
+++ b/bugsnag-plugin-android-apphang/src/main/java/com/bugsnag/android/internal/LooperMonitorThread.kt
@@ -69,11 +69,11 @@ internal class LooperMonitorThread(
 
             if (timeSinceLastHeartbeat >= appHangThresholdMillis) {
                 reportAppHang(timeSinceLastHeartbeat)
-            } else {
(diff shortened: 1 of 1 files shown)


YYChen01988 (approved):
LGTM

tomlongridge (approved):

PR #2427, opened 2026-06-04
Added New phase for NDK Plugin checks.

SB-sachinp:
In the diagnostic snapshot, we added phase-level instrumentation for NDK plugin load:

- plugin_init
- library_resolve
- load_library
- link_native
- post_init

Each phase emits structured fields:

- phase_start_ns
- phase_end_ns
- phase_duration_ns
- thread_name
- is_main_thread
- outcome (ok / error)
- error_class (only on error)

Plus an extra resolve check to improve root-cause quality:

- resolved_library_path
- resolved_library_exists

This gives exact failure/latency location instead of one opaque startup block.

- Improves NDK load diagnostics to split `queue_wait_ns`, `native_load_ns`, and `caller_blocked_ns`, making delays easier to attribute.
- Adds explicit retry reporting for native load (`load_retry_attempted`, `load_retry_succeeded`, first/second attempt outcomes).
- Captures original linker error message(s) alongside error class to better diagnose ABI/dependency/ELF/linker issues.
- Enhances library path diagnostics with mapped name, source (`nativeLibraryDir` vs fallback), confidence (`definitive`), file existence, and optional size.

### DIFF
--- a/bugsnag-android-core/src/main/java/com/bugsnag/android/LibraryLoader.java
+++ b/bugsnag-android-core/src/main/java/com/bugsnag/android/LibraryLoader.java
@@ -7,6 +7,76 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 class LibraryLoader {
 
+    static final class LoadLibraryReport {
+        final boolean loaded;
+        final boolean retried;
+        final boolean retrySucceeded;
+        final String firstErrorClass;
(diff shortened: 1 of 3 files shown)


SB-adityap (dismissed):
Done

SB-jigneshR (changes requested):
Blocking as it has specific changes.

PR #2430, opened 2026-06-18
updated change log

SB-sachinp:
### Enhancements

* **Security Hardening**: Enabled `_FORTIFY_SOURCE=2` and stack protector (`-fstack-protector-all`) for all native libraries (`libbugsnag-ndk.so`, `libbugsnag-plugin-android-anr.so`, `libbugsnag-root-detection.so`) to improve runtime buffer overflow detection.


SB-adityap (approved):
Ok

